PR #45, opened 2016-10-16
bound function should return package limits

etano:
Previously the bound function returned the sum of the package limits and the DRAM limits. This double counts DRAM if, as stated in Platform.hpp, the bound function should return the RAPL bounds.

### DIFF
--- a/src/RAPLPlatform.cpp
+++ b/src/RAPLPlatform.cpp
@@ -141,15 +141,7 @@ namespace geopm
 
     void RAPLPlatform::bound(double &upper_bound, double &lower_bound)
     {
-        double min_pkg;
-        double max_pkg;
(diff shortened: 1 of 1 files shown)


sssylvester:
The RAPL PKG domain does not include DRAM power. There is a RAPL PP0 domain that is supposed to include both PKG and DRAM, but it is broken and will not be supported in the future.

cmcantalupo:
Posted review to gerrit here:

https://review.gerrithub.io/#/c/298493/

etano:
Ok, so I suppose I was confused since there's some conflicting material around the web about RAPL. For example Figure 1 here provides a 3rd explanation: https://software.intel.com/en-us/articles/intel-power-governor

Issue #3, opened 2016-05-17
out-of-place-testing


sssylvester:
Building out of place works correctly, but testing out of place
does not, gtest links are broken and the license header check
does not work.

cmcantalupo:
Seems like ronn man page generation does not work with out of place builds either.

cmcantalupo:
This still doesn't work.

PR #169, opened 2018-02-19
Modifications performed to make geopm compile out-of source

luigigenovese:
I have performed some modifications on the build system to make geopm able to compile out-of-source.
Do not hesitate in correcting/amending them in the Makefile.am

Also I have removed the VERSION file from the .gitignore as it seems to be needed. I have added a empty file in the repository for the same reason.

This might be interesting in view of compiling the geopm together with an software package suite.

Many thanks

Luigi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,7 +120,6 @@
 /test/fortran/fortran_links/*
 /test_integration/geopm_test_integration
 /test_integration/geopm_test_integration.pyc
-/VERSION
 /doc/*.aux
 /doc/*.bbl
 /doc/*.blg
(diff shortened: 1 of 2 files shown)


cmcantalupo:
Thanks for your contribution Luigi!  The VERSION file is generated by running autogen.sh if the file is not present.  If the file is present but empty, then the VERSION file is not generated and the version is essentially set to the empty string.  The feature of not generating the VERSION file every time autogen.sh is run allows for the user to override the version by creating the file by hand, however, having an empty file breaks the build:

https://travis-ci.org/geopm/geopm/builds/343330834?utm_source=github_status&utm_medium=notification

Since the VERSION file is generated, it seems it should be in the .gitignore.  

I do like the changes to avoid the relative path references when creating the man pages with ronn.  I think those changes to the Makefile.am in isolation would improve the flexibility of the build.  There is an open issue for enabling out of place testing (one of our very first issues: https://github.com/geopm/geopm/issues/3).  I believe that the make check target still does not work out of place, but I haven't tested recently.  

Can you test this patch without the empty VERSION file and without the modification to .gitignore?  Are these changes somehow required for building out of place?

Issue #383, opened 2018-10-24
Refactor MSRIOGroup for easier extensibility


dianarg:
We would like a way to have an MSRIOGroup pull MSR descriptions from multiple files.
- Architectural registers are shared between multiple platforms. 
- Some data might change depending on SKU.

dianarg:
Consensus seems to be that moving the data from the msr_*cpp files into JSON format would be a good implementation for this.

dianarg:
https://github.com/geopm/geopm/commit/99a75f04feebf8ac271c6d452457d40ebfd38199
https://github.com/geopm/geopm/commit/ba2a071814871c78b67eebdf563ba307ce735d7c
https://github.com/geopm/geopm/commit/1b203661cce6b7881f697b1ada4ec3d62624afc9

Issue #281, opened 2018-06-25
Check RAPL_POWER_UNIT register and throw if values don't match expected


dianarg:
The scalar factors for power, energy, and time window have been hardcoded to the values observed for MSR_RAPL_POWER_UNIT.  If we encounter a machine where these units don't match, we should throw an error because all power and energy numbers will be off.

PR #497, opened 2019-01-25
Revert energy status units for DRAM on Broadwell

slabasan:
I believe the DRAM energy usage numbers for Broadwell and Haswell should be using 15.3 microJoules, instead of the value coming from RAPL_POWER_UNIT.


dianarg:
Thanks Stephanie.  Related to #281

dianarg:
Hi @slabasan, Can you confirm whether this number is supposed to be the one from MSR_RAPL_POWER_UNIT?  In which case I think it should be the same for MSR_PKG_ENERGY_STATUS.  

Also, on Quartz at least, I am still seeing the value of this register is 6.10352e-05.  If it's different on different systems, we will need to update how these MSRs find their units. #383 may be required

PR #502, opened 2019-01-30
Make tutorial shell scripts compatabile with more shell variants

cw585968:
- to eliminate syntax violation when `$GEOPM_LAUNCHER` is not set to srun or aprun and we want to use `$MPIEXEC` directly, we use quotes around the `$MPIEXEC` variable within the if test
- to use a more compatible syntax for string comparisons, we use single instead of double equal signs within the if test
- to use a more compatible syntax for sourcing files, we replace the shell specific `source` command with the dot `.` command

All changes can be reproduced by applying the following `sed` commands from within the GEOPM root directory.

```sd
$ sed -i 's/^elif \[ $MPIEXEC \]; then$/elif \[ "$MPIEXEC" \]; then/g' tutorial/tutorial_*.sh tutorial/agent/agent_tutorial.sh
$ sed -i 's/==/=/g' tutorial/tutorial_*.sh tutorial/agent/agent_tutorial.sh
$ sed -i "s/^source/./g" tutorial/tutorial_*.sh tutorial/agent/agent_tutorial.sh tutorial/tutorial_build_*.sh
```

### DIFF
--- a/tutorial/agent/agent_tutorial.sh
+++ b/tutorial/agent/agent_tutorial.sh
@@ -32,7 +32,7 @@
 #
 
 set err=0
-source ../tutorial_env.sh
+. ../tutorial_env.sh
 
 export PATH=$GEOPM_BINDIR:$PATH
 export PYTHONPATH=$GEOPMPY_PKGDIR:$PYTHONPATH
@@ -49,7 +49,7 @@ echo "Redirecting output to example.stdout and example.stderr."
 # launch geopm controller as an MPI process
 # create a report file
 # create trace files
-if [ "$GEOPM_LAUNCHER" == "srun" ]; then
+if [ "$GEOPM_LAUNCHER" = "srun" ]; then
     # Use GEOPM launcher wrapper script with SLURM's srun
     geopmlaunch srun \
                 -N 2 \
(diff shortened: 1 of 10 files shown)


cw585968:
Hm... 🤔  Looking at the assertions which failed (see below), there only seems to be some timing issues. Could this happen due to a heavily loaded test machine? Maybe someone could assess, whether the changes in this pull request even interfere with the failed tests.

```text
test/PowerGovernorAgentTest.cpp:148: Failure
The difference between 0.005 and geopm_time_diff(&start_time, &end_time) is 0.0014112640000000006, which exceeds 0.0001, where
0.005 evaluates to 0.0050000000000000001,
geopm_time_diff(&start_time, &end_time) evaluates to 0.0064112640000000007, and
0.0001 evaluates to 0.0001.
[  FAILED  ] PowerGovernorAgentTest.wait (12 ms)
[...]
[ RUN      ] TimeIOGroupTest.read_signal
test/TimeIOGroupTest.cpp:161: Failure
The difference between time1 - time0 and 1.0 is 0.0019916349999999916, which exceeds 0.001, where
time1 - time0 evaluates to 1.001991635,
1.0 evaluates to 1, and
0.001 evaluates to 0.001.
[  FAILED  ] TimeIOGroupTest.read_signal (1002 ms)
```

dianarg:
Thanks your contribution!  We will pull it over to gerrit for code review.

Sometimes those timing tests can fail depending on the conditions in Travis, so retriggering did indeed fix it.

cmcantalupo:
Review is posted here:

https://review.gerrithub.io/c/geopm/geopm/+/445603

PR #613, opened 2019-05-23
Check that MPIEXEC is set first before checking value of GEOPM_LAUNCHER

slabasan:
Even if MPIEXEC is set, the `tutorial_env.sh` script sets `GEOPM_LAUNCHER`, which is checked first in the if statement of the tutorial script.

### DIFF
--- a/tutorial/README.md
+++ b/tutorial/README.md
@@ -170,7 +170,7 @@ progress values recorded can be seen in the trace output.
 
 6. The model application
 ------------------------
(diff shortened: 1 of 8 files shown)


bgeltz:
FYI that I've pulled this into a Gerrit review here:
https://review.gerrithub.io/c/geopm/geopm/+/458023

I'm working on #614, and I think this needs to be merged before any of the other changes I'm doing in #614.

PR #888, opened 2020-01-23
Story 806

### LINKED ISSUE #874: As a system administator, I want to verify that the GEOPM static policy plugin applies limits to non-GEOPM jobs
> estimate 5

### Use Case
A system admin or site admin may want to apply limits such as a power cap or maximum frequency to all jobs on the system, whether they use GEOPM or not.  This can be enabled through the use of a resource manager plugin similar to https://github.com/geopm/geopm-slurm/blob/master/src/geopm_spank.c that uses the geopm_agent_enforce_policy() feature of GEOPM.  An administrator may want to test whether the plugin is installed and configured correctly with the agent specified in GEOPM's environment files.  Two common use cases should be covered:
- power limit using either PowerBalancerAgent or PowerGovernorAgent
- maximum or fixed frequency using either FrequencyMapAgent or EnergyEfficientAgent

### Done Criteria
- A test that runs a job script with a high-power operation without any GEOPM controller and reports power cap or frequency cap as detected by reading the corresponding MSRs.  The test should check these values against the expected values using the configuration specified in the environment-[default | override] file.

### LINKED ISSUE #884: Add geopmadmin tool to display location of GEOPM configuration files
> estimate 3

Feature enables end user to know were to install configuration files for default and override values for environment.

Parent story #806

Done criterion
--------------
- Documentation in geopmagent(1) describing command line option.
- Unit test for class that implements the feature.
- Integration test that shows that the output matches what is written in config.log.

dianarg:
Resolves #884 
Resolves #874


tutorial/admin/README.md
```diff
@@ -0,0 +1,91 @@
...
+2. Set the energy efficient agent as the default agent
+------------------------------------------------------
+When users run jobs with GEOPM, the built-in default agent is the monitor
+(no controls, monitoring only).  To change this default to the energy
+efficient agent so that jobs launched with GEOPM optimize energy efficiency,
+see the script "setup_default_energy_efficient.sh".  In this example, jobs
+not using GEOPM will run not have their frequency changed.  The test
+script XXXX can be used to check whether this setup is working.
+
+3. Restrict power limit for jobs that do not use the GEOPM runtime
```
cmcantalupo: Section header talks about the power limit, but the text uses the EE agent and talks about CPU frequency

tutorial/admin/setup_fixed_frequency.sh
```diff
@@ -0,0 +1,70 @@
...
+#     "FREQ_MAX": "NAN",
+#     "PERF_MARGIN": "NAN",
+#     "FREQ_FIXED": 1800000000
+#   }
+# for a system where the sticker frequency is 2.1 GHz and the target
+# frequency is 1.8 GHz.
+
+# Set the GEOPM configuration to use this policy file and the
+# energy efficient agent.
+geopmadmin --magic > $(geopmagent --config-default)
```
cmcantalupo: This is actually even more magic.  I think the --config-default would be the path the the environment default json.  We really want to populate the policy file that is referred to by the environment default.

dianarg: That should be handled by the geopmagent command above on 51.  We need two json files; what's missing is a command to generate the one below.

PR #895, opened 2020-01-27
Story 668: Introduce Policy class

dianarg:


test/Makefile.mk
```diff
@@ -342,6 +342,7 @@ GTEST_TESTS = test/gtest_links/AgentFactoryTest.static_info_monitor \
               test/gtest_links/TreeCommLevelTest.receive_up_complete \
               test/gtest_links/TreeCommLevelTest.receive_up_incomplete \
               test/gtest_links/TreeCommLevelTest.send_down \
+              test/gtest_links/TreeCommLevelTest.send_down_zero_value \
               test/gtest_links/TreeCommLevelTest.send_up \
```
dannosliwcd: New PolicyTest cases need to be added

ronn/geopm_agent_c.3.ronn
```diff
@@ -117,7 +117,9 @@ zero upon success; see [ERRORS][] section below for details.
     creates a JSON file based on the provided _policy_array_ that can be used
     for statically controlling the job policy for _agent_name_.  The output is
     stored in the _json_string_ buffer.  The number of bytes reserved for the
-    buffer is specified in _json_string_max_.
+    buffer is specified in _json_string_max_.  The size of the _policy_array_
+    must equal the total number of policy values available for the agent.
```
dannosliwcd: Optional: Suggest using the `_partial` variant when the caller desires to provide a subset of all values.

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+{
+    /// @brief The Policy class is used to handle operations on a
+    ///        policy as a vector of doubles, such as comparison and
+    ///        formatting.
+    class Policy
+    {
+        public:
+            Policy(const std::vector<std::string> &names,
+                   const std::vector<double> &values);
+            //Policy(std::intiastd::initializer_list<double> list);
```
dannosliwcd: leftover comment

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+            Policy(const Policy &other);// = default;
```
dannosliwcd: leftover comment

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+            /// @brief Equality comparison operator.  Trailing NANs are
+            ///        not considered when checking for equality.
+            bool operator==(const Policy &other) const;
+            bool operator!=(const Policy &other) const;
+
+            /// @brief Format the Policy vector as a character-delimited
+            ///        list.
+            std::string to_string(const std::string &delimiter) const;
+
+            // todo: from string?  e.g. "180,NAN,3"
```
dannosliwcd: leftover comment

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+            double &operator[](size_t index);
+
+            /// @brief Access a value of the policy by name or add if missing.
+            double &operator[](const std::string &name);
+
+            /// @brief Access a value of the policy by index and throw if missing.
+            double at(size_t index) const;
+
+            /// @brief Access a value of the policy by name and throw if missing.
+            double at(const std::string &name) const;
```
dannosliwcd: One group returns by reference, the other returns by value. Is this intentional?

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+            /// @brief Convert the policy values to a std::vector
+            std::vector<double> to_vector(void) const;
+
+            /// @brief Fill in NAN for missing values up until Policy
+            ///        reaches the given size.  The new size must be
+            ///        greater than or equal to the current size.
+            void pad_nan_to(size_t size);
+
+            // todo: Policy Policy::convert(const std::vector<std::string> &policy_names);
```
dannosliwcd: leftover comment

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+            // which creates a policy for a different agent, with values for common policy names transferred.
+
+
+        private:
+            //std::vector<double> m_values;
```
dannosliwcd: leftover comment

src/Policy.hpp
dannosliwcd: is copy-assign also needed?

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+            virtual ~Policy() = default;
+
+            /// @brief Returns the number of values in the policy.
+            size_t size(void) const;
+
+            /// @brief Returns the vector of policy names.
+            std::vector<std::string> policy_names(void) const;
+
+            /// @brief Access a value of the policy by index or add if missing.
+            double &operator[](size_t index);
```
dannosliwcd: Adds if missing and in-bounds, throws if missing and out of bounds

src/Policy.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+            /// @brief Access a value of the policy by name or add if missing.
```
dannosliwcd: Adds if missing and in the name set, throws if missing and not a valid name

src/Policy.cpp
```diff
@@ -0,0 +1,252 @@
...
+            else {
+                temp << geopm::string_format_double(m_values.at(m_names.at(ii)));
+            }
+        }
+        return temp.str();
+    }
+
+    std::string Policy::to_json(void) const
+    {
+        std::stringstream output_str;
```
dannosliwcd: ->`std::ostringstream`

src/Policy.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    void Policy::pad_nan_to(size_t size)
+    {
+        if (size < m_values.size()) {
+            throw Exception("Policy::pad_to_nan(): size of policy cannot be reduced.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        for (size_t ii = m_values.size(); ii < size; ++ii) {
+            m_values[m_names.at(ii)] = NAN;
```
dannosliwcd: Should this be checking whether the values are already set?

E.g. may cause issues when padding to 4 (p3 would be overwritten with NAN): `['p1', 'p2', 'p3', 'p4'] -> {'p1': 1, 'p3': 3}`

dannosliwcd (changes requested):

src/Policy.hpp
dianarg: I didn't realize that std::map .at() also has a reference version.  It looks a little weird to me.  This was intended to match the read-only behavior of at().

In other words std::map allows:
```
std::map<std::string, double> pol {{"one", 1}};
pol.at("one") = 2;
```

src/Policy.hpp
dianarg: I'll add more ctors to get closer to std::map

src/Policy.cpp
dianarg: Also missing check that size < m_names.size().  This seems to work okay but a unit test that checks values don't get overwritten.

dannosliwcd (dismissed):
LGTM in its current state.

We had an offline discussion where you proposed some usability changes regarding construction of this class, implicitly adding internal NANs vs leaving gaps in the data structure, and whether to_json/from_json should be symmetric. I can re-review if you plan to add those changes in this PR. If you prefer to save those issues for a separate issue, I think that's fine too.

PR #896, opened 2020-01-28
Add scenarios for override settings

dianarg:


cmcantalupo (commented):
Seems like a good start.  I was thinking that we would name the scripts after the numbered step (e.g. setup_1.sh things may not break down exactly like that.

tutorial/admin/README.md
```diff
@@ -12,7 +12,71 @@ The directory contains scripts that can be used to configure systems
 for use with the GEOPM runtime.  The scripts associated with each
 section are numbered accordingly.
 
-0. Enable users to do research with the GEOPM runtime
+0. Set up a system to use the GEOPM static policy plugin for SLURM
+-------------------------------------------------------------------
+GEOPM provides two libraries: libgeopmpolicy contains tools for interacting
+with hardware signals and controls, such as geopmread, and the supporting
+library functions.  libgeopm contains all the same functions and additionally
+provides tools for launching and interacting with MPI applications. The slurm
```
bgeltz: s/slurm/SLURM

tutorial/admin/README.md
```diff
@@ -12,7 +12,71 @@ The directory contains scripts that can be used to configure systems
...
+        ```
+    - Run configure with a local prefix that will be used to build the plugin:
+        ```
+        ./configure --disable-mpi --disable-fortran --prefix=$HOME/build/geopm-no-mpi
+        ```
+    - Build and install for use when building the slurm plugin
+        ```
+        make -j && make install
+        ```
+    - Build an RPM to be used to install geopmpolicy on the compute nodes:
```
bgeltz: s/geopmpolicy/libgeopmpolicy

tutorial/admin/README.md
```diff
@@ -12,7 +12,71 @@ The directory contains scripts that can be used to configure systems
...
+        ```
+B.  Install the libgeopmpolicy package in the compute nodes using the RPM.
+The library must be accessible to the root user when the plugin runs,
+for example in /usr/lib64, and this version of the library should be built
+against a toolchain available in the system default paths.
+
+C.  Install libspank_geopm.so* into /usr/lib64/slurm on the compute nodes.
+
+D.  Create or update plugstack.conf in /etc/slurm on the compute nodes
+only with the following:
```
bgeltz: s/only with/to contain

tutorial/admin/README.md
```diff
@@ -12,7 +12,71 @@ The directory contains scripts that can be used to configure systems
...
+    ```
+    optional  libgeopm_spank.so
+    ```
+The head node should not need plugstack.conf; or if present, it should not
+contain any reference to libgeopm_spank.so.
+
+E.  Update the Slurm configuration (/etc/slurm/slurm.conf) as follows:
+    - Update PluginDir to contain "/usr/lib64/slurm"
+    - Make sure that PlugStackConfig is not set.
```
bgeltz: This is OK imo.  If the admin wants to put the plugstack.conf somewhere other than the default, that's OK as long as libgeopm_spank.so is present in the file.

tutorial/admin/setup_default_energy_efficient.sh
```diff
@@ -41,29 +41,40 @@
...
 # This file should look similar to the following:
 # {
 #    "GEOPM_AGENT": "energy_efficient",
 #    "GEOPM_POLICY": "/etc/geopm/energy_efficient_policy.json"
 # }
+
+# Example sanity checks of the configuration
+#
+# Fixed frequency for non-GEOPM jobs stays at max
+#   > srun --reservation=plugin_freq_cap geopmread MSR::PERF_CTL:FREQ board 0
```
bgeltz: Consider adding something that explains about the --reservation flag.  Otherwise I'd remove it.

bgeltz (changes requested):
A few little things.

bgeltz (approved):

PR #898, opened 2020-01-29
Add integration test of static policy plugin

### LINKED ISSUE #874: As a system administator, I want to verify that the GEOPM static policy plugin applies limits to non-GEOPM jobs

bakerbrandond:
- Asserts that expected site policy is enforced on non-GEOPM jobs
- Fixes #874

Change-Id: I32b4197323e2814e20ba122ab4f32b6ca2eef604

### DIFF
--- a/src/Environment.hpp
+++ b/src/Environment.hpp
@@ -74,6 +74,7 @@ namespace geopm
             virtual bool do_ompt(void) const = 0;
             virtual std::string default_config_path(void) const = 0;
             virtual std::string override_config_path(void) const = 0;
+            static std::map<std::string, std::string> parse_environment_file(const std::string &env_file_path);
     };
 
     class EnvironmentImp : public Environment
@@ -107,13 +108,16 @@ namespace geopm
             bool do_profile() const override;
             int timeout(void) const override;
             int debug_attach(void) const override;
-            static std::set<std::string> get_all_vars();
+            static std::set<std::string> get_all_vars(void);
             bool do_ompt(void) const override;
             std::string default_config_path(void) const override;
             std::string override_config_path(void) const override;
+            static void parse_environment_file(const std::string &settings_path,
+                                               const std::set<std::string> &all_names,
+                                               const std::set<std::string> &user_defined_names,
+                                               std::map<std::string, std::string> &name_value_map);
         protected:
-            void parse_environment();
-            void parse_environment_file(const std::string &settings_path);
+            void parse_environment(void);
             bool is_set(const std::string &env_var) const;
             std::string lookup(const std::string &env_var) const;
             const std::set<std::string> m_all_names;
--- a/test_integration/test_plugin_static_policy.py
+++ b/test_integration/test_plugin_static_policy.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+#
+#  Copyright (c) 2015, 2016, 2017, 2018, 2019, Intel Corporation
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#      * Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#
+#      * Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in
+#        the documentation and/or other materials provided with the
+#        distribution.
+#
+#      * Neither the name of Intel Corporation nor the names of its
+#        contributors may be used to endorse or promote products derived
+#        from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+#  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+#  A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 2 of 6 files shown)


test_integration/test_plugin_static_policy.py
```diff
@@ -0,0 +1,115 @@
...
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from test_integration import geopm_context
+import geopmpy.io
+from test_integration import geopm_test_launcher
+from test_integration import util
+import geopmpy.topo
+
+# todo this function results in test launcher
+# invocation, and is called 3 times (1 by class and
+# in 2 decorator tests).  may be useful to add once logic
```
bgeltz: Is there an issue open for this?

bgeltz: Is there an issue open for this open already?

test_integration/test_plugin_static_policy.py
```diff
@@ -0,0 +1,115 @@
...
+
+    """
+    @classmethod
+    def setUpClass(cls):
+        cls._tmp_files = []
+        cls._geopmadminagent = geopmadminagent()
+        cls._geopmadminagentpolicy = geopmadminagentpolicy()
+
+    @unittest.skipUnless(geopmadminagent() in ['frequency_map', 'energy_efficient'],
+                         'Requires environment defaut/override to be configured to cap frequency.')
```
bgeltz: Spelling: defau*L*t/override

test_integration/test_plugin_static_policy.py
```diff
@@ -0,0 +1,115 @@
...
+        elif self._geopmadminagent == 'energy_efficient':
+            test_freq = self._geopmadminagentpolicy['FREQ_FIXED']
+        current_freq = geopm_test_launcher.geopmread("MSR::PERF_CTL:FREQ board 0")
+        self.assertEqual(test_freq, current_freq)
+
+    @unittest.skipUnless(geopmadminagent() in ['power_governor','power_balancer'],
+                         'Requires environment defaut/override to be configured to cap power.')
+    def test_power_cap_enforced(self):
+        test_power = self._geopmadminagentpolicy['POWER_PACKAGE_LIMIT_TOTAL']
+        # todo need test_launcher to execute topo on compute node
```
bgeltz: Issue for this too please.

test_integration/test_plugin_static_policy.py
```diff
@@ -0,0 +1,115 @@
...
+        num_pkg = geopmpy.topo.num_domain('package')
+        current_power = num_pkg * geopm_test_launcher.geopmread("MSR::PKG_POWER_LIMIT:PL1_POWER_LIMIT board 0")
```
bgeltz: It isn't safe to assume that the PL1 limit is the same on both sockets if multisocket.

I was going to suggest you use num_pkg to loop over the packages, and use geopmread to read the limit on the individual sockets.  But I see you're reading the board limit...

I think this should fail:  reading the board limit should read each package and sum them I think.  There's a bug somewhere here, I think it's in how geopmread sums (or not) the values for the board domain if this logic works.

bgeltz (changes requested):
A few nits, and a (I think) a bonafide bug.

test_integration/test_plugin_static_policy.py
bakerbrandond: inherited from here: https://github.com/geopm/geopm/blob/dev/test_integration/test_static_policy.py#L173-L180
shall we file a bug?

test_integration/test_plugin_static_policy.py
bakerbrandond: added a decorator to the test class to require the test be ran in an allocation to resolve this (same as the static (non plugin) policy integration test)

test_integration/test_plugin_static_policy.py
bakerbrandond: There is a subtlety here that we both missed.  The policy name is: POWER_PACKAGE_LIMIT_TOTAL, as in total for all of the packages on the board.  That's why sampling board works but yes this implies they packages split the allotted power equally.  Is a non uniform distribution of power across packages expected to appear in the wild?

src/Environment.hpp
```diff
@@ -107,13 +108,16 @@ namespace geopm
             bool do_profile() const override;
             int timeout(void) const override;
             int debug_attach(void) const override;
-            static std::set<std::string> get_all_vars();
+            static std::set<std::string> get_all_vars(void);
             bool do_ompt(void) const override;
             std::string default_config_path(void) const override;
             std::string override_config_path(void) const override;
+            static void parse_environment_file(const std::string &settings_path,
```
cmcantalupo, line 115: This static method could be private right?

cmcantalupo (commented):
That should work

src/Environment.hpp
bakerbrandond: Nope the Environment (base class) needs access to this static public member of the EnvironmentImp class line 86 in green of this patch

bgeltz (commented):
The integration test looks good now.  Thanks for fixing that up.  I think that bit is g2g.

I assume you and Chris are hashing out the rest, so I didn't approve.

PR #899, opened 2020-01-31
Update EE agent to save energy assuming recent hardware

### LINKED ISSUE #865: Modify the energy efficient agent so that it changes frequency upon region entry and reduces controller wait if hardware supports fast p-state transitions
> estimate 3

Energy efficient agent should check if the hardware can support fast p-state transitions.  If it can, the control loop should be reduced to 2 ms and the number of samples required to transition frequency should be reduced to zero.  
SEE COMMENTS BELOW

Requires: #864 

Parent story: #855

bakerbrandond:
Fixes #865.

Change-Id: I8d44ddea22bcce0bddfc3517266a1e1b6f294c66


dianarg (commented):
👍

bgeltz (approved):

PR #900, opened 2020-01-31
Finish implementation of Admin class geompadmin tool and unit testing

### LINKED ISSUE #891: geopmadmin tool to aid system administrators in checking configuration files
> estimate 5

Parent Story #806

cmcantalupo:
Fixes #891

### DIFF
--- a/src/Admin.cpp
+++ b/src/Admin.cpp
@@ -31,37 +31,64 @@
  */
 
 #include "config.h"
-#include <sstream>
+
+#include <cmath>
+#include <cpuid.h>
 #include <memory>
+#include <sstream>
+
 #include "Admin.hpp"
+#include "Agent.hpp"
 #include "Environment.hpp"
 #include "Exception.hpp"
+#include "FilePolicy.hpp"
+#include "Helper.hpp"
 #include "MSRIOGroup.hpp"
+#include "OptionParser.hpp"
+
 
 namespace geopm
 {
     Admin::Admin()
         : Admin(geopm::environment().default_config_path(),
                 geopm::environment().override_config_path(),
-                nullptr)
+                MSRIOGroup::cpuid())
     {
 
     }
 
     Admin::Admin(const std::string &default_config_path,
                  const std::string &override_config_path,
-                 std::shared_ptr<MSRIOGroup> msriog)
+                 int cpuid_local)
         : m_default_config_path(default_config_path)
         , m_override_config_path(override_config_path)
-        , m_msriog(msriog)
+        , m_cpuid_local(cpuid_local)
     {
 
     }
 
-    std::string Admin::main(bool do_default,
-                            bool do_override,
-                            bool do_whitelist,
-                            int cpuid)
+    void Admin::main(int argc, const char **argv, std::ostream &std_out, std::ostream &std_err)
+    {
+        OptionParser par = parser(std_out, std_err);
+        bool early_exit = par.parse(argc, argv);
+        if (!early_exit) {
+            auto pos_args = par.get_positional_args();
+            if (!pos_args.empty()) {
+                throw geopm::Exception("The following positional argument(s) are in error: " +
+                                       geopm::string_join(pos_args, " "),
+                                       EINVAL, __FILE__, __LINE__);
+            }
+            std_out << run(par.is_set("default"),
+                           par.is_set("override"),
+                           par.is_set("whitelist"),
+                           std::stoi(par.get_value("cpuid"), NULL, 16));
+        }
+    }
+
+    std::string Admin::run(bool do_default,
+                           bool do_override,
+                           bool do_whitelist,
+                           int cpuid)
     {
         int action_count = 0;
         action_count += do_default;
@@ -88,6 +115,22 @@ namespace geopm
         return result;
     }
 
+    OptionParser Admin::parser(std::ostream &std_out, std::ostream &std_err)
+    {
+        OptionParser result{"geopmadmin", std_out, std_err, ""};
+        result.add_option("default", 'd', "config-default", false,
+                          "print the path of the GEOPM default configuration file");
+        result.add_option("override", 'o', "config-override", false,
+                          "print the path of the GEOPM override configuration file");
+        result.add_option("whitelist", 'w', "msr-whitelist", false,
+                          "print the minimum msr-safe whitelist required by GEOPM");
+        result.add_option("cpuid", 'c', "cpuid", "-1",
+                          "cpuid in hexidecimal for whitelist (default is current platform)");
+        result.add_example_usage("");
+        result.add_example_usage("[--config-default|--config-override|--msr-whitelist] [--cpuid]");
+        return result;
+    }
+
     std::string Admin::default_config(void)
     {
         return m_default_config_path + "\n";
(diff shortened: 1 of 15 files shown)


src/Admin.cpp
```diff
@@ -100,23 +143,118 @@ namespace geopm
...
+        return geopm::MSRIOGroup::msr_whitelist(cpuid);
+    }
+
+    std::vector<std::string> Admin::dup_keys(const std::map<std::string, std::string> &map_a,
+                                             const std::map<std::string, std::string> &map_b)
+    {
+        std::vector<std::string> result;
+        auto a_it = map_a.begin();
+        auto b_it = map_b.begin();
+        while (a_it != map_a.end() && b_it != map_b.end()) {
```
bakerbrandond, line 158: This looks like it is incomplete if a has less keys than b, no?

bakerbrandond: Maybe create 2 sets of keys (looping thru both maps independently) and return the intersection?

src/Admin.cpp
```diff
@@ -100,23 +143,118 @@ namespace geopm
...
+        // Check for parameters that are defined in both files
+        std::vector<std::string> overlap = dup_keys(default_map, override_map);
+        if (overlap.size()) {
+            throw Exception("Admin::check_node: "
+                            "parameter(s) defined in both the override and default files: \"" +
+                             string_join(overlap, "\", \"") + "\"\n",
+                             EINVAL, __FILE__, __LINE__);
+        }
+        // Combine settings
+        std::map<std::string, std::string> config_map(override_map);
```
bakerbrandond, line 189: I know we assert that there are no overlaps so order doesn't matter, but default first here would give me the warm and fuzzies and mirror what happens in the Environment class.

src/Admin.hpp
```diff
@@ -30,30 +30,53 @@
...
 #include <memory>
 #include <string>
+#include <vector>
+#include <map>
 
 namespace geopm
 {
-    class MSRIOGroup;
+    class OptionParser;
+    class Environment;
```
bakerbrandond: rm?

test/MSRIOGroupTest.cpp
```diff
@@ -83,11 +83,11 @@ class MSRIOGroupTest : public :: testing :: Test
         void mock_enable_fixed_counters(void);
 };
 
-class MockMSRIO : public geopm::MSRIOImp
+class MockForGroupMSRIO : public geopm::MSRIOImp
```
bakerbrandond: I don't think I agree/understand the reason for these changes... We are worsening the coverage of the real MSRIOGroup class... No?

bakerbrandond (changes requested):
Few comments.  We have to make sure to merge my story-806 branch PR first

src/Admin.cpp
```diff
@@ -100,23 +143,118 @@ namespace geopm
...
+            if (override_map.find(config.first) != override_map.end()) {
+                result << " (override)\n";
+            }
+            else {
+                result << " (default)\n";
+            }
+        }
+        if (policy_vals.size()) {
+            result << "\n";
+            result << "AGENT POLICY\n"
```
dianarg, line 247: Might want to revisit this with the Policy class refactoring.  If the original JSON file has gaps in the policy values, they get filled in with NAN, so this may have more keys listed than are in the original file.

src/Helper.cpp
```diff
@@ -128,6 +129,20 @@ namespace geopm
...
+                            const std::string &delim)
+    {
+        std::ostringstream result;
+        if (list.size() > 0) {
+            auto back_it = list.end() - 1;
+            for (auto str_it = list.begin(); str_it != back_it; ++str_it) {
+                result << *str_it << delim;
+            }
+            result << *back_it;
+        }
```
dianarg: Given this is a vector not a list, could be simpler to just do ` for i=0..list.size()` and check  i!=0 for delimiter

test/EndpointTest.cpp
```diff
@@ -241,7 +241,7 @@ TEST_F(EndpointTest, get_hostnames)
         hostlist << host << "\n";
     }
     hostlist.close();
-    strncpy(data->hostlist_path, hostlist_path.c_str(), GEOPM_ENDPOINT_HOSTLIST_PATH_MAX);
+    strncpy(data->hostlist_path, hostlist_path.c_str(), GEOPM_ENDPOINT_HOSTLIST_PATH_MAX - 1);
     strncpy(data->agent, "monitor", GEOPM_ENDPOINT_AGENT_NAME_MAX);
```
dianarg, line 245: here too?  there are a bunch more if you grep for strncpy

src/Admin.cpp
dianarg: I think it will work as long as keys are ordered with < (should be true for std::map).

src/Admin.cpp
```diff
@@ -100,23 +143,118 @@ namespace geopm
...
-        throw geopm::Exception("Check has not been implemented yet",
-                               GEOPM_ERROR_NOT_IMPLEMENTED, __FILE__, __LINE__);
-        return "";
+        std::map<std::string, std::string> default_map =
+            Environment::parse_environment_file(m_default_config_path);
+        std::map<std::string, std::string> override_map =
+            Environment::parse_environment_file(m_override_config_path);
+        // Check for parameters that are defined in both files
+        std::vector<std::string> overlap = dup_keys(default_map, override_map);
+        if (overlap.size()) {
```
dianarg: `overlap.size() > 0`

src/Admin.cpp
cmcantalupo: Should work to find all duplicates.  Do not have to keep looking once one of the maps has been exhausted.

src/Admin.cpp
cmcantalupo: This copies all of the elements from override_map into config_map, and then inserts any keys that are in the default map and *not* in the config_map into the config map.  I.e. this does replicate the behavior in Environment.

test/MSRIOGroupTest.cpp
cmcantalupo: This change is no longer necessary.  It was introduced because I had introduced a different object that I called a "MockMSRIO" in an earlier version.  I got rid of that mock in a newer revision, so this name can be changed back.

src/Admin.cpp
cmcantalupo: Okay I was worried about what changes from the Policy class I might be needing here.

test/EndpointTest.cpp
cmcantalupo: I just fixed the ones that gcc8 was complaining about.

src/Admin.cpp
cmcantalupo: overlap.size() > 0U?

tomato tomato.  I can change, but I don't like it.

src/Helper.cpp
cmcantalupo: Probably would compile to the same object code, but this avoids a branch in the loop.

src/Admin.hpp
cmcantalupo: yes thanks!

src/Helper.cpp
cmcantalupo: The bigger issue is this function has no unit test and no doxygen :(

dianarg (approved):

PR #923, opened 2020-02-06
Add timed_stream model region for geopmbench

### LINKED ISSUE #863: Modify geopmbench so that it can support short running regions
> estimate 5

Introduce a stream region that at startup calibrates its execution by time rather than amount of memory.  This can be executed without threading on a subset of cores to provide a CPU bound short running region with predictable execution time.

Parent story #855

dianarg:
- Learned big-o may be highly variable run-to-run without
  explictly setting OMP_NUM_THREADS.
- Resolves #863.

### DIFF
--- a/src/ModelRegion.cpp
+++ b/src/ModelRegion.cpp
@@ -43,6 +43,7 @@
 #include "All2allModelRegion.hpp"
 #include "DGEMMModelRegion.hpp"
 #include "StreamModelRegion.hpp"
+#include "TimedStreamModelRegion.hpp"
 #include "SpinModelRegion.hpp"
 #include "IgnoreModelRegion.hpp"
 
@@ -83,6 +84,9 @@ namespace geopm
         }
         else if (name_check(name, "stream")) {
             return geopm::make_unique<StreamModelRegion>(big_o, verbosity, do_imbalance, do_progress, do_unmarked);
+        }
+        else if (name_check(name, "timed_stream")) {
+            return geopm::make_unique<TimedStreamModelRegion>(big_o, verbosity, do_imbalance, do_progress, do_unmarked);
         }
         else if (name_check(name, "all2all")) {
             return geopm::make_unique<All2allModelRegion>(big_o, verbosity, do_imbalance, do_progress, do_unmarked);
--- a/Makefile.am
+++ b/Makefile.am
@@ -669,6 +669,8 @@ model_source_files = src/All2allModelRegion.cpp \
                      src/SpinModelRegion.hpp \
                      src/StreamModelRegion.cpp \
                      src/StreamModelRegion.hpp \
+                     src/TimedStreamModelRegion.cpp \
+                     src/TimedStreamModelRegion.hpp \
                      # end
 
 if ENABLE_MPI
--- a/copying_headers/MANIFEST.BSD3-intel
+++ b/copying_headers/MANIFEST.BSD3-intel
@@ -225,6 +225,8 @@ src/SpinModelRegion.cpp
 src/SpinModelRegion.hpp
 src/StreamModelRegion.cpp
 src/StreamModelRegion.hpp
+src/TimedStreamModelRegion.cpp
+src/TimedStreamModelRegion.hpp
 src/TimeIOGroup.cpp
 src/TimeIOGroup.hpp
 src/Tracer.cpp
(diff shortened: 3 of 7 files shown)


src/ModelRegion.cpp
```diff
@@ -427,11 +436,80 @@ namespace geopm
...
+            std::cerr << "Warning: <geopm> timed_stream region may run out of memory for large big-o values." << std::endl;
+        }
+
+        if (m_verbosity) {
+            std::cout << "Calibrating timed_stream region to " << big_o_in << " seconds.  Please wait..." << std::endl;
+        }
+        double measured_time = 0;
+        bool done = false;
+        double new_big_o = big_o_in;
+        double lower = 0.0;
```
bgeltz: Per TravisCI, `lower` is set but not used.

src/ModelRegion.cpp
```diff
@@ -427,11 +436,80 @@ namespace geopm
...
+                                                   int verbosity,
+                                                   bool do_imbalance,
+                                                   bool do_progress,
+                                                   bool do_unmarked)
+        : StreamModelRegion(big_o_in, verbosity, do_imbalance, do_progress, do_unmarked)
+    {
+        m_name = "timed_stream";
+
+        if (big_o_in > 10.0) {
+            /// @todo: adjust to reuse same buffer when big-o would exhaust system memory. This depends on number of ranks.
```
bgeltz: Is this future work?
Is there a future work issue open for this?

src/ModelRegion.cpp
```diff
@@ -427,11 +436,80 @@ namespace geopm
...
+            geopm_time(&start);
+            run();
+            measured_time = geopm_time_since(&start);
+            if (m_verbosity) {
+                std::cout << "stream big-o=" << new_big_o << ", "
+                          << "runtime=" << measured_time << "s" << std::endl;
+            }
+
+            if (measured_time > big_o_in + EPS) {
+                // if too long, search between current big_o and lower bound
```
bgeltz: `and lower bound` does not seem to be satisfied here.

src/ModelRegion.cpp
```diff
@@ -427,11 +436,80 @@ namespace geopm
...
+            std::cout << "Calibrating timed_stream region to " << big_o_in << " seconds.  Please wait..." << std::endl;
+        }
+        double measured_time = 0;
+        bool done = false;
+        double new_big_o = big_o_in;
+        double lower = 0.0;
+        const double EPS = big_o_in / 100.0;
+        const int MAX_ITERATIONS = 20;
+        int iteration_count = 0;
+        double step_size = new_big_o;
```
bgeltz: It looks like the intention of the algorithm w.r.t. evaluating the search space is to:
1. Halve the the step size if it ran too long.
2. If it ran too short, add the step size to the big O and re-run.
a. If it was too short on the first eval, big O is doubled.  Otherwise, big-O is set to <= 0.5 `big_o_in`.

I just want to be sure I'm groking that right.  A comment would be nice.

bgeltz (changes requested):

src/ModelRegion.cpp
dianarg: Ideally we would do this for all regions types, not just this new one.  The current behavior is to die without explanation.  I'll open a new issue and remove this todo

src/ModelRegion.cpp
dianarg: Yes that's correct.  I'm not 100% sure if this is the best algorithm. The original version was increasing the big-o by (0.5 * (upper-lower)) but this is simpler and seems to converge faster when the initial big-o is too small. I haven't done very many experiments with different big-o values and OMP_NUM_THREADS combinations.

src/ModelRegion.cpp
```diff
@@ -407,7 +412,11 @@ namespace geopm
         }
     }
 
-    StreamModelRegion::StreamModelRegion(double big_o_in, int verbosity, bool do_imbalance, bool do_progress, bool do_unmarked)
+    StreamModelRegion::StreamModelRegion(double big_o_in,
```
bakerbrandond: StreamModelRegion ctor needs to be updated to take name as parameter to be used properly by TimedStreamModelRegion.  m_name being updated at 446 below is too late.  Subsequently the name in the report is not "timed_stream" as expected.

PR #932, opened 2020-02-12
Fixes for energy efficient analysis

dianarg:
- Fix for range of frequencies used in parsing files during skip launch.
- Remove enable-turbo option
- Fix detected max frequency to use PlatformIO
- Fix for report/trace cache file names with python3


bgeltz (changes requested):
ronn updates please

bgeltz (changes requested):
Thanks for the docu update.  We need to bottom out on the integration testing issues, then this is g2g

PR #934, opened 2020-02-13
Recent hw fixup

bakerbrandond:
These changes are to patch up changes that were introduced in f4bfb179.
+ Update FrequencyMap agent gate to 2ms such that it can be compared to EE agent.
+ Update EE agent, removing minimum region runtime requirement

### DIFF
--- a/src/EnergyEfficientAgent.cpp
+++ b/src/EnergyEfficientAgent.cpp
@@ -67,7 +67,6 @@ namespace geopm
                                                std::map<uint64_t, std::shared_ptr<EnergyEfficientRegion> > region_map)
         : M_PRECISION(16)
         , M_WAIT_SEC(0.002)
-        , M_MIN_LEARNING_RUNTIME(M_WAIT_SEC * 10)
         , M_POLICY_PERF_MARGIN_DEFAULT(0.10)  // max 10% performance degradation
         , m_platform_io(plat_io)
         , m_platform_topo(topo)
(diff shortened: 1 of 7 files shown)


dianarg (approved):
LGTM

src/EnergyEfficientAgent.cpp
```diff
@@ -249,10 +248,6 @@ namespace geopm
                                         "(): region exit before entry detected.",
                                         GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
                     }
-                    if (last_region_info.runtime != 0.0 &&
-                        last_region_info.runtime < M_MIN_LEARNING_RUNTIME) {
-                        last_region_it->second->disable();
```
cmcantalupo, line 254: looks like this is the only use of the EnergyEfficientRegion::disable() api.  This method should be deleted from the class.  What was the check for `last_region_info.runtime != 0.0` for?  Why is deleting it okay?

src/FrequencyMapAgent.cpp
```diff
@@ -297,7 +297,7 @@ namespace geopm
 
     void FrequencyMapAgent::wait(void)
     {
-        static double M_WAIT_SEC = 0.005;
+        static double M_WAIT_SEC = 0.002;
```
cmcantalupo: not related to the change, but why is this non-member variable prefixe with M_?

src/EnergyEfficientAgent.cpp
bakerbrandond: 0.0 is the starting runtime for regions prior to first observation, the check was there to prevent disabling a region before it has ran.

yes this is the only place disable is used, there is a todo in the commit message.  I was unsure of how big of a change we wanted here an now.  if the removal of the api is a blocker for merge then I can take care of that now.

cmcantalupo (approved):
LGTM

PR #935, opened 2020-02-13
Add include guard check to travis

dianarg:
- Fix headers for OMPT and PolicyStoreImp.


cmcantalupo (commented):
Looks good to me.  Similarly nice check would be to look for `#include "conflg.h"` in *.c and *.cpp files. I'm okay with the early bail out, but I don't have a strong opinion.

PR #938, opened 2020-02-14
Adds three model regions: scaling, reduce and barrier

### LINKED ISSUE #863: Modify geopmbench so that it can support short running regions

### LINKED ISSUE #922: YAML-backed RawReport cannot read frequency_map agent reports
The yaml-based RawReport reader is unable to read frequency_map agent reports. These reports contain a "Frequency map" section where each entry is prefixed with a '\t' character, instead of spaces. Example:

```
Host: mcfly7
Frequency map:
	0x00000000725e8066:3.7000000000000000e+09
	0x00000000f01e38a3:3.7000000000000000e+09
```

YAML doesn't allow tabs for structural indentation, and the RawReport constructor raises a yaml error:

```
yaml.scanner.ScannerError: while scanning for the next token
found character '\t' that cannot start any token
```

One fix may be to modify the frequency_map agent to prefix these lines with spaces instead of tabs.

### LINKED ISSUE #926: Add a Model region whose completion isn't time-based and uses MPI
> estimate 3

Parent story #918

cmcantalupo:
- Fixes #926 
- Fixes #922
- Fixes #863 
- Scaling region has an integration test.
- Inspection of scaling region results looks reasonable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,6 +143,7 @@
 /test_integration/test_ee_stream_dgemm_mix
 /test_integration/test_omp_outer_loop
 /test_integration/test_profile_policy
+/test_integration/test_scaling_region
 /test_integration/test_static_policy
 /tutorial/*.o
 /tutorial/Exception.hpp
--- a/src/ReduceModelRegion.cpp
+++ b/src/ReduceModelRegion.cpp
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "ReduceModelRegion.hpp"
+
+#include <iostream>
+#include <vector>
+#include <mpi.h>
+
+#include "Exception.hpp"
+
+namespace geopm
+{
+    ReduceModelRegion::ReduceModelRegion(double big_o_in,
+                                         int verbosity,
+                                         bool do_imbalance,
+                                         bool do_progress,
+                                         bool do_unmarked)
+        : ModelRegion(verbosity)
+    {
+        big_o(big_o_in);
+    }
+
+    void ReduceModelRegion::big_o(double big_o)
+    {
+        m_num_elem = 67108864 * big_o; // 64 MiB per big_o
+        m_send_buffer.resize(m_num_elem);
+        m_recv_buffer.resize(m_num_elem);
+        std::fill(m_send_buffer.begin(), m_send_buffer.end(), 1.0);
+        std::fill(m_recv_buffer.begin(), m_recv_buffer.end(), 0.0);
+    }
+
+    void ReduceModelRegion::run(void)
+    {
+        int num_rank = 0;
+        int err = MPI_Comm_size(MPI_COMM_WORLD, &num_rank);
+        if (err) {
+            throw Exception("MPI_Comm_size", err, __FILE__, __LINE__);
+        }
+        if (m_verbosity != 0) {
+            std::cout << "Executing reduce\n";
+        }
+        err = MPI_Allreduce(m_send_buffer.data(), m_recv_buffer.data(),
+                            m_num_elem, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+        if (err) {
+            throw Exception("MPI_Allreduce", err, __FILE__, __LINE__);
+        }
+    }
+}
--- a/src/ScalingModelRegion.cpp
+++ b/src/ScalingModelRegion.cpp
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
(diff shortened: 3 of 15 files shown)


src/ReduceModelRegion.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    void ReduceModelRegion::run(void)
+    {
+        int num_rank = 0;
+        int err = MPI_Comm_size(MPI_COMM_WORLD, &num_rank);
+        if (err) {
+            throw Exception("MPI_Comm_size", err, __FILE__, __LINE__);
+        }
+        std::vector<double> send_buffer(m_num_elem, 1.0);
```
cmcantalupo: Allocation in constructor would be better.

test_integration/test_scaling_region.py
```diff
@@ -0,0 +1,194 @@
...
+
+        """
+        script_dir = os.path.dirname(os.path.realpath(__file__))
+        return os.path.join(script_dir, '.libs', 'test_scaling_region')
+
+    def get_exec_args(self):
+        return []
+
+
+#@util.skip_unless_cpufreq()
```
cmcantalupo: Maybe get rid of #?

src/ScalingModelRegion.cpp
```diff
@@ -0,0 +1,158 @@
...
+            for (size_t it = 0; it != repeat; ++it) {
+                run_atom();
+            }
+            atom_time[trial_idx] = geopm_time_since(&time_0) / repeat;
+        }
+        std::sort(atom_time.begin(), atom_time.end());
+        double median_atom_time = atom_time[median_idx];
+        m_num_atom = big_o_in / median_atom_time;
+        m_num_atom = m_num_atom ? m_num_atom : 1;
+        m_norm = 1.0 / m_num_atom;
```
dianarg, line 140: where does this get used?

test_integration/test_scaling_region.py
```diff
@@ -0,0 +1,194 @@
...
+                            0xba9639ec,
+                            0x61d2be85,
+                            0x28eec3a2,
+                            0xd2b7c6a6,
+                            0x9b8bbb81,
+                            0x40cf3ce8,
+                            0x09f341cf,
+                            0xb1914011,
+                            0xf8ad3d36,
+                            0x07949283]
```
dianarg, line 125: is there some way to discover these from the report?

src/ReduceModelRegion.cpp
```diff
@@ -0,0 +1,77 @@
...
+        std::vector<double> recv_buffer(m_num_elem, 0.0);
```
dianarg: make members

test_integration/test_scaling_region.py
```diff
@@ -0,0 +1,194 @@
...
+
+    """
+    def write(self):
+        """No configuration files are required.
+
+        """
+        pass
+
+    def get_exec_path(self):
+        """Path to bencmark
```
dianarg: benchmark

src/ScalingModelRegion.cpp
dianarg: used by base class :(

dianarg (dismissed):
Ok as is

.gitignore
```diff
@@ -140,9 +140,11 @@
 /test_integration/*.trace*
 /test_integration/geopm_test_integration
 /test_integration/geopm_test_integration.pyc
+/test_integration/test_ee_scaling_spin_mix
 /test_integration/test_ee_stream_dgemm_mix
 /test_integration/test_omp_outer_loop
 /test_integration/test_profile_policy
+/test_integration/test_scaling_region
```
bakerbrandond, line 146: Missing 'ee' and 'mix'?

test_integration/geopm_test_launcher.py
```diff
@@ -283,12 +283,15 @@ def set_num_cpu(self):
             # Don't use hyper-threads.
             self._num_cpu = cpu_thread_core_socket[2] * cpu_thread_core_socket[3] - 1
 
-    def set_cpu_per_rank(self):
-        try:
-            rank_per_node = int(math.ceil(self._num_rank / self._num_node))
-            self._cpu_per_rank = int(self._num_cpu // rank_per_node)
-        except (AttributeError, TypeError):
-            pass
+    def set_cpu_per_rank(self, cpu_per_rank=None):
```
bakerbrandond, line 290: No longer need this changes, right?

test_integration/test_ee_scaling_spin_mix.cpp
```diff
@@ -0,0 +1,83 @@
...
+        geopm::Profile &prof = geopm::Profile::default_profile();
+        uint64_t network_spin_rid = prof->region("network_spin", GEOPM_REGION_HINT_NETWORK);
+        std::unique_ptr<geopm::ModelRegion> net_spin_model(geopm::ModelRegion::model_region("spin", net_spin_big_o * mix_idx, is_verbose));
+        for (int rep_idx = 0; rep_idx != repeat; ++rep_idx) {
+            scaling_model->run();
+            prof->enter(network_spin_rid);
+            net_spin_model->run();
+            prof->exit(network_spin_rid);
+        }
+        MPI_Barrier(MPI_COMM_WORLD);
```
bakerbrandond: When do we start using the new barrier model region?

test_integration/test_ee_scaling_spin_mix.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env python
```
bakerbrandond: This is just a copy of the steam dgemm mix py file... Needs to be updated with the asserts in my integration test

bakerbrandond (dismissed):

cmcantalupo (commented):
done

src/ReduceModelRegion.cpp
cmcantalupo: done

test_integration/test_scaling_region.py
cmcantalupo: done

.gitignore
cmcantalupo: they are in ABC order

test_integration/geopm_test_launcher.py
cmcantalupo: We don't need them, but they also don't hurt.  Seems liek a feature we may need later.

test_integration/test_ee_scaling_spin_mix.cpp
cmcantalupo: its kind of useless unless you are using geopmbench.  Maybe we should delete the BarrierModelRegion.  It is just damaging our test coverage.

test_integration/test_ee_scaling_spin_mix.py
cmcantalupo: Removed top commit from this PR

test_integration/test_scaling_region.py
cmcantalupo: Maybe we will remember later if it fails?

cmcantalupo:
I think this is now good to go.  Hopefully travis will finish soon.

test_integration/geopm_test_launcher.py
bgeltz: Some context for what was happening here would be useful.

test_integration/test_scaling_region.py
bgeltz: This should be removed if it's not necessary.

test_integration/test_scaling_region.py
bgeltz: Why is this array of hashes required?  This is gross.

bgeltz (changes requested):

test_integration/geopm_test_launcher.py
cmcantalupo: Would you prefer to delete this patch @bgeltz?  It is not needed for the feature, but without it you cannot control the threads per rank when using the test_launcher.

test_integration/test_scaling_region.py
dianarg: Can be uncommented; the fix for this decorator is in dev

src/ScalingModelRegion.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <cstdint>
+#include "ModelRegion.hpp"
+
+namespace geopm
+{
+    class ScalingModelRegion : public ModelRegion
+    {
+        public:
+            ScalingModelRegion(double big_o_in,
+                              int verbosity,
```
dianarg: ws

test_integration/geopm_test_launcher.py
bgeltz: I'm fine with it here, I just did not understand what it was fixing.

test_integration/test_scaling_region.py
```diff
@@ -0,0 +1,194 @@
...
+                time = report.get_field(region, 'runtime', 'sec')
+                freq_time.append((freq, time))
+            freq_time.sort()
+            time = list(zip(*freq_time)[1])
+            paired_time = zip(time[0:-1], time[1:])
+            for pt in paired_time:
+                self.assertGreater(pt[0], pt[1])
+
+if __name__ == '__main__':
+    try:
```
dianarg, line 189: remind me why we need this.  Doesn't setUpClass only get called once?

test_integration/test_scaling_region.py
cmcantalupo: We don't have a python interface for geopm_region_hash().  I could run with the monitor just to get a report that has the hashes in it, but that seems worse.  I just run the examples/geopm_hash on the names in the test application. This is test code.

test_integration/test_scaling_region.py
cmcantalupo: Okay

test_integration/test_scaling_region.py
cmcantalupo: The skip launch is to enable rerun of the tests without executing the benchmark.

bgeltz (approved):

PR #942, opened 2020-02-19
Revised outlier detection scripts.

alawibaba:
Total revamp of the outlier detection scripts.


bakerbrandond (commented):
Just took a first look at this and I will need to sit down with you to fully grok.  Who, other than you is familiar with this code?  Maybe we should have them involved in review as well.

Should this be moved to the admin tutorials dir?
Could we provide the app + run scripts needed to produce the traces and report files being analyzed?
Should that app + run script also reside in admin tutorials dir?

PR #952, opened 2020-02-21
Add hooks for TimedScalingModelRegion and implementation

cmcantalupo:
- Small fixup to ScalingModelRegion in one patch.
- Hooks added in second patch.  Commit message describes features.
- Implementation for TimedScalingModelRegion is just the constructor.
- Added an integration test.

### DIFF
--- a/src/SpinModelRegion.cpp
+++ b/src/SpinModelRegion.cpp
@@ -71,6 +71,11 @@ namespace geopm
         m_big_o = big_o_in;
     }
 
+    void SpinModelRegion::set_atom(std::function<void(void)> run_atom)
+    {
+        m_run_atom = run_atom;
+    }
+
     void SpinModelRegion::run(void)
     {
         if (m_big_o != 0.0) {
@@ -80,12 +85,14 @@ namespace geopm
             ModelRegion::region_enter();
             for (uint64_t i = 0 ; i < m_num_progress_updates; ++i) {
                 ModelRegion::loop_enter(i);
-
                 double timeout = 0.0;
                 struct geopm_time_s start = {{0,0}};
                 struct geopm_time_s curr = {{0,0}};
                 (void)geopm_time(&start);
                 while (timeout < m_delay) {
+                    if (m_run_atom) {
+                        m_run_atom();
+                    }
                     (void)geopm_time(&curr);
                     timeout = geopm_time_diff(&start, &curr);
                 }
--- a/src/SpinModelRegion.cpp
+++ b/src/SpinModelRegion.cpp
@@ -71,9 +71,9 @@ namespace geopm
         m_big_o = big_o_in;
     }
 
-    void SpinModelRegion::set_atom(std::function<void(void)> run_atom)
+    void SpinModelRegion::run_atom(void)
     {
-        m_run_atom = run_atom;
+
     }
 
     void SpinModelRegion::run(void)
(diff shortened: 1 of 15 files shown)


dannosliwcd (dismissed):
LGTM. Nice way to reuse a lot of model region code.

src/SpinModelRegion.cpp
```diff
@@ -71,6 +71,11 @@ namespace geopm
         m_big_o = big_o_in;
     }
 
+    void SpinModelRegion::set_atom(std::function<void(void)>run_atom)
```
bgeltz: std::function<void(void)> run_atom

There's a space missing between the std:function declaration and the name.

src/SpinModelRegion.hpp
```diff
@@ -49,8 +50,10 @@ namespace geopm
             virtual ~SpinModelRegion();
             void big_o(double big_o);
             void run(void);
+            void set_atom(std::function<void(void)>run_atom);
```
bgeltz: Missing space here too.

src/SpinModelRegion.cpp
```diff
@@ -80,12 +85,14 @@ namespace geopm
             ModelRegion::region_enter();
             for (uint64_t i = 0 ; i < m_num_progress_updates; ++i) {
                 ModelRegion::loop_enter(i);
-
                 double timeout = 0.0;
                 struct geopm_time_s start = {{0,0}};
                 struct geopm_time_s curr = {{0,0}};
                 (void)geopm_time(&start);
                 while (timeout < m_delay) {
+                    if (m_run_atom) {
```
bgeltz: ```m_run_atom``` is not initialized in the SpinModelRegion constructor.  I'm wondering if it would be more safe to explicitly set it to ```NULL``` there and then make this check if it is still ```NULL```.

bgeltz (changes requested):
Nits and a comment about ```m_run_atom```.

src/SpinModelRegion.cpp
cmcantalupo: It's not a function pointer, its a std::function.  It has bool method:

https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool

src/SpinModelRegion.cpp
cmcantalupo: kk

src/SpinModelRegion.cpp
bgeltz: Fancy.  I didn't realize it worked that way.

bgeltz (dismissed):

bakerbrandond (dismissed):
please change the name for factory to scaling_timed

src/TimedScalingModelRegion.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace geopm
+{
+    TimedScalingModelRegion::TimedScalingModelRegion(double big_o_in,
+                                                     int verbosity,
+                                                     bool do_imbalance,
+                                                     bool do_progress,
+                                                     bool do_unmarked)
+        : SpinModelRegion(big_o_in, verbosity, do_imbalance, do_progress, do_unmarked)
+        , m_scaling_model(std::make_shared<ScalingModelRegion>(1, 0, false, false, true))
+        , m_run_atom([this](){this->m_scaling_model->run_atom();})
```
bgeltz: After discussion, it would be great if this lambda wasn't necessary.  This can be accomplished by creating a run_atom in the base class...

bgeltz (changes requested):

src/TimedScalingModelRegion.cpp
cmcantalupo: Done

bgeltz (approved):

PR #953, opened 2020-02-21
Add composite signal support to PlatformIO.read_signal

### LINKED ISSUE #827: There is no way to perform multiple power measurements.
The singleton geopm::platform_io, once a push_signal is performed, never resets the flag m_is_active to false. A work around is to instantiate a non-public class:

const auto platform_io = std::make_unique<geopm::PlatformIOImp>();
const auto idx = platform_io->push_signal(signal_name, domain_type, domain_index);
[...]

This code is used in a daemon started by systemd, that reads on-demand, i.e. not in a cli tool.

bgeltz:
Fixes #827.

### DIFF
--- a/src/Agg.cpp
+++ b/src/Agg.cpp
@@ -36,6 +36,7 @@
 
 #include <algorithm>
 #include <numeric>
+#include <functional>
 
 #include "geopm_internal.h"
 #include "geopm_hash.h"
@@ -49,6 +50,15 @@ namespace geopm
         double result = NAN;
         if (operand.size()) {
             result = std::accumulate(operand.begin(), operand.end(), 0.0);
+        }
+        return result;
+    }
+
+    double Agg::difference(const std::vector<double> &operand)
+    {
+        double result = NAN;
+        if (operand.size()) {
+            result = std::accumulate(operand.begin() + 1, operand.end(), operand[0], std::minus<double>());
         }
         return result;
     }
--- a/src/PlatformIO.cpp
+++ b/src/PlatformIO.cpp
@@ -180,7 +180,7 @@ namespace geopm
                 is_found = true;
             }
             if (signal_name == "TEMPERATURE_PACKAGE") {
-                result = signal_domain_type("TEMPERATURE_PKG_UNDER");
+                result = signal_domain_type("TEMPERATURE_PACKAGE_UNDER");
                 is_found = true;
             }
         }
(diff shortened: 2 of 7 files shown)


src/Agg.cpp
```diff
@@ -53,6 +54,15 @@ namespace geopm
         return result;
     }
 
+    double Agg::difference(const std::vector<double> &operand)
```
dianarg, line 57: These Agg functions are intended to handle any size vector, but for the combined signal, the indices in the vector have special meaning.  For CombinedSignal, the injected function expects a vector of a specific size. So I don't think this function can replace that lambda, and it's not really general-purpose like the other functions in Agg.

src/PlatformIO.cpp
```diff
@@ -300,22 +301,12 @@ namespace geopm
...
                                      {max_idx, under_idx},
-                                     std::unique_ptr<CombinedSignal>(
-                                         new CombinedSignal(
-                                             [] (const std::vector<double> &val) -> double {
-#ifdef GEOPM_DEBUG
-                                                 if (val.size() != 2) {
-                                                     throw Exception("PlatformIOImp::push_signal_temperature(): expected two values to subtract for temperature.",
-                                                                     GEOPM_ERROR_LOGIC, __FILE__, __LINE__);
-                                                 }
-#endif
```
dianarg, line 316: See above; this check should be preserved.

src/Agg.cpp
dianarg: If you want to avoid repeating the stubtracting lambda, maybe you can add another CombinedSignal derived type (e.g. 'DifferenceCombinedSignal') and override the sample() method.  This would mean that the CombinedSignal base class with injected function is never used, which might be a good thing.

dianarg (changes requested):

src/Agg.cpp
bgeltz: I originally went down the path of creating a DifferenceCombined signal subclass, and didn't like the result.  It was a bit more code to do it that way.  I think there are other spots where we're using the CombinedSignal base class with the constructor default of Agg::sum and with the injected function (e.g. in push_combined_signal) so I'm not sure we can get away from calling CombinedSignal this way entirely.  Adding the Agg function resulted in more clean/tidy code in my opinion.

Agreed that for the combined signal of ```TEMPERATURE``` the ordering of the signals is important for this to work.  This new Agg::difference function will work OK for any size vector.  With the present example of ```TEMPERATURE``` it only makes sense with 2 elements, but I could see a future use case where we try to calculate left over power with something like ```TDP_POWER - PACKAGE_POWER - DRAM_POWER``` or similar.  I think there's value in adding this Agg function for those situations (though they may be rare).

I haven't come across any cases in my testing where the lambda fixed a problem with Agg::difference.

src/PlatformIO.cpp
bgeltz: I'm not sure why this check is necessary since we can see the size of the vector on line 307.

src/Agg.cpp
bgeltz: Let me try to redo this with the DifferenceCombinedSignal and we can compare.

bgeltz:
Abandoning this in favor of #955.

PR #962, opened 2020-03-02
Update EE agent to ignore regions with ignore hint

### LINKED ISSUE #828: EE Agent is not ignoring regions marked as GEOPM_REGION_HINT_IGNORE
These regions should be ignored and pinned to max freq as dictated by policy?

bakerbrandond:
- Fixes #828.

### DIFF
--- a/src/EnergyEfficientAgent.cpp
+++ b/src/EnergyEfficientAgent.cpp
@@ -194,6 +194,9 @@ namespace geopm
             if (GEOPM_REGION_HASH_UNMARKED == hash) {
                 m_target_freq[ctl_idx] = m_freq_governor->get_frequency_max();
             }
+            else if (GEOPM_REGION_HINT_IGNORE == hint) {
+                m_target_freq[ctl_idx] = m_freq_governor->get_frequency_max();
+            }
             else if (GEOPM_REGION_HINT_NETWORK == hint) {
                 m_target_freq[ctl_idx] = m_freq_governor->get_frequency_min();
             }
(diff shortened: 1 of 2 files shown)


src/EnergyEfficientAgent.cpp
```diff
@@ -228,7 +231,8 @@ namespace geopm
             if (m_last_region_info[ctl_idx].hash != current_region_info.hash ||
                 m_last_region_info[ctl_idx].count != current_region_info.count) {
                 if (current_region_info.hash != GEOPM_REGION_HASH_UNMARKED &&
-                    current_region_info.hint != GEOPM_REGION_HINT_NETWORK) {
+                    current_region_info.hint != GEOPM_REGION_HINT_NETWORK &&
```
cmcantalupo: might be nice to have a private method:

```
EnergyEfficientAgent::do_learning(uint64_t hash, uint64_t hint)
{
    return (hash != GEOPM_REGION_UNMARKED &&
            hint != GEOPM_REGION_HINT_NETWORK &&
            hint != GEOPM_REGION_HINT_IGNORE);
}
```
That way this logic doesn't get repeated twice.

bakerbrandond: done

src/EnergyEfficientAgent.hpp
```diff
@@ -83,6 +83,7 @@ namespace geopm
             static std::unique_ptr<Agent> make_plugin(void);
             static std::vector<std::string> policy_names(void);
             static std::vector<std::string> sample_names(void);
+            bool do_learning(uint64_t hash, uint64_t hint) const;
```
cmcantalupo: Private? I think the question is, can we get unit test coverage without making this public and adding unit tests for the interface directly?  If there is already good coverage of the logic, seems like it should be private.

cmcantalupo (approved):

PR #988, opened 2020-03-11
Refactoring patches for endpoint and policystore

dianarg:
A few changes to clean up code related to geopmcd implementation and testing.

- Infinite timeout for Endpoint agent attach wait
- Changes to PolicyStore API to make agent consistently the first parameter
- Rename test_static_policy to test_enforce_policy
- Add Endpoint::wait_for_agent_detach()


src/Endpoint.cpp
```diff
@@ -194,14 +194,26 @@ namespace geopm
...
             }
         }
     }
 
+    void EndpointImp::wait_for_agent_detach(double timeout)
+    {
+        std::string agent = get_agent();
+        double sample_age = 0;
+        std::vector<double> sample(m_num_sample);
+        while (m_continue_loop && agent != "" && sample_age < timeout) {
```
cmcantalupo: here timeout == 0 -> do not wait, in wait_for_agent_attach timeout == 0 -> wait forever.  This seems a little confusing.

test/EndpointTest.cpp
```diff
@@ -304,3 +304,27 @@ TEST_F(EndpointTest, wait_stops_when_agent_attaches)
     EXPECT_NE(result, std::future_status::timeout);
     mio.close();
 }
+
```
cmcantalupo, line 334: Does it make sense to test timeout==0 for the two wait methods?

src/Endpoint.cpp
dianarg: The intention was that timeout==0 means do not wait.  I'll add a test for that.

dianarg: So the condition needs to check the value of agent in wait_for_agent_attach

src/Endpoint.cpp
```diff
@@ -205,12 +205,16 @@ namespace geopm
...
-        double sample_age = 0;
-        std::vector<double> sample(m_num_sample);
-        while (m_continue_loop && agent != "" && sample_age < timeout) {
+        geopm_time_s start;
+        geopm_time(&start);
+        while (m_continue_loop && agent != "") {
             agent = get_agent();
-            sample_age = read_sample(sample);
             std::this_thread::sleep_for(std::chrono::milliseconds(100));
+            if (timeout >= 0 && geopm_time_since(&start) >= timeout) {
```
cmcantalupo, line 213: Not sure I understand what the check:

```timeout >= 0```

is doing.  Seems redundant.

cmcantalupo (commented):
Little nit, but I think this makes sense.

cmcantalupo (approved):

PR #989, opened 2020-03-11
Move Profile and OMPT classes into libgeopm

### LINKED ISSUE #980: Cryptic error messages with debug builds of geopmread
When running ```geopmread``` with a debug build, using the wrong domain or the wrong domain index for a given signal results in a very cryptic error message.  The following cases result in the same error:

1. Using an index that is out-of-bounds for a given domain results in the MPIComm error.
```geopmread POWER_DRAM board_memory 2```
2. Using a domain that is too granular for the given signal results in the MPIComm error.
```geopmread POWER_DRAM core 0```

```
  what():  <geopm> Invalid argument: PluginFactory::make_plugin(): name: "MPIComm" has not been previously registered: at geopm/src/PluginFactory.hpp:85
Aborted
```

With release builds, you see the proper error message:
```
$ geopmread POWER_DRAM core 0
Error: cannot read signal: <geopm> Invalid argument: PlatformIOImp::read_signal(): domain 2 is not valid for signal "ENERGY_DRAM": at geopm/src/PlatformIO.cpp:575

$ geopmread POWER_DRAM board_memory 2
Error: cannot read signal: <geopm> Invalid argument: PlatformIOImp::read_signal(): domain_idx is out of range: at geopm/src/PlatformIO.cpp:516
```

### PR DESCRIPTION
**Pull request by dianarg**
- Profile has a dependency on Comm, and the only Comm that can be
  constructed is MPIComm.
- Fixes #980

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -463,7 +463,6 @@ libgeopmpolicy_la_SOURCES = contrib/json11/json11.cpp \
                             src/CSV.hpp \
                             src/DebugIOGroup.cpp \
                             src/DebugIOGroup.hpp \
-                            src/DefaultProfile.cpp \
                             src/ELF.cpp \
                             src/ELF.hpp \
                             src/Endpoint.cpp \
@@ -514,8 +513,6 @@ libgeopmpolicy_la_SOURCES = contrib/json11/json11.cpp \
                             src/MSRSignalImp.hpp \
                             src/MonitorAgent.cpp \
                             src/MonitorAgent.hpp \
-                            src/OMPT.cpp \
-                            src/OMPT.hpp \
                             src/OptionParser.cpp \
                             src/OptionParser.hpp \
                             src/PlatformIO.cpp \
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by bgeltz — DISMISSED** [review]

File: Makefile.am [hunk 1]
```diff
@@ -630,15 +619,29 @@ if ENABLE_MPI
...
+                          src/Profile.hpp \
                           src/geopm_ctl.h \
                           src/geopm_mpi_comm_split.h \
                           src/geopm_pmpi.c \
                           src/geopm_pmpi_helper.cpp \
                           src/geopm_pmpi.h \
                           # end
+
+if ENABLE_OMPT
+    libgeopm_la_SOURCES += src/OMPT.cpp \
```
cmcantalupo: would be nice, in theory to have unit test for OMPT.cpp that didn't require openmp.  There is no explicit dependence on OpenMP in the OMPT class.  It is all in the ompt_callback.cpp.

Makefile.am
```diff
@@ -630,15 +619,29 @@ if ENABLE_MPI
...
+                           src/OMPT.hpp \
+                           src/ompt_callback.cpp \
+                           # end
+else
+    EXTRA_DIST += src/ompt_callback.cpp
```
cmcantalupo: need to put OMPT.[ch]pp into the EXTRA_DIST also

dianarg: What about the rest of the libgeopm sources?

cmcantalupo (commented):
LGTM

cmcantalupo (approved):

PR #995, opened 2020-03-17
Task 967

### LINKED ISSUE #967: Explore DVFS inefficiencies and impact on instruction per cycle rate
> estimate 5

The Energy Efficient agent tests (test_ee_short_region_slop) have shown some impact to IPC when DVFS is used.  

Done criterion: Write scripts for producing reference data for future performance tests that explore this effect and aid in understanding its root cause.
- Using two-region app from #947 and measure IPC in trace; produce plot(s) of the following metrics over time: IPC, requested frequency, achieved core frequency, uncore frequency, MSR::PERF_CTL:FREQ  
- Rerun existing script with ioctl writes and produce same set of plot from #944 for comparison.

Parent story #978

cmcantalupo:
- Fixes #967 from github issues
- Adds trace generation and plotting for test_ee_short_region_slop test

### DIFF
--- a/test_integration/test_ee_short_region_slop.cpp
+++ b/test_integration/test_ee_short_region_slop.cpp
@@ -50,7 +50,7 @@ int main(int argc, char **argv)
     bool is_verbose = false;
     geopm::Profile &prof = geopm::Profile::default_profile();
     // Run twelve trials with region duration ranging from 100 us - 400 ms
-    size_t num_duration = 12;
+    size_t num_duration = 7;
     double duration = 0.2048;
     double repeat = 200; // Each trial takes 41 seconds and the
                             // whole execution takes 16 minutes at
--- a/test_integration/test_ee_short_region_slop.cpp
+++ b/test_integration/test_ee_short_region_slop.cpp
@@ -65,16 +65,27 @@ int main(int argc, char **argv)
         // Rename model regions
         std::string scaling_name = "scaling_" + std::to_string(duration_idx);
         std::string timed_name = "timed_" + std::to_string(duration_idx);
+        std::string barrier_scaling_name = "barrier_scaling_" + std::to_string(duration_idx);
+        std::string barrier_timed_name = "barrier_timed_" + std::to_string(duration_idx);
+
         uint64_t scaling_rid = prof.region(scaling_name, GEOPM_REGION_HINT_UNKNOWN);
         uint64_t timed_rid = prof.region(timed_name, GEOPM_REGION_HINT_UNKNOWN);
+        uint64_t barrier_scaling_rid = prof.region(barrier_scaling_name, GEOPM_REGION_HINT_UNKNOWN);
+        uint64_t barrier_timed_rid = prof.region(barrier_timed_name, GEOPM_REGION_HINT_UNKNOWN);
         // Execute the regions back to back repeatedly
         for (int idx = 0; idx != repeat; ++idx) {
             prof.enter(scaling_rid);
             model_scaling->run();
             prof.exit(scaling_rid);
+            prof.enter(barrier_scaling_rid);
+            MPI_Barrier(MPI_COMM_WORLD);
+            prof.exit(barrier_scaling_rid);
             prof.enter(timed_rid);
             model_timed->run();
             prof.exit(timed_rid);
+            prof.enter(barrier_timed_rid);
+            MPI_Barrier(MPI_COMM_WORLD);
+            prof.exit(barrier_timed_rid);
         }
         repeat *= 2;
         duration /= 2.0;
--- a/test_integration/test_ee_short_region_slop.cpp
+++ b/test_integration/test_ee_short_region_slop.cpp
@@ -49,12 +49,9 @@ int main(int argc, char **argv)
     MPI_Init(&argc, &argv);
     bool is_verbose = false;
     geopm::Profile &prof = geopm::Profile::default_profile();
-    // Run twelve trials with region duration ranging from 100 us - 400 ms
     size_t num_duration = 7;
     double duration = 0.2048;
-    double repeat = 200; // Each trial takes 41 seconds and the
-                            // whole execution takes 16 minutes at
-                            // sticker frequency.
+    double repeat = 200; // Each trial takes 41 seconds at sticker frequency
     for (size_t duration_idx = 0; duration_idx != num_duration; ++duration_idx) {
         // Create scaling and scaling_timed model regions
         std::unique_ptr<geopm::ModelRegion> model_scaling(
--- a/test_integration/test_ee_short_region_slop.py
+++ b/test_integration/test_ee_short_region_slop.py
@@ -172,8 +172,8 @@ class TestIntegration_ee_short_region_slop(unittest.TestCase):
                                                       'frequency_map',
                                                       agent_conf_dynamic_dict)
 
-            report_signals = 'INSTRUCTIONS_RETIRED,CYCLES_REFERENCE,CYCLES_THREAD,MSR::UNCORE_PERF_STATUS:FREQ'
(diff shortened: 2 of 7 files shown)


test_integration/test_ee_short_region_slop.cpp
```diff
@@ -50,7 +50,7 @@ int main(int argc, char **argv)
     bool is_verbose = false;
     geopm::Profile &prof = geopm::Profile::default_profile();
     // Run twelve trials with region duration ranging from 100 us - 400 ms
```
dianarg: comment is no longer correct

test_integration/test_ee_short_region_slop.py
```diff
@@ -70,6 +72,13 @@
     from test_integration import util
     geopmpy.error.exc_clear()
 
+# Globals controlling uniform y axis limits in all plots
```
dianarg, line 82: Could be moved into test class?  Plotting functions also would need to move

test_integration/test_ee_short_region_slop.py
```diff
@@ -204,71 +241,89 @@ def tearDownClass(cls):
...
+                if len(in_path) == 0:
+                    raise RuntimeError('No trace files found')
+                in_path = in_path[0]
+                trace = read_trace(in_path)
+                # Get the first and last time that the regions of interest were executed
+                begin_time = trace['TIME'].loc[(trace['REGION_HASH'] == begin_region_hash) &
+                                               (trace['REGION_COUNT'] == region_count[0])].iloc[0]
+                end_time = trace['TIME'].loc[(trace['REGION_HASH'] == end_region_hash) &
+                                             (trace['REGION_COUNT'] == region_count[-1])].iloc[-1]
+                duration = end_time - begin_time
```
dianarg: duration is a confusing name
just used for width of plot

test_integration/test_ee_short_region_slop.py
```diff
@@ -325,11 +364,135 @@ def extract_data_region(report, cols, key):
...
+
+def select_time_range(trace, duration, begin_time=None):
+    if begin_time is None:
+        begin_time = trace['TIME'].iloc[0]
+    end_time = begin_time + duration
+    return trace.loc[(trace['TIME'] >= begin_time) &
+                     (trace['TIME'] < end_time)]
+
+
+def generate_trace_region_plot(trace, out_path, region_name, region_count):
```
dianarg: add "overlay" to name?

test_integration/test_scaling_region.py
```diff
@@ -93,14 +94,8 @@ def setUpClass(cls):
...
         # scaling_region_0, scaling_region_1, ... , scaling_region_30
-        cls._region_hash = [0x1f2b4ef8, 0xed40cdfb, 0xfe103e0f, 0x0c7bbd0c,
-                            0xd8b1d9e7, 0x2ada5ae4, 0x398aa910, 0xcbe12a13,
-                            0x95f21637, 0x67999534, 0xea0548e2, 0xa33935c5,
-                            0x787db2ac, 0x3141cf8b, 0xcb18ca8f, 0x8224b7a8,
-                            0x596030c1, 0x105c4de6, 0xa83e4c38, 0xe102311f,
-                            0xf3aa44cb, 0xba9639ec, 0x61d2be85, 0x28eec3a2,
-                            0xd2b7c6a6, 0x9b8bbb81, 0x40cf3ce8, 0x09f341cf,
-                            0xb1914011, 0xf8ad3d36, 0x07949283]
+        cls._region_hash = [geopmpy.hash.crc32_str('scaling_region_{}'.format(idx))
```
dianarg, line 97: yay

test_integration/test_ee_short_region_slop.py
```diff
@@ -325,11 +364,135 @@ def extract_data_region(report, cols, key):
...
+def read_trace(path):
+    converters={'REGION_HASH': lambda xx: int(xx, 16)}
+    return pandas.read_csv(path, sep='|', comment='#', converters=converters)
+
+
+def get_ipc(trace):
+    di = trace['INSTRUCTIONS_RETIRED'].diff()
+    dc = trace['CYCLES_THREAD'].diff()
+    return di / dc
+
```
bgeltz, line 432: need another \n

test_integration/test_ee_short_region_slop.py
```diff
@@ -325,11 +364,135 @@ def extract_data_region(report, cols, key):
...
+def get_freq(trace):
+    return trace['FREQUENCY']
+
```
bgeltz, line 436: need another \n

test_integration/test_ee_short_region_slop.py
```diff
@@ -325,11 +364,135 @@ def extract_data_region(report, cols, key):
...
+def select_region_count(trace, region_count):
+    return trace.loc[trace['REGION_COUNT'] == region_count]
+
+
+def select_region(trace, region_prefix, extension=''):
+    """e.g.:
+    >>> select_region(trace, 'scaling_2')
+    >>> select_region(trace, 'scaling_', range(12))
+    >>> select_region(trace, '', ['scaling_3', 'timed_3'])
```
bgeltz, line 458: Nice.

test_integration/test_ee_short_region_slop.py
```diff
@@ -325,11 +364,135 @@ def extract_data_region(report, cols, key):
...
+    if type(extension) is str:
+        extension = [extension]
+    region_hash = []
+    for ext in extension:
+        region_name = '{}{}'.format(region_prefix, ext)
+        region_hash.append(geopmpy.hash.crc32_str(region_name))
+    return trace.loc[trace['REGION_HASH'].isin(region_hash)]
+
+
+def select_time_range(trace, duration, begin_time=None):
```
bgeltz: I like these helper funcs.  I'm wondering how we should organize a #import with all of them.

test_integration/test_ee_short_region_slop.py
```diff
@@ -325,11 +364,135 @@ def extract_data_region(report, cols, key):
...
+        plt.xlabel('time since region start (sec)')
+        plt.ylabel('CPU freq (Hz)')
+        plt.ylim(g_plot_freq_lim)
+        freq_legend.append('count {}'.format(rc))
+    plt.subplot(2,1,1)
+    plt.legend(ipc_legend)
+    plt.subplot(2,1,2)
+    plt.legend(freq_legend)
+    plt.savefig(out_path)
+    plt.close()
```
bgeltz: What do you have against \n's ? 😆

bgeltz (changes requested):
This is pretty slick with all the plotting stuff.  Tiny whitespace changes needed, and could you add to the scaling model region patch commit msg some more about why you're doing that?

test_integration/test_ee_short_region_slop.py
cmcantalupo: I'm going to leave these separate for now with the idea that some other test may want to reuse the plotting routines.  We can make these bounds a default if that happens.

test_integration/test_ee_short_region_slop.py
cmcantalupo: Seems like a good idea for another PR.

test_integration/test_ee_short_region_slop.py
cmcantalupo: I have removed the patch that has the memory alignment change.  I tried to refactor the blob into a loop, but it wasn't worth the hassle.

dianarg (dismissed):
I don't see any reason not to merge this as is now.  In the future, we might want to convert the sbatch script to a more generic launch script similar to the tutorials.

bgeltz (dismissed):

dianarg (approved):

Issue #1009, opened 2020-03-25
Frequency map may choose wrong frequency based on hint


cmcantalupo:
The algorithm for the frequency map agent sample the region hint value the first time a region is encountered.  The hint may change throughout the region exection, but these changes would not effect the frequency selected by these agents which is fixed based on the first time sample.  If a compute bound region were to begin with a brief call to MPI, the frequency would be set to the minimum for the entire execution of the region.

bgeltz:
> If a compute bound region were to begin with a brief call to MPI, the frequency would be set to the minimum for the entire execution of the region.

I'm not sure how to resolve this bit here about the *minimum* being used for the whole region with this other blurb in #1051 about the *maximum*:
>Write an integration test that calls MPI within a user defined region that has a compute hint directly after region entry.
Inspect the report and show that the region is assigned the maximum frequency when running the frequency map agent

The FrequencyMapAgent sets compute regions to max frequency.  Is that what you meant @cmcantalupo ?

PR #1011, opened 2020-03-25
Do not set the frequency based on the hint in the frequency map agent.

cmcantalupo:
- Related to #1009.
- Better fix might be to change the frequency with the hint if the hash is not in the policy map but do not associate the frequency choice with the region.


cmcantalupo:
DO NOT MERGE

This solution needs work (for one thing it breaks the unit tests)

dianarg:
Superceded by #1089

PR #1013, opened 2020-03-25
Fix pthread launch

dianarg:
- Moves initialization of Profile (handshake with application) out of the constructor and calls it explicitly in geopm_pmpi_init
- Fix for test_profile_policy

### DIFF
--- a/src/Profile.cpp
+++ b/src/Profile.cpp
@@ -62,23 +62,31 @@
 
 namespace geopm
 {
-    ProfileImp::ProfileImp(const std::string &prof_name, const std::string &key_base,
-                           const std::string &report, double timeout, bool do_region_barrier,
-                           std::unique_ptr<Comm> comm, std::unique_ptr<ControlMessage> ctl_msg,
-                           const PlatformTopo &topo, std::unique_ptr<ProfileTable> table,
+    ProfileImp::ProfileImp(const std::string &prof_name,
+                           const std::string &key_base,
+                           const std::string &report,
+                           double timeout,
+                           bool do_region_barrier,
+                           std::unique_ptr<Comm> comm,
+                           std::unique_ptr<ControlMessage> ctl_msg,
+                           const PlatformTopo &topo,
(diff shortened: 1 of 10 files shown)


test_integration/geopm_test_integration.py
```diff
@@ -200,7 +200,7 @@ def test_report_and_trace_generation(self):
             trace = self._output.get_trace_data(node_name=nn)
             self.assertNotEqual(0, len(trace))
 
-    @unittest.skipUnless('mr-fusion' in socket.gethostname(), "This test only enabled on known working systems.")
+    @unittest.skipIf(socket.gethostname() in [], "This test is disabled enabled on some systems.")
```
bgeltz: ```in []``` seems wrong.

src/Profile.cpp
```diff
@@ -572,4 +617,9 @@ namespace geopm
     {
         return m_tprof_table;
     }
+
+    void ProfileImp::enable_pmpi(void)
+    {
+
+    }
```
bgeltz, line 624: Is this supposed to be empty?

src/Profile.cpp
```diff
@@ -110,21 +126,29 @@ namespace geopm
...
 #endif
-        std::string sample_key(key_base + "-sample");
-        std::string tprof_key(key_base + "-tprof");
+        std::string sample_key(m_key_base + "-sample");
+        std::string tprof_key(m_key_base + "-tprof");
         int shm_num_rank = 0;
 
-        init_prof_comm(std::move(comm), shm_num_rank);
+        init_prof_comm(std::move(m_comm), shm_num_rank);
+        std::string step;
```
bgeltz, line 134: Thanks for adding this.  It will help with future debug here.

bgeltz (changes requested):

src/Profile.cpp
dianarg: yes, one reason why I don't love this workaround.  Only DefaultProfile implements any behavior

bgeltz (approved):

Issue #999, opened 2020-03-18
As a GEOPM developer, I would like to educate users of system-D on how to building site-specific RPMs


sidjana:
> estimate 2

# Done Criteria:
(a) get account access to system-D
(b) get the spec files ready

PR #1019, opened 2020-03-30
Add spec files for TOSS

dianarg:
- Part of #999

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -330,6 +330,8 @@ EXTRA_DIST = $(TUTORIAL_DIST) \
              scripts/outlier/README.md \
              scripts/outlier/outlier_detection.py \
              scripts/outlier/theta_nodelist_broken.txt \
+             specs/geopm-gcc-mvapich2-toss.spec \
+             specs/geopm-intel-mvapich2-toss.spec \
              travis_ci.sh \
              test_hw/Makefile \
              test_hw/geopm_time.h \
--- a/specs/geopm-gcc-mvapich2-toss.spec
+++ b/specs/geopm-gcc-mvapich2-toss.spec
@@ -1,0 +1,148 @@
+#  Copyright (c) 2020, Intel Corporation
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#      * Redistributions of source code must retain the above copyright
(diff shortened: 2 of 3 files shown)


specs/geopm-gcc-mvapich2-quartz.spec
```diff
@@ -0,0 +1,148 @@
...
+resource manager to extend the hierarchy above the individual MPI
+application and enable the management of system power resources for
+multiple MPI jobs and multiple users by the system resource manager.
+
+%prep
+
+%setup -q -n %{pname}-%{version}
+
+%build
+module load %{compiler_name}/%{compiler_version}
```
cmcantalupo: maybe precede with a ```module purge```?

specs/geopm-intel-mvapich2-quartz.spec
```diff
@@ -0,0 +1,149 @@
...
+resource manager to extend the hierarchy above the individual MPI
+application and enable the management of system power resources for
+multiple MPI jobs and multiple users by the system resource manager.
+
+%prep
+
+%setup -q -n %{pname}-%{version}
+
+%build
+module load %{compiler_name}/%{compiler_version}
```
cmcantalupo: maybe precede with a ```module purge```?

slabasan (dismissed):
+1, thanks!

Makefile.am
```diff
@@ -330,6 +330,8 @@ EXTRA_DIST = $(TUTORIAL_DIST) \
              scripts/outlier/README.md \
              scripts/outlier/outlier_detection.py \
              scripts/outlier/theta_nodelist_broken.txt \
+             specs/geopm-gcc-mvapich2-quartz.spec \
```
amarathe84: Since the intended use of these spec files is to build GEOPM RPMs for TOSS (not Quartz specifically), you could change the spec file names from *-quartz to *-toss (or no suffix). I was successfully able to build RPMs on Catalyst using the same spec files. We assume that we should be able to build RPMs on other systems running TOSS since all of them have the same root directory structures.

amarathe84 (approved):
Thanks for putting this together! The changes look good. I tested the PR patch with local changes and the RPMs build correctly. I also tested the GEOPM package using small-scale tests.

cmcantalupo (approved):

PR #1034, opened 2020-04-09
Add temperature signals in MSRIOGroup

dianarg:
- Adds MSR::TEMPERATURE_CORE and MSR::TEMPERATURE_PACKAGE with logic similar to what is in PlatformIO.
- PlatformIO aliases are still present to allow comparison with the new signals.
- In the future, all signal types will be handled the same way without the need for specific name checks.

### DIFF
--- a/src/MSRIOGroup.cpp
+++ b/src/MSRIOGroup.cpp
@@ -39,6 +39,7 @@
 #include <utility>
 #include <iomanip>
 #include <iostream>
+#include <set>
 
 #include "contrib/json11/json11.hpp"
 
@@ -53,6 +54,7 @@
 #include "MSRIOImp.hpp"
 #include "PlatformTopo.hpp"
 #include "Helper.hpp"
+#include "CombinedSignal.hpp"
 #include "config.h"
 
 using json11::Json;
@@ -67,6 +69,12 @@ namespace geopm
     const std::string snb_msr_json(void);
     const std::string skx_msr_json(void);
     static std::vector<std::unique_ptr<MSR> > init_msr_arr(int cpu_id);
+
+    // @todo: temporary workaround for combined signals
+    const std::set<std::string> MSRIOGroup::m_combined_signal_names = {
+        "MSR::TEMPERATURE_CORE",
+        "MSR::TEMPERATURE_PACKAGE"
+    };
 
     MSRIOGroup::MSRIOGroup()
         : MSRIOGroup(platform_topo(), std::unique_ptr<MSRIO>(new MSRIOImp), cpuid(), geopm_sched_num_cpu())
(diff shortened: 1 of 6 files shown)


src/MSRIOGroup.cpp
```diff
@@ -261,39 +291,63 @@ namespace geopm
         }
         std::set<int> cpu_idx = m_platform_topo.domain_nested(GEOPM_DOMAIN_CPU,
                                                               domain_type, domain_idx);
-
+        int rep_cpu = *(cpu_idx.begin());
```
bgeltz, line 294: What all is going on here with the rep_cpu usage?  Could you explain a bit what the intended functionality is here?

I think rep_cpu for combined signals here means that only 1 of the cores will have it's MSRs read for determining the value that's returned.  Is that right?  I'm curious if that means that if you request temperature at the package level, that it only reads one core?

bgeltz (approved):
LGTM so far.
I see ```/// @todo remove with combined signal refactoring``` in a few places, so I'm looking forward to the refactor.  The tests, and the current code keeping the functionality we have working looks good though.

One question about the new nested domain handling.

src/MSRIOGroup.cpp
```diff
@@ -262,39 +291,63 @@ namespace geopm
...
+                int rep_core = *(cores.begin());
+                int max_idx = push_signal("MSR::TEMPERATURE_TARGET:PROCHOT_MIN", GEOPM_DOMAIN_CORE, rep_core);
+                int under_idx = -1;
+                if (signal_name == "MSR::TEMPERATURE_CORE") {
+                    under_idx = push_signal("MSR::THERM_STATUS:DIGITAL_READOUT", GEOPM_DOMAIN_CORE, domain_idx);
+                }
+                else if (signal_name =="MSR::TEMPERATURE_PACKAGE") {
+                    under_idx = push_signal("MSR::PACKAGE_THERM_STATUS:DIGITAL_READOUT", GEOPM_DOMAIN_PACKAGE, domain_idx);
+                }
+                std::unique_ptr<CombinedSignal> signal(new DifferenceCombinedSignal);
```
cmcantalupo, line 329: Do we use make_unique() for creating objects with the default constructor?

src/MSRIOGroup.cpp
```diff
@@ -68,6 +70,12 @@ namespace geopm
...
     static std::vector<std::unique_ptr<MSR> > init_msr_arr(int cpu_id);
 
+    // @todo: temporary workaround for combined signals
+    const std::set<std::string> MSRIOGroup::m_combined_signal_names = {
+        "MSR::TEMPERATURE_CORE",
+        "MSR::TEMPERATURE_PACKAGE"
+    };
+
     MSRIOGroup::MSRIOGroup()
         : MSRIOGroup(platform_topo(), std::unique_ptr<MSRIO>(new MSRIOImp), cpuid(), geopm_sched_num_cpu())
```
cmcantalupo, line 80: This is the same pattern as below without the make_unique() call.

src/MSRIOGroup.hpp
```diff
@@ -171,6 +172,11 @@ namespace geopm
             std::map<std::string, std::string> m_signal_desc_map;
             std::map<std::string, std::string> m_control_desc_map;
             std::map<std::string, int> m_signal_units_map;
+
+            /// @todo remove with combined signal refactoring
```
cmcantalupo, line 176: Looking forward to seeing the combined signal refactoring.  The type for m_combined_signals is quite a container.

cmcantalupo (commented):
Looks like a good start to me.  Looking forward to the refactoring that will clean up all of the todos.  I like the unit tests first.

cmcantalupo (approved):

src/MSRIOGroup.cpp
dianarg: Yes that's my understanding of how this is supposed to work.

PR #1039, opened 2020-04-13
Signal interface and derived classes

### LINKED ISSUE #1038: Add Signal classes to encapsulate MSR to double conversion logic
> estimate 5

Add a Signal interface that can be used for raw MSRs, MSR fields, and signals that are a combination of other signals.

Parent story #435

dianarg:
Adds classes that will be used to support MSRIOGroup refactoring.  These classes will replace MSRSignal, CombinedSignal & friends, and parts of the MSR + MSREncode usage in MSRIOGroup.

Resolves #1038

### DIFF
--- a/src/DerivativeSignal.cpp
+++ b/src/DerivativeSignal.cpp
@@ -53,24 +53,7 @@ namespace geopm
         , m_sleep_time(sleep_time)
     {
         GEOPM_DEBUG_ASSERT(m_time_sig && m_y_sig,
-                           "underlying Signals cannot be null.");
-    }
-
-    DerivativeSignal::DerivativeSignal(const DerivativeSignal &other)
-        : m_time_sig(other.m_time_sig->clone())
-        , m_y_sig(other.m_y_sig->clone())
-        , M_NUM_SAMPLE_HISTORY(other.M_NUM_SAMPLE_HISTORY)
-        , m_history(M_NUM_SAMPLE_HISTORY)
-        , m_derivative_num_fit(0)
-        , m_is_batch_ready(false)
-        , m_sleep_time(other.m_sleep_time)
-    {
-
-    }
-
-    std::unique_ptr<Signal> DerivativeSignal::clone(void) const
-    {
-        return geopm::make_unique<DerivativeSignal>(*this);
+                           "Signal pointers for time_sig and y_sig cannot be null.");
     }
 
     void DerivativeSignal::setup_batch(void)
@@ -129,7 +112,7 @@ namespace geopm
         return compute_next(m_history, m_derivative_num_fit, time, signal);
     }
 
-    double DerivativeSignal::read(void)
+    double DerivativeSignal::read(void) const
     {
         double result = NAN;
         CircularBuffer<m_sample_s> temp_history(M_NUM_SAMPLE_HISTORY);
@@ -139,8 +122,8 @@ namespace geopm
             double signal = m_y_sig->read();
             result = compute_next(temp_history, num_fit, time, signal);
             if (ii < M_NUM_SAMPLE_HISTORY) {
-                usleep(m_sleep_time);            }
-
+                usleep(m_sleep_time);
+            }
         }
         return result;
     }
--- a/src/DerivativeSignal.cpp
+++ b/src/DerivativeSignal.cpp
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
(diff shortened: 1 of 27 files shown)


src/RawMSRSignal.cpp
```diff
@@ -0,0 +1,102 @@
...
+    {
+        return geopm::make_unique<RawMSRSignal>(*this);
+    }
+
+    void RawMSRSignal::setup_batch(void)
+    {
+        GEOPM_DEBUG_ASSERT(m_msrio != nullptr, "no valid MSRIO object.");
+
+        if (!m_is_batch_ready) {
+            m_data = m_msrio->add_read(m_cpu, m_offset);
```
dianarg, line 61: One change we discussed was to have this data pointer not be exposed here and add a sample() method to MSRIO.

src/DerivativeSignal.cpp
```diff
@@ -0,0 +1,148 @@
...
+        : m_time_sig(time_sig)
+        , m_y_sig(y_sig)
+        , M_NUM_SAMPLE_HISTORY(num_sample_history)
+        , m_history(M_NUM_SAMPLE_HISTORY)
+        , m_derivative_num_fit(0)
+        , m_is_batch_ready(false)
+        , m_sleep_time(sleep_time)
+    {
+        GEOPM_DEBUG_ASSERT(m_time_sig && m_y_sig,
+                           "underlying Signals cannot be null.");
```
bgeltz: Underlying signals...

src/DerivativeSignal.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace geopm
+{
+    class DerivativeSignal : public Signal
+    {
+        public:
+            DerivativeSignal(std::shared_ptr<Signal> time_sig,
+                             std::shared_ptr<Signal> y_sig,
+                             int read_loops, double sleep_time);
+            DerivativeSignal(const DerivativeSignal &other);
```
bgeltz: Is the copy constructor used anywhere?

src/DerivativeSignal.cpp
```diff
@@ -0,0 +1,148 @@
...
+    {
+        double result = NAN;
+        CircularBuffer<m_sample_s> temp_history(M_NUM_SAMPLE_HISTORY);
+        int num_fit = 0;
+        for (int ii = 0; ii < M_NUM_SAMPLE_HISTORY; ++ii) {
+            double time = m_time_sig->read();
+            double signal = m_y_sig->read();
+            result = compute_next(temp_history, num_fit, time, signal);
+            if (ii < M_NUM_SAMPLE_HISTORY) {
+                usleep(m_sleep_time);            }
```
bgeltz: Closing curly is in the wrong spot.

src/DifferenceSignal.cpp
```diff
@@ -0,0 +1,87 @@
...
+namespace geopm
+{
+    DifferenceSignal::DifferenceSignal(std::shared_ptr<Signal> minuend,
+                                       std::shared_ptr<Signal> subtrahend)
+        : m_minuend(minuend)
+        , m_subtrahend(subtrahend)
+        , m_is_batch_ready(false)
+    {
+        GEOPM_DEBUG_ASSERT(m_minuend && m_subtrahend,
+                           "underlying Signals cannot be null.");
```
bgeltz: Same capitalization issues as before

src/DifferenceSignal.hpp
```diff
@@ -0,0 +1,62 @@
...
+namespace geopm
+{
+    /// A composite signal used by an IOGroup to produce a signal as
+    /// the difference of two signals.
+    class DifferenceSignal : public Signal
+    {
+        public:
+            DifferenceSignal(std::shared_ptr<Signal> minuend,
+                             std::shared_ptr<Signal> subtrahend);
+            DifferenceSignal(const DifferenceSignal &other);
```
bgeltz: Is the copy constructor used?

src/MSRFieldSignal.hpp
```diff
@@ -0,0 +1,91 @@
...
+    /// MSREncode.  The enum for the function comes from the MSR class.
+    class MSRFieldSignal : public Signal
+    {
+        public:
+            MSRFieldSignal(std::shared_ptr<Signal> raw_msr,
+                           int begin_bit,
+                           int end_bit,
+                           int function,
+                           double scalar);
+            MSRFieldSignal(const MSRFieldSignal &other);
```
bgeltz: ?

bgeltz (changes requested):
Looks mostly good to me.

A few nits on capitalization and curly brace placement.

Do we need all those copy constructors for the various signal types?  I didn't see where they were used.

src/DerivativeSignal.cpp
dianarg: Maybe something like "Signal pointers for time_sig and y_sig cannot be null" ?

src/DerivativeSignal.hpp
dianarg: It's used in clone()

dianarg (commented):
The copy constructors are needed because when the MSRIOGroup implements read_signal(), it will need to copy a shared_ptr<Signal> (case class pointer) but we need to know which copy constructor to call.  Also, some of the Signal types have state so the copy constructors are non-trivial.

bgeltz (approved):

PR #1066, opened 2020-04-23
Fix for EE Short Slop test low IPC and calibration issues

lhlawson:
Changes to the EE short slop test and ScalingModelRegion code to address Low IPC throughout the EE short slop test and calibration issues that lead to short scaling model region

- Fix for Low IPC throughout test: Addition of memory alignment to ScalingModelRegion constructor for arrays operated on in run_atom().  
- Fix for Low IPC during instantiation leading to short ScalingModelRegion: Addition of 2000 run_atom calls pre-calibration in the big_o code to guarantee calibration occurs after low IPC region.

### DIFF
--- a/src/ScalingModelRegion.cpp
+++ b/src/ScalingModelRegion.cpp
@@ -38,6 +38,7 @@
 #include <fstream>
 #include <sstream>
 #include <algorithm>
+#include <stdlib.h>
 
 #include "geopm.h"
 #include "geopm_time.h"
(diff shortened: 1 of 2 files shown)


test_integration/test_ee_short_region_slop.cpp
```diff
@@ -53,11 +53,11 @@ int main(int argc, char **argv)
     double duration = 0.2048;
     double repeat = 200; // Each trial takes 41 seconds at sticker frequency
     for (size_t duration_idx = 0; duration_idx != num_duration; ++duration_idx) {
-        // Create scaling and scaling_timed model regions
```
cmcantalupo: Is this reversal required for performance?  If so I'd like to understand why, and if not, we should revert this part of the change.

cmcantalupo (commented):
I rewrote your history a bit and force pushed to your fork.

cmcantalupo (approved):

PR #1086, opened 2020-05-01
Fix parsing issue in geopmpy.io when raw MSR registers are added to the trace

### LINKED ISSUE #1028: If raw MSR signals are added to the trace file then the geopmpy.io.trace() object fails to parse
Raw MSR signals have a '#' character in their names.  The trace parsing code uses a pandas option to treat '#' characters like comments, so the trace header is not properly parsed.  This leads to a confusing error message about applying the format function.  This issue is fixed in a similar bit of code here:

https://github.com/geopm/geopm/commit/afd2106ba933f42f192d7e56c4dcc0d5959d0d70

cmcantalupo:
- Raw MSR names contain the '#' character, but this was being treated
  like a comment character by the parser to skip the header.
- Fixes #1028 from github issues.

### DIFF
--- a/scripts/geopmpy/io.py
+++ b/scripts/geopmpy/io.py
@@ -948,6 +948,9 @@ class Trace(object):
         old_governor_headers = {'power_budget': 'POWER_BUDGET'}
         old_headers.update(old_governor_headers)
 
+        # Need to determine how many lines are in the header
+        # explicitly.  We cannot use '#' as a comment character since
+        # it occurs in raw MSR signal names.
         skiprows = 0
         with open(trace_path) as fid:
             for ll in fid:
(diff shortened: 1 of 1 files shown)


scripts/geopmpy/io.py
```diff
@@ -948,7 +948,14 @@ def __init__(self, trace_path, use_agent=True):
         old_governor_headers = {'power_budget': 'POWER_BUDGET'}
         old_headers.update(old_governor_headers)
 
-        column_headers = pandas.read_csv(trace_path, sep='|', comment='#', nrows=0, encoding='utf-8').columns.tolist()
+        skiprows = 0
```
bgeltz, line 954: Please put a comment in here explaining why we have to do this dance on skipping the rows that start with '#' instead of using the pandas ```comment=``` equals option.  My fear is that we'll come back in here some day, look at this new block and think it's not necessary and reintroduce the bug you're fixing.

bgeltz (changes requested):
LGTM save for missing a comment to explain why.

bgeltz (approved):

Issue #795, opened 2019-11-05
Signal alias for value of frequency control setting


dianarg:
Reading "MSR::PERF_CTL:FREQ" to get the actual controlled frequency setting is a common enough use case that it makes sense to have an alias.

PR #1089, opened 2020-05-01
Update FrequencyMapAgent and tests to match documentation

### LINKED ISSUE #1009: Frequency map may choose wrong frequency based on hint

### LINKED ISSUE #1052: Update Frequency Map agent to follow new man page requirements
> estimate 5

Parent story #1046

Done criteria:
- Implement features described in #1049 
- Make sure that test #1051 passes for the FrequencyMapAgent
- Add unit tests

dianarg:
- Uncore can not be set to fixed frequency through the policy.
- NAN is not allowed for mapped regions.
- Removed consideration of hints.
- Fixed up integration test.
- Resolves #1052.
- Fixes #1009.

Change-Id: Ia09fe225d939cce5c3ae6c9ccf42d6e2b96dac2c
Signed-off-by: Diana Guttman <diana.r.guttman@intel.com>

### DIFF
--- a/ronn/geopm_agent_frequency_map.7.ronn
+++ b/ronn/geopm_agent_frequency_map.7.ronn
@@ -73,7 +73,9 @@ name (see **geopm(7)**).  This name can also be passed to the
                   specified with a CPU frequency that is allowed by
                   the system.  Failure to provide this parameter,
                   setting it to an out of range value or specifying
-                  NAN will result in a runtime error.
+                  NAN will result in a runtime error, unless the policy
+                  is all NAN, in which case the agent will not exert
+                  any control.
 
   `FREQ_UNCORE`: The operating frequency for the uncore clock in units
                  of Hz.  If specified the uncore clock will operate
@@ -124,10 +126,8 @@ raised.
 
 
 ## REPORT EXTENSIONS
-The assigned frequency for each region is added to each region
-including those that were not in the policy map (which will show the
-policy default CPU frequency).
-
+The assigned frequency is added to each region's section of the report
+if that region was explicitly assigned a frequency in the policy map.
 
 ## CONTROL LOOP RATE
 The agent gates the control loop to sample the region hash and control
--- a/ronn/geopm_agent_frequency_map.7.ronn
+++ b/ronn/geopm_agent_frequency_map.7.ronn
@@ -73,9 +73,7 @@ name (see **geopm(7)**).  This name can also be passed to the
                   specified with a CPU frequency that is allowed by
                   the system.  Failure to provide this parameter,
                   setting it to an out of range value or specifying
-                  NAN will result in a runtime error, unless the policy
-                  is all NAN, in which case the agent will not exert
-                  any control.
+                  NAN will result in a runtime error.
 
   `FREQ_UNCORE`: The operating frequency for the uncore clock in units
                  of Hz.  If specified the uncore clock will operate
--- a/src/FrequencyMapAgent.cpp
+++ b/src/FrequencyMapAgent.cpp
@@ -37,6 +37,7 @@
 #include <iomanip>
 #include <iostream>
 #include <utility>
+#include <algorithm>
 
 #include "contrib/json11/json11.hpp"
 
@@ -49,60 +50,40 @@
 #include "FrequencyGovernor.hpp"
 #include "Helper.hpp"
 #include "Exception.hpp"
+#include "geopm_debug.hpp"
 #include "config.h"
 
 using json11::Json;
 
 namespace geopm
 {
-    static std::map<uint64_t, double> parse_env_map(void)
-    {
-        std::map<uint64_t, double> frequency_map;
-        std::string env_map_str = environment().frequency_map();
-        if (env_map_str.length()) {
-            std::cerr << "Warning: <geopm> Use of the GEOPM_FREQUENCY_MAP "
-                         "environment variable is deprecated. Frequency maps "
-                         "will only be set via "
-                      << FrequencyMapAgent::plugin_name()
-                      << " agent policies in the future.\n";
-            std::string err;
-            Json root = Json::parse(env_map_str, err);
-            if (!err.empty() || !root.is_object()) {
-                throw Exception("FrequencyMapAgent::" + std::string(__func__) + "(): detected a malformed json config file: " + err,
-                                GEOPM_ERROR_FILE_PARSE, __FILE__, __LINE__);
-            }
-            for (const auto &obj : root.object_items()) {
-                if (obj.second.type() != Json::NUMBER) {
-                    throw Exception("FrequencyMapAgent::" + std::string(__func__) +
-                                    ": Region best-fit frequency must be a number",
-                                    GEOPM_ERROR_FILE_PARSE, __FILE__, __LINE__);
-                }
-                uint64_t hash = geopm_crc32_str(obj.first.c_str());
-                frequency_map[hash] = obj.second.number_value();
-            }
-        }
-        return frequency_map;
-    }
-
     FrequencyMapAgent::FrequencyMapAgent()
-        : FrequencyMapAgent(platform_io(), platform_topo(), FrequencyGovernor::make_shared(), parse_env_map())
+        : FrequencyMapAgent(platform_io(), platform_topo())
     {
 
     }
 
-    FrequencyMapAgent::FrequencyMapAgent(PlatformIO &plat_io, const PlatformTopo &topo,
-                                         std::shared_ptr<FrequencyGovernor> gov, std::map<uint64_t, double> frequency_map)
+    FrequencyMapAgent::FrequencyMapAgent(PlatformIO &plat_io, const PlatformTopo &topo)
         : M_PRECISION(16)
         , M_WAIT_SEC(0.002)
         , m_platform_io(plat_io)
         , m_platform_topo(topo)
-        , m_freq_governor(gov)
-        , m_hash_freq_map(frequency_map)
-        , m_last_wait(GEOPM_TIME_REF)
-        , m_level(-1)
+        , m_wait_time(GEOPM_TIME_REF)
+        , m_uncore_min_ctl_idx(-1)
+        , m_uncore_max_ctl_idx(-1)
+        , m_last_uncore_freq(NAN)
         , m_num_children(0)
         , m_is_policy_updated(false)
-        , m_is_initialized_with_map(!frequency_map.empty())
+        , m_do_write_batch(false)
+        , m_is_adjust_initialized(false)
+        , m_freq_ctl_domain_type(GEOPM_DOMAIN_INVALID)
+        , m_num_freq_ctl_domain(0)
+        , m_core_freq_min(NAN)
+        , m_core_freq_max(NAN)
+        , m_uncore_init_min(NAN)
+        , m_uncore_init_max(NAN)
+        , m_default_freq(NAN)
+        , m_uncore_freq(NAN)
     {
 
     }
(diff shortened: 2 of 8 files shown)


ronn/geopm_agent_frequency_map.7.ronn
```diff
@@ -73,7 +73,9 @@ name (see **geopm(7)**).  This name can also be passed to the
                   specified with a CPU frequency that is allowed by
                   the system.  Failure to provide this parameter,
                   setting it to an out of range value or specifying
-                  NAN will result in a runtime error.
+                  NAN will result in a runtime error, unless the policy
+                  is all NAN, in which case the agent will not exert
+                  any control.
```
bgeltz: Why allow an all NAN policy?  Why have this agent not exert control?  We have the monitor for this use case.

ronn/geopm_agent_frequency_map.7.ronn
```diff
@@ -124,10 +126,8 @@ raised.
 
 
 ## REPORT EXTENSIONS
-The assigned frequency for each region is added to each region
-including those that were not in the policy map (which will show the
-policy default CPU frequency).
-
+The assigned frequency is added to each region's section of the report
+if that region was explicitly assigned a frequency in the policy map.
```
bgeltz, line 128: Why remove the feature of having frequencies for all regions show up in the report, including those not in the policy map?

src/FrequencyMapAgent.cpp
```diff
@@ -169,36 +167,52 @@ namespace geopm
...
+            m_is_policy_updated = false;
+            return;
+        }
 
+        std::map<uint64_t, double> old_freq_map = m_hash_freq_map;
+        m_hash_freq_map.clear();
         for (auto it = policy.begin() + M_POLICY_FIRST_HASH;
-             it != policy.end() && std::next(it) != policy.end(); std::advance(it, 2)) {
+             it != policy.end() && std::next(it) != policy.end();
+             std::advance(it, 2)) {
```
bgeltz, line 195: Why?

Also, it looks like there's a dangling curly brace at EOL.

src/FrequencyMapAgent.cpp
```diff
@@ -169,36 +167,52 @@ namespace geopm
...
-                        }
-                    }
-                }
+
+                // Not valid to have NAN freq for hash.
+                // This is a logic error because it is checked by
+                // validate policy, which the controller should
+                // call before this function.
+                GEOPM_DEBUG_ASSERT(!std::isnan(freq),
+                                   "mapped region with no frequency assigned.");
```
bgeltz, line 205: Shouldn't it be impossible to hit this assert?

src/FrequencyMapAgent.cpp
```diff
@@ -169,36 +167,52 @@ namespace geopm
...
+                m_hash_freq_map[hash] = freq;
             }
         }
+        m_is_policy_updated = false;
+        // check if policy changed
+        if (m_default_freq != policy[M_POLICY_FREQ_DEFAULT]) {
+            m_is_policy_updated = true;
+            m_default_freq = policy[M_POLICY_FREQ_DEFAULT];
+        }
+        if (m_hash_freq_map != old_freq_map) {
```
bgeltz, line 215: I'm surprised this works.

src/FrequencyMapAgent.cpp
```diff
@@ -242,71 +256,83 @@ namespace geopm
     void FrequencyMapAgent::adjust_platform(const std::vector<double> &in_policy)
     {
         update_policy(in_policy);
+
+        m_do_write_batch = false;
+
+        if (!m_is_adjust_initialized) {
+            // adjust all controls once in case not applied by policy
+            for (size_t ctl_idx = 0; ctl_idx < (size_t) m_num_freq_ctl_domain; ++ctl_idx) {
+                // @todo: this is bad; agent should be able to use aliases
```
bgeltz, line 272: Is there a bug open?

test_integration/test_frequency_hint_usage.py
```diff
@@ -50,10 +50,11 @@
 import geopmpy.error
 
 from test_integration import util
+from test_integration import geopm_test_launcher
```
bgeltz, line 53: Why?

If this is wrong, it's wrong in the template too.

test_integration/test_frequency_hint_usage.py
```diff
@@ -98,13 +99,13 @@ def setUpClass(cls):
 
         cls._fmap_report_path = 'test_{}_fmap.report'.format(test_name)
         cls._fmap_trace_path = 'test_{}_fmap.trace'.format(test_name)
-        cls._fmap_agent_conf_path = test_name + '-fmap-agent-config.json'
+        cls._fmap_agent_conf_path = 'test_' + test_name + '-fmap-agent-config.json'
```
bgeltz, line 102: Wrong in the template too.

test_integration/test_frequency_hint_usage.py
```diff
@@ -98,13 +99,13 @@ def setUpClass(cls):
...
 
         cls._ee_report_path = 'test_{}_ee.report'.format(test_name)
         cls._ee_trace_path = 'test_{}_ee.trace'.format(test_name)
-        cls._ee_agent_conf_path = test_name + '-ee-agent-config.json'
+        cls._ee_agent_conf_path = 'test_' + test_name + '-ee-agent-config.json'
 
-        geopmpy.error.exc_clear() # Clear out exception record for python 2 support
+        geopmpy.error.exc_clear()  # Clear out exception record for python 2 support
```
bgeltz, line 108: Why the space?

test_integration/test_frequency_hint_usage.py
```diff
@@ -138,7 +139,7 @@ def setUpClass(cls):
                                                         time_limit=time_limit)
             launcher.set_num_node(num_node)
             launcher.set_num_rank(num_rank)
-            launcher.run(test_name)
+            launcher.run('test_' + test_name)
```
bgeltz, line 142: Wrong in the template too.

test_integration/test_frequency_hint_usage.py
```diff
@@ -207,12 +206,13 @@ def test_frequency_sane_ee(self):
 
         # EE agent should assign not the minimum frequency for regions
         # with the COMPUTE hint.
-        self.assertNotEqual(self._freq_min, assigned_freq, msg='Expected assigned frequency to not be the min from policy since this region has the COMPUTE hint. ({} == {}).'.format(self._freq_min, assigned_freq))
-        util.assertNotNear(self, self._freq_min, achieved_freq, msg='Expected achieved frequency to NOT be near min from policy. ({} ~= {})'.format(self._freq_min, achieved_freq))
+        self.assertNotEqual(self._freq_min, assigned_freq,
+                            msg='Expected assigned frequency to not be the min from policy since this region has the COMPUTE hint. ({} == {}).'.format(self._freq_min, assigned_freq))
+        util.assertNotNear(self, self._freq_min, achieved_freq,
+                           msg='Expected achieved frequency to NOT be near min from policy. ({} ~= {})'.format(self._freq_min, achieved_freq))
```
bgeltz, line 215: I don't see anything changed here besides putting msg on it's own line.

test_integration/test_frequency_hint_usage.py
```diff
@@ -207,12 +206,13 @@ def test_frequency_sane_ee(self):
...
 
 
 if __name__ == '__main__':
     # Call do_launch to clear non-pyunit command line option
     util.do_launch()
     unittest.main()
```
bgeltz, line 221: This newline is vital. :P

I'm noticing a trend where patches I write introduce stuff like this, and patches that Chris does remove them...

bgeltz (changes requested):

ronn/geopm_agent_frequency_map.7.ronn
cmcantalupo: This issue is around the case where a non-root controller starts up and has not yet received a policy from it's parent.  In this case, we don't want the agent to do anything until a policy with values arrives.

src/FrequencyMapAgent.cpp
dianarg: This is just an extra newline.  The advance by two is because this part of the policy has pairs of hash + frequency.

src/FrequencyMapAgent.cpp
dianarg: Yes: #795

src/FrequencyMapAgent.cpp
dianarg: Yes if you are using the controller; since this is a public interface, you can still hit it through unit testing.

src/FrequencyMapAgent.cpp
dianarg: Me too!

ronn/geopm_agent_frequency_map.7.ronn
cmcantalupo: The problem is that the Agent interface is kind of not awesome.  The agent is not given a list of all regions to report on, it has to observe them in a sample to know about them.  For this reason, we will not be able to provide the keys for the map returned by the Agent::report_region() api:  https://github.com/geopm/geopm/blob/dev/src/Agent.hpp#L121 for short running regions. In at least the case of a short running region not in the policy map, the agent is unable to meet the requirement.  So if it isn't reliably there, I said why put there for the unmapped regions.

test_integration/test_frequency_hint_usage.py
dianarg: pretty sure this is a pep8 style guideline; my linter was complaining.

src/FrequencyMapAgent.cpp
cmcantalupo: Why advance two elements?  One is the hash one is the frequency.  Not sure about the dangling curly, it was there before the change.

src/FrequencyMapAgent.cpp
cmcantalupo: That is why it is a debug assert.

src/FrequencyMapAgent.cpp
cmcantalupo: ?

test_integration/test_frequency_hint_usage.py
cmcantalupo: The problem is that the test fixtures use the launcher in this test to get the min and max frequencies.  This is not generally going to be the case.

ronn/geopm_agent_frequency_map.7.ronn
bgeltz: If a NAN policy comes through the Endpoint, this is not an error.  We need to consider this use case.

Do we expect the policy writer to actually write an all NAN's policy?

If the agent gets a non-all-NAN's policy, from that point forward should the agent throw if it receives an all-NAN policy?

Change the agent to throw that it will not accept an all NANs policy after it has received a valid policy.  Eventually even the initial all-NAN policy should be an error (this should be a TODO in the source).  Remove the documentation that all-NANs is valid from the man page.

ronn/geopm_agent_frequency_map.7.ronn
bgeltz: Please make it so that if any region runs long enough to get the default frequency from the policy, that the default frequency shows up in the individual region's section in the report (not just up in the top policy section).

test_integration/test_frequency_hint_usage.py
bgeltz: Bug opened for this: #1090

bgeltz (changes requested):

bgeltz (approved):

PR #1092, opened 2020-05-04
Add checks for invalid agent or policy in PolicyStore

### LINKED ISSUE #775: PolicyStore python interface should not allow invalid policies to be set
For example, set_default('monitor', [float('nan')] should raise an error because the monitor does not take any policies.  Not sure if this also needs to be checked in underlying C and C++ methods.

Parent story #916

### LINKED ISSUE #985: PolicyStore python interface allows policies to be saved for unknown agents
Related to #775 

The methods to store default and best policies do not check if the string passed for the agent name is a real agent known to the factory.
Will also require #764 to be fixed.

Parent story #916

dianarg:
- Fixes #775
- Fixes #985

### DIFF
--- a/src/PolicyStoreImp.cpp
+++ b/src/PolicyStoreImp.cpp
@@ -300,6 +300,13 @@ namespace geopm
                                   const std::string &profile_name,
                                   const std::vector<double> &policy)
     {
+        // check that the policy is valid for the agent
+        size_t num_policy = Agent::num_policy(agent_name);
+        if (policy.size() > num_policy) {
+            throw Exception("PolicyStoreImp::set_best(): invalid policy for " + agent_name + " agent.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
         begin_transaction_or_throw(m_database);
         {
             // Remove existing policy values for this record in case the new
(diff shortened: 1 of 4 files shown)


src/PolicyStoreImp.cpp
```diff
@@ -300,6 +300,13 @@ namespace geopm
                                   const std::string &profile_name,
                                   const std::vector<double> &policy)
     {
+        // check that the policy is valid for the agent
+        size_t num_policy = Agent::num_policy(agent_name);
+        if (policy.size() > num_policy) {
```
bgeltz, line 305: Is it possible that policy.size() would be < num_policy?  Is that also an error condition?

src/PolicyStoreImp.cpp
```diff
@@ -338,6 +345,13 @@ namespace geopm
     void PolicyStoreImp::set_default(const std::string &agent_name,
                                      const std::vector<double> &policy)
     {
+        // check that the policy is valid for the agent
+        size_t num_policy = Agent::num_policy(agent_name);
+        if (policy.size() > num_policy) {
```
bgeltz, line 350: Same question on size here.

bgeltz (changes requested):

src/PolicyStoreImp.cpp
dianarg: This is allowed, and the missing policy values will be omitted from the policystore.  There is a test of this behavior on line 145 of PolicyStoreTest.cpp

bgeltz (approved):

PR #1094, opened 2020-05-04
Fix energy accounting for application totals in report

### LINKED ISSUE #1093: Energy totals in report incorrect
```
Application Totals:
    runtime (sec): 10.0534
    package-energy (joules): 1235.31
    dram-energy (joules): -65427.2
    power (watts): 122.874
    network-time (sec): 0
    ignore-time (sec): 0
    geopmctl memory HWM: 39324 kB
    geopmctl network BW (B/sec): 0
```

The ApplicationIO class uses read_signal() not sample() to calculate the energy values.  The new implementation doesn't borrow roll-over data from the read_batch() calls to service the read_signal() calls.  This make sense, but the ApplicationIO must be changed to use sample for these caluclations.

cmcantalupo:
- Fixes #1093 from github issues
- The ApplicationIO class must use the PlatformIO sample()
  API not read_signal() in order to properly deal with rollover
  of the energy counters issues in energy counters.

### DIFF
--- a/src/ApplicationIO.cpp
+++ b/src/ApplicationIO.cpp
@@ -76,9 +76,8 @@ namespace geopm
         , m_is_connected(false)
         , m_rank_per_node(-1)
         , m_epoch_regulator(std::move(epoch_regulator))
-        , m_start_energy_pkg(NAN)
-        , m_start_energy_dram(NAN)
     {
+
     }
 
     ApplicationIOImp::~ApplicationIOImp()
@@ -100,9 +99,6 @@ namespace geopm
                 platform_io().register_iogroup(geopm::make_unique<ProfileIOGroup>(m_profile_io_sample, *m_epoch_regulator));
             }
             m_is_connected = true;
-
-            m_start_energy_pkg = current_energy_pkg();
-            m_start_energy_dram = current_energy_dram();
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/ApplicationIO.cpp
```diff
@@ -101,8 +104,10 @@ namespace geopm
             }
             m_is_connected = true;
 
-            m_start_energy_pkg = current_energy_pkg();
-            m_start_energy_dram = current_energy_dram();
+            m_energy_pkg_idx = m_platform_io.push_signal("ENERGY_PACKAGE", GEOPM_DOMAIN_BOARD, 0);
+            m_energy_dram_idx = m_platform_io.push_signal("ENERGY_DRAM", GEOPM_DOMAIN_BOARD, 0);
+            m_start_energy_pkg = m_platform_io.sample(m_energy_pkg_idx);
```
cmcantalupo: This is a non-starter.  We cannot sample at this point in the stack.  We will not be able to push any signals other than the ProfileIO signals.  The right solution is to push_signal() and call read_signal() here.  The call at the end should be to sample().

src/Reporter.cpp
```diff
@@ -288,13 +303,14 @@ namespace geopm
         // extra runtimes for epoch region
         report << "    epoch-runtime-ignore (sec): " << application_io.total_epoch_runtime_ignore() << std::endl;
 
-        double total_runtime = application_io.total_app_runtime();
-        double app_energy_pkg = application_io.total_app_energy_pkg();
+        double total_runtime = m_platform_io.sample(m_app_time_signal_idx);
```
dianarg: assumes start time is 0

test/ApplicationIOTest.cpp
```diff
@@ -77,6 +79,9 @@ void ApplicationIOTest::SetUp()
     m_epoch_regulator = new MockEpochRuntimeRegulator;
     auto tmp_reg = std::unique_ptr<MockEpochRuntimeRegulator>(m_epoch_regulator);
 
+    m_energy_pkg_idx = 18;
```
dianarg: leftovers here and on 69-70

test/ReporterTest.cpp
```diff
@@ -203,13 +209,19 @@ TEST_F(ReporterTest, generate)
...
-    EXPECT_CALL(m_application_io, total_app_runtime()).WillOnce(Return(56));
-    EXPECT_CALL(m_application_io, total_app_energy_pkg()).WillOnce(Return(2222));
-    EXPECT_CALL(m_application_io, total_app_energy_dram()).WillOnce(Return(2222));
     EXPECT_CALL(m_application_io, total_app_runtime_mpi()).WillOnce(Return(45));
     EXPECT_CALL(m_application_io, total_app_runtime_ignore()).WillOnce(Return(0.7));
     EXPECT_CALL(m_application_io, total_epoch_runtime_ignore()).WillRepeatedly(Return(0.7));
     EXPECT_CALL(m_application_io, total_epoch_runtime()).WillOnce(Return(70.0));
+    EXPECT_CALL(m_platform_io, sample(M_TIME_IDX))
+        .WillOnce(Return(56));
+    EXPECT_CALL(*m_agg, read_batch);
```
dianarg, line 216: move up above sample

bgeltz (approved):

PR #1097, opened 2020-05-07
Fix lazy pio registration

cmcantalupo:
Merge after #1094

### DIFF
--- a/src/PlatformIO.cpp
+++ b/src/PlatformIO.cpp
@@ -78,7 +78,7 @@ namespace geopm
         if (m_iogroup_list.size() == 0) {
             for (const auto &it : IOGroup::iogroup_names()) {
                 try {
-                    if (it != "PROFILE") {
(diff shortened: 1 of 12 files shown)


src/PlatformIO.cpp
```diff
@@ -77,7 +78,12 @@ namespace geopm
         if (m_iogroup_list.size() == 0) {
             for (const auto &it : IOGroup::iogroup_names()) {
                 try {
-                    register_iogroup(IOGroup::make_unique(it));
+                    if (it != "PROFILE") {
```
cmcantalupo: Use ProfileIOGroup::plugin_name()

cmcantalupo (commented):
Have ProfileIOGroup return no signal names until connect happens.  This will avoid publishing signals that are not yet supported.

dianarg (approved):

PR #1101, opened 2020-05-07
Clean up mocking in MSRIOGroup tests

### LINKED ISSUE #1060: Clean up MSRIOGroup unit tests
> estimate 3

The existing tests use a MockMSRIO that inherits from the real one, which requires using real MSR offset and pwrite to implement the tests.  These test would be easier to write if they used a true mock for MSRIO and injected values for the underlying MSR reads.

The current tests could be kept in some form as integration tests.

Done criteria
--------------
- Complete testing of all signal aliases, both read_signal and read_batch+sample.
- Add unit tests of power signals, which cannot be supported with the current scheme; depends on #1042 
- Add tests of JSON parsing to Signal or similar object; depends on #435 
  + Might also want to add validation for arch/cpuid in JSON parsing

dianarg:
- Resolves #1060

### DIFF
--- a/src/MSRIOGroup.cpp
+++ b/src/MSRIOGroup.cpp
@@ -82,12 +82,12 @@ namespace geopm
         "Manual for information about this MSR";
 
     MSRIOGroup::MSRIOGroup()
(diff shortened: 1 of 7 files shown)


cmcantalupo (commented):
LGTM

cmcantalupo (approved):
LGTM

PR #1114, opened 2020-05-20
Issues 1110 and 1111

### LINKED ISSUE #1110: Create a singleton object that holds the EpochRuntimeRegulator and the ProfileSampler
> estimate 3

For the first iteration we will introduce a new Singleton object called the KProfileSamper.   This will have four methods to start with:
```
class KProfileSampler
{
    public:
        static KProfileSampler& profile_sampler(void);
        void set_sampler(std::shared_ptr<ProfileSamper>);
        std::shared_ptr<ProfileSampler> get_sampler(void);
        void set_regulator(std::shared_ptr<EpochRuntimeRegulator>);
        std::shared_ptr<EpochRuntimeRegulator> get_regulator(void);
    private:
        KProfileSampler() = default;
        virtual ~KProfileSampler() = default;
};
```

Done criterion:
- Implement interface above with error conditions for multiple set calls.
- Write unit tests.
- Replace all references to the ProfileSampler or the EpochRuntimeRegulator with singleton accesses KProfileSampler::profile_sampler()...

Parent story: #1109

### LINKED ISSUE #1111: Add two APIs to the KProfileSampler class for dumping profile data and accessing it
> estimate 5

This higher level API will use memory managed by the KProfileSampler object to cache the output from ProfileSampler::dump().  There will be an accessor method that can see the last update and can be called many times.  The dump method will be called once to update the cache for all users.

Done criterion:
- Add methods as described above to the KProfileSampler
- Add unit tests
- Disperse responsibilities currently handled in ApplicationIO::update() to their respective objects
- Make sure integration tests don't show regression.

Parent story: #1109

cmcantalupo:
- Fixes #1110 from github issues.
- Fixes #1111 from github issues.

### DIFF
--- a/src/ApplicationIO.cpp
+++ b/src/ApplicationIO.cpp
@@ -35,6 +35,7 @@
 #include <utility>
 
 #include "Exception.hpp"
+#include "ApplicationSampler.hpp"
 #include "EpochRuntimeRegulator.hpp"
 #include "PlatformIO.hpp"
 #include "PlatformTopo.hpp"
@@ -55,29 +56,26 @@ namespace geopm
 
     ApplicationIOImp::ApplicationIOImp(const std::string &shm_key)
         : ApplicationIOImp(shm_key,
-                           geopm::make_unique<ProfileSamplerImp>(M_SHMEM_REGION_SIZE),
-                           nullptr, nullptr,
+                           ApplicationSampler::application_sampler(),
                            platform_io(), platform_topo())
     {
 
     }
 
     ApplicationIOImp::ApplicationIOImp(const std::string &shm_key,
-                                       std::unique_ptr<ProfileSampler> sampler,
-                                       std::shared_ptr<ProfileIOSample> pio_sample,
-                                       std::shared_ptr<EpochRuntimeRegulator> epoch_regulator,
+                                       ApplicationSampler &application_sampler,
                                        PlatformIO &platform_io,
                                        const PlatformTopo &platform_topo)
-        : m_sampler(std::move(sampler))
-        , m_profile_io_sample(pio_sample)
-        , m_platform_io(platform_io)
+        : m_platform_io(platform_io)
         , m_platform_topo(platform_topo)
         , m_thread_progress(m_platform_topo.num_domain(GEOPM_DOMAIN_CPU))
         , m_is_connected(false)
-        , m_rank_per_node(-1)
-        , m_epoch_regulator(epoch_regulator)
+        , m_application_sampler(application_sampler)
     {
-
+        if (m_application_sampler.get_io_sample() == nullptr) {
+            auto sampler = std::make_shared<ProfileSamplerImp>(M_SHMEM_REGION_SIZE);
+            m_application_sampler.set_sampler(sampler);
+        }
     }
 
     ApplicationIOImp::~ApplicationIOImp()
@@ -88,14 +86,17 @@ namespace geopm
     void ApplicationIOImp::connect(void)
     {
         if (!m_is_connected) {
-            m_sampler->initialize();
-            m_rank_per_node = m_sampler->rank_per_node();
-            m_prof_sample.resize(m_sampler->capacity());
-            std::vector<int> cpu_rank = m_sampler->cpu_rank();
-            if (m_profile_io_sample == nullptr) {
-                m_epoch_regulator = std::make_shared<EpochRuntimeRegulatorImp>(m_rank_per_node, m_platform_io, m_platform_topo);
-                m_profile_io_sample = std::make_shared<ProfileIOSampleImp>(cpu_rank, m_epoch_regulator);
-                platform_io().get_profileio()->connect(m_profile_io_sample, m_epoch_regulator);
+            m_application_sampler.get_sampler()->initialize();
+            m_prof_sample.resize(m_application_sampler.get_sampler()->capacity());
+
+            if (m_application_sampler.get_io_sample() == nullptr) {
+                auto rank_per_node = m_application_sampler.get_sampler()->rank_per_node();
+                auto epoch_regulator = std::make_shared<EpochRuntimeRegulatorImp>(rank_per_node, m_platform_io, m_platform_topo);
+                m_application_sampler.set_regulator(epoch_regulator);
+
+                auto profile_io_sample = std::make_shared<ProfileIOSampleImp>(m_application_sampler); // This could probably be moved
+                m_application_sampler.set_io_sample(profile_io_sample);
+                platform_io().get_profileio()->connect();
             }
             m_is_connected = true;
         }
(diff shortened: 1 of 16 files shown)


src/ApplicationSampler.cpp
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+
+    void ApplicationSamplerImp::time_zero(const geopm_time_s &start_time)
+    {
+        m_time_zero = start_time;
+    }
+
+
+    void ApplicationSamplerImp::update_records(void)
```
cmcantalupo, line 189: Function should probably be broken down into smaller private methods for each different kind of geopm_message_s.

src/ApplicationSampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+            void update_records_progress(const geopm_prof_message_s &msg);
+            m_process_s &get_process(int process);
+            std::shared_ptr<ProfileSampler> m_sampler;
+            std::shared_ptr<EpochRuntimeRegulator> m_regulator;
+            std::shared_ptr<ProfileIOSample> m_io_sample;
+            struct geopm_time_s m_time_zero;
+            std::vector<m_record_s> m_record_buffer;
+            std::map<int, m_process_s> m_process_map;
+    };
+
```
dianarg, line 81: acres of whitespace

src/ApplicationSampler.hpp
```diff
@@ -0,0 +1,165 @@
...
+            /// a record will represent different data.  The
+            /// description of the "signal" field in the m_record_s is
+            /// given for each enum value.
+            enum m_event_e {
+                M_EVENT_REGION_ENTRY,     /// EVENT: The application has entered a region.
+                                          /// SIGNAL: The hash of the entered region.
+                M_EVENT_REGION_EXIT,      /// EVENT: The application has exited a region.
+                                          /// SIGNAL: The hash of the exited region.
+                ///
+                /// @todo SUPPORT FOR EVENTS BELOW IS FUTURE WORK
```
dianarg, line 72: move down; some are supported

src/ApplicationSampler.hpp
```diff
@@ -0,0 +1,165 @@
...
+                ///
+                M_EVENT_EPOCH_COUNT,      /// EVENT: The application has begun a new epoch.
+                                          /// SIGNAL: The number new epochs begun (1, 2,...)
```
dianarg: number of

src/ApplicationSampler.hpp
```diff
@@ -0,0 +1,165 @@
...
+                M_EVENT_REGISTER_PROFILE, /// EVENT: The application has started up and all
```
dianarg: maybe an event for removed a profile - when application has finished

also might want an event for writing reports; report name?

src/ApplicationSampler.hpp
```diff
@@ -0,0 +1,165 @@
...
+                                          ///        all processes associated with the
+                                          ///        application identify their profile name.
+                                          /// SIGNAL: The hash of the profile name unique to
+                                          ///         the application.
+                M_EVENT_CLAIM_CPU,        /// EVENT: The application has started up.  Each
```
dianarg: what happens if CPU is already claimed? -> error
need to be robust to missing release events

src/ApplicationSampler.hpp
```diff
@@ -0,0 +1,165 @@
...
+            ///        then create a reverse mapping of hashes
+            ///        recieved in prior events to their string name
+            ///        for reporting purposes.
+            ///
+            /// @param [in] name_key The signal from the name key
+            ///        event record.
+            /// @return Vector of all strings hashed by the process
+            ///         where the hash was returned as part of an
+            ///         event.
+            virtual std::vector<std::string> get_names(uint64_t name_key) const = 0;
```
dianarg: could be map from hash to string

src/ApplicationSampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+    void ApplicationSamplerImp::update_records_exit(const geopm_prof_message_s &msg)
+    {
+        // Set common values for all events
+        double time = geopm_time_diff(&m_time_zero, &(msg.timestamp));
+        int process = msg.rank;
+        /// Handle outer region entry and exit
+        double hash = (double)geopm_region_id_hash(msg.region_id);
+        int event = M_EVENT_REGION_EXIT;
+        uint64_t hint = GEOPM_REGION_HINT_UNKNOWN;
```
dianarg: could move down next to hint event

test/ApplicationSamplerTest.cpp
```diff
@@ -0,0 +1,259 @@
...
+    EXPECT_EQ(0, result[3].process);
+    EXPECT_EQ(ApplicationSampler::M_EVENT_HINT, result[3].event);
+    EXPECT_EQ(GEOPM_REGION_HINT_UNKNOWN, result[3].signal);
+}
+
+TEST_F(ApplicationSamplerTest, with_mpi)
+{
+    uint64_t region_id = 0xabcd | GEOPM_REGION_HINT_COMPUTE;
+    std::vector<struct geopm_prof_message_s> message_buffer {
+        {.rank=234,
```
dianarg: could represent this data as a table without field names

src/ApplicationIO.cpp
```diff
@@ -88,23 +86,26 @@ namespace geopm
...
-                platform_io().get_profileio()->connect(m_profile_io_sample, m_epoch_regulator);
+            m_application_sampler.get_sampler()->initialize();
+            m_prof_sample.resize(m_application_sampler.get_sampler()->capacity());
+
+            if (m_application_sampler.get_io_sample() == nullptr) {
+                auto rank_per_node = m_application_sampler.get_sampler()->rank_per_node();
+                auto epoch_regulator = std::make_shared<EpochRuntimeRegulatorImp>(rank_per_node, m_platform_io, m_platform_topo);
+                m_application_sampler.set_regulator(epoch_regulator);
+
+                auto profile_io_sample = std::make_shared<ProfileIOSampleImp>(m_application_sampler); // This could probably be moved
```
dianarg: should just use singleton.  not sure if two objects having shared_ptr to each other will be a problem.

src/ApplicationSampler.hpp
```diff
@@ -0,0 +1,165 @@
...
+            struct m_record_s {
+                /// @brief Elapsed time since time zero when event was
+                ///        recorded.
+                double time;
+                /// @brief The process identifier where event occured.
+                int process;
+                /// @brief One of the m_event_e event types.
+                int event;
+                /// @brief The signal associated with the event type.
+                double signal;
```
dianarg: could be union of double or int

dianarg: or maybe this should just be an integer since progress will go away

src/ApplicationSampler.hpp
cmcantalupo: Maybe these are the same event?  I have created

M_EVENT_PROFILE

and

M_EVENT_REPORT

src/ApplicationSampler.hpp
cmcantalupo: In current implementation, the controller throws if two ranks are affinitized to the same CPU.

bgeltz (dismissed):

bgeltz (approved):

dianarg (approved):
LGTM

PR #1116, opened 2020-05-21
Centralize definition of time zero

cmcantalupo:


bgeltz (approved):

bgeltz (changes requested):
Something looks fundamentally broken in the integration tests.

Code LGTM though.

bgeltz (dismissed):
The exit cookie thing is gross.  I'm wondering if it would be better to just write out {0,0} in all the spots?  Anyway, approved.

bgeltz (approved):

PR #1133, opened 2020-05-29
Introducing the ProcessEpoch interface

### LINKED ISSUE #1122: Design interface for EpochIOGroup supporting classes
> estimate 1

We will have different implementations for determining the Epoch signals.  We want to isolate those different methods into a class used by the EpochIOGroup.  There will be one of these objects per process.  The interfaces that this will support need to be defined.  The aggregation of per-process information will not be the responsibility of this helper class.

Parent story: #1121

Done criteria:
Merged PR with header for ProcessEpoch interface including doxygen comments.

Pseudo code below
```
class ProcessEpoch
{
    void update(const ApplicationSampler::m_record_s &record_it); 
    double last_epoch_runtime(void);
    double last_epoch_runtime_network(void);
    double last_epoch_runtime_ignore(void);
    int epoch_count(void);
};


/// EpochIOGroup implementation using ProcessEpoch map.
std::map<int, std::shared_ptr<ProcessEpoch> m_process_map;

epoch_count[]
for (auto &it : ApplicationSamper::application_sampler().get_records()) {
    m_process_map[it.process].update(it);
}

for (auto proc_it : m_process_map) {
    cpu_set = magic.cpu_set(proc_it.first);
    int ic = proc_it.second->epoch_count();
    for (cpu_it : cpu_set) {
        epoch_count[cpu_it] = ic;
    }
}
```

cmcantalupo:
- Fixes #1122 from github issues.

### DIFF
--- a/src/ApplicationSampler.hpp
+++ b/src/ApplicationSampler.hpp
@@ -35,6 +35,8 @@
 
 #include <memory>
 #include <vector>
+#include <map>
+#include <cstdint>
 
 #include "geopm_time.h"
 
--- a/src/ApplicationSampler.hpp
+++ b/src/ApplicationSampler.hpp
@@ -33,10 +33,10 @@
 #ifndef APPLICATIONSAMPLER_HPP_INCLUDE
 #define APPLICATIONSAMPLER_HPP_INCLUDE
 
+#include <cstdint>
 #include <memory>
 #include <vector>
 #include <map>
-#include <cstdint>
 
 #include "geopm_time.h"
 
@@ -102,7 +102,7 @@ namespace geopm
                 /// @brief Elapsed time since time zero when event was
                 ///        recorded.
                 double time;
-                /// @brief The process identifier where event occured.
+                /// @brief The process identifier where event occurred.
                 int process;
                 /// @brief One of the m_event_e event types.
                 int event;
--- a/src/ProcessEpoch.hpp
+++ b/src/ProcessEpoch.hpp
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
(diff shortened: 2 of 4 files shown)


src/ApplicationSampler.hpp
```diff
@@ -35,6 +35,8 @@
 
 #include <memory>
 #include <vector>
+#include <map>
+#include <cstdint>
```
dianarg: move before C++ headers

src/ProcessEpoch.hpp
```diff
@@ -0,0 +1,106 @@
...
+            ///        updates.
+            ///
+            ///  Factory method to create a unique pointer to a
+            ///  ProcessEpoch object that uses the M_EVENT_EPOCH_COUNT
+            ///  and M_EVENT_HINT events to determine the epoch
+            ///  related signals.
+            ///
+            ///  @return Unique pointer to a ProcessEpoch base class
+            ///          object that processes epoch events.
+            static std::unique_ptr<ProcessEpoch> make_unique(void);
```
dianarg, line 65: will move into derived class, right?

src/ProcessEpoch.hpp
```diff
@@ -0,0 +1,106 @@
...
+            virtual ~ProcessEpoch() = default;
+            /// @brief Process application sampler record to update
+            ///        signal data.
+            ///
+            /// @param record A reference to a record that was queued
+            ///        by the process tracked by this object.  It is
+            ///        the callers responsiblity to filter out records
+            ///        that are sent from other processes.
+            virtual void update(const ApplicationSampler::m_record_s &record) = 0;
+            /// @brief The number of epoch events that have occured
```
dianarg: occurred

dianarg: maybe 'completed', since the count starts at zero?

src/ProcessEpoch.hpp
```diff
@@ -0,0 +1,106 @@
...
+            ///        for the process that is tracked by this object.
+            ///
+            /// @return Zero based counter of the number of epoch
+            ///         events.
+            virtual int epoch_count(void) const = 0;
+            /// @brief The total runtime that elapsed between the last
```
dianarg, line 86: If this is supposed to return NAN before the first epoch completes, we should mention that here

dianarg (approved):

PR #1135, opened 2020-05-30
Add ProcessEpochImp

### LINKED ISSUE #1123: Write unit tests and implementation for RecordProcessEpoch
> estimate 3

The RecordProcessEpoch will be responsible for processing of records that were generated with the geopm_prof_epoch() call on the application side.

Done criteria:
Unit tests pass and branch merged.

Parent story: #1121

dianarg:
- Resolves #1123

### DIFF
--- a/src/ProcessEpochImp.cpp
+++ b/src/ProcessEpochImp.cpp
@@ -36,11 +36,12 @@
 
 #include "geopm.h"
 #include "ProcessEpochImp.hpp"
+#include "Exception.hpp"
 
 namespace geopm
 {
     ProcessEpochImp::ProcessEpochImp()
-        : m_epoch_count(-1)
+        : m_epoch_count(0)
         , m_last_epoch_time(NAN)
         , m_last_runtime(NAN)
         , m_curr_hint(GEOPM_REGION_HINT_UNKNOWN)
@@ -50,7 +51,6 @@ namespace geopm
         reset_hint_map(m_last_hint_runtime, NAN);
     }
 
-    // todo: might want value input too
     void ProcessEpochImp::reset_hint_map(std::map<uint64_t, double> &hint_map,
                                          double value)
     {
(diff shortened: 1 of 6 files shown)


src/ProcessEpochImp.hpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    class ProcessEpochImp : public ProcessEpoch
+    {
+        public:
+            ProcessEpochImp();
+            virtual ~ProcessEpochImp() = default;
+
+            void update(const ApplicationSampler::m_record_s &record) override;
+            double last_epoch_runtime(void) const override;
+            double last_epoch_runtime_network(void) const override;
```
cmcantalupo, line 48: could be

```
double last_epoch_runtime_hint(uint64_t hint);
```

given the implementation.

src/ProcessEpochImp.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+#include <cmath>
+
+#include "geopm.h"
+#include "ProcessEpochImp.hpp"
+
+namespace geopm
+{
+    ProcessEpochImp::ProcessEpochImp()
+        : m_epoch_count(-1)
```
cmcantalupo: Maybe should be zero based until we generate the signal in the IOGroup.  Leaves fewer places to change later if we redefine what the signal means.

test/ProcessEpochImpTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+void ProcessEpochImpTest::update_all(void)
+{
+    for (const auto &rec : m_records) {
+        m_process.update(rec);
+    }
+}
+
+// shorter names for the enum event types
+enum {
```
cmcantalupo, line 64: nice touch, would be nice to copy in the filter tests.

src/ProcessEpochImp.cpp
```diff
@@ -0,0 +1,124 @@
...
+        hint_map[GEOPM_REGION_HINT_PARALLEL] = value;
+        hint_map[GEOPM_REGION_HINT_IGNORE] = value;
+    }
+
+    void ProcessEpochImp::update(const ApplicationSampler::m_record_s &record)
+    {
+        switch (record.event) {
+            case ApplicationSampler::M_EVENT_EPOCH_COUNT:
+                // update count
+                m_epoch_count = record.signal;
```
cmcantalupo: Do we want to do anything about checking that record.signal == m_epoch_count + 1?  Not sure what the right response should be if they are not contiguous.  Might consider dividing last runtime by the difference, but I'm not sure that makes sense.

src/ProcessEpochImp.cpp
```diff
@@ -0,0 +1,124 @@
...
+        hint_map[GEOPM_REGION_HINT_NETWORK] = value;
+        hint_map[GEOPM_REGION_HINT_IO] = value;
+        hint_map[GEOPM_REGION_HINT_SERIAL] = value;
+        hint_map[GEOPM_REGION_HINT_PARALLEL] = value;
+        hint_map[GEOPM_REGION_HINT_IGNORE] = value;
+    }
+
+    void ProcessEpochImp::update(const ApplicationSampler::m_record_s &record)
+    {
+        switch (record.event) {
```
cmcantalupo: Could be an if() else if() construction since there are only two enums, but the case statement is okay.  Might be nice to have these two cases implementations in private methods anyway.

cmcantalupo (dismissed):
Looks pretty good to me.  A couple of comments, but I'm open to discussion about the suggested changes. Looks pretty tidy, and I like the tests.

src/ProcessEpochImp.cpp
dianarg: possibly this could be checked by the "validate" filter, but that would mean we have to keep two separate epoch counts.

cmcantalupo (approved):

PR #1139, opened 2020-06-03
Add base class and one implementation of a record filter

cmcantalupo:
- Add doxygen comments to filter classes
- Implemented simple epoch inference with the
  ProxyEpochRecordFilter
- Filtering done in the ApplicationSampler
- Add plumbing to add record filter to ApplicationSampler

Change-Id: Id5bc5b97d3cc2d5f5953d3580f6f403bb59452ce

### DIFF
--- a/src/ApplicationSampler.cpp
+++ b/src/ApplicationSampler.cpp
@@ -33,6 +33,7 @@
 #include "config.h"
 
 #include <map>
+#include <functional>
 
 #include "ApplicationSampler.hpp"
 #include "ProfileSampler.hpp"
@@ -48,7 +49,7 @@ namespace geopm
     {
         public:
             ApplicationSamplerImp();
-            ApplicationSamplerImp(const std::string &filter_name);
+            ApplicationSamplerImp(std::function<std::unique_ptr<RecordFilter>()> filter_factory);
             virtual ~ApplicationSamplerImp() = default;
             void time_zero(const geopm_time_s &start_time) override;
             void update_records(void) override;
@@ -81,7 +82,7 @@ namespace geopm
             struct geopm_time_s m_time_zero;
             std::vector<m_record_s> m_record_buffer;
             std::map<int, m_process_s> m_process_map;
-            std::string m_filter_name;
+            std::function<std::unique_ptr<RecordFilter>()> m_filter_factory;
             bool m_is_filtered;
     };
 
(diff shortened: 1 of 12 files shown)


src/ProxyEpochRecordFilter.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+#ifndef PROXYEPOCHRECORDFILTER_HPP_INCLUDE
+#define PROXYEPOCHRECORDFILTER_HPP_INCLUDE
+
+#include "RecordFilter.hpp"
+
+namespace geopm
+{
+    /// @brief Filter that can be used in place of the
+    ///        EpochRecordFilter to synthesize epoch events from a
```
dianarg: EpochRecordFilter is going away, right?

src/ProxyEpochRecordFilter.hpp
```diff
@@ -0,0 +1,99 @@
...
+    ///        sequence of region entry events.  The filter also
+    ///        passes hint events through.
```
dianarg: update to indicate all events except epoch are passed though

src/RecordFilter.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "config.h"
+
+#include "RecordFilter.hpp"
+#include "ProxyEpochRecordFilter.hpp"
+#include "Helper.hpp"
+#include "Exception.hpp"
+
+namespace geopm
+{
+    std::unique_ptr<RecordFilter> RecordFilter::make_unique(const std::string &name)
```
dianarg, line 42: please add tests for this

src/ApplicationSampler.cpp
```diff
@@ -86,7 +92,15 @@ namespace geopm
     }
 
     ApplicationSamplerImp::ApplicationSamplerImp()
+        : ApplicationSamplerImp(environment().record_filter())
+    {
+
+    }
+
+    ApplicationSamplerImp::ApplicationSamplerImp(const std::string &filter_name)
```
dianarg: it might be easier to test the passthrough if RecordFilters are injected instead of the name

dianarg (changes requested):

src/ApplicationSampler.cpp
cmcantalupo: We need a way to generate record filters for each process after construction.  That is facilitated with a string now.  I'm not sure how we want to mock out the RecordFilter::make_unique() call, which is where we will run into testing problems.

cmcantalupo: maybe:

```
ApplicationSamplerImp(std::function<std::unique_ptr<RecordFilter>()> filter_factory);
```

test/RecordFilterTest.cpp
```diff
@@ -0,0 +1,95 @@
...
+TEST_F(RecordFilterTest, invalid_filter_name)
+{
+    GEOPM_EXPECT_THROW_MESSAGE(RecordFilter::make_unique("invalid_filter_name"),
+                               GEOPM_ERROR_INVALID,"parse name");
+}
+
+TEST_F(RecordFilterTest, make_proxy_epoch)
+{
+    std::shared_ptr<RecordFilter> filter = RecordFilter::make_unique("proxy_epoch,0xabcd1234");
+    EXPECT_TRUE(filter);
```
dianarg: maybe EXPECT_NO_THROW would be clearer?

dianarg (dismissed):

dianarg (approved):

PR #1140, opened 2020-06-04
Move filter string parsing into derived class

cmcantalupo:


dianarg (dismissed):

src/ProxyEpochRecordFilter.cpp
```diff
@@ -34,9 +34,98 @@
 
 #include "ProxyEpochRecordFilter.hpp"
 #include "Exception.hpp"
+#include "Helper.hpp"
 
 namespace geopm
 {
+    static uint64_t parse_region_hash(std::string filter_name)
```
dianarg: inputs could be const&

bgeltz (approved):

PR #1144, opened 2020-06-05
Implement and test the ValidateRecordFilter.

### LINKED ISSUE #1132: Implement validation of event stream
> estimate 2

Parent story: #1121 

The current proposal is that this is one or more filter type(s) that checks validity and consistency of the event stream. For example:
- from the point of view of a single ProcessEpoch object, it is a logic error if we see a different process id, even though this object doesn't use that field.
- signal values for the HINT type should be in bounds of the hint enum.
- check matching region entry/exit pairs.
- Epoch count must be strictly increasing. No duplicate counts, no going backwards, no gaps.
- timestamps for a given process must be non-decreasing

cmcantalupo:
Fixes #1132

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -591,6 +591,8 @@ libgeopmpolicy_la_SOURCES = contrib/json11/json11.cpp \
                             src/TreeComm.hpp \
                             src/TreeCommLevel.cpp \
                             src/TreeCommLevel.hpp \
+                            src/ValidateRecord.cpp \
+                            src/ValidateRecord.hpp \
                             src/geopm.h \
                             src/geopm_agent.h \
                             src/geopm_endpoint.h \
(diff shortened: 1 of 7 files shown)


src/ValidateRecordFilter.cpp
```diff
@@ -0,0 +1,116 @@
...
+    std::vector<ApplicationSampler::m_record_s> ValidateRecordFilter::filter(const ApplicationSampler::m_record_s &record)
+    {
+        if (m_is_empty) {
+            m_time = record.time;
+            m_process = record.process;
+            m_epoch_count = 0;
+            m_region_hash = GEOPM_REGION_HASH_INVALID;
+            m_is_empty = false;
+        }
+        if (record.process != m_process) {
```
dianarg: so we will have a filter per process?  I think this means this filter needs to live in the IOGroup

bgeltz (approved):

PR #1146, opened 2020-06-05
Integration test of EPOCH signals

dianarg:


test_integration/geopm_test_launcher.py
```diff
@@ -199,7 +199,7 @@ def check_run(self, test_name):
                                                          self._node_list, self._host_file)
             launcher.run(stdout=outfile, stderr=outfile)
 
-    def run(self, test_name, include_geopm_policy=True):
+    def run(self, test_name, include_geopm_policy=True, add_geopm_args=[]):
```
cmcantalupo, line 202: the back door is open!

test_integration/test_epoch_inference.cpp
```diff
@@ -0,0 +1,85 @@
...
+        for (int arg_idx = 1; arg_idx < argc; ++arg_idx) {
+            std::string arg(argv[arg_idx]);
+            if (arg == "--verbose" || arg == "-v") {
+                is_verbose = true;
+            }
+        }
+    }
+
+    std::unique_ptr<geopm::ModelRegion> all2all(
+        geopm::ModelRegion::model_region("all2all", 1.0, is_verbose));
```
cmcantalupo, line 64: do we want "all2all-unmarked" "spin-unmarked" etc?

test_integration/test_epoch_inference.cpp
```diff
@@ -0,0 +1,85 @@
...
+    std::unique_ptr<geopm::ModelRegion> spin(
+        geopm::ModelRegion::model_region("spin", 1.0, is_verbose));
+    std::unique_ptr<geopm::ModelRegion> ignore(
+        geopm::ModelRegion::model_region("spin", 0.5, is_verbose));
+
+    geopm::Profile &prof = geopm::Profile::default_profile();
+    uint64_t ignore_rid = prof.region("ignore", GEOPM_REGION_HINT_IGNORE);
```
cmcantalupo: not sure i follow why we need the special ignore_rid

cmcantalupo (dismissed):
Looks good to me.  I like the back door, per-run parameters for the test launcher.

test_integration/geopm_test_launcher.py
dianarg: I would really like to fix up the test launcher.  There are multiple different ways of handling the geopm command line args.

test_integration/test_epoch_inference.py
```diff
@@ -0,0 +1,295 @@
...
+                                                            agent_conf,
+                                                            report_path,
+                                                            trace_path,
+                                                            time_limit=time_limit,
+                                                            trace_signals=trace_signals)
+                launcher.set_num_node(num_node)
+                launcher.set_num_rank(num_rank)
+                # Run the test application
+                geopm_args = []
+                # TODO: would be nice if this env variable could take name instead of hash
```
cmcantalupo: It does!

cmcantalupo (dismissed):
Could change the hash's to strings, but also okay as it is.  Might be nice to have a test binary that doesn't link to geopm but still shows the epoch signal through just PMPI callbacks, but there are a lot of things that would be nice.

cmcantalupo (approved):

PR #1148, opened 2020-06-08
Change epoch and app total runtime test comparison

dianarg:
- App total runtime is not strictly greater than epoch total because
  app total uses sampled values (i.e. sync runtime).


cmcantalupo (approved):
LGTM

PR #1151, opened 2020-06-10
Remove geopmanalysis

### LINKED ISSUE #1150: Remove deprecated geopmanalysis tool and tests
Use of this tool will be replaced by integration tests

dianarg:
- Resolves #1150


cmcantalupo (approved):
A bit sad to drop this feature.  Would be nice to bring back some of the logic for use in integration tests. It may also be nice to use some of of the logic when trying to generate application specific frequency map agent policies as a feature of the policy store.  Lets bring it back opportunistically.

PR #1159, opened 2020-06-11
Fix docu issues around signal/control descriptions

bgeltz:
Follow-up work from the previous sprint.

### DIFF
--- a/src/msr_data_knl.cpp
+++ b/src/msr_data_knl.cpp
@@ -89,7 +89,7 @@ namespace geopm
                     "scalar":    1e8,
                     "writeable": false,
                     "aggregation": "average",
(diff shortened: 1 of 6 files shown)


cmcantalupo (approved):

src/msr_data_knl.cpp
```diff
@@ -1,4 +1,4 @@
-/*
+*
```
cmcantalupo: Ooops

cmcantalupo (changes requested):
Dropped a comment character which broke the build.

dianarg (approved):
LGTM

cmcantalupo (approved):

PR #1164, opened 2020-06-18
Move enum helpers

cmcantalupo:
Requires merging #1163 first.

### DIFF
--- a/src/ProfileTracer.cpp
+++ b/src/ProfileTracer.cpp
@@ -98,7 +98,7 @@ namespace geopm
             // This is a call to format the event column
             // Store the event type for the next call
             event_type = value;
-            result = ApplicationSampler::event_name((int)value);
+            result = event_name((int)value);
             // The next call will format the signal column
             is_signal = true;
         }
@@ -113,7 +113,7 @@ namespace geopm
                     result = string_format_integer(value);
                     break;
                 case EVENT_HINT:
-                    result = string_format_hint(value);
+                    result = hint_name(value);
                     break;
                 default:
                     result = "INVALID";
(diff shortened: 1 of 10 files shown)


src/record.cpp
```diff
@@ -0,0 +1,115 @@
...
+        };
+        auto it = result_map.find(hint);
+        if (it == result_map.end()) {
+            throw Exception("Profile::hint_hame(): Invalid hint",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        return it->second;
+    }
+
+    uint64_t hint_type(std::string hint_name)
```
bgeltz: const std::string&

src/ProfileTracer.cpp
```diff
@@ -98,7 +98,7 @@ namespace geopm
             // This is a call to format the event column
             // Store the event type for the next call
             event_type = value;
-            result = ApplicationSampler::event_name((int)value);
+            result = event_name((int)value);
```
cmcantalupo, line 101: doesn't need a cast any more.

src/record.hpp
```diff
@@ -85,6 +86,23 @@ namespace geopm
                                  ///         parameter).
     };
 
+    /// @brief Format an event_e type as a string.
+    std::string event_name(int event_type);
```
bgeltz, line 90: If these methods are going to be called a million times a second for populating the profile trace, should you be using string references wherever possible?

bgeltz (changes requested):

src/record.hpp
cmcantalupo: Not for return values, that is resolved by return value optimization:

https://en.cppreference.com/w/cpp/language/copy_elision

src/record.cpp
cmcantalupo: thanks!

bgeltz (dismissed):

bgeltz (approved):

PR #1165, opened 2020-06-18
Proxy test tutorial 2

cmcantalupo:
Adds a unit test for the proxy epoch record filter that uses the profile trace from tutorial 2


test/ProxyEpochRecordFilterTest.cpp
```diff
@@ -240,3 +241,259 @@ TEST_F(ProxyEpochRecordFilterTest, parse_name)
...
+36.855053916|0|HINT|NETWORK
+36.855350322|0|REGION_EXIT|0x00000000d88e02a5
+36.855350322|0|HINT|UNKNOWN
+)";
+    MockApplicationSampler app;
+    ProxyEpochRecordFilter perf(0x9803a79a, 1, 0);
+    uint64_t epoch_count = 0;
+    bool is_epoch = false;
+    app.inject_records(tutorial_2_prof_trace);
+    for (double time = 0.0; time < 38.0; time+=1.0) {
```
cmcantalupo: time+=1.0 -> time += 1.0

test/ProxyEpochRecordFilterTest.cpp
```diff
@@ -240,3 +241,259 @@ TEST_F(ProxyEpochRecordFilterTest, parse_name)
         region_hash, calls_per_epoch, startup_count),
         GEOPM_ERROR_INVALID, "Unable to parse parameter startup_count");
 }
+
+TEST_F(ProxyEpochRecordFilterTest, parse_tutorial_2)
+{
+    const std::string tutorial_2_prof_trace = R"(# geopm_version: 1.1.0+dev241g195e2ed
```
bgeltz: It would be nice if these huge blocks of text were in a file that was read in at test time.  It would be much easier to maintain that should the trace format ever change again.

Also keeps this source file from getting bloated.

bgeltz (dismissed):

bgeltz (dismissed):

configure.ac
```diff
@@ -585,6 +585,9 @@ else
 fi
 AC_SUBST([ENABLE_BETA])
 
+GEOPM_SOURCE_DIR=`readlink -f $srcdir`
```
bgeltz: Unless backticks are required because of autoconf, please use $() instead.

Backticks for subshells have been deprecated.

test/ProxyEpochRecordFilterTest.cpp
```diff
@@ -64,6 +68,8 @@ void ProxyEpochRecordFilterTest::SetUp()
     m_out_events = {
         geopm::EVENT_EPOCH_COUNT,
     };
+    m_tutorial_2_prof_trace_path = GEOPM_SOURCE_DIR;
+    m_tutorial_2_prof_trace_path += "/test/ProxyEpochRecordFilterTest.tutorial_2_profile_trace";
```
bgeltz, line 72: In the future, we may want to have other tests that need trace file data as input.  When that happens, it would be a nice refactor to have one class member like ```std::string m_prof_trace_path``` and then have the specific tests populate the rest of the path to the data then parse.

bgeltz (changes requested):

bgeltz (approved):

PR #1204, opened 2020-07-15
Update test_frequency_sane_fmap integration test

### LINKED ISSUE #1155: Intermittent failures of test_frequency_sane_fmap integration test
There is a race condition where if the controller's first sample of the region is prior to the hint change due to the entry into a network sub-region then the report includes the map for the region otherwise it does not.  For this reason the failure of the integration test is intermittent.

Proposed solution: 
1. Make the agent aware of the MPI regions that we care about in the runtime.  The agent should not care about the hint any longer.

2. If we are not in a user marked region and we encounter an MPI region, do what the agent(s) normally does with the MPI regions.

bakerbrandond:
- There is a race condition where if the controller's first sample of
the region is prior to the hint change due to the entry into a network
sub-region then the report includes the map for the region otherwise it
does not.
- Update the test to supress the KeyError when learning doesn't occur
due to the race.
- Fixes #1155

Change-Id: Iee50d47c3b1b740f6cf4c7f328104e7ef80770b3

Summary of change.

- Details of change; please include all interfaces added, removed, or modified.
- Fixes #XXX from github issues.

### DIFF
--- a/src/FrequencyMapAgent.cpp
+++ b/src/FrequencyMapAgent.cpp
@@ -291,8 +291,8 @@ namespace geopm
                 freq = it->second;
             }
             else {
+                m_default_freq_hash.insert(curr_hash);
                 freq = m_default_freq;
-                m_hash_freq_map[curr_hash] = m_default_freq;
             }
             if (m_last_freq[ctl_idx] != freq) {
                 m_last_freq[ctl_idx] = freq;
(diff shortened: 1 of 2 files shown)


cmcantalupo:
This seems like a bug, not a test issue.  Is it not possible to fix the bug?

cmcantalupo (approved):

PR #1205, opened 2020-07-15
Issue #1080

### LINKED ISSUE #1080: Energy efficient agent may choose wrong frequency based on hint
The algorithm for the the energy efficient agent sample the region hint value the first time a region is encountered. The hint may change throughout the region exection, but these changes would not effect the frequency selected by these agents which is fixed based on the first time sample. If a compute bound region were to begin with a brief call to MPI, the frequency would be set to the minimum for the entire execution of the region.

Possible Duplicate: #744

bakerbrandond:
Fixes #1080

### DIFF
--- a/src/EnergyEfficientAgent.cpp
+++ b/src/EnergyEfficientAgent.cpp
@@ -226,10 +226,9 @@ namespace geopm
                 .hint = (uint64_t)m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_HINT][ctl_idx]),
                 .runtime = m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_RUNTIME][ctl_idx]),
                 .count = (uint64_t)m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_COUNT][ctl_idx])};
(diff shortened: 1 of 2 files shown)


src/EnergyEfficientAgent.cpp
```diff
@@ -215,6 +215,19 @@ namespace geopm
...
     }
 
+    bool EnergyEfficientAgent::is_region_boundary(const struct m_region_info_s &last_region_info, const struct m_region_info_s &current_region_info)
+    {
+        bool ret = false;
+        if (last_region_info.hash != current_region_info.hash || // region hash change
+            (last_region_info.hash == current_region_info.hash &&
+             last_region_info.count != current_region_info.count) || // same hash, new count
+            (last_region_info.hash == current_region_info.hash &&
+             last_region_info.hint != current_region_info.hint)) { // same hash, new hint
```
bgeltz, line 225: With this change in place, does this mean that if MPI is nested within a user-marked region that the nested MPI is treated as a wholly separate region?

It seems that way for the last case in the if where the hash is the same but the hint has changed.

Issue #1194, opened 2020-07-08
Write a simple integration test using model regions to introduce rank-level imbalance across sockets


cmcantalupo:
> estimate 3

This integration test is to show the functionality of feature #1191, and will shall fail prior to merging the solution for #1191 and pass after the feature is properly implemented.  The application will use the shmem comm split feature to identify ranks on the same node.  

Done criteria:
- Application is implemented that splits the ranks into socket 0 and 1 ranks and gives and imbalanced amount of work to the sockets.
- Write an integration test that run a governor / balancer experiment on the above application on a single node and tests that the balancer provides better performance than the governor and that the governor shows imbalance.
- Another test is implemented that runs on two nodes and gives imbalance across sockets only and has similar assertions as the above test.

Parent story: #1191

PR #1207, opened 2020-07-17
Power balancer test

dianarg:
- Update existing test_power_balancer as standalone test
- Add another app with socket imbalance to the test
- Some clean up of --skip-launch option
- Added --show-details option for dumping raw data used by test
- Part of #1194


integration/test/test_power_balancer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    MPI_Comm shared_comm;
+    int err = geopm_comm_split_shared(MPI_COMM_WORLD, NULL, &shared_comm);
+    if (err) {
+        throw geopm::Exception("geopm_comm_split_shared()", err, __FILE__, __LINE__);
+    }
+    int shared_rank;
+    err = MPI_Comm_rank(shared_comm, &shared_rank);
+    if (err) {
+        throw geopm::Exception("MPI_Comm_rank()", err, __FILE__, __LINE__);
+    }
```
bgeltz: What purpose does this code serve?  Please comment.

integration/test/geopm_test_integration.py
```diff
@@ -79,6 +79,7 @@ def create_frequency_map_policy(max_freq, frequency_map):
     return policy
 
 
+@util.skip_unless_do_launch()
```
bgeltz, line 82: When running with skip-launch, does this skip everything in this class now?

integration/test/test_power_balancer.py
```diff
@@ -0,0 +1,243 @@
...
+            power_data = power_data.loc[(power_data != 0).all(axis=1)]  # Will drop any row that is all 0's
+
+            pkg_energy_cols = [s for s in power_data.keys() if 'ENERGY_PACKAGE' in s]
+            dram_energy_cols = [s for s in power_data.keys() if 'ENERGY_DRAM' in s]
+            power_data['SOCKET_POWER'] = power_data[pkg_energy_cols].sum(axis=1) / power_data['ELAPSED_TIME']
+            power_data['DRAM_POWER'] = power_data[dram_energy_cols].sum(axis=1) / power_data['ELAPSED_TIME']
+            power_data['COMBINED_POWER'] = power_data['SOCKET_POWER'] + power_data['DRAM_POWER']
+
+            pandas.set_option('display.width', 100)
+            # launcher.write_log(name, 'Power stats from {} {} :\n{}'.format(agent, nn, power_data.describe()))
```
bgeltz: Maybe this should just be uncommented and left in all the time.  Having this data from the nightlies might be useful.  I think you could also wrap it in an if check for skip launch if you don't want to pollute the output log when you're running that way.

integration/test/test_power_balancer.py
```diff
@@ -0,0 +1,243 @@
...
+                    app_conf.set_loop_count(loop_count)
+                    # Update app config with imbalance
+                    alloc_nodes = geopm_test_launcher.TestLauncher.get_alloc_nodes()
+                    for nn in range(len(alloc_nodes) // 2):
+                        app_conf.append_imbalance(alloc_nodes[nn], 0.5)
+                elif app_name == 'socket_imbalance':
+                    app_conf = cls.AppConf()
+                else:
+                    raise RuntimeError('No application config for app name {}'.format(app_name))
+                for agent in cls._agent_list:
```
bgeltz, line 125: In the run scripts I've had on the big clusters, I've been putting a 60 second sleep between Governor and Balancer runs in the hope that it's enough time to cool the node.  Otherwise we may see thermal related throttling that impacts this test.

bgeltz (changes requested):

integration/test/geopm_test_integration.py
dianarg: yes.  When we want to run all the tests/run nightly regressions, we shouldn't try to use skip-launch for now.

integration/test/test_power_balancer.cpp
dianarg: I think this is boilerplate from the test template.  @cmcantalupo do you know what this is for?

integration/test/test_power_balancer.py
```diff
@@ -0,0 +1,243 @@
...
+import glob
+import pandas
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+from integration.test import geopm_context
+import geopmpy.io
+import geopmpy.error
+
+from integration.test import util
+if True: #not util.do_launch():
```
dianarg: fixme

integration/test/test_power_balancer.py
dianarg: I can add that here.  One of the done criteria for this story is to add an experiment type to abstract out this launch config.  The 60sec sleep should be in there too (probably between power caps too)

integration/test/test_power_balancer.py
dianarg: It can be printed to stdout if you pass --show-details (on 186-187), but it won't be in the launcher log.  After splitting test into a launch step and analysis step, the launcher object doesn't actually exist in this function

integration/test/test_power_balancer.cpp
dianarg: Just going to delete these lines.  it doesn't do anything

bgeltz (approved):

PR #1208, opened 2020-07-21
Alawibaba/outlier test

alawibaba:
Outlier detection integration test for large clusters.

Is the integration test style still preferred for administrator tasks?

I'm assuming that I'll also need to squash commits before this could be merged, but I figured I'd shoot this over for a once-over.


bakerbrandond (commented):
Took a quick peek and had a question regarding the invocation of this new test.  Should this be added to the integration test binary?  If so may wanna enumerate it here: https://github.com/geopm/geopm/blob/dev/integration/test/__init__.py

bakerbrandond:
Also update the manifest appropriately, didn't look into your build failure but I assume that's what Travis is complaining about.

PR #1217, opened 2020-07-28
Introducing script to generate bug stat visualizations

cmcantalupo:
- Uses github API to query bug information
- For now it Popen's curl, would be better
  to use the requests module.
- Even when authentication seemed to be happening
  with requests, I was still being severly rate
  limited.
- The command line curl binary does not have this problem.
- It may have to do with proxy support, not sure.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -255,6 +255,7 @@ EXTRA_DIST = $(TUTORIAL_DIST) \
              geopm-tutorial.tar.gz \
              geopm.spec \
              geopm.spec.in \
+             indicators/indicators.py \
              pull_request_template.md \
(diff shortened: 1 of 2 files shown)


bgeltz (dismissed):
Some comments explaining what's going on (or a README) would be great in the next version.

sidjana:
plots look great!
Might help to add some artificial stalls to limit the query rate on the github API - should help limit the number of empty strings returned by the github servers.

cmcantalupo:
Cleaned up the colors and added a 100 ms stall between queries.

bgeltz (approved):

PR #1225, opened 2020-08-05
Issue 1048

bakerbrandond:


src/EnergyEfficientAgent.cpp
```diff
@@ -223,14 +221,15 @@ namespace geopm
         for (size_t ctl_idx = 0; ctl_idx < (size_t) m_num_freq_ctl_domain; ++ctl_idx) {
             struct m_region_info_s current_region_info {
                 .hash = (uint64_t)m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_HASH][ctl_idx]),
-                .hint = (uint64_t)m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_HINT][ctl_idx]),
+                .hint = GEOPM_REGION_HINT_UNKNOWN,
```
cmcantalupo: if the 'hint" field is not used, can we just remove it from the m_region_info_s?

src/ApplicationSampler.hpp
```diff
@@ -54,6 +55,10 @@ namespace geopm
         public:
             /// @brief Singleton accessor for the application sampler.
             static ApplicationSampler &application_sampler(void);
+            /// @brief Returns set of region hashes associated with
+            ///        application network functions.
+            /// @return Set of netowrk function region hashes.
+            static std::set<uint64_t> network_funcs_hash(void);
```
cmcantalupo: region_hash_network()?

src/EnergyEfficientAgent.cpp
```diff
@@ -223,14 +221,15 @@ namespace geopm
...
                 .runtime = m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_RUNTIME][ctl_idx]),
                 .count = (uint64_t)m_platform_io.sample(m_signal_idx[M_SIGNAL_REGION_COUNT][ctl_idx])};
             // If region hash has changed, or region count changed for the same region
             // update current region (entry)
             if (m_last_region_info[ctl_idx].hash != current_region_info.hash ||
                 m_last_region_info[ctl_idx].count != current_region_info.count) {
-                if (do_learning(current_region_info.hash, current_region_info.hint)) {
+                if ((current_region_info.hash != GEOPM_REGION_HASH_UNMARKED &&
```
cmcantalupo: is this special case for unmarked an optimization?  I don't think this check is needed.

src/EnergyEfficientAgent.cpp
```diff
@@ -242,11 +241,13 @@ namespace geopm
                 }
                 /// update previous region (exit)
                 struct m_region_info_s last_region_info = m_last_region_info[ctl_idx];
-                if (do_learning(last_region_info.hash, last_region_info.hint)) {
+                if ((last_region_info.hash != GEOPM_REGION_HASH_UNMARKED &&
```
cmcantalupo: Same as above, why check for unmarked here?

cmcantalupo (changes requested):

src/EnergyEfficientAgent.cpp
bakerbrandond: This hash check was preserved from the deleted do_learning.  I kept it because otherwise this would be a behavior change as we don't want to learn for the unmarked buckets of time

src/EnergyEfficientAgent.cpp
bakerbrandond: done

src/ApplicationSampler.hpp
bakerbrandond: done

src/EnergyEfficientAgent.cpp
```diff
@@ -190,14 +192,10 @@ namespace geopm
...
             uint64_t hash = m_last_region_info[ctl_idx].hash;
-            uint64_t hint = m_last_region_info[ctl_idx].hint;
             if (GEOPM_REGION_HASH_UNMARKED == hash) {
                 m_target_freq[ctl_idx] = m_freq_governor->get_frequency_max();
             }
-            else if (GEOPM_REGION_HINT_IGNORE == hint) {
-                m_target_freq[ctl_idx] = m_freq_governor->get_frequency_max();
-            }
-            else if (GEOPM_REGION_HINT_NETWORK == hint) {
+            else if (!(m_network_hash.find(hash) == m_network_hash.end())) {
```
cmcantalupo: Might be more readable as:

```
else if (m_network_hash.find(hash) != m_network_hash.end()) {
```

src/EnergyEfficientAgent.cpp
cmcantalupo: Okay thanks, makes sense.  When I suggested getting rid of the `do_learning()` interface, I did not realize that the check for unmarked was included in the function in addition to the find() result.  A different option to the do_learning() API would be  to have two sets of hashes one with the list of all low frequency regions, and another for all regions where learning should not occur.  The second set would contain the first.  I think this would be clearer when reading the code.  That is to say there would be two sets of uint64_t as members:
```
m_low_freq_hash = ApplicationSampler::region_hash_network();
m_no_learning_hash = m_low_freq_hash;
m_no_learning_hash.insert(GEOPM_REGION_HASH_UNMARKED);
```

src/ApplicationSampler.cpp
```diff
@@ -53,6 +54,49 @@ namespace geopm
...
+                                             "MPI_Waitany",
+                                             "MPI_Wait",
+                                             "MPI_Waitsome",
+                                             "MPI_Exscan",
+                                             "MPI_Recv",
+                                             "MPI_Send",
+                                             "MPI_Sendrecv",
+                                             "MPI_Sendrecv_replace",
+                                             "MPI_Ssend"};
+        for (auto const &func_name : network_funcs) {
```
cmcantalupo, line 101: There are many simple ways to avoid hashing these values on every call.

cmcantalupo: https://github.com/cmcantalupo/geopm/commit/05db103194c97356c91208298e6225d6d632d094

cmcantalupo (approved):

PR #1229, opened 2020-08-06
Initial updates for board accelerator support

lhlawson:
This PR is intended to provide basic support of the NVIDIA Management Library for monitoring and controlling NVIDIA accelerators.  It is primarily focused on discovering and enumerating the NVML devices and reporting their topology/relationship to other components in the system.  Beyond this it includes API calls in the NVMLDevicePool to support the control and monitoring of NVML Devices.

Tested with icpc (ICC) 19.0.5.281 20190815 and CUDA Toolkit version 10.2.

Summary of change.
- Addition of an AcceleratorTopo class to provide static topology information as required by PlatformTopo
- Addition of an NVMLAcceleratorTopo implementation of the AcceleratorTopo class to provide the same for NVML devices
- Addition of an NVMLDevicePool class to act as a thin wrapper around NVML API calls of interest
- Modification of PlatformTopo to use the AcceleratorTopo in all GEOPM_DOMAIN_BOARD_ACCELERATOR cases
- Addition of Unit tests to test and mock the basic features and known corner cases of the above changes.
- Modification of the configure.ac to provide --enable-nvml and --with-libnvml options for enabling the NVML accelerator classes

Summary of additional changes in response to review.
- Modification of NVMLDevicePool to take number of cpus as an input
- Addition of MockAcceleratorTopo
- Modification of PlatformTopo to take AcceleratorTopo in construction for testing with MockAcceleratorTopo

High level overview:
image

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -134,6 +134,19 @@ fi
 [enable_procfs="1"]
 )
 
+AC_ARG_ENABLE([nvml],
+  [AS_HELP_STRING([--enable-nvml], [Enable GEOPM to use the NVIDIA NVML driver to support NVIDIA board accelerators])],
+[if test "x$enable_nvml" = "xno" ; then
+  enable_nvml="0"
+else
+  enable_nvml="1"
+fi
+],
+[enable_nvml="0"]
+)
+AC_SUBST([enable_nvml])
+AM_CONDITIONAL([ENABLE_NVML], [test "x$enable_nvml" = "x1"])
+
 AC_ARG_ENABLE([cnl-iogroup],
   [AS_HELP_STRING([--enable-cnl-iogroup], [Enable the CNL IOGroup])],
 [if test "x$enable_cnl_iogroup" = "xno" ; then
@@ -175,6 +188,14 @@ if test "x$enable_msrsafe_ioctl_write" = "x1" ; then
   AC_DEFINE([GEOPM_MSRSAFE_IOCTL_WRITE], [ ], [Enables use of msr-safe ioctl feature for writing (broken as of msr-safe version 1.2.0)])
 fi
 AC_SUBST([enable_msrsafe_ioctl_write])
+
+if test "x$enable_nvml" = "x1" ; then
+  ENABLE_NVML=True
+  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of NVIDIA NVML driver to support NVIDIA Board acellerators])
+else
+  ENABLE_NVML=False
+fi
+AC_SUBST([ENABLE_NVML])
 
 AC_ARG_WITH([sqlite3], [AS_HELP_STRING([--with-sqlite3=PATH],
             [specify directory for installed sqlite3 package.])])
@@ -269,6 +290,14 @@ fi
 ],
 [enable_ronn="1"]
 )
+
+AC_ARG_WITH([libnvml], [AS_HELP_STRING([--with-libnvml=PATH],
+            [specify directory for installed libnvml package.])])
+if test "x$with_libnvml" != x; then
+  EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -I$with_libnvml/include"
+  LD_LIBRARY_PATH="$with_libnvml/lib:$LD_LIBRARY_PATH"
+  EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_libnvml/lib"
+fi
 
 AC_SUBST([enable_ronn])
 AM_CONDITIONAL([ENABLE_RONN], [test "x$enable_ronn" = "x1"])
@@ -466,6 +495,16 @@ AC_CHECK_HEADER([gelf.h], [], [
     echo "missing gelf.h: ELF library is required, use --with-libelf or --with-libelf-include to specify location"
     exit -1])
 
+if test "x$enable_nvml" = "x1" ; then
+    AC_SEARCH_LIBS([nvmlSystemGetCudaDriverVersion], [nvidia-ml])
+    AC_CHECK_LIB([nvidia-ml], [nvmlSystemGetCudaDriverVersion], [], [
+        echo "missing libnvidia-ml: NVML library is required, use --with-libnvml to specify location"
+        exit -1])
+    AC_CHECK_HEADER([nvml.h], [], [
+        echo "missing nvml.h: NVML header is required, use --with-libnvml to specify location"
+        exit -1])
+fi
+
 # Checks for header files.
 AC_CHECK_HEADERS([arpa/inet.h fcntl.h float.h inttypes.h malloc.h netdb.h stddef.h stdint.h stdlib.h string.h sys/socket.h unistd.h])
 
@@ -647,4 +686,5 @@ AC_MSG_RESULT([ronn               : ${enable_ronn}])
 AC_MSG_RESULT([ompt               : ${enable_ompt}])
 AC_MSG_RESULT([beta               : ${enable_beta}])
 AC_MSG_RESULT([bloat              : ${enable_bloat}])
+AC_MSG_RESULT([nvml               : ${enable_nvml}])
 AC_MSG_RESULT([===============================================================================])
--- a/configure.ac
+++ b/configure.ac
@@ -135,7 +135,7 @@ fi
 )
 
 AC_ARG_ENABLE([nvml],
-  [AS_HELP_STRING([--enable-nvml], [Enables use of the NVML driver to support NVML board accelerators])],
+  [AS_HELP_STRING([--enable-nvml], [Enables use of the NVML library to support NVML board accelerators])],
 [if test "x$enable_nvml" = "xno" ; then
   enable_nvml="0"
 else
@@ -191,11 +191,10 @@ AC_SUBST([enable_msrsafe_ioctl_write])
 
 if test "x$enable_nvml" = "x1" ; then
   ENABLE_NVML=True
-  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of the NVML driver to support NVML board accelerators])
+  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of the NVML library to support NVML board accelerators])
 else
   ENABLE_NVML=False
 fi
-AC_SUBST([ENABLE_NVML])
 
 AC_ARG_WITH([sqlite3], [AS_HELP_STRING([--with-sqlite3=PATH],
             [specify directory for installed sqlite3 package.])])
--- a/configure.ac
+++ b/configure.ac
@@ -135,7 +135,7 @@ fi
 )
 
 AC_ARG_ENABLE([nvml],
-  [AS_HELP_STRING([--enable-nvml], [Enable GEOPM to use the NVIDIA NVML driver to support NVIDIA board accelerators])],
+  [AS_HELP_STRING([--enable-nvml], [Enables use of the NVML driver to support NVML board accelerators])],
 [if test "x$enable_nvml" = "xno" ; then
   enable_nvml="0"
 else
@@ -191,7 +191,7 @@ AC_SUBST([enable_msrsafe_ioctl_write])
 
 if test "x$enable_nvml" = "x1" ; then
   ENABLE_NVML=True
-  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of NVIDIA NVML driver to support NVIDIA Board acellerators])
+  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of the NVML driver to support NVML board accelerators])
 else
   ENABLE_NVML=False
 fi
--- a/src/AcceleratorTopo.cpp
+++ b/src/AcceleratorTopo.cpp
@@ -56,7 +56,7 @@ namespace geopm
         return 0;
     }
 
-    std::set<int> AcceleratorTopo::ideal_cpu_affinitization(int domain_idx) const
+    std::set<int> AcceleratorTopo::cpu_affinity_ideal(int domain_idx) const
     {
         return {};
     }
--- a/src/AcceleratorTopo.cpp
+++ b/src/AcceleratorTopo.cpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include "config.h"
+#include "Exception.hpp"
+#include "NVMLAcceleratorTopo.hpp"
+
+namespace geopm
+{
+
+    const AcceleratorTopo &accelerator_topo(void)
+    {
+#ifdef GEOPM_ENABLE_NVML
+        static NVMLAcceleratorTopo instance;
+#else
+        static AcceleratorTopo instance;
+#endif
+        return instance;
+    }
+
+    int AcceleratorTopo::num_accelerator() const
+    {
+        return 0;
+    }
+
+    std::set<int> AcceleratorTopo::ideal_cpu_affinitization(int domain_idx) const
+    {
+        return {};
+    }
+}
--- a/src/AcceleratorTopo.cpp
+++ b/src/AcceleratorTopo.cpp
@@ -41,7 +41,6 @@
 
 namespace geopm
 {
-
     const AcceleratorTopo &accelerator_topo(void)
     {
 #ifdef GEOPM_ENABLE_NVML
@@ -52,7 +51,7 @@ namespace geopm
         return instance;
     }
 
-    int AcceleratorTopo::num_accelerator() const
+    int AcceleratorTopo::num_accelerator(void) const
     {
         return 0;
     }
--- a/src/AcceleratorTopo.hpp
+++ b/src/AcceleratorTopo.hpp
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef ACCELERATORTOPO_HPP_INCLUDE
+#define ACCELERATORTOPO_HPP_INCLUDE
+
+#include <cstdint>
+
+namespace geopm
+{
+    class AcceleratorTopo
+    {
+        public:
+            AcceleratorTopo() = default;
+            virtual ~AcceleratorTopo() = default;
+            /// @brief Number of accelerators on the platform.
+            virtual int num_accelerator() const;
+            /// @brief CPU Affinitization set for a particular accelerator
+            /// @param [in] domain_idx The index indicating a particular
+            ///        accelerator
+            virtual std::set<int> ideal_cpu_affinitization(int domain_idx) const;
+
+    };
+
+    const AcceleratorTopo &accelerator_topo(void);
+}
+#endif
--- a/src/AcceleratorTopo.hpp
+++ b/src/AcceleratorTopo.hpp
@@ -47,7 +47,7 @@ namespace geopm
             /// @brief CPU Affinitization set for a particular accelerator
             /// @param [in] domain_idx The index indicating a particular
             ///        accelerator
-            virtual std::set<int> ideal_cpu_affinitization(int domain_idx) const;
+            virtual std::set<int> cpu_affinity_ideal(int domain_idx) const;
     };
 
     const AcceleratorTopo &accelerator_topo(void);
--- a/src/AcceleratorTopo.hpp
+++ b/src/AcceleratorTopo.hpp
@@ -43,12 +43,11 @@ namespace geopm
             AcceleratorTopo() = default;
             virtual ~AcceleratorTopo() = default;
             /// @brief Number of accelerators on the platform.
-            virtual int num_accelerator() const;
+            virtual int num_accelerator(void) const;
             /// @brief CPU Affinitization set for a particular accelerator
             /// @param [in] domain_idx The index indicating a particular
             ///        accelerator
             virtual std::set<int> ideal_cpu_affinitization(int domain_idx) const;
-
     };
 
     const AcceleratorTopo &accelerator_topo(void);
--- a/src/NVMLAcceleratorTopo.cpp
+++ b/src/NVMLAcceleratorTopo.cpp
@@ -50,9 +50,8 @@ namespace geopm
 
     NVMLAcceleratorTopo::NVMLAcceleratorTopo(const NVMLDevicePool &device_pool, const int num_cpu)
         : m_nvml_device_pool(device_pool)
+        , m_num_accelerator(m_nvml_device_pool.num_accelerator())
     {
-        m_num_accelerator = m_nvml_device_pool.num_accelerator();
-
         if (m_num_accelerator == 0) {
             std::cerr << "Warning: <geopm> NVMLAcceleratorTopo: No NVML accelerators detected.\n";
         }
@@ -63,11 +62,11 @@ namespace geopm
 
             CPU_ZERO_S(CPU_ALLOC_SIZE(num_cpu), affinitized_cpuset);
 
-            m_ideal_cpu_affinitization.resize(m_num_accelerator);
+            m_cpu_affinity_ideal.resize(m_num_accelerator);
 
             // Cache ideal affinitization due to the overhead associated with the NVML calls
             for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
-                ideal_affinitization_mask_vec.push_back(m_nvml_device_pool.ideal_cpu_affinitization_mask(accel_idx));
+                ideal_affinitization_mask_vec.push_back(m_nvml_device_pool.cpu_affinity_ideal_mask(accel_idx));
             }
 
             /// @todo: As an optimization this may be replacable with CPU_OR of all masks in ideal_affinitzation_mask_vec
@@ -99,10 +98,12 @@ namespace geopm
                 // for the accelerator with the fewest possible CPUs in the accelerator mask
                 for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
                     unsigned int accelerator_cpu_count = 0;
-                    int cpu_idx = 0;
-                    while (cpu_idx < num_cpu && accelerator_cpu_count < num_cpu_per_accelerator) {
+                    for (int cpu_idx = 0;
+                         cpu_idx != num_cpu &&
+                         accelerator_cpu_count < num_cpu_per_accelerator;
+                         ++cpu_idx) {
                         if (CPU_ISSET(cpu_idx, ideal_affinitization_mask_vec.at(accel_idx))) {
-                            m_ideal_cpu_affinitization.at(accel_idx).insert(cpu_idx);
+                            m_cpu_affinity_ideal.at(accel_idx).insert(cpu_idx);
                             --cpu_remaining;
                             ++accelerator_cpu_count;
 
@@ -111,7 +112,6 @@ namespace geopm
                                 CPU_CLR(cpu_idx, ideal_affinitization_mask_vec.at(accel_idx_inner));
                             }
                         }
-                        ++cpu_idx;
                     }
                 }
             }
@@ -130,13 +130,13 @@ namespace geopm
         return m_num_accelerator;
     }
 
-    std::set<int> NVMLAcceleratorTopo::ideal_cpu_affinitization(int accel_idx) const
+    std::set<int> NVMLAcceleratorTopo::cpu_affinity_ideal(int accel_idx) const
     {
         if (accel_idx < 0 || accel_idx >= m_num_accelerator) {
             throw Exception("NVMLAcceleratorTopo::" + std::string(__func__) + ": accel_idx " +
                             std::to_string(accel_idx) + " is out of range",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        return m_ideal_cpu_affinitization.at(accel_idx);
+        return m_cpu_affinity_ideal.at(accel_idx);
     }
 }
--- a/src/NVMLAcceleratorTopo.cpp
+++ b/src/NVMLAcceleratorTopo.cpp
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <map>
+
+#include "config.h"
+#include "Exception.hpp"
+#include "NVMLDevicePool.hpp"
+#include "NVMLAcceleratorTopo.hpp"
+
+namespace geopm
+{
+
+    NVMLAcceleratorTopo::NVMLAcceleratorTopo()
+                                                  // TODO: can this be extracted from platform topo
+        : NVMLAcceleratorTopo(nvml_device_pool(), geopm_sched_num_cpu())
+    {
+    }
+
+    NVMLAcceleratorTopo::NVMLAcceleratorTopo(const NVMLDevicePool &device_pool, const int num_cpu)
+        : m_nvml_device_pool(device_pool)
+    {
+        m_num_accelerator = m_nvml_device_pool.num_accelerator();
+
+        if (m_num_accelerator == 0) {
+            std::cerr << "Warning: <geopm> NVMLAcceleratorTopo: No NVIDIA accelerators detected.\n";
+        }
+        else {
+            int cpu_remaining = 0;
+            std::vector<cpu_set_t *> ideal_affinitization_mask_vec;
+            cpu_set_t *affinitized_cpuset = NULL;
+
+            affinitized_cpuset = CPU_ALLOC(num_cpu);
+            CPU_ZERO_S(CPU_ALLOC_SIZE(num_cpu), affinitized_cpuset);
+
+            m_ideal_cpu_affinitization.resize(m_num_accelerator);
+
+            // Cache ideal affinitization due to the overhead associated with the NVML calls
+            for(int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
+                ideal_affinitization_mask_vec.push_back(m_nvml_device_pool.ideal_cpu_affinitization_mask(accel_idx));
+            }
+
+            // TODO: As an optimization this may be replacable with CPU_OR of all masks in ideal_affinitzation_mask_vec
+            //       and CPU_COUNT of the output
+            for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
+                for (int cpu_idx = 0; cpu_idx < num_cpu; cpu_idx++) {
+                    if (CPU_ISSET(cpu_idx, ideal_affinitization_mask_vec.at(accel_idx))) {
+                        if (CPU_ISSET(cpu_idx, affinitized_cpuset) == 0) {
+                            //if this is in this accelerator mask and has not
+                            //been picked by another accelerator
+                            CPU_SET(cpu_idx, affinitized_cpuset);
+                            ++cpu_remaining;
+                        }
+                    }
+                }
+            }
+
+            // In order to handle systems where the number of CPUs are not evenly divisble by the number of GPUs 
+            // a two pass process is used.  This does not guarantee affinitization is successful, but covers many
+            // common cases
+            for(int affinitization_attempts = 0; affinitization_attempts < 2; ++affinitization_attempts) {
+                unsigned int num_cpu_per_accelerator = cpu_remaining / m_num_accelerator;
+                if (num_cpu_per_accelerator == 0) {
+                    num_cpu_per_accelerator = cpu_remaining % m_num_accelerator;
+                }
+
+                // This is a greedy approach for mapping CPUs to accelerators, and as such may result in some CPUs
+                // not being affinitized at all.  A potential improvement is to always determine affinity
+                // for the accelerator with the fewest possible CPUs in the accelerator mask
+                for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
+                    unsigned int accelerator_cpu_count = 0;
+
+                    int cpu_idx = 0;
+                    while (cpu_idx < num_cpu && accelerator_cpu_count < num_cpu_per_accelerator) {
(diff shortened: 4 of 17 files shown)


configure.ac
```diff
@@ -270,6 +291,17 @@ fi
 [enable_ronn="1"]
 )
 
+AC_SUBST([enable_doc])
+AM_CONDITIONAL([ENABLE_DOC], [test "x$enable_doc" = "x1"])
```
bgeltz: This is leftover old code.  It must be from a bad rebase.  I think it can be removed.

README.md
```diff
@@ -1 +0,0 @@
-README
```
bgeltz: Not sure what happened, but this should remain how it was.

```README.md``` is a symlink to ```README```.  All ```README``` changes should happen in ```README```, not ```README.md```.

File: src/AcceleratorTopo.hpp [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+    {
+        public:
+            AcceleratorTopo() = default;
+            virtual ~AcceleratorTopo() = default;
+            /// @brief Number of accelerators on the platform.
+            virtual int num_accelerator() const;
+            /// @brief CPU Affinitization set for a particular accelerator
+            /// @param [in] domain_idx The index indicating a particular
+            ///        accelerator
+            virtual std::set<int> ideal_cpu_affinitization(int domain_idx) const;
```
bgeltz: I expected to see ```cpu_set_t```'s here for the return type.  Why ```std::set<int>```?

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "Exception.hpp"
+#include "NVMLDevicePool.hpp"
+#include "NVMLAcceleratorTopo.hpp"
+
+namespace geopm
+{
+
+    NVMLAcceleratorTopo::NVMLAcceleratorTopo()
+                                                  // TODO: can this be extracted from platform topo
+        : NVMLAcceleratorTopo(nvml_device_pool(), geopm_sched_num_cpu())
```
bgeltz: We could in theory pass the number of CPUs from PlatformTopo through the constructor instead of calling ```geopm_sched_num_cpu()```.  Might want to discuss this in a live group review.

src/AcceleratorTopo.hpp
bgeltz: Why not vector of cpu_sets?

src/NVMLAcceleratorTopo.hpp
```diff
@@ -0,0 +1,61 @@
...
+    class NVMLDevicePool;
+
+    class NVMLAcceleratorTopo : public AcceleratorTopo
+    {
+        public:
+            NVMLAcceleratorTopo();
+            NVMLAcceleratorTopo(const NVMLDevicePool &device_pool, const int num_cpu);
+            virtual ~NVMLAcceleratorTopo() = default;
+            virtual int num_accelerator() const override;
+            virtual std::set<int> ideal_cpu_affinitization(int accel_idx) const override;
```
bgeltz: std::set<cpu_set_t>?

src/NVMLAcceleratorTopo.hpp
```diff
@@ -0,0 +1,61 @@
...
+        private:
+            const NVMLDevicePool &m_nvml_device_pool;
+
+            std::vector<std::set<int> > m_ideal_cpu_affinitization;
```
bgeltz: Why not std::vector<cpu_set_t>?

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+                if (num_cpu_per_accelerator == 0) {
+                    num_cpu_per_accelerator = cpu_remaining % m_num_accelerator;
+                }
+
+                // This is a greedy approach for mapping CPUs to accelerators, and as such may result in some CPUs
+                // not being affinitized at all.  A potential improvement is to always determine affinity
+                // for the accelerator with the fewest possible CPUs in the accelerator mask
+                for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
+                    unsigned int accelerator_cpu_count = 0;
+
```
bgeltz: rm wsp

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            if(cpu_remaining != 0){
+                throw Exception("NVMLAcceleratorTopo::" + std::string(__func__) +
+                                ": Failed to affinitize all valid CPUs to Accelerators.  " +
+                                std::to_string(cpu_remaining) + " CPUs remain unassociated with any accelerator.",
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+
```
bgeltz: rm wsp

src/NVMLDevicePoolImp.hpp
```diff
@@ -0,0 +1,76 @@
...
+            virtual void frequency_reset_control(int accel_idx) const override;
+            virtual void power_control(int accel_idx, int setting) const override;
+
+        private:
+            virtual void check_accel_range(int accel_idx) const;
+
+            unsigned int m_num_accelerator;
+            std::vector<nvmlDevice_t> m_nvml_device;
+
+            unsigned int m_max_contexts;
```
bgeltz: Make this a ```const unsigned int``` or a ```const size_t```.

Also, any bonafide const vars should be all caps: M_MAX_CONTEXTS.

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    const NVMLDevicePool &nvml_device_pool(void)
+    {
+        static NVMLDevicePoolImp instance;
+        return instance;
+    }
+
+    NVMLDevicePoolImp::NVMLDevicePoolImp()
+    {
+        m_max_contexts = 64;
```
bgeltz: See the comment in the header.  Make this a const, and set it in the initializer list in the constructor.

See ```M_PRECISION``` in the FrequencyMapAgent.[ch]pp for more details.

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+        if (nvml_result != NVML_SUCCESS) {
+            throw Exception("NVMLDevicePool::" + std::string(__func__) +
+                            ": NVML failed to query device count.  Error: " +
+                            nvmlErrorString(nvml_result),
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        //Acquire device handle for each NVIDIA Accelerator
+        m_nvml_device.resize(m_num_accelerator);
+        for (unsigned int accel_idx = 0; accel_idx < m_num_accelerator; ++accel_idx){
```
bgeltz: Open curly for the body of the for loop should have a space between it and the closing paren for the loop criteria:
```...++accel_idx) {```

Please check your other loops.

File: src/NVMLDevicePool.cpp [hunk 12]
```diff
@@ -0,0 +1,465 @@
...
+        }
+
+        return result;
+    }
+
+    uint64_t NVMLDevicePoolImp::throttle_reasons(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        uint64_t result = 0;
+        unsigned long long temp_ull;
```
bgeltz: Defining this temp is really necessary?

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+            else {
+                throw Exception("NVMLDevicePool::" + std::string(__func__) +
+                                ": NVML failed to acquire running processes for accelerator " +
+                                std::to_string(accel_idx) + ".  Error: " + nvmlErrorString(nvml_result),
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+        return result;
+    }
+
```
bgeltz, line 334: rm \n

test/Makefile.mk
```diff
@@ -476,6 +477,20 @@ if ENABLE_OMPT
...
+                   test/gtest_links/NVMLAcceleratorTopoTest.no_gpu_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.mutex_affinitization_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.equidistant_affinitization_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.n1_superset_n_affinitization_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.greedbuster_affinitization_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.hpe_6500_affinitization_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.uneven_affinitization_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.high_cpu_count_config \
+                   test/gtest_links/NVMLAcceleratorTopoTest.high_cpu_count_gaps_config \
+                   # end
```
bgeltz, line 491: Nice!  Lots of tests.

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+{
+    int num_accelerator = 0;
+    int num_cpu = 40;
+
+    EXPECT_CALL(*m_device_pool, num_accelerator()).WillOnce(Return(num_accelerator));
+
+    NVMLAcceleratorTopo topo(*m_device_pool, num_cpu);
+    EXPECT_EQ(0, topo.num_accelerator());
+
+    EXPECT_THROW(topo.ideal_cpu_affinitization(0), geopm::Exception);
```
bgeltz: If you want to get fancy, we have ``` GEOPM_EXPECT_THROW_MESSAGE``` which allows you to look for a particular substring in the exception text.  It's more specific than just catching any geopm::Exception that occurs here.

There's a ton of examples of usage in the test/ dir.

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    for (int accel_idx = 0; accel_idx < topo.num_accelerator(); ++accel_idx) {
+        ASSERT_THAT(topo.ideal_cpu_affinitization(accel_idx), cpus_allowed_set[accel_idx]);
+    }
+}
+
+//Test case: CPU count that is not evenly divisible by the accelerator count
+TEST_F(NVMLAcceleratorTopoTest, uneven_affinitization_config)
+{
+
```
bgeltz: rm wsp

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+    accel_bitmask[7][1] = 0xffffffffffffffff;
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        EXPECT_CALL(*m_device_pool, ideal_cpu_affinitization_mask(accel_idx)).WillOnce(Return((cpu_set_t *)&accel_bitmask[accel_idx]));
+    }
+    EXPECT_CALL(*m_device_pool, num_accelerator()).WillOnce(Return(num_accelerator));
+
+    NVMLAcceleratorTopo topo(*m_device_pool, num_cpu);
+
+    EXPECT_EQ(8, topo.num_accelerator());
```
bgeltz: You could just say ```EXPECT_EQ(num_acelerator, topo.num_accelerator());```.  No need to hardcode the 8.

test/PlatformTopoTest.cpp
```diff
@@ -566,6 +582,15 @@ TEST_F(PlatformTopoTest, bdx_domain_nested)
                                         GEOPM_DOMAIN_BOARD, 0);
     EXPECT_EQ(idx_set_expect, idx_set_actual);
 
+    // Board Accelerator
+#ifdef GEOPM_ENABLE_NVML
+    EXPECT_NO_THROW(topo.domain_nested(GEOPM_DOMAIN_CPU,
+                                       GEOPM_DOMAIN_BOARD_ACCELERATOR, 0));
+#else
+    EXPECT_THROW(topo.domain_nested(GEOPM_DOMAIN_CPU,
```
bgeltz: Could also consider using ```GEOPM_EXPECT_THROW_MESSAGE``` here too.

bgeltz (changes requested):
Lots of nits and a few real changes requested.

I think if everyone gives this a once-over, then we do a live session for final review this should go quick.

configure.ac
lhlawson: Agreed, this isn't an intentional inclusion.

README.md
lhlawson: This definitely wasn't an intended modification.

src/AcceleratorTopo.hpp
lhlawson: PlatformTopo domain_cpus calls for a std::set<int>, this is a translation function that provides this format based upon the parsing of the NVML cpu_set parsing that is done when the NVMLAcceleratorTopo is instantiated.  This was meant to help encapsulate the NVML specific affinitization (or any other accelerator type approach later on) in *AcceleratorTopo vs having per accelerator type mapping handling in PlatformTopo

src/NVMLAcceleratorTopo.cpp
lhlawson: Either approach seems reasonable to me.  There is also the possibility of modifying the code to solely operate on number of affinitized CPUs, which is provided by the NVMLDevicePool (which calls geopm_sched_num_cpu() as well)

src/NVMLAcceleratorTopo.hpp
lhlawson: PlatformTopo domain_cpus calls for a std::set<int>, this is a translation function that provides this format based upon the parsing of the NVML cpu_set parsing that is done when the NVMLAcceleratorTopo is instantiated.  This was meant to help encapsulate the NVML specific affinitization (or any other accelerator type approach later on) in *AcceleratorTopo vs having per accelerator type mapping handling in PlatformTopo

src/NVMLAcceleratorTopo.hpp
lhlawson: PlatformTopo domain_cpus calls for a std::set<int>, this is a translation function that provides this format based upon the parsing of the NVML cpu_set parsing that is done when the NVMLAcceleratorTopo is instantiated.  This was meant to help encapsulate the NVML specific affinitization (or any other accelerator type approach later on) in *AcceleratorTopo vs having per accelerator type mapping handling in PlatformTopo

src/NVMLDevicePoolImp.hpp
lhlawson: Original intent was to update m_max_contexts if the call to  nvmlDeviceGetComputeRunningProcesses() returned NVML_ERROR_INSUFFICIENT_SIZE.  This functionality is not currently present, so whether this should be a member var at all is also questionable.

configure.ac
```diff
@@ -176,6 +189,14 @@ if test "x$enable_msrsafe_ioctl_write" = "x1" ; then
 fi
 AC_SUBST([enable_msrsafe_ioctl_write])
 
+if test "x$enable_nvml" = "x1" ; then
+  ENABLE_NVML=True
+  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of NVIDIA NVML driver to support NVIDIA Board acellerators])
```
dianarg: accelerators

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+    {
+    }
+
+    NVMLAcceleratorTopo::NVMLAcceleratorTopo(const NVMLDevicePool &device_pool, const int num_cpu)
+        : m_nvml_device_pool(device_pool)
+    {
+        m_num_accelerator = m_nvml_device_pool.num_accelerator();
+
+        if (m_num_accelerator == 0) {
+            std::cerr << "Warning: <geopm> NVMLAcceleratorTopo: No NVIDIA accelerators detected.\n";
```
dianarg: Not sure if we need to be careful about other company's trademarks in code

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+        nvmlReturn_t nvml_result;
+
+        //Initialize NVML
+        nvml_result = nvmlInit();
+        if (nvml_result != NVML_SUCCESS) {
+            throw Exception("NVMLDevicePool::" + std::string(__func__) +
+                            ": NVML failed to initialize.  Error: " +
+                            nvmlErrorString(nvml_result),
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
```
dianarg: Might want a helper function for these checks

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    cpu_set_t *NVMLDevicePoolImp::ideal_cpu_affinitization_mask(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+
+        cpu_set_t *accel_cpuset = NULL;
+        //TODO: can this be extracted from platform topo
+        const int num_cpu = geopm_sched_num_cpu();
```
dianarg: Agreed, and inject in constructor

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+        nvml_result = nvmlDeviceGetClock(m_nvml_device.at(accel_idx), NVML_CLOCK_SM, NVML_CLOCK_ID_CURRENT,
+                                         &temp);
+        if (nvml_result == NVML_SUCCESS) {
+            result = (uint64_t)temp;
+        }
+        else {
+            throw Exception("NVMLDevicePool::" + std::string(__func__) +
+                            ": NVML failed to get SM Frequency for accelerator " +
+                            std::to_string(accel_idx) + ".  Error: " + nvmlErrorString(nvml_result),
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
dianarg: same helper function to check NVML_SUCCESS

dianarg: more below

src/NVMLDevicePool.cpp
dianarg: If you reorder the if to put the exception first, can get rid of `result` and return the cast temp_ull directly

src/PlatformTopo.cpp
```diff
@@ -313,6 +330,11 @@ namespace geopm
             // To support mapping CPU signals to DRAM domain (e.g. power)
             result = true;
         }
+        else if (outer_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR &&
+                 inner_domain == GEOPM_DOMAIN_CPU) {
+            // To support mapping CPU signals to ACCELERATOR domain (e.g. power)
```
dianarg: is "(e.g. power)" relevant here?

test/AcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,72 @@
...
+        void TearDown();
+};
+
+void AcceleratorTopoTest::SetUp()
+{
+}
+
+void AcceleratorTopoTest::TearDown()
+{
+}
```
dianarg: You can remove all of this and use TEST() instead of TEST_F() if you don't need the class for anything

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+    accel_bitmask[3] = 0xfffff00000;
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        EXPECT_CALL(*m_device_pool, ideal_cpu_affinitization_mask(accel_idx)).WillOnce(Return((cpu_set_t *) &accel_bitmask[accel_idx]));
+    }
+
+    EXPECT_CALL(*m_device_pool, num_accelerator()).WillOnce(Return(num_accelerator));
+
+    NVMLAcceleratorTopo topo(*m_device_pool, num_cpu);
+    EXPECT_EQ(4, topo.num_accelerator());
```
dianarg: these checks could use num_accelerator variable from above; more below

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+    NVMLAcceleratorTopo topo(*m_device_pool, num_cpu);
+    EXPECT_EQ(0, topo.num_accelerator());
+
+    EXPECT_THROW(topo.ideal_cpu_affinitization(0), geopm::Exception);
+}
+
+//Test case: The HPE SX40 default system configuration
+TEST_F(NVMLAcceleratorTopoTest, hpe_sx40_default_config)
+{
+    int num_accelerator = 4;
```
dianarg: const

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+    accel_bitmask[3] = 0x00000003ff;
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        EXPECT_CALL(*m_device_pool, ideal_cpu_affinitization_mask(accel_idx)).WillOnce(Return((cpu_set_t *)&accel_bitmask[accel_idx]));
+    }
+    EXPECT_CALL(*m_device_pool, num_accelerator()).WillOnce(Return(num_accelerator));
+
+    EXPECT_THROW(NVMLAcceleratorTopo topo(*m_device_pool, num_cpu), geopm::Exception);
+
+    //TODO: If we move away from the greedy affinitization approach this is the expected result.
```
dianarg: commented code.  Just test for current behavior.  you can also open an issue to change this behavior later.

test/NVMLAcceleratorTopoTest.cpp
dianarg: more below

test/NVMLAcceleratorTopoTest.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        EXPECT_CALL(*m_device_pool, ideal_cpu_affinitization_mask(accel_idx)).WillOnce(Return((cpu_set_t *)&accel_bitmask[accel_idx]));
+    }
+    EXPECT_CALL(*m_device_pool, num_accelerator()).WillOnce(Return(num_accelerator));
+
+    NVMLAcceleratorTopo topo(*m_device_pool, num_cpu);
+
+    EXPECT_EQ(8, topo.num_accelerator());
+    std::set<int> cpus_allowed_set[topo.num_accelerator()];
```
dianarg: num_accelerator instead of function call

test/PlatformTopoTest.cpp
```diff
@@ -267,7 +268,11 @@ TEST_F(PlatformTopoTest, hsw_num_domain)
     /// @todo when implemented, add tests for each platform
     EXPECT_EQ(0, topo.num_domain(GEOPM_DOMAIN_BOARD_NIC));
     EXPECT_EQ(0, topo.num_domain(GEOPM_DOMAIN_PACKAGE_NIC));
+#ifdef GEOPM_ENABLE_NVML
+    EXPECT_GE(topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR), 0);
```
dianarg: is GE intended here?  I would expect this to always return 0 anyway without any additional mocks

test/PlatformTopoTest.cpp
```diff
@@ -376,10 +386,16 @@ TEST_F(PlatformTopoTest, bdx_domain_idx)
     for (auto cpu_idx : cpu_set_node1) {
         EXPECT_EQ(1, topo.domain_idx(GEOPM_DOMAIN_BOARD_MEMORY, cpu_idx));
     }
+
+#ifdef GEOPM_ENABLE_NVML
+    EXPECT_GE(topo.domain_idx(GEOPM_DOMAIN_BOARD_ACCELERATOR, 0), -1);
```
dianarg: is GE intended?

src/AcceleratorTopo.hpp
bgeltz: OK.

src/AcceleratorTopo.hpp
```diff
@@ -0,0 +1,56 @@
...
+
```
bgeltz: wsp

src/AcceleratorTopo.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <sstream>
+#include <string>
+
+#include "config.h"
+#include "Exception.hpp"
+#include "NVMLAcceleratorTopo.hpp"
+
+namespace geopm
+{
+
```
bgeltz: wsp

src/AcceleratorTopo.cpp
```diff
@@ -0,0 +1,64 @@
...
+    {
+#ifdef GEOPM_ENABLE_NVML
+        static NVMLAcceleratorTopo instance;
+#else
+        static AcceleratorTopo instance;
+#endif
+        return instance;
+    }
+
+    int AcceleratorTopo::num_accelerator() const
```
bgeltz: ```int AcceleratorTopo::num_accelerator(void) const```
Also in the header.

src/NVMLAcceleratorTopo.hpp
```diff
@@ -0,0 +1,61 @@
...
+    {
+        public:
+            NVMLAcceleratorTopo();
+            NVMLAcceleratorTopo(const NVMLDevicePool &device_pool, const int num_cpu);
+            virtual ~NVMLAcceleratorTopo() = default;
+            virtual int num_accelerator() const override;
+            virtual std::set<int> ideal_cpu_affinitization(int accel_idx) const override;
+        private:
+            const NVMLDevicePool &m_nvml_device_pool;
+
```
bgeltz: wsp

src/NVMLAcceleratorTopo.cpp
bgeltz: After discussion, this is fine for now.  Any objects that are instantiated downstream from here should be updated to use the num_cpus from here and *not* call ```geopm_sched_num_cpu()``` directly.

Separately Diana's going to open an issue to get all the other calls to ```geopm_sched_num_cpu()``` out of the codebase where possible.

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+        }
+        else {
+            int cpu_remaining = 0;
+            std::vector<cpu_set_t *> ideal_affinitization_mask_vec;
+            cpu_set_t *affinitized_cpuset = NULL;
+
+            affinitized_cpuset = CPU_ALLOC(num_cpu);
```
bgeltz: These lines can be combined.

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+            CPU_ZERO_S(CPU_ALLOC_SIZE(num_cpu), affinitized_cpuset);
+
+            m_ideal_cpu_affinitization.resize(m_num_accelerator);
+
+            // Cache ideal affinitization due to the overhead associated with the NVML calls
+            for(int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
```
bgeltz: ```for (```

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+                ideal_affinitization_mask_vec.push_back(m_nvml_device_pool.ideal_cpu_affinitization_mask(accel_idx));
+            }
+
+            // TODO: As an optimization this may be replacable with CPU_OR of all masks in ideal_affinitzation_mask_vec
```
bgeltz: ```/// @todo ```

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+                            for (int accel_idx_inner = 0; accel_idx_inner <  m_num_accelerator; ++accel_idx_inner) {
+                                CPU_CLR(cpu_idx, ideal_affinitization_mask_vec.at(accel_idx_inner));
+                            }
+                        }
+                        ++cpu_idx;
+                    }
+                }
+            }
+
+            if(cpu_remaining != 0){
```
bgeltz: add wsp plz

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,148 @@
...
+                            ++cpu_remaining;
+                        }
+                    }
+                }
+            }
+
+            // In order to handle systems where the number of CPUs are not evenly divisble by the number of GPUs 
+            // a two pass process is used.  This does not guarantee affinitization is successful, but covers many
+            // common cases
+            for(int affinitization_attempts = 0; affinitization_attempts < 2; ++affinitization_attempts) {
```
bgeltz: ```for (```

src/NVMLDevicePool.cpp
bgeltz: ```nvml_check_ok(nvml_result, <GEOPM_EXCEPTION_TYPE>, "<ERROR_STRING>")```

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+    }
+
+    NVMLDevicePoolImp::~NVMLDevicePoolImp()
+    {
+        //Shutdown NVML
+        nvmlReturn_t nvml_result;
+        nvml_result = nvmlShutdown();
```
bgeltz: These lines can be combined.

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+        nvmlProcessInfo_t *process_info_list;
+        process_info_list = new nvmlProcessInfo_t[temp];
+
+        nvml_result = nvmlDeviceGetComputeRunningProcesses(m_nvml_device[accel_idx], &temp, &process_info_list[0]);
+        if (nvml_result == NVML_SUCCESS) {
+            for (int i = 0; i<temp; i++) {
+                result.push_back(process_info_list[i].pid);
+            }
+        }
+        else {
```
bgeltz: ```else if {```
Unindent where applicable below.

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+            if (nvml_result == NVML_ERROR_INSUFFICIENT_SIZE) {
```
bgeltz, line 313: Please put a comment here as well that this is the retry block after temp has been updated from the first call.

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+                process_info_list = new nvmlProcessInfo_t[temp];
+                nvml_result = nvmlDeviceGetComputeRunningProcesses(m_nvml_device[accel_idx], &temp, &process_info_list[0]);
+
+                if (nvml_result == NVML_SUCCESS) {
+                    for (int i = 0; i<temp; i++) {
+                        result.push_back(process_info_list[i].pid);
+                    }
+                }
+                else {
+                    throw Exception("NVMLDevicePool::" + std::string(__func__) +
```
bgeltz: Retrying once wasn't enough, increase max_contexts again.

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,465 @@
...
+                                ": NVML failed to acquire running processes for accelerator " +
+                                std::to_string(accel_idx) + ".  Error: " + nvmlErrorString(nvml_result),
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+        return result;
+    }
+
+
+    void NVMLDevicePoolImp::frequency_control(int accel_idx, int setting) const
```
bgeltz: Change the API to match what the NVML spec is.
```(int accel_idx, int min_freq, int max_freq)```

Also rename to frequency_control_sm.

File: src/NVMLDevicePool.cpp [hunk 46]
```diff
@@ -0,0 +1,465 @@
...
+            throw Exception("NVMLDevicePool::" + std::string(__func__) +
+                            ": NVML failed to get CPU Affinity bitmask for accelerator " +
+                            std::to_string(accel_idx) + ".  Error: " + nvmlErrorString(nvml_result),
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        return accel_cpuset;
+    }
+
+    uint64_t NVMLDevicePoolImp::frequency_status(int accel_idx) const
```
bgeltz: rename to frequency_status_sm

src/PlatformTopo.cpp
```diff
@@ -258,12 +268,19 @@ namespace geopm
                         ++numa_idx;
                     }
                     break;
+                case GEOPM_DOMAIN_BOARD_ACCELERATOR:
+                    for(int accel_idx = 0; accel_idx <  m_accelerator_topo.num_accelerator(); ++accel_idx) {
```
bgeltz: ```for(int accel_idx = 0; (accel_idx <  m_accelerator_topo.num_accelerator()) && result == -1; ++accel_idx) {```

src/PlatformTopo.cpp
```diff
@@ -258,12 +268,19 @@ namespace geopm
...
+                        }
+                    }
+                    break;
                 case GEOPM_DOMAIN_PACKAGE_MEMORY:
                 case GEOPM_DOMAIN_BOARD_NIC:
                 case GEOPM_DOMAIN_PACKAGE_NIC:
-                case GEOPM_DOMAIN_BOARD_ACCELERATOR:
                 case GEOPM_DOMAIN_PACKAGE_ACCELERATOR:
-                    /// @todo Add support for package memory NIC and accelerators to domain_idx() method.
+                    // @todo Add support for package memory NIC and accelerators to domain_idx() method.
```
bgeltz: Keep it triple slash, and package acclerators.

test/MockNVMLDevicePool.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+class MockNVMLDevicePool : public geopm::NVMLDevicePool
+{
+    public:
+        MOCK_CONST_METHOD0(num_accelerator,
+                           int(void));
+        MOCK_CONST_METHOD1(ideal_cpu_affinitization_mask,
+                           cpu_set_t *(int));
+        MOCK_CONST_METHOD1(frequency_status,
+                          uint64_t(int));
```
bgeltz: Needs 1 space

test/PlatformTopoTest.cpp
bgeltz: A mocked accelerator topo would allow for this to be tested without invoking the actual NVML bits.

bgeltz (changes requested):

test/PlatformTopoTest.cpp
lhlawson: PlatformTopoTests with a mocked accelerator slated for a later PR since this bundles nicely with fixes for failures when running on a system with NVML Lib support but without the driver.  Switched for EXPECT_NO_THROW for now.

src/NVMLDevicePool.cpp
```diff
@@ -64,37 +65,36 @@ namespace geopm
...
                             ": NVML failed to initialize.  Error: " +
                             nvmlErrorString(nvml_result),
-                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
 
-        //Query number of NVIDIA Accelerators
+        //Query number of NVML Accelerators
         nvml_result = nvmlDeviceGetCount(&m_num_accelerator);
         if (nvml_result != NVML_SUCCESS) {
```
dianarg: could be check_nvml_result?

src/NVMLDevicePool.cpp
```diff
@@ -136,251 +138,170 @@ namespace geopm
...
-        if (NVML_SUCCESS != nvml_result) {
-            throw Exception("NVMLDevicePool::" + std::string(__func__) +
-                            ": NVML failed to get CPU Affinity bitmask for accelerator " +
-                            std::to_string(accel_idx) + ".  Error: " + nvmlErrorString(nvml_result),
-                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-        }
-
+        check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
+                          ": NVML failed to get CPU Affinity bitmask for accelerator " +
+                          std::to_string(accel_idx) + ".  Error: " + nvmlErrorString(nvml_result));
```
dianarg: `nvmlErrorString(nvml_result)` could probably move inside the function too.  Either way is fine.

dianarg (dismissed):

test/PlatformTopoTest.cpp
lhlawson: On second thought, a first pass of PlatformTopo using a mocked accelerator topo is now included in the latest commit for review as part of this PR.  This does not resolve all failures that may occur when running on a system with NVML Lib support but without the driver, however it does address the issue above and most of the PlatformTopoTest cases in that situation.

src/NVMLDevicePool.cpp
lhlawson: Good catch, concur.

configure.ac
```diff
@@ -176,6 +189,14 @@ if test "x$enable_msrsafe_ioctl_write" = "x1" ; then
 fi
 AC_SUBST([enable_msrsafe_ioctl_write])
 
+if test "x$enable_nvml" = "x1" ; then
+  ENABLE_NVML=True
+  AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of the NVML driver to support NVML board accelerators])
```
cmcantalupo: "Enables use of the NVML library to support NVML board accelerators"

configure.ac
```diff
@@ -176,6 +189,14 @@ if test "x$enable_msrsafe_ioctl_write" = "x1" ; then
...
+else
+  ENABLE_NVML=False
+fi
+AC_SUBST([ENABLE_NVML])
```
cmcantalupo: ENABLE_NVML is already defined as an AM_CONDITIONAL, it should not also be a make variable.   

```
diff --git a/configure.ac b/configure.ac
index 640f447..7749d8d 100644
--- a/configure.ac
+++ b/configure.ac
@@ -190,12 +190,8 @@ fi
 AC_SUBST([enable_msrsafe_ioctl_write])
 
 if test "x$enable_nvml" = "x1" ; then
-  ENABLE_NVML=True
   AC_DEFINE([GEOPM_ENABLE_NVML], [ ], [Enables use of the NVML driver to suppor
-else
-  ENABLE_NVML=False
 fi
-AC_SUBST([ENABLE_NVML])
 
 AC_ARG_WITH([sqlite3], [AS_HELP_STRING([--with-sqlite3=PATH],
             [specify directory for installed sqlite3 package.])])
```

src/AcceleratorTopo.hpp
```diff
@@ -0,0 +1,55 @@
...
+    {
+        public:
+            AcceleratorTopo() = default;
+            virtual ~AcceleratorTopo() = default;
+            /// @brief Number of accelerators on the platform.
+            virtual int num_accelerator(void) const;
+            /// @brief CPU Affinitization set for a particular accelerator
+            /// @param [in] domain_idx The index indicating a particular
+            ///        accelerator
+            virtual std::set<int> ideal_cpu_affinitization(int domain_idx) const;
```
cmcantalupo: cpu_affinity_ideal() maybe a better match for other names in GEOPM

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+    NVMLAcceleratorTopo::NVMLAcceleratorTopo()
+        : NVMLAcceleratorTopo(nvml_device_pool(geopm_sched_num_cpu()), geopm_sched_num_cpu())
+    {
+    }
+
+    NVMLAcceleratorTopo::NVMLAcceleratorTopo(const NVMLDevicePool &device_pool, const int num_cpu)
+        : m_nvml_device_pool(device_pool)
+    {
+        m_num_accelerator = m_nvml_device_pool.num_accelerator();
```
cmcantalupo: put in initializer list

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    num_cpu_per_accelerator = cpu_remaining % m_num_accelerator;
+                }
+
+                // This is a greedy approach for mapping CPUs to accelerators, and as such may result in some CPUs
+                // not being affinitized at all.  A potential improvement is to always determine affinity
+                // for the accelerator with the fewest possible CPUs in the accelerator mask
+                for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
+                    unsigned int accelerator_cpu_count = 0;
+                    int cpu_idx = 0;
+                    while (cpu_idx < num_cpu && accelerator_cpu_count < num_cpu_per_accelerator) {
```
cmcantalupo:
```
for (cpu_idx = 0;
     cpu_idx != num_cpu &&
     accelerator_cpu_count < num_cpu_per_accelerator;
     ++cpu_idx) {

File: src/NVMLAcceleratorTopo.cpp [hunk 57]
```diff
@@ -0,0 +1,142 @@
...
+                            m_ideal_cpu_affinitization.at(accel_idx).insert(cpu_idx);
+                            --cpu_remaining;
+                            ++accelerator_cpu_count;
+
+                            // Remove this CPU from the affinity mask of all Accelerators
+                            for (int accel_idx_inner = 0; accel_idx_inner <  m_num_accelerator; ++accel_idx_inner) {
+                                CPU_CLR(cpu_idx, ideal_affinitization_mask_vec.at(accel_idx_inner));
+                            }
+                        }
+                        ++cpu_idx;
```
cmcantalupo: remove

src/NVMLAcceleratorTopo.cpp
```diff
@@ -0,0 +1,142 @@
...
+                if (num_cpu_per_accelerator == 0) {
+                    num_cpu_per_accelerator = cpu_remaining % m_num_accelerator;
+                }
+
+                // This is a greedy approach for mapping CPUs to accelerators, and as such may result in some CPUs
+                // not being affinitized at all.  A potential improvement is to always determine affinity
+                // for the accelerator with the fewest possible CPUs in the accelerator mask
+                for (int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
+                    unsigned int accelerator_cpu_count = 0;
+                    int cpu_idx = 0;
```
cmcantalupo: remove

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,367 @@
...
+            std::cerr << "NVMLDevicePool::" << __func__ <<  ": NVML failed to shutdown."
+                      << "  Error: " <<  nvmlErrorString(nvml_result) << std::endl;
+        }
+#endif
+    }
+
+    void NVMLDevicePoolImp::check_nvml_result(nvmlReturn_t nvml_result, int error, std::string message) const
+    {
+        if (nvml_result != NVML_SUCCESS) {
+            throw Exception(message + "  Error: " + nvmlErrorString(nvml_result), error, __FILE__, __LINE__);
```
cmcantalupo: `__FILE__` and `__LINE__` should be parameters to this function (otherwise it will always print NVMLDevicePool.cpp:98)

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,367 @@
...
+            throw Exception("NVMLDevicePool::" + std::string(__func__) + ": accel_idx " +
+                            std::to_string(accel_idx) + "  is out of range",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+    }
+
+    cpu_set_t *NVMLDevicePoolImp::ideal_cpu_affinitization_mask(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        unsigned int cpu_set_size = CPU_ALLOC_SIZE(M_NUM_CPU)/sizeof(unsigned long);
```
cmcantalupo: spaces around binary operators `CPU_ALLOC_SIZE(M_NUM_CPU) / sizeof(unsigned long);`

src/NVMLDevicePool.cpp
```diff
@@ -0,0 +1,367 @@
...
+        check_accel_range(accel_idx);
+        nvmlUtilization_t result;
+        nvmlReturn_t nvml_result;
+
+        nvml_result =  nvmlDeviceGetUtilizationRates(m_nvml_device.at(accel_idx), &result);
+        check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
+                          ": NVML failed to get memory utilization for accelerator " +
+                          std::to_string(accel_idx) + ".");
+
+        return (double)result.memory;
```
cmcantalupo: This is in bytes?

src/NVMLDevicePool.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace geopm
+{
+
+    class NVMLDevicePool
+    {
+        public:
+            NVMLDevicePool() = default;
+            virtual ~NVMLDevicePool() = default;
+            /// @brief Number of accelerators on the platform.
```
cmcantalupo, line 52: add @return to every method that returns a value, and describe the output

src/PlatformTopo.cpp
```diff
@@ -258,12 +277,19 @@ namespace geopm
                         ++numa_idx;
                     }
                     break;
+                case GEOPM_DOMAIN_BOARD_ACCELERATOR:
+                    for(int accel_idx = 0; (accel_idx <  m_accelerator_topo.num_accelerator()) && (result == -1); ++accel_idx) {
+                        std::set <int> affin = m_accelerator_topo.ideal_cpu_affinitization(accel_idx);
```
cmcantalupo: `std::set<int>`

cmcantalupo (changes requested):
Big patch, in general nice work!  Made some comments some of which are nits, but the documentation for the returns is important.

src/NVMLDevicePool.cpp
lhlawson: https://docs.nvidia.com/deploy/nvml-api/structnvmlUtilization__t.html#structnvmlUtilization__t
The result here is a struct that contain GPU and Memory Utilization as an unsigned int with values from 0-100.  

This (and the gpu utilizaiton) should probably be returned as uint64_t, with casting to double to convert the scale to 0.0-1.0 in the IOGroup.

src/NVMLDevicePool.cpp
lhlawson: __LINE__ should definitely be a parameter, fixed in latest commit.  All NVML Handling is supposed to occur in this file, which leads to me wanting to keep __FILE__ as is.

src/NVMLDevicePool.hpp
```diff
@@ -50,64 +50,79 @@ namespace geopm
...
             //         in MHz
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator streaming multiproccesor frequency in MHz
             virtual uint64_t frequency_status_sm(int accel_idx) const = 0;
             /// @brief Get the NVML device utilization metric
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
-            virtual double utilization(int accel_idx) const = 0;
+            /// @return Accelerator streaming multiprocessor utilization
```
cmcantalupo, line 69: Is this a fraction from 0 - 1.0?

src/NVMLDevicePool.hpp
```diff
@@ -50,64 +50,79 @@ namespace geopm
...
             virtual uint64_t power (int accel_idx) const = 0;
             /// @brief Get the NVML device memory subsystem frequency in MHz
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator memory frequency in MHz
             virtual uint64_t frequency_status_mem(int accel_idx) const = 0;
             /// @brief Get the current NVML device clock throttle reasons
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator clock throttle reasons
```
cmcantalupo: Is there an "clock throttle reasons" enum somewhere that you can reference in the documentation?

src/NVMLDevicePool.hpp
```diff
@@ -50,64 +50,79 @@ namespace geopm
...
             //         an NVML device in millijoules
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator energy consumption in millijoules
             virtual uint64_t energy(int accel_idx) const = 0;
             /// @brief Get the current performance state of an NVML
             //         device
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator performance state
```
cmcantalupo: Any more information about what a "performance state" means?

src/NVMLDevicePool.hpp
```diff
@@ -50,64 +50,79 @@ namespace geopm
...
             virtual uint64_t performance_state(int accel_idx) const = 0;
             /// @brief Get the pcie rx throughput over a 20ms period for
             //         an NVML device
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator pcie rx throughput in kilobytes per second
```
cmcantalupo: can you define rx and tx if easy?

src/NVMLDevicePool.hpp
```diff
@@ -50,64 +50,79 @@ namespace geopm
...
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
+            /// @return Accelerator pcie tx throughput in kilobytes per second
             virtual uint64_t throughput_tx_pcie(int accel_idx) const = 0;
             /// @brief Get the NVML device memory Utilization metric
             /// @param [in] accel_idx The index indicating a particular
             ///        accelerator
-            virtual double utilization_mem(int accel_idx) const = 0;
-            /// @brief Get the list of PIDs actively running on an NVML
+            /// @return Accelerator memory utilization
```
cmcantalupo: What are the units?

cmcantalupo (commented):
Just a few more clarifying questions about the device pool interface documentation.  Thanks!

src/NVMLDevicePool.hpp
lhlawson: Percentage as a whole number 0-100.

src/NVMLDevicePool.hpp
lhlawson: Receive and transmit, respectively.

src/NVMLDevicePool.hpp
lhlawson: The possible states are defined here: https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceEnumvs.html#group__nvmlDeviceEnumvs_1g9de0cdfc67dfc2704b86344a7d5dc4fb

Non GPGPU description available here: 
https://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/group__gpupstate.html

The best guess from other documentation is that this ties to a preset min/max frequency range.

src/NVMLDevicePool.hpp
lhlawson: These are currently a series of defines in the NVML API
https://docs.nvidia.com/deploy/nvml-api/group__nvmlClocksThrottleReasons.html#group__nvmlClocksThrottleReasons
There is not currently an enum in GEOPM for this, but we can point to nvml.h

src/NVMLDevicePool.hpp
lhlawson: Percentage represented as a whole number 0 to 100

bgeltz (approved):

cmcantalupo (approved):

PR #1231, opened 2020-08-07
Remove skip launch and show details cli from test_power_balancer.py

cmcantalupo:
- The code should be refactored as an experiment to enable these features.


dianarg (approved):
I agree about refactoring. The information from show-details should be covered already by the balancer_comparison script in integration/experiment

PR #1232, opened 2020-08-10
Experiment scripts for frequency sweep

dianarg:
- Frequency sweep helpers and run script for DGEMM
- Runtime/energy vs frequency plot based on frequency sweep

### DIFF
--- a/integration/experiment/frequency_sweep/Makefile.mk
+++ b/integration/experiment/frequency_sweep/Makefile.mk
@@ -33,5 +33,6 @@ EXTRA_DIST += integration/experiment/frequency_sweep/__init__.py \
               integration/experiment/frequency_sweep/frequency_sweep.py \
               integration/experiment/frequency_sweep/gen_frequency_map.py \
               integration/experiment/frequency_sweep/gen_plot_runtime_energy.py \
+              integration/experiment/frequency_sweep/gen_region_summary.py \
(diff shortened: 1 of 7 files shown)


integration/apps/geopmbench.py
```diff
@@ -0,0 +1,76 @@
+import os
```
bgeltz: Missing license header.

integration/apps/geopmbench.py
```diff
@@ -0,0 +1,76 @@
...
+import textwrap
+
+from . import apps
```
bgeltz: Why not just ```import apps```?

bgeltz (commented):
I think I'd benefit from a review session on all of this.  I didn't realize there was so much new code.

integration/apps/geopmbench.py
dianarg: That pattern didn't work unless CWD was in the PYTHONPATH.  We probably need to look through all the files and make sure we import in a consistent way.  There are also some differences between python2 and 3.

dianarg:
This PR is one patch on top of #1233

cmcantalupo (dismissed):
LGTM.

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #1246, opened 2020-08-18
NVML Accelerator IO Group Addition

lhlawson:
This PR is a follow on to [PR 1229](https://github.com/geopm/geopm/pull/1229#).  It is intended to provide basic monitoring and control of NVML devices through the addition of an NVMLIOGroup that uses the classes added in the previous PR  With the addition of this IOGroup users should be able to set NVML device power & frequency limits using geopmread/write and monitor workloads using the monitoring agent and trace signals launch option.  PR should also allow for the creation of NVML agents but such agents are not included at this time.  

Note that the NVML Driver may require root permissions by default for the controls added to the NVMLIOGroup.  

Summary of change.
- Addition of an NVMLIOGroup class to provide an interface to the NVMLDevicePool, description of signals, and parsing of NVMLDevicePool results into SI units where possible
- Modification of the IOGroup class to include the NVMLIOGroup when appropriate configure options are used (--enable-nvml introduced in PR 1229)
- Addition of NVMLIOGroupTest.cpp to test the NVMLIOGroup functionality.  

High level overview:
image
Gray items were introduced in PR 1229

### DIFF
--- a/src/NVMLDevicePool.cpp
+++ b/src/NVMLDevicePool.cpp
@@ -174,6 +174,20 @@ namespace geopm
         return (uint64_t)result;
     }
 
+    uint64_t NVMLDevicePoolImp::power_limit(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        unsigned int result;
+        nvmlReturn_t nvml_result;
+
+        nvml_result = nvmlDeviceGetPowerManagementLimit(m_nvml_device.at(accel_idx), &result);
+        check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
+                          ": NVML failed to get power limit for accelerator " +
+                          std::to_string(accel_idx) + ".", __LINE__);
+
+        return (uint64_t)result;
+    }
+
     uint64_t NVMLDevicePoolImp::frequency_status_mem(int accel_idx) const
     {
         check_accel_range(accel_idx);
--- a/src/NVMLDevicePool.cpp
+++ b/src/NVMLDevicePool.cpp
@@ -137,9 +137,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetClock(m_nvml_device.at(accel_idx), NVML_CLOCK_SM, NVML_CLOCK_ID_CURRENT,
+        nvmlReturn_t nvml_result = nvmlDeviceGetClock(m_nvml_device.at(accel_idx), NVML_CLOCK_SM, NVML_CLOCK_ID_CURRENT,
                                          &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get SM Frequency for accelerator " +
@@ -151,9 +150,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         nvmlUtilization_t result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result =  nvmlDeviceGetUtilizationRates(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result =  nvmlDeviceGetUtilizationRates(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get GPU Utilization for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -164,9 +162,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetPowerUsage(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetPowerUsage(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get power for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -178,9 +175,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetPowerManagementLimit(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetPowerManagementLimit(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get power limit for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -192,9 +188,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetClock(m_nvml_device.at(accel_idx), NVML_CLOCK_MEM, NVML_CLOCK_ID_CURRENT, &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetClock(m_nvml_device.at(accel_idx), NVML_CLOCK_MEM, NVML_CLOCK_ID_CURRENT, &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get Memory Frequency for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -206,9 +201,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned long long result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetCurrentClocksThrottleReasons(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetCurrentClocksThrottleReasons(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get current clock throttle reasosn for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -220,9 +214,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result =  nvmlDeviceGetTemperature(m_nvml_device.at(accel_idx), NVML_TEMPERATURE_GPU, &result);
+        nvmlReturn_t nvml_result =  nvmlDeviceGetTemperature(m_nvml_device.at(accel_idx), NVML_TEMPERATURE_GPU, &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get temperature for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -234,9 +227,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned long long result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result =  nvmlDeviceGetTotalEnergyConsumption(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result =  nvmlDeviceGetTotalEnergyConsumption(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get energy for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -248,9 +240,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         nvmlPstates_t result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetPerformanceState(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetPerformanceState(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get performance state for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -262,9 +253,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetPcieThroughput(m_nvml_device.at(accel_idx), NVML_PCIE_UTIL_RX_BYTES, &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetPcieThroughput(m_nvml_device.at(accel_idx), NVML_PCIE_UTIL_RX_BYTES, &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get PCIE received throughput rate for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -276,9 +266,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         unsigned int result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceGetPcieThroughput(m_nvml_device.at(accel_idx), NVML_PCIE_UTIL_TX_BYTES, &result);
+        nvmlReturn_t nvml_result = nvmlDeviceGetPcieThroughput(m_nvml_device.at(accel_idx), NVML_PCIE_UTIL_TX_BYTES, &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get PCIE transmitted throughput rate for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -290,9 +279,8 @@ namespace geopm
     {
         check_accel_range(accel_idx);
         nvmlUtilization_t result;
-        nvmlReturn_t nvml_result;
 
-        nvml_result =  nvmlDeviceGetUtilizationRates(m_nvml_device.at(accel_idx), &result);
+        nvmlReturn_t nvml_result =  nvmlDeviceGetUtilizationRates(m_nvml_device.at(accel_idx), &result);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to get memory utilization for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -306,11 +294,10 @@ namespace geopm
         std::vector<int> result;
         unsigned int temp = M_MAX_CONTEXTS;
 
-        nvmlReturn_t nvml_result;
-        nvmlProcessInfo_t *process_info_list;
+                nvmlProcessInfo_t *process_info_list;
         process_info_list = new nvmlProcessInfo_t[temp];
 
-        nvml_result = nvmlDeviceGetComputeRunningProcesses(m_nvml_device[accel_idx], &temp, &process_info_list[0]);
+        nvmlReturn_t nvml_result = nvmlDeviceGetComputeRunningProcesses(m_nvml_device[accel_idx], &temp, &process_info_list[0]);
         if (nvml_result == NVML_SUCCESS) {
             for (int i = 0; i<temp; i++) {
                 result.push_back(process_info_list[i].pid);
@@ -349,9 +336,8 @@ namespace geopm
     void NVMLDevicePoolImp::frequency_control_sm(int accel_idx, int min_freq, int max_freq) const
     {
         check_accel_range(accel_idx);
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceSetGpuLockedClocks(m_nvml_device[accel_idx], (unsigned int) min_freq, (unsigned int) max_freq);
+        nvmlReturn_t nvml_result = nvmlDeviceSetGpuLockedClocks(m_nvml_device[accel_idx], (unsigned int) min_freq, (unsigned int) max_freq);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to set sm frequency for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -360,9 +346,8 @@ namespace geopm
     void NVMLDevicePoolImp::frequency_reset_control(int accel_idx) const
     {
         check_accel_range(accel_idx);
-        nvmlReturn_t nvml_result;
 
-        nvml_result =  nvmlDeviceResetGpuLockedClocks(m_nvml_device[accel_idx]);
+        nvmlReturn_t nvml_result =  nvmlDeviceResetGpuLockedClocks(m_nvml_device[accel_idx]);
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to reset sm frequency for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
@@ -371,9 +356,8 @@ namespace geopm
     void NVMLDevicePoolImp::power_control(int accel_idx, int setting) const
     {
         check_accel_range(accel_idx);
-        nvmlReturn_t nvml_result;
 
-        nvml_result = nvmlDeviceSetPowerManagementLimit(m_nvml_device.at(accel_idx), (unsigned int) (setting));
+        nvmlReturn_t nvml_result = nvmlDeviceSetPowerManagementLimit(m_nvml_device.at(accel_idx), (unsigned int) (setting));
         check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
                           ": NVML failed to set power limit for accelerator " +
                           std::to_string(accel_idx) + ".", __LINE__);
--- a/src/NVMLIOGroup.cpp
+++ b/src/NVMLIOGroup.cpp
@@ -67,7 +67,7 @@ namespace geopm
                                   string_format_double
                                   }},
                               {"NVML::UTILIZATION_ACCELERATOR", {
-                                  "Percentage of time the accelerator operated on a kernel in the last set of driver samples",
+                                  "Fraction of time the accelerator operated on a kernel in the last set of driver samples",
                                   {},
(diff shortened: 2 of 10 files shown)


src/NVMLDevicePool.cpp
```diff
@@ -174,6 +174,20 @@ namespace geopm
         return (uint64_t)result;
     }
 
+    uint64_t NVMLDevicePoolImp::power_limit(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        unsigned int result;
+        nvmlReturn_t nvml_result;
+
+        nvml_result = nvmlDeviceGetPowerManagementLimit(m_nvml_device.at(accel_idx), &result);
```
dianarg: can combine declaration and init

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+                                  }},
+                              {"NVML::THROTTLE_REASONS", {
+                                  "Accelerator clock throttling reasons",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"NVML::TEMPERATURE", {
+                                  "Accelerator temperature",
```
dianarg: in degrees C

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+                                  }},
+                              {"NVML::UTILIZATION_MEMORY", {
+                                  "Percentage of time the accelerator memory was accessed in the last set of driver samples",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::max,
+                                  string_format_double
+                                  }}
+						     })
+        , m_control_available({{"NVML::FREQUENCY", {
```
dianarg: We might want this name to be distinct from the signal to avoid confusion

dianarg: Same with POWER_LIMIT

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+
+        return result;
+    }
+
+    // Parse and update saved values for signals
+    void NVMLIOGroup::read_batch(void)
+    {
+        m_is_batch_read = true;
+        for (auto &sv : m_signal_available) {
+            if (sv.first == "NVML::CPU_ACCELERATOR_ACTIVE_AFFINITIZATION") {
```
dianarg, line 400: everything within this if branch could be in a helper function

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+
+                // Parse PID to CPU affinitzation and use process list --> accelerator mapping to get CPU --> accelerator
+                cpu_set_t *proc_cpuset = NULL;
+                proc_cpuset = CPU_ALLOC(m_platform_topo.num_domain(GEOPM_DOMAIN_CPU));
+                if (proc_cpuset != NULL) {
+                    for (auto &proc : process_accelerator_map) {
+                        for (int domain_idx = 0; domain_idx < sv.second.signals.size(); ++domain_idx) {
+                            if (sv.second.signals.at(domain_idx)->m_do_read) {
+                                if (sched_getaffinity(proc.first, CPU_ALLOC_SIZE(m_platform_topo.num_domain(GEOPM_DOMAIN_CPU)), proc_cpuset) == 0) {
+                                    if (CPU_ISSET(domain_idx, proc_cpuset)) {
```
dianarg: combine with &&?

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+    // Save a setting to be written by a future write_batch()
+    void NVMLIOGroup::adjust(int batch_idx, double setting)
+    {
+        if (batch_idx < 0 || (unsigned)batch_idx >= m_control_pushed.size()) {
+            throw Exception("NVMLIOGroup::" + std::string(__func__) + "(): batch_idx " +std::to_string(batch_idx)+ " out of range",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        m_control_pushed.at(batch_idx)->m_setting = setting;
+        m_control_pushed.at(batch_idx)->m_do_write = true;
```
dianarg: maybe m_is_adjusted

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+        else if (signal_name == "NVML::PCIE_RX_THROUGHPUT") {
+            result = (double) m_nvml_device_pool.throughput_rx_pcie(domain_idx)*1024;
+        }
+        else if (signal_name == "NVML::PCIE_TX_THROUGHPUT") {
+            result = (double) m_nvml_device_pool.throughput_tx_pcie(domain_idx)*1024;
+        }
+        else if (signal_name == "NVML::UTILIZATION_MEMORY") {
+            result = (double) m_nvml_device_pool.utilization_mem(domain_idx)/100;
+        }
+        else if (signal_name == "NVML::CPU_ACCELERATOR_ACTIVE_AFFINITIZATION") {
```
dianarg, line 514: is this repeated code with what's in read_batch?

src/NVMLIOGroup.cpp
dianarg: Can this call read_signal() instead?

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+        for (int domain_idx = 0; domain_idx < m_platform_topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR); ++domain_idx) {
+            m_initial_power_limit.push_back(m_nvml_device_pool.power_limit(domain_idx));
+        }
+    }
+
+    // Implemented to allow an IOGroup to restore previously saved
+    // platform settings
+    void NVMLIOGroup::restore_control(void)
+    {
+        /// @todo: Usage of the NVML API for setting frequency, power, etc requires root privileges.
```
dianarg, line 577: I don't love commented code.. maybe this can live in a separate PR

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+    // A user-friendly description of each signal
+    std::string NVMLIOGroup::signal_description(const std::string &signal_name) const
+    {
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("NVMLIOGroup::" + std::string(__func__) + ": signal_name " + signal_name +
+                            " not valid for NVMLIOGroup.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        std::string result = "";
+        result = m_signal_available.at(signal_name).m_description;
```
dianarg: combine declare and init, or just return

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+    {
+        if (!is_valid_control(control_name)) {
+            throw Exception("NVMLIOGroup::" + std::string(__func__) + ": " + control_name +
+                            "not valid for NVMLIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        std::string result = "";
+        result = m_control_available.at(control_name).m_description;
+        return result;
```
dianarg: return one line

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+    }
+
+    // Function used by the factory to create objects of this type
+    std::unique_ptr<IOGroup> NVMLIOGroup::make_plugin(void)
+    {
+        return std::unique_ptr<IOGroup>(new NVMLIOGroup);
+    }
+
+    void NVMLIOGroup::register_signal_alias(const std::string &alias_name,
+                                           const std::string &signal_name)
```
dianarg: ws

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,702 @@
...
+    // Name used for registration with the IOGroup factory
+    std::string NVMLIOGroup::plugin_name(void)
+    {
+        return "nvml";
+    }
+
+    // Function used by the factory to create objects of this type
+    std::unique_ptr<IOGroup> NVMLIOGroup::make_plugin(void)
+    {
+        return std::unique_ptr<IOGroup>(new NVMLIOGroup);
```
dianarg: you should be able to use geopm::make_unique here, since the return type is base class pointer

dianarg: unfortunately I see we have a bunch of occurrences of this pattern in other places still

src/NVMLIOGroup.hpp
```diff
@@ -0,0 +1,123 @@
...
+#define NVMLIOGROUP_HPP_INCLUDE
+
+#include <map>
+#include <vector>
+#include <string>
+#include <memory>
+
+#include "PluginFactory.hpp"
+#include "IOGroup.hpp"
+#include "NVMLDevicePool.hpp"
```
dianarg: you should be able to forward-declare and remove this include (like PlatformTopo)

src/NVMLIOGroup.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+#ifndef NVMLIOGROUP_HPP_INCLUDE
+#define NVMLIOGROUP_HPP_INCLUDE
+
+#include <map>
+#include <vector>
+#include <string>
+#include <memory>
+
+#include "PluginFactory.hpp"
```
dianarg: is PluginFactory needed here?

test/NVMLIOGroupTest.cpp
```diff
@@ -0,0 +1,330 @@
...
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        batch_idx.push_back(nvml_io.push_control("NVML::FREQUENCY", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx));
+        batch_idx.push_back(nvml_io.push_control("FREQUENCY_ACCELERATOR_CONTROL", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx));
+        batch_idx.push_back(nvml_io.push_control("NVML::FREQUENCY_RESET", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx));
+        batch_idx.push_back(nvml_io.push_control("NVML::POWER_LIMIT", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx));
+        batch_idx.push_back(nvml_io.push_control("POWER_ACCELERATOR_LIMIT", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx));
+    }
+
+    for (int idx = 0; idx < batch_idx.size(); ++idx) {
+        // Given that we are mocking NVMLDevicePool the actual setting here doesn't matter
```
dianarg, line 148: you can use `EXPECT_CALL(*m_device_pool, <function>(<value>))` to check that the value is getting passed through correctly.  You would need to save the batch indexes above to look at each control independently

test/NVMLIOGroupTest.cpp
```diff
@@ -0,0 +1,330 @@
...
+    EXPECT_NO_THROW(nvml_io.write_batch());
+}
+
+TEST_F(NVMLIOGroupTest, write_control)
+{
+    const int num_accelerator = m_platform_topo->num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR);
+    NVMLIOGroup nvml_io(*m_platform_topo, *m_device_pool);
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        EXPECT_NO_THROW(nvml_io.write_control("NVML::FREQUENCY", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx, 1530000000));
```
dianarg: similar EXPECT_CALL() usage here

test/NVMLIOGroupTest.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+    NVMLIOGroup nvml_io(*m_platform_topo, *m_device_pool);
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        batch_idx.push_back(nvml_io.push_signal("NVML::FREQUENCY", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx));
+    }
+    nvml_io.read_batch();
+
+    for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
+        double frequency = nvml_io.read_signal("NVML::FREQUENCY", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx);
```
dianarg, line 203: since there is a read_signal test below, this test can just be about read_batch + sample

test/NVMLIOGroupTest.cpp
```diff
@@ -0,0 +1,330 @@
...
+    std::vector<double> mock_freq_mem = {877, 877, 877, 877};
+    std::vector<double> mock_throttle_reasons = {0, 1, 3, 128};
+    std::vector<double> mock_temperature = {45, 60, 68, 92};
+    std::vector<double> mock_energy = {630000, 280000, 470000, 950000};
+    std::vector<double> mock_performance_state = {0, 2, 3, 5};
+    std::vector<double> mock_pcie_rx_throughput = {4000, 3000, 2000, 0};
+    std::vector<double> mock_pcie_tx_throughput = {2000, 3000, 4000, 100};
+    std::vector<double> mock_utilization_mem = {25, 50, 100, 75};
+
+    /// @todo: Read /proc/sys/kernel/pid_max to guarantee that we never hit an actual PID in testing?
```
dianarg: what bad thing can happen if we do?

test/NVMLIOGroupTest.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+        double pcie_tx_throughput = nvml_io.read_signal("NVML::PCIE_TX_THROUGHPUT", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx);
+        EXPECT_DOUBLE_EQ(pcie_tx_throughput, mock_pcie_tx_throughput.at(accel_idx)*1024);
+
+        double utilization_mem = nvml_io.read_signal("NVML::UTILIZATION_MEMORY", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx);
+        EXPECT_DOUBLE_EQ(utilization_mem, mock_utilization_mem.at(accel_idx)/100);
+    }
+
+    for (int cpu_idx = 0; cpu_idx < num_cpu; ++cpu_idx) {
+        // FIXME: The most complex signal is the cpu accelerator active afifinitzation signal, which is currently
```
dianarg, line 307: maybe there is a need for another mockable object to provide this

test/NVMLIOGroupTest.cpp
dianarg: There's no requirement in the agent interface that read_signal() and sample() provide the same value.  Mostly we want to test that sample reflects the expected value after read_batch (and throws before).  You could also have a second round of read_batch+sample like the following:
```
{
EXPECT_CALL(*m_device_pool, .... (3));
read_batch();
EXPECT_EQ(3, sample());
EXPECT_EQ(3, sample());
}
{
EXPECT_CALL(*m_device_pool, .... (5));
read_batch();
EXPECT_EQ(5, sample());
EXPECT_EQ(5, sample());
}
```

dianarg (changes requested):
Looking pretty good so far

test/NVMLIOGroupTest.cpp
lhlawson: In the initial test case I believe there would have been a failure since we wouldn't see the expected result when calling NVML::CPU_ACCELERATOR_ACTIVE_AFFINITIZATION, but that currently only EXPECTS_NO_THROW, so this is likely no longer necessary

src/NVMLIOGroup.cpp
lhlawson: This (and the read_signal version) can be broken into helper functions, but I don't believe we should have read_batch call the read_signal version.

Splitting the read_batch and read_signal implementation was an attempt at optimizing the read_batch case.  The underlying NVML call for this signal is nvmlDeviceGetComputeRunningProcesses, which can be costly (it was clocked at 0.5 to 2ms per call, and worst case we call it twice per accelerator).  This returns a list of PIDs associated with each Accelerator.  

In the read_batch case we know that we are looking for this info for all CPUs, so we can cache the PID list at the start of read_batch and then find the CPU-Accelerator mapping for all CPUs.  So it should take num accelerators multiplied by ~0.5 to 2ms + map processing time.

In the read_signal case we don't know that we are looking for more than one CPU, or how long it has been between signal reads, so we call the NVMLDevicePool (and thereby nvmldDeviceGetComputeRunningProcesses) every time.  This difference is key since it will take num cpus multiplied by num accelerators multiplied by ~0.5 to 2ms + map processing time to do the same work as read_batch.

src/NVMLIOGroup.cpp
lhlawson: Moving repeated code to accelerator_process_map() and cpu_accelerator_affinity() helper functions in next commit.

src/NVMLDevicePool.cpp
lhlawson: Change made throughout NVMLDevicePool

src/NVMLIOGroup.cpp
lhlawson: added "in degrees Celsius"

src/NVMLIOGroup.cpp
lhlawson: All controls updated to end with _CONTROL for clarity

src/NVMLIOGroup.cpp
lhlawson: No long an issue with moving this code in part to helper functions

src/NVMLIOGroup.hpp
lhlawson: Nope!

test/NVMLIOGroupTest.cpp
lhlawson: This was following the pattern from the TimeIOGroupTest for read_signal matching read_batch.  I'm amenable to removing it if need be.  
A second batch read with values differing from the first has been added.

src/NVMLIOGroup.cpp
```diff
@@ -337,60 +338,67 @@ namespace geopm
...
+    // before using them elsewhere
+    std::map<pid_t, int> NVMLIOGroup::accelerator_process_map(void) const
+    {
+        std::map<pid_t,int> accelerator_pid_map;
+
+        for (int accel_idx = 0; accel_idx < m_platform_topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR); ++accel_idx) {
+            std::vector<int> active_process_list = m_nvml_device_pool.active_process_list(accel_idx);
+            for (auto proc_itr : active_process_list) {
+                // If a process is associated with multiple accelerators we have no good means of
+                // signaling the user beyond providing an error value (-1).
```
dianarg, line 355: Comment doesn't seem to match.  Aren't there three cases here (0, 1, >1) ?  Is the first case intended to be count == 1?

dianarg (changes requested):

src/NVMLIOGroup.cpp
lhlawson: The expected return from this function is a map of PIDs to accelerator id (the NVML Device index), not the count of accelerators.  The expected values for the map are -1 to N, where N is the number of accelerators. 

As we iterate over the active process list for each accelerator we add a key-value pair into the map, but if at any point the .count function returns that a matching key already exists in the map for this PID we set the value for that PID to -1 to indicate there isn't a single accelerator mapping that we can report (read_signal & sample require a single value to return)

The intent is that for the following PID to accelerator index values:
123, 0
124, 1
124, 2
125, 3

We will return
123: 0
124: -1
125: 3

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  string_format_double
+                                  }},
+                              {"NVML::UTILIZATION_MEMORY", {
+                                  "Percentage of time the accelerator memory was accessed in the last set of driver samples",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::max,
+                                  string_format_double
+                                  }}
+						     })
```
dianarg: indentation

src/NVMLIOGroup.cpp
dianarg: got it

dianarg (dismissed):

dianarg (dismissed):

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"NVML::THROTTLE_REASONS", {
+                                  "Accelerator clock throttling reasons",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::average,
```
cmcantalupo: Is reasons an enum?  If so it shouldn't be averaged.

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  }},
+                              {"NVML::TOTAL_ENERGY_CONSUMPTION", {
+                                  "Accelerator energy consumption in joules since the driver was loaded",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::sum,
+                                  string_format_double
+                                  }},
+                              {"NVML::PERFORMANCE_STATE", {
+                                  "Accelerator performance state",
```
cmcantalupo: What is a performance state, and should it be averaged?  A bit more descriptions would be nice, units?

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"NVML::PCIE_RX_THROUGHPUT", {
+                                  "Accelerator PCIE receive throughput in bytes per second over a 20 millisecond period",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::average,
```
cmcantalupo: average or sum?

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  string_format_double
+                                  }},
+                              {"NVML::CPU_ACCELERATOR_ACTIVE_AFFINITIZATION", {
+                                  "Returns the associated accelerator for a given CPU as determined by running processes."
+                                  "\n  If no accelerators map to the CPU then NAN is returned"
+                                  "\n  If multiple accelerators map to the CPU -1 is returned",
+                                  {},
+                                  GEOPM_DOMAIN_CPU,
+                                  Agg::average,
```
cmcantalupo: expect same

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  string_format_double
+                                  }},
+                              {"NVML::UTILIZATION_MEMORY", {
+                                  "Percentage of time the accelerator memory was accessed in the last set of driver samples",
```
cmcantalupo: Fraction rather than percentage would be better.

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::max,
+                                  string_format_double
+                                  }}
+                             })
+        , m_control_available({{"NVML::FREQUENCY_CONTROL", {
+                                    "Sets streaming multiprocessor frequency min and max to the same limit",
```
cmcantalupo: Units

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                    {},
+                                    GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                    Agg::average,
+                                    string_format_double
+                                    }},
+                               {"NVML::FREQUENCY_RESET_CONTROL", {
+                                    "Resets streaming multiprocessor frequency min and max limits to default values",
```
cmcantalupo: How is the parameter used, if not used, please specify.  Perhaps it would be better to just enforce using a getter and a setter?

src/NVMLIOGroup.cpp
```diff
@@ -0,0 +1,673 @@
...
+                                    {},
+                                    GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                    Agg::average,
+                                    string_format_double
+                                    }},
+                               {"NVML::POWER_LIMIT_CONTROL", {
+                                    "Sets accelerator power limit",
+                                    {},
+                                    GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                    Agg::average,
```
cmcantalupo: sum

cmcantalupo (changes requested):
Please address feedback from last review before merging.  I pressed the wrong button before.

src/NVMLIOGroup.cpp
lhlawson: NAN is used to represent a CPU not being affinitized to an accelerator when reading at the CPU level.  -1 is used to indicate a CPU is associated with multiple GPUs when reading at the CPU level.  If we read a CORE with two CPUs that are affinitized to two different GPUs we will return NAN with expect_same and not -1.  

The cleanest solution might be to change to expect_same and modify the returns to NAN --> affinitization with multiple GPUs and -1 --> no affinitization.

src/NVMLIOGroup.cpp
lhlawson: It is not used, updated to reflect that.

src/NVMLIOGroup.cpp
lhlawson: Fixed.  Description change only, signals was already reporting fraction.  UTILIZATION_ACCELERATOR also updated

src/NVMLIOGroup.cpp
lhlawson: I've swapped the returns on CPU_ACCELERATOR_AFFINITZATION such that -1 means not affinitized, NAN means affinitized to multiple accelerators.  This pulls per CPU usage in line with aggregation usage as outlined above.

src/NVMLIOGroup.cpp
lhlawson: Changed to sum.

src/NVMLIOGroup.cpp
lhlawson: Changed to expect_same, description updated

src/NVMLIOGroup.cpp
lhlawson: It is.  Changed to expect_same.

test/NVMLIOGroupTest.cpp
lhlawson: Tracked as #1274

src/NVMLIOGroup.hpp
```diff
@@ -80,8 +80,8 @@ namespace geopm
             void register_signal_alias(const std::string &alias_name, const std::string &signal_name);
             void register_control_alias(const std::string &alias_name, const std::string &control_name);
 
-            std::map<pid_t, int> accelerator_process_map(void) const;
-            double cpu_accelerator_affinity(int cpu_idx, std::map<pid_t, int> process_map) const;
+            std::map<pid_t, double> accelerator_process_map(void) const;
```
cmcantalupo, line 83: Why did all of these integers become doubles?

cmcantalupo (commented):
Since the method is private, I suppose it doesn't really matter if the intermediate values are stored as integers or doubles, but I'm not sure what motivated the change to double.  Using an integer seems like a more natural storage type for the internal data structure.  Please look here:

https://github.com/cmcantalupo/geopm/tree/public-lhlawson-nvml-io

for a suggestion for error checking sched_getaffinity().  I did not test, so please let me know if that makes sense.

src/NVMLIOGroup.hpp
lhlawson: This change was made to more cleanly pass along NAN for cases of CPUs associated with multiple accelerators, to keep it in line with the expect_same aggregation semantics.

lhlawson:
https://github.com/cmcantalupo/geopm/tree/public-lhlawson-nvml-io

The general cleanup looks fine but error checking of sched_getaffinity should be more nuance.  A common use case appears to be running the CUDA MPS server which some workloads use before the job starts. In this case, GEOPM would call sched_getaffinity on a process ID running on the GPU associated with another user id, and receive the EPERM error code.  In addition theres the chance that between querying the GPU for PIDs and getting to this checking the process finishes, which could result in ESRCH.  

I think we're fine to modify this to error checking for EINVAL and EFAULT, but exclude the above cases.

cmcantalupo (approved):

PR #1247, opened 2020-08-19
Changes to app launch

dianarg:
- Add run_id to setup()
- Add custom geopm command line args to AppConf
- Add nekbone frequency sweep
- Small changes to frequency sweep analsysis scripts
- Figure of Merit parsing

### DIFF
--- a/integration/apps/apps.py
+++ b/integration/apps/apps.py
@@ -80,7 +80,7 @@ class AppConf(object):
     def parse_fom(self, log_path):
         return None
 
-    def make_bash(self, output_dir, run_id):
+    def make_bash(self, run_id):
         # setup has side effects; call before get_exec_args
         setup = self.setup(run_id)
         app_params = self.get_exec_args()
@@ -88,19 +88,15 @@ class AppConf(object):
             app_params = ' '.join(self.get_exec_args())
 
         script = '''#!/bin/bash\n'''
-        # TODO: cd interferes with AgentConf ability to correctly write the agent policy file
-        # and later be read by the controller
         script += textwrap.dedent('''\
-            # cd {output_dir}
             {setup}
             {app_exec} {app_params}
             {cleanup}
-        '''.format(output_dir=output_dir,
-                   setup=setup,
+        '''.format(setup=setup,
                    app_exec=self.get_exec_path(),
                    app_params=app_params,
                    cleanup=self.cleanup()))
-        bash_file = os.path.join(output_dir, '{}.sh'.format(self.name()))
+        bash_file = '{}.sh'.format(self.name())
         with open(bash_file, 'w') as ofile:
             ofile.write(script)
         # read, write, execute for owner
--- a/integration/apps/apps.py
+++ b/integration/apps/apps.py
@@ -37,7 +37,7 @@ import textwrap
 class AppConf(object):
     """
     An object that contains all details needed to run an application.
-    When used with experiment run scripts, setup(), get_exec_path(),
+    When used with experiment run scripts, setup_iteration(), get_exec_path(),
     get_exec_args(), and cleanup() will be used to construct command
     line arguments to the launcher.
     """
@@ -57,7 +57,7 @@ class AppConf(object):
         ''' Hardware threads per rank required by the application. '''
         return None
 
-    def setup(self):
+    def setup(self, run_id):
         ''' Any steps to be run prior to running one iteration of the application. '''
         return ''
 
@@ -69,6 +69,10 @@ class AppConf(object):
         ''' Command line arguments to the application. '''
         return self._exec_args
 
+    def get_custom_geopm_args(self):
+        ''' Additional geopm arguments required for the app, such as --geopm-ompt-disable.'''
+        return []
+
     def cleanup(self):
         ''' Any steps to be run after running one iteration of the application. '''
         return ''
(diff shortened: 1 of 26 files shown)


integration/apps/apps.py
```diff
@@ -37,7 +37,7 @@
 class AppConf(object):
     """
     An object that contains all details needed to run an application.
-    When used with experiment run scripts, setup(), get_exec_path(),
+    When used with experiment run scripts, setup_iteration(), get_exec_path(),
```
dianarg: just setup()

integration/apps/apps.py
```diff
@@ -69,26 +69,34 @@ def get_exec_args(self):
...
-    def make_bash(self, output_dir):
+    def make_bash(self, output_dir, run_id):
+        # setup has side effects; call before get_exec_args
+        setup = self.setup(run_id)
         app_params = self.get_exec_args()
         if type(app_params) is list:
             app_params = ' '.join(self.get_exec_args())
 
         script = '''#!/bin/bash\n'''
+        # TODO: cd interferes with AgentConf ability to correctly write the agent policy file
```
dianarg: check if this is still true; problem is with controller read, not agent conf write

integration/apps/nekbone/nekbone.py
```diff
@@ -73,6 +73,10 @@ def get_exec_path(self):
     def get_exec_args(self):
         return ['ex1']
 
+    def get_custom_geopm_args(self):
+        return ['--geopm-ompt-disable',
+                '--geopm-hyperthreads-disable']
```
bgeltz, line 78: I have not been running with this.  @cmcantalupo is this only present to avoid having to explicitly specify OMP_NUM_THREADS?  If yes, I'm fine with this.

integration/experiment/util.py
```diff
@@ -154,3 +179,9 @@ def launch_run(agent_conf, app_conf, output_dir, extra_cli_args, log_path=None,
         with open(log_path, 'w') as outfile:
             # TODO: some apps print to stderr..... use stderr=outfile
             launcher.run(stdout=outfile)
+
+    # Get app-reported figure of merit
+    fom = app_conf.parse_fom(log_path)
+    # Append to report????????????????
+    with open(report_path, 'a') as report:
+        report.write('\nFigure of Merit: {}'.format(fom))
```
bgeltz: '\nFigure of Merit: {}\n'

scripts/geopmpy/io.py
```diff
@@ -1571,6 +1575,8 @@ def _try_float(val):
                     header[top_key] = top_val
 
             _add_column('all', 'host')
+            _add_column('app', 'FOM')
```
bgeltz: Please make sure this works when the FOM is not present in the report.

bgeltz (changes requested):

integration/apps/nekbone/nekbone.py
cmcantalupo: @bgeltz  that was the motivation.

scripts/geopmpy/io.py
```diff
@@ -1383,8 +1385,16 @@ def dump_json(self, path):
     def meta_data(self):
         return copy.deepcopy(self._raw_dict['GEOPM Meta Data'])
 
+    def figure_of_merit(self):
+        result = None
+        try:
+            result = copy.deepcopy(self._raw_dict['Figure of Merit'])
```
cmcantalupo, line 1391: If the type of result is string or float, I don't think think copy is required.

cmcantalupo (dismissed):

integration/apps/apps.py
bgeltz: Is this still a problem?

scripts/geopmpy/io.py
```diff
@@ -1606,6 +1619,7 @@ def _try_float(val):
                 epoch_df = epoch_df.append(epoch_row, ignore_index=True)
 
                 app_row = copy.deepcopy(header)
+                app_row['FOM'] = figure_of_merit
```
bgeltz: I'm wondering if this should even be added if figure_of_merit() returned None.  This would potentially add a lot of useless data if None was returned.

bgeltz (changes requested):

scripts/geopmpy/io.py
dianarg: yes, it adds an entire column of None

dianarg: I can remove this; any analysis script that uses the FOM will have to check if it's there (but the analysis won't work anyway if it's all None)

integration/experiment/power_sweep/run_power_sweep_dgemm_tiny.py
```diff
@@ -57,7 +57,7 @@
     mach = machine.init_output_dir(output_dir)
 
     # application parameters
-    app_conf = geopmbench.TinyAppConf(output_dir)
+    app_conf = geopmbench.TinyAppConf()
     num_rank = num_nodes * app_conf.get_rank_per_node()
```
bgeltz: I think ```num_rank``` is unused and can be removed.

bgeltz (changes requested):
Tiny change, then I think this is g2g.

bgeltz (approved):

PR #1260, opened 2020-08-25
Add power balancer efficiency run script for dgemm and nekbone

cmcantalupo:


dianarg (commented):
LGTM, waiting on nekbone results

dianarg (approved):

PR #1263, opened 2020-08-25
Add support to the experiment infrastructure for studying barriers before collectives

bgeltz:
- Add special nekbone binary with injected barriers before MPI_Allreduce.
- Clean up nekbone patch files.
- Add safety check around ```rm -fr``` usage.
- Write messages to stdout on job step start and job completion.
- Use git to track app source changes.

### DIFF
--- a/integration/apps/nekbone/0003-Link-w-GEOPM.patch
+++ b/integration/apps/nekbone/0003-Link-w-GEOPM.patch
@@ -1,7 +1,7 @@
-From a193a2176c9b2aa82bd28a3dd8d767c3c51b2407 Mon Sep 17 00:00:00 2001
+From 7ddb3453e08be80148a4e032abb9490d620ac795 Mon Sep 17 00:00:00 2001
 From: Brad Geltz <brad.geltz@intel.com>
-Date: Thu, 13 Aug 2020 17:00:11 -0700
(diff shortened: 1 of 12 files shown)


bgeltz:
It's probably a good idea to do a live review of the changes I made to the nekbone patch files.

integration/apps/nekbone/0003-Link-w-GEOPM.patch
```diff
@@ -33,20 +33,23 @@ Subject: [PATCH 2/7] Link w/GEOPM
...
 diff --git a/test/example1/makenek-intel b/test/example1/makenek-intel
-index fa6e4f3..15eab37 100755
+index a5cce80..491621b 100755
 --- a/test/example1/makenek-intel
 +++ b/test/example1/makenek-intel
 @@ -4,12 +4,13 @@
  
  # source path
  SOURCE_ROOT="../../src"
-+GEOPM_ROOT="${GEOPM_PREFIX:?Please set GEOPM_PREFIX.}" # Defined in intelCRADA/env.sh
```
dianarg, line 49: not sure if we might to bring this pattern back at some point

dianarg (approved):
I tested these with some overnight runs and confirmed the runtime is being shifted from Allreduce into Barrier

PR #1267, opened 2020-08-27
Add script to plot different profiles against a baseline

### LINKED ISSUE #1240: As a user of the integration infrastructure, I would like to be able to visualize changes to energy and runtime for different reports based on profile name
> estimate 3

The script takes an --output-dir containing reports.  It globs out *.report and groups these report files by their profile name after stripping off any trailing numbers from the string as iterations of the same experiment.  The command line takes a list of profile names in the order that they will appear on the x-axis of the plots described below (without the iteration number).  Statistics are drawn from all reports that have a profile name that matches the command line arguments so that matches are considered iterations of the same experiment.

The visualization of energy and runtime is similar to #1238, except on the x-axis is an unique tag that identifies the policy.

Done criteria:
- Script implementing visualization above is committed to dev.

dianarg:
- Resolves #1240

### DIFF
--- a/integration/experiment/energy_efficiency/gen_plot_profile_comparison.py
+++ b/integration/experiment/energy_efficiency/gen_plot_profile_comparison.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python
+#
+#  Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#      * Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#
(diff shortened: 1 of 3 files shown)


integration/experiment/energy_efficiency/gen_plot_profile_comparison.py
```diff
@@ -0,0 +1,224 @@
...
+    common_args.add_output_dir(parser)
+    common_args.add_show_details(parser)
+
+    parser.add_argument('--baseline', dest='baseline',
+                        action='store', default=None,
+                        help='baseline profile name')
+    parser.add_argument('--targets', dest='targets',
+                        action='store', default=None,
+                        help='comma-separated list of profile names to compare')
+    parser.add_argument('--show-profiles', dest='show_profiles',
```
dianarg: --list-profiles

cmcantalupo (approved):
LGTM

PR #1270, opened 2020-08-28
Issue 1266

bakerbrandond:


Makefile.am
```diff
@@ -211,10 +211,12 @@ else
 ronn_man_noinst += $(ronn_man_beta)
 endif
 
+if ENABLE_RONN
 ronn_man = $(ronn_man_inst) $(ronn_man_noinst)
```
cmcantalupo, line 215: Without ronn_man, how does the target on 

https://github.com/geopm/geopm/pull/1270/files#diff-c949f93d03f44a4217d7a138f9e2e54aL860

work?  Which build features have been tested when configured with --disable-ronn?  Please check that when configured with --disable-ronn that the basic targets like make install and make dist either are successful, or give a reasonable error message.

PR #1272, opened 2020-09-01
Commit for AED PDP, new AED knob and AED test fixes (new PR)

fkeceli:
- Implemented the AED progressive dynamic-programming table optimization implemented in StringEditPeriodicityDetector class which speeds up the algorithm without changing its output. (Formerly known as online AED)
- Added a new knob in the StringEditEpochRecordFilter class: min_detectable_period.
- Some AED related tests were not included in the test/Makefile. Adding these tests to the Makefile exposed broken tests, fixed those tests.

### DIFF
--- a/src/EditDistPeriodicityDetector.cpp
+++ b/src/EditDistPeriodicityDetector.cpp
@@ -36,6 +36,7 @@
 #include <vector>
 #include <cmath>
 #include <algorithm>
+#include <limits>
 
 #include "record.hpp"
 #include "geopm_debug.hpp"
@@ -45,8 +46,11 @@ namespace geopm
 {
     EditDistPeriodicityDetector::EditDistPeriodicityDetector(int history_buffer_size)
         : m_history_buffer(history_buffer_size)
+        , m_history_buffer_size(history_buffer_size)
         , m_period(-1)
         , m_score(-1)
+        , m_record_count(0)
+        , m_DP(history_buffer_size * history_buffer_size * history_buffer_size)
     {
 
     }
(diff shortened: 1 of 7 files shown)


src/EditDistPeriodicityDetector.cpp
```diff
@@ -55,60 +59,89 @@ namespace geopm
...
+        m_DP[((ii % m_history_buffer_size)  * m_history_buffer_size
+            + (jj % m_history_buffer_size)) * m_history_buffer_size
+            + (mm % m_history_buffer_size)] = val;
+    }
+
+    uint32_t EditDistPeriodicityDetector::Dget(int ii, int jj, int mm) {
+        if (ii <= m_record_count - m_history_buffer_size) {
+            // This value is supposed to be INF but not so large that it gets wrapped around when
+            // a small value is added to it.
+            return std::numeric_limits<uint32_t>::max()/2;
```
dianarg: could still be a named constant if this is used in a few places

dianarg (commented):
Consider bringing back the constant for very large value.  Otherwise looks fine to me

src/EditDistPeriodicityDetector.cpp
fkeceli: It is only used in this function so I think this is fine: less global constants to keep track of.

cmcantalupo: Avoid early returns, and define constant at top of function for repeated use within the function please.

src/EditDistPeriodicityDetector.cpp
```diff
@@ -55,60 +59,89 @@ namespace geopm
...
+            for (int ii = std::max({1, m_record_count - m_history_buffer_size}); ii < mm + 1; ++ii) {
+                int term = 2;
+                if ((m_record_count - (ii - 1) <= num_recs_in_hist) &&
+                    (m_history_buffer.value(num_recs_in_hist - (m_record_count - (ii - 1))) ==
+                     m_history_buffer.value(num_recs_in_hist - 1)))
+                {
+                    term = 0;
                 }
+                Dset(ii, m_record_count - mm, mm,
+                        std::min({Dget(ii - 1, m_record_count - mm    , mm) + 1,
```
cmcantalupo: use a intermediate variable to store result of the min() computation.

src/EditDistPeriodicityDetector.cpp
```diff
@@ -55,60 +59,89 @@ namespace geopm
     {
         if (record.event == EVENT_REGION_ENTRY) {
             m_history_buffer.insert(record.signal);
+            ++m_record_count;
             calc_period();
         }
     }
 
+    void EditDistPeriodicityDetector::Dset(int ii, int jj, int mm, uint32_t val) {
+        m_DP[((ii % m_history_buffer_size)  * m_history_buffer_size
```
cmcantalupo, line 68: ((ii % m_history_buffer_size) * m_history_buffer_size * m_history_buffer_size

src/EditDistPeriodicityDetector.cpp
```diff
@@ -55,60 +59,89 @@ namespace geopm
...
+            // This value is supposed to be INF but not so large that it gets wrapped around when
+            // a small value is added to it.
+            return std::numeric_limits<uint32_t>::max()/2;
+        }
+        if (mm <= m_record_count - m_history_buffer_size) {
+            // This value is supposed to be INF but not so large that it gets wrapped around when
+            // a small value is added to it.
+            return std::numeric_limits<uint32_t>::max()/2;
+        }
+        return m_DP[((ii % m_history_buffer_size)  * m_history_buffer_size
```
cmcantalupo: ((ii % m_history_buffer_size) * m_history_buffer_size * m_history_buffer_size

src/EditDistPeriodicityDetector.cpp
```diff
@@ -55,60 +59,89 @@ namespace geopm
...
-                                get_history_value(mm + jj - 1)) ?
-                               2 : 0;
-                    DD[ii][jj][mm] = std::min({DD[ii - 1][jj    ][mm] + 1,
-                                               DD[ii    ][jj - 1][mm] + 1,
-                                               DD[ii - 1][jj - 1][mm] + term});
+        for (int mm = std::max({1, m_record_count - m_history_buffer_size}); mm < m_record_count; ++mm) {
+            for (int ii = std::max({1, m_record_count - m_history_buffer_size}); ii < mm + 1; ++ii) {
+                int term = 2;
+                if ((m_record_count - (ii - 1) <= num_recs_in_hist) &&
+                    (m_history_buffer.value(num_recs_in_hist - (m_record_count - (ii - 1))) ==
```
cmcantalupo: some local paramters to break this line up would make it easier to read.

cmcantalupo (changes requested):
Some nits, but the 3 dimensional array indexing seems incorrect, and I'm a little confused why that isn't breaking any tests.

src/EditDistPeriodicityDetector.cpp
fkeceli: This is done.

src/EditDistPeriodicityDetector.cpp
fkeceli: Fixed this.

src/EditDistPeriodicityDetector.cpp
fkeceli: Done.

src/EditDistPeriodicityDetector.cpp
fkeceli: Wrote these indices more explicitly to easier understand.

cmcantalupo (approved):

Issue #1235, opened 2020-08-13
Small changes to plots and analysis


dianarg:
- [x] launch_power_sweep.py should use ranks from AppConf
- [x] Inconsistent naming of plotting script files in power_sweep
- [x] remove non-standard comments from frequency map output
- [x] hash inputs to frequency map agent must be numbers, not hex
- [x] gen_region_summary.py does not put output files in output-dir
- [x] Add error bars to freq sweep line plot
- [x] frequency sweep plot should put png files in figures subdir
- [x] Use better colors for profile comparison plot (not the same colors as balancer comparison)
- [x] Add "monitor" to name of monitor run scripts
- [x] gen scripts should not use parse_known_args()

dianarg:
Moved remaining tasks to #1326. #1327, #1328

PR #1287, opened 2020-09-03
Integration small fixes

dianarg:
- Part of #1235 
- Stricter command line options for gen scripts
- Make script names and output dirs consistent
- Add error bars to frequency sweep plot


integration/experiment/common_args.py
```diff
@@ -86,3 +86,9 @@ def add_iterations(parser):
     parser.add_argument('--iterations', dest='iterations',
                         action='store', type=int, default=2,
                         help='number of iterations to launch')
+
+
+def add_use_stdev(parser):
```
cmcantalupo, line 91: Maybe would be nicer to select the error bar type?

--error-bar=stdev

or

--error-bar=minmax

I think it depends on how many options we expect to have the error bar choice.

dianarg: maybe we can change this when we add a third option?  otherwise, it looks like we can use 'choices' parameter from argparse to validate this option

dianarg: also, would the default then be no error bars, or default to minmax?

bgeltz: My vote is for minmax, but I'm fine with this coming in later.

bgeltz (approved):

integration/experiment/common_args.py
dianarg: I'll work on a separate PR to change the option to string and refactor the repeated code for error bars

PR #1292, opened 2020-09-09
Workaround PyTables lack of unicode support

bgeltz:
- The current version of PyTables does not support unicode strings when
  creating HDF5 files.
- https://github.com/PyTables/PyTables/issues/333
- This is fixed in the latest release for Python 3, but it is
  not yet available in pip.  It's unclear if this will ever be
  fixed in Pytables for Python 2.
- https://www.pytables.org/MIGRATING_TO_3.x.html
- https://docs.h5py.org/en/stable/strings.html
- This will be needed for backwards compatability for older PyTables
  installations.
- This fixes the issue for both Python 2 and 3.

Signed-off-by: Brad Geltz <brad.geltz@intel.com>

### DIFF
--- a/scripts/geopmpy/io.py
+++ b/scripts/geopmpy/io.py
@@ -1452,6 +1452,7 @@ class RawReportCollection(object):
         self._reports_df = pandas.DataFrame()
         self._app_reports_df = pandas.DataFrame()
         self._epoch_reports_df = pandas.DataFrame()
+        self._meta_data = None
         self.load_reports(report_paths, dir_name, verbose, do_cache)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


scripts/geopmpy/io.py
```diff
@@ -1505,14 +1506,27 @@ def load_reports(self, reports, dir_name, verbose, do_cache):
...
+                    except TypeError as error:
+                        if verbose:
+                            sys.stdout.write('Applying workaround for strings in HDF5 files... ')
+                        for key, val in self._meta_data.items():
+                            if type(val) is dict: # Do not try to convert 'Policy', it has been removed
+                                continue
+                            self._reports_df[key] = self._reports_df[key].astype('S')
+                            self._app_reports_df[key] = self._app_reports_df[key].astype('S')
+                            self._epoch_reports_df[key] = self._epoch_reports_df[key].astype('S')
+                        continue # Try to write HDF5 files again
```
bgeltz: Move to function.  Change if to be if not, indent self.* lines under if.

scripts/geopmpy/io.py
```diff
@@ -1505,14 +1506,27 @@ def load_reports(self, reports, dir_name, verbose, do_cache):
...
                 # Cache report dataframe
-                try:
-                    if verbose:
-                        sys.stdout.write('Generating HDF5 files... ')
-                    self._reports_df.to_hdf(report_h5_name, 'report', format='table')
-                    self._app_reports_df.to_hdf(report_h5_name, 'app_report', format='table', append=True)
-                    self._epoch_reports_df.to_hdf(report_h5_name, 'epoch_report', format='table', append=True)
-                except ImportError as error:
-                    sys.stderr.write('Warning: <geopm> geopmpy.io: Unable to write HDF5 file: {}\n'.format(str(error)))
+                while True:
```
bgeltz: Use variable: ```while not done:```

dianarg (approved):

PR #1293, opened 2020-09-09
Update the way run scripts populate the argument parser

cmcantalupo:
There is a lot of repreated code in the run scripts for adding command line arguments.   This change puts this code in a common place in common_args.py.  Another change is to pass the args object directly to the launch functions rather than breaking out the members.  Finally, a basic main() function has been added for experiments to use which takes an app config as input.

### DIFF
--- a/integration/experiment/common_args.py
+++ b/integration/experiment/common_args.py
@@ -44,84 +44,84 @@ def setup_run_args(parser):
     add_cool_off_time(parser)
 
 
-def add_output_dir(parser, default='.'):
+def add_output_dir(parser):
     parser.add_argument('--output-dir', dest='output_dir',
-                        action='store', default=default,
+                        action='store', default='.',
                         help='location for reports and other output files')
 
 
-def add_trial_count(parser, default=2):
+def add_trial_count(parser):
     parser.add_argument('--trial-count', dest='trial_count',
-                        action='store', type=int, default=default,
+                        action='store', type=int, default=2,
                         help='number of experiment trials to launch')
 
 
-def add_node_count(parser, default=1):
+def add_node_count(parser):
     parser.add_argument('--node-count', dest='node_count',
-                        default=default, type=int,
+                        default=1, type=int,
                         help='number of nodes to use for launch')
 
 
-def add_show_details(parser, default=False):
+def add_show_details(parser):
         parser.add_argument('--show-details', dest='show_details',
-                            action='store_true', default=default,
+                            action='store_true', default=False,
                             help='print additional data analysis details')
 
 
-def add_min_power(parser, default=None):
+def add_min_power(parser):
     parser.add_argument('--min-power', dest='min_power',
-                        action='store', type=float, default=default,
+                        action='store', type=float, default=None,
                         help='bottom power limit for the sweep')
 
 
-def add_max_power(parser, default=None):
+def add_max_power(parser):
     parser.add_argument('--max-power', dest='max_power',
-                        action='store', type=float, default=default,
+                        action='store', type=float, default=None,
                         help='top power limit for the sweep')
 
 
-def add_step_power(parser, default=10):
+def add_step_power(parser):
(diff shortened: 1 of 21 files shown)


integration/experiment/common_args.py
```diff
@@ -34,15 +34,36 @@
...
-def add_nodes(parser):
-    parser.add_argument('--nodes', dest='nodes',
+def add_trial_count(parser):
+    parser.add_argument('-t', '--trial-count', dest='trial_count',
+                        action='store', type=int, default=2,
+                        help='number of experiment trials to launch')
+
+
+def add_node_count(parser):
+    parser.add_argument('-n', '--node-count', dest='node_count',
```
dianarg: Do we want to leave open the possibility of forwarding -n to the launch command?  Same question for other short form args.

integration/experiment/power_sweep/power_sweep.py
```diff
@@ -41,6 +41,21 @@
 import geopmpy.io
 
 from experiment import launch_util
+from experiment import common_args
+
+def setup_run_args(parser, cool_off_time=60):
+    common_args.add_output_dir(parser)
```
dianarg: is this deliberately not using common_args.setup_run_args()?

dianarg: if different defaults is the problem, it could forward them to that function (maybe a dict?)

cmcantalupo: promoting the cool_off_time default to be locally configured is the issue that cause the change in the pattern.  I'm not sure its is a huge problem as it is.

integration/experiment/common_args.py
cmcantalupo: yeah, think we just want to get rid of the short form arguments.

integration/experiment/common_args.py
```diff
@@ -34,61 +34,94 @@
...
+    """Add common arguments for all run scripts:
+       --output-dir --node-count --trial-count --cool-off-time
+    """
+    add_output_dir(parser)
+    add_node_count(parser)
+    add_trial_count(parser)
+    add_cool_off_time(parser)
+
+
+def add_output_dir(parser, default='.'):
```
cmcantalupo: Enabling override of defaults upon call to add_*() may not be all that useful since we can always call parser.set_defaults() after the fact.  Maybe we should not make that change.

dianarg: agreed, there should be one way to override defaults

integration/experiment/energy_efficiency/barrier_frequency_sweep.py
```diff
@@ -76,28 +82,44 @@ def launch_configs(app_conf_ref, app_conf, default_freq, sweep_freqs, barrier_ha
...
+def launch(app_conf_init, args, experiment_cli_args):
+    mach = machine.init_output_dir(args.output_dir)
+    freq_range = frequency_sweep.setup_frequency_bounds(mach,
+                                                        args.min_frequency,
+                                                        args.max_frequency,
+                                                        args.step_frequency,
+                                                        add_turbo_step=True)
+    barrier_hash = geopmpy.hash.crc32_str('MPI_Barrier')
+    default_freq = max(freq_range)
+    targets = launch_configs(app_conf_ref=app_conf_init(add_barriers=False),
```
dianarg: I'd prefer these still come in as separate AppConf objects, unless we are requiring that all AppConf constructors use the same add_barriers name.  I could see a situation where the app owner chose to have different classes

dianarg (changes requested):

integration/experiment/energy_efficiency/barrier_frequency_sweep.py
cmcantalupo: Right, just have two input arguments to launch().  That was a brain fart late yesterday.

dianarg (dismissed):

dianarg (approved):

PR #1306, opened 2020-09-16
Add hpcg to smoke test lists

cmcantalupo:


integration/smoke/run_smoke.sh
```diff
@@ -47,21 +47,21 @@ function run_all {
 }
 
 function run_all_monitor {
-    APPLICATIONS="dgemm dgemm_tiny nekbone minife"
+    APPLICATIONS="dgemm dgemm_tiny nekbone minife hpcg hpcg_mkl"
```
cmcantalupo, line 50: Think we will want to get rid of "hpcg_mkl" once the change to remove the option is merged.

PR #1311, opened 2020-09-21
Use absolute path for RawReportCollection h5 name

### LINKED ISSUE #1258: hdf5 cache files for AppOutput and RawReportCollection can overlap
These two classes use the same method to produce their cache file names, so running both parsing classes on the same report files can cause problems when the .h5 file is used.  These two classes should not be used at the same time, but a workaround is to use do_cache=False.

dianarg:
- Fixes #1258

### DIFF
--- a/scripts/geopmpy/io.py
+++ b/scripts/geopmpy/io.py
@@ -1454,6 +1454,16 @@ class RawReportCollection(object):
         self._epoch_reports_df = pandas.DataFrame()
         self.load_reports(report_paths, dir_name, verbose, do_cache)
 
+    @staticmethod
+    def make_h5_name(paths, outdir):
+        paths_str = str([os.path.abspath(rr) for rr in paths])
+        try:
+            h5_id = hashlib.shake_256(paths_str.encode()).hexdigest(14)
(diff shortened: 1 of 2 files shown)


scripts/geopmpy/io.py
```diff
@@ -1454,6 +1454,16 @@ def __init__(self, report_paths, dir_name='.', verbose=True, do_cache=True):
         self._epoch_reports_df = pandas.DataFrame()
         self.load_reports(report_paths, dir_name, verbose, do_cache)
 
+    @staticmethod
+    def make_h5_name(paths, outdir):
+        paths_str = str([os.path.abspath(rr) for rr in paths])
+        try:
+            h5_id = hashlib.shake_256(paths_str.encode()).hexdigest(14)
+        except AttributeError:
+            h5_id = hash(paths_str)
```
bgeltz: ```hashlib.sha256("test".encode()).hexdigest()[:14]```
That seems to work in Py 2 and 3.

bgeltz (changes requested):

bgeltz (approved):

PR #1318, opened 2020-09-25
Add support for ICX using SKX MSR definitions

### LINKED ISSUE #1304: As a user of PlatformIO I would like support for the Ice Lake architecture
> estimate 2

Done Criteria:
-----------------
-- Use mapping from SKX for ICX as was done for HSX and BDX
-- Add new test case for MSRIOGroupTest.
-- Check geopmread/geopmwrite
-- Check BIOS enablement

cmcantalupo:
- Fixes #1304

### DIFF
--- a/test/MSRIOGroupTest.cpp
+++ b/test/MSRIOGroupTest.cpp
@@ -109,6 +109,7 @@ TEST_F(MSRIOGroupTest, supported_cpuid)
         MSRIOGroup::M_CPUID_BDX,
         MSRIOGroup::M_CPUID_KNL,
         MSRIOGroup::M_CPUID_SKX,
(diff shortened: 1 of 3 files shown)


test/MSRIOGroupTest.cpp
```diff
@@ -109,6 +109,7 @@ TEST_F(MSRIOGroupTest, supported_cpuid)
         MSRIOGroup::M_CPUID_BDX,
         MSRIOGroup::M_CPUID_KNL,
         MSRIOGroup::M_CPUID_SKX,
```
dianarg, line 111: this is the test that's failing in CI

dianarg (approved):

PR #1325, opened 2020-09-29
Fix for HPCG build script

dianarg:
- git am hangs when patch list is empty

### DIFF
--- a/integration/apps/hpcg/download_and_build.sh
+++ b/integration/apps/hpcg/download_and_build.sh
@@ -31,6 +31,7 @@
 #
 
 set -e
+set -x
 
 TARGETS="mcfly quartz xeon"
 if [ $# -ne 1 ]; then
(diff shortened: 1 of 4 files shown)


integration/apps/hpcg/download_and_build.sh
```diff
@@ -83,7 +84,9 @@ git add -A
 git commit --no-edit -s -m "Initial commit"
 
 # Apply patches
-git am ${BUILD_PATCHES}
+if [ "${BUILD_PATCHES}" != "" ]; then
```
cmcantalupo:
```
if [ ! -z "$BUILD_PATCHES" ]; then
```

integration/apps/hpcg/download_and_build.sh
```diff
@@ -31,6 +31,7 @@
 #
 
 set -e
+set -x
```
cmcantalupo, line 34: I'm not opposed to this change, but set -x is not present in the other download and build scripts.  Maybe it should be.

cmcantalupo (approved):

PR #1351, opened 2020-10-02
Nekbone fixups and code style alignment

bgeltz:
Incorporates Nekbone changes to only checkout the trunk branch and use the main branch.
Use POSIX compliant function declarations.
Align variable naming with other BASH scripts in the repo.

### DIFF
--- a/integration/apps/build_func.sh
+++ b/integration/apps/build_func.sh
@@ -32,14 +32,14 @@
 
 # Clean out old versions of application source and warn user
 clean_source() {
-    local dirname=$1
+    local DIRNAME=${1}
     # Clear out old versions:
-    if [ -d "$dirname" ]; then
-        echo "WARNING: Previous source directory detected at ./$dirname"
+    if [ -d "${DIRNAME}" ]; then
+        echo "WARNING: Previous source directory detected at ./${DIRNAME}"
         read -p "OK to delete and rebuild? (y/n) " -n 1 -r
         echo
         if [[ ${REPLY} =~ ^[Yy]$ ]]; then
-            rm -rf $dirname
+            rm -rf ${DIRNAME}
         else
             echo "Not OK.  Stopping."
             exit 1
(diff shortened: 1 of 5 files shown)


integration/apps/build_func.sh
```diff
@@ -48,44 +48,45 @@ function clean_source {
...
     fi
     cd -
 }
 
 # Get the source archive from local cache or web page
-function get_archive {
-    local archive=$1
-    if [ ! -f $archive ]; then
-        if [ -f "$GEOPM_APPS_SRCDIR/$archive" ]; then
-            cp $"$GEOPM_APPS_SRCDIR/$archive" .
```
bgeltz, line 75: Looks like this was never tested.

integration/apps/nekbone/build.sh
```diff
@@ -32,31 +32,23 @@
...
-    dirname=$1
-    basedir=$PWD
-    tmpdir=$(mktemp -d)
-    cd $tmpdir
-    svn checkout https://repocafe.cels.anl.gov/repos/nekbone
-    mv nekbone/trunk/nekbone $basedir/$dirname
-    cd -
-    rm -rf $tmpdir
+get_nekbone_source() {
+    svn checkout https://repocafe.cels.anl.gov/repos/nekbone/trunk/nekbone ${1}
```
cmcantalupo: The goal with the tmp directory was to get rid of the SVN data.  Be nice to keep that feature.

cmcantalupo (commented):
LGTM.  I my rule of thumb is to all caps exported bash variables and down-case local variables.  I'm also not very consistent with that rule.  If you want to all caps everything, I guess that is okay too.  With the tmp checkout my goal was to get rid of the svn data (so we don't have a git repo inside of an svn repo).

integration/apps/nekbone/build.sh
cmcantalupo: Also, I've been trying to name $1 in the function with a local variable so it is clear what it is.

cmcantalupo (approved):

PR #1356, opened 2020-10-05
Update amg build to use new profiling patch

cmcantalupo:
Patch that was in the AMG directory did not apply cleanly to new source.  This patch integrates geopm profiling with the internal AMG profiling (from hypre).


dianarg (approved):

dianarg (changes requested):
Need to fix makefile?

dianarg (approved):

PR #1357, opened 2020-10-06
Switch AMG to build to using github version and add epoch call

cmcantalupo:
Small updates to amg benchmark in integration/apps


cmcantalupo:
Duplicate PR

PR #1359, opened 2020-10-07
Figure of merit and perf per watt metrics in plots

dianarg:
- Add --performance-metric option for plots
- Figure-of-merit performance uses perf. per watt instead of total energy
- Fixes for error bars in balancer comparison
- TODO: still need to test all combinations of options, especially for balancer comparison speedup and normalize


cmcantalupo (commented):
Looks good to me.  I'd prefer that we specify the error bar as a string rather than a bool (to enable more ways of plotting errors), but we can leave that for another patch.  I'll give it another look in the morning.

cmcantalupo (approved):

PR #1362, opened 2020-10-08
Update integration apps to use build environment

bgeltz:
- Minor cleanup and separation of changes for MiniFE.
- Add SLURM run script generator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,6 +121,7 @@
 /src/*.lo
 /src/*.o
 /stamp-h1
+*.sbatch
 *.swp
 /tags
 /test-driver
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@
 /googletest-release-1.8.1.tar.gz
 /gtest-all.o
 /install-sh
+/integration/apps/hpcg/hpcg/
 /integration/apps/nekbone/nekbone/
 /integration/apps/nasft/nasft/nas_ft
 /integration/apps/nasft/nasft/*.o
--- a/integration/apps/minife/0004-Add-Epoch-markup.patch
+++ b/integration/apps/minife/0004-Add-Epoch-markup.patch
@@ -1,3 +1,8 @@
+From 3ebeb9799662a3414e74aff5e0e95884e83945ee Mon Sep 17 00:00:00 2001
+From: Brad Geltz <brad.geltz@intel.com>
+Date: Wed, 7 Oct 2020 15:54:58 -0700
+Subject: [PATCH 4/5] Add Epoch markup
+#
 #  Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
 #
 #  Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -121,6 +122,7 @@
 /src/*.lo
 /src/*.o
 /stamp-h1
+*.sbatch
```
dianarg, line 125: aren't we planning to add some sbatch scripts?

integration/apps/minife/0005-Add-region-markup.patch
```diff
@@ -34,59 +34,20 @@ Subject: [PATCH] Adding geopm markup to CORAL version of miniFE.
...
- # GNU CC=mpicc
- 
- # BG/Q
--CXX=mpixlcxx_r
--CC=mpixlc_r
--
-+# CXX=mpixlcxx_r
-+# CC=mpixlc_r
-+#
-+CXX=mpicxx
```
dianarg, line 78: Should be $MPICC etc?

.gitignore
dianarg: nevermind, found the generator

integration/experiment/gen_slurm.sh
```diff
@@ -0,0 +1,92 @@
...
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+# GEOPM SBATCH run script generator for SLURM
+
+# Check command line args
+if [ $# -ne 4 ]; then
+    echo "Usage: $0 <num_nodes> <app> <exp_dir> <exp_type>."
+    echo "       ex. $ ./gen_slurm.sh 1 nekbone monitor monitor"
```
cmcantalupo: maybe we could have exp_type default to exp_dir if not provided?

integration/experiment/gen_slurm.sh
```diff
@@ -0,0 +1,92 @@
...
+OUTPUT_DIR=\${GEOPM_WORKDIR}/\${SLURM_JOB_NAME}_\${SLURM_JOBID}
+
+${GEOPM_SOURCE}/integration/experiment/${EXP_DIR}/run_${EXP_TYPE}_${APP}.py \\
+    --nodes=\${SLURM_NNODES} \\
+    --output-dir=\${OUTPUT_DIR} \\
+    # end
+
+EOF
+
+uniq ${SCRIPT_NAME} .${SCRIPT_NAME}
```
cmcantalupo, line 101: why are duplicate lines expected?

integration/apps/minife/0004-Add-Epoch-markup.patch
```diff
@@ -0,0 +1,69 @@
...
+
+diff --git a/src/cg_solve.hpp b/src/cg_solve.hpp
+index 69812c4..d43c6cd 100644
+--- a/src/cg_solve.hpp
++++ b/src/cg_solve.hpp
+@@ -36,6 +36,10 @@
+
+ #include <outstream.hpp>
+
++#ifndef NO_GEOPM
```
cmcantalupo, line 50: we have
```
#ifdef GEOPM
```
in other apps not sure it is very important to be consistent.

cmcantalupo (commented):
Looks pretty good.  The bash bash generation is a little weird, but seems like a good idea.

integration/apps/minife/0005-Add-region-markup.patch
bgeltz: Yeah, you're correct.  I was looking at Makefile.intel.openmp, and I was trying to use that instead of the BG/Q based regular Makefile.  But I just checked, and the binary that's built doesn't utilize MPI.  I'll fix this up.

integration/experiment/gen_slurm.sh
bgeltz: Some of the env vars above (e.g. ${SBATCH_ACCOUNT_LINE}) may expand to nothing since they're not always needed and may not be defined.

dianarg (dismissed):

integration/experiment/gen_slurm.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Check command line args
+if [ $# -eq 3 ]; then
+    # ex. ./gen_slurm.sh 1 nekbone monitor
+    NUM_NODES=$1
+    APP=$2
+    EXP_DIR=$3
+    EXP_TYPE=$3
+elif [ $# -ne 4 ]; then
+    echo "Usage: $0 <num_nodes> <app> <exp_dir & exp_type>"
```
cmcantalupo: [exp_type]
if exp_type not provided exp_dir is used for

cmcantalupo (approved):

PR #1363, opened 2020-10-09
Add total node memory to machine file

dianarg:


integration/experiment/machine.py
```diff
@@ -46,10 +46,15 @@ def load(self):
...
                 data = json.load(fid)
-                if 'signals' in data and 'topo' in data:
+                try:
                     self.signals = data['signals']
                     self.topo = data['topo']
-                else:
+                    try:
+                        self.meminfo = data['meminfo']
+                    except:
+                        #  :( :(
```
cmcantalupo: I think it isn't so sad to support backward compatibility.  Might want a different comment.

integration/experiment/util.py
```diff
@@ -146,3 +146,25 @@ def geopmread_domain():
...
+        raise err
+    output = stdout.getvalue()
+    result = {}
+    for line in output.strip().split('\n'):
+        if 'MemTotal' in line:
+            key, total, units = tuple(line.split())
+            if units != 'kB':
+                raise RuntimeError("Unknown how to convert units: {}".format(units))
+            # Note: in spite of 'kB' label, meminfo displays kibibytes
+            result['MemTotal'] = float(total) / 1024
```
cmcantalupo: Not sure what the best units are here, but I suppose MiB is okay.  Could also just report bytes since we are casting it as a float anyway.

cmcantalupo (dismissed):

cmcantalupo (commented):
A couple of comments, but in general lgtm

integration/experiment/util.py
dianarg: bytes might be better since there is already some ambiguity of MiB vs MB

cmcantalupo (approved):

PR #1365, opened 2020-10-11
HPL netlib CPU version integration and monitor experiment.

fkeceli:
Netlib version is not the most performant version but it is a good start.

### DIFF
--- a/integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
+++ b/integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
@@ -4,6 +4,38 @@ Date: Fri, 9 Oct 2020 08:54:30 -0700
 Subject: [PATCH 2/3] Changed LAinc/lib to the correct paths and -openmp flag
  to -qopenmp
 
+#  Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#      * Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#
+#      * Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in
+#        the documentation and/or other materials provided with the
+#        distribution.
+#
+#      * Neither the name of Intel Corporation nor the names of its
+#        contributors may be used to endorse or promote products derived
+#        from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+#  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+#  A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+#  OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+#  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+Signed-off-by: Fuat Keceli <fuat.keceli@intel.com>
 ---
  Make.Linux_Intel64 | 6 +++---
  1 file changed, 3 insertions(+), 3 deletions(-)
--- a/integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
+++ b/integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
@@ -1,7 +1,7 @@
-From 28146d103eaa9b2d03dd21c971c2fcf10cdda702 Mon Sep 17 00:00:00 2001
+From 4a214908473fa022e407974591b8faaab22a2526 Mon Sep 17 00:00:00 2001
 From: Fuat Keceli <fuat.keceli@intel.com>
 Date: Fri, 9 Oct 2020 08:54:30 -0700
-Subject: [PATCH 2/3] Changed LAinc/lib to the correct paths and -openmp flag
+Subject: [PATCH 2/4] Changed LAinc/lib to the correct paths and -openmp flag
  to -qopenmp
 
 #  Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
--- a/integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
+++ b/integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
@@ -1,0 +1,39 @@
+From 28146d103eaa9b2d03dd21c971c2fcf10cdda702 Mon Sep 17 00:00:00 2001
+From: Fuat Keceli <fuat.keceli@intel.com>
+Date: Fri, 9 Oct 2020 08:54:30 -0700
+Subject: [PATCH 2/3] Changed LAinc/lib to the correct paths and -openmp flag
+ to -qopenmp
+
+---
+ Make.Linux_Intel64 | 6 +++---
+ 1 file changed, 3 insertions(+), 3 deletions(-)
+
+diff --git a/Make.Linux_Intel64 b/Make.Linux_Intel64
+index 47661c2..205a459 100644
+--- a/Make.Linux_Intel64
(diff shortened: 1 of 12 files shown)


integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
```diff
@@ -0,0 +1,71 @@
...
+ endif
+ ifndef  LAlib
+-LAlib        = -L$(LAdir)/mkl/lib/intel64 \
++LAlib        = -L$(LAdir)/lib/intel64 \
+                -Wl,--start-group \
+                $(LAdir)/lib/intel64/libmkl_intel_lp64.a \
+                $(LAdir)/lib/intel64/libmkl_intel_thread.a \
+@@ -177,7 +177,7 @@ HPL_DEFS     = $(F2CDEFS) $(HPL_OPTS) $(HPL_INCLUDES)
+ #
+ CC       = mpiicc
```
bgeltz, line 62: The compilers should all be updated to reference the build environment variables:
https://github.com/geopm/geopm/blob/dev/integration/config/build_env.sh#L54-61

Here this means:
```CC       = ${MPICC}```

Not sure if there are other compilers being used (Fortran?) but they should also be updated.

File: integration/apps/hpl_cpu/0003-Changed-the-definiton-of-the-TOPdir-in-Makefile.patch [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+diff --git a/Make.Linux_Intel64 b/Make.Linux_Intel64
+index 205a459..87d7191 100644
+--- a/Make.Linux_Intel64
++++ b/Make.Linux_Intel64
+@@ -67,7 +67,7 @@ ARCH         = Linux_Intel64
+ # - HPL Directory Structure / HPL library ------------------------------
+ # ----------------------------------------------------------------------
+ #
+-TOPdir       = $(HOME)/hpl
++TOPdir       := $(shell git rev-parse --show-toplevel)
```
bgeltz, line 51: Nice.  I like this trick.

integration/apps/hpl_cpu/hpl_cpu.py
```diff
@@ -0,0 +1,155 @@
...
+        }
+        if num_nodes not in process_grid_ratios:
+            raise RuntimeError(f"Number of nodes {num_nodes} is not defined for HPL.")
+        self.P = process_grid_ratios[num_nodes]['P']
+        self.Q = process_grid_ratios[num_nodes]['Q']
+
+        self.N = int(round(math.sqrt(dram_for_app / 8)))
+        self._cpu_per_rank = cores_per_node
+
+        print(f'DRAM reserved for APP: {dram_for_app} Cores for app: {cores_per_node} N={self.N}')
```
bgeltz: Please use sys.stdout.write() if you need to print to stdout.  print() is deprecated.

integration/apps/hpl_cpu/hpl_cpu.py
```diff
@@ -0,0 +1,155 @@
...
+        # should not be needed since --geopm-hyperthreads-disable is set below.
+        # OMP_NUM_THREADS: # number of cores per node. Technically this should
+        # not be needed since setting _cpu_per_rank will prompt the launcher to
+        # choose that many cores.
+        # MKL_NUM_THREADS: # matches with above, effective if linked against MKL.
+        env_file = textwrap.dedent(f'''
+        export KMP_HW_SUBSET=1t
+        export OMP_NUM_THREADS={self._cpu_per_rank}
+        export MKL_NUM_THREADS={self._cpu_per_rank}
+        EOF
```
bgeltz: If these aren't needed, I'd remove them.  I'd explain in the README about the nuances on these and the interactions with the launcher.

Also, if MKL_NUM_THREADS is not something that's set by default anywhere (I don't believe it is), it might be something else we want to explore setting in the launcher for performance reasons in other apps.  What happens if MKL_NUM_THREADS is not set here?

integration/apps/hpl_cpu/hpl_cpu.py
```diff
@@ -0,0 +1,155 @@
...
+        ''')
+
+        return f'cat > hpl_env.sh << EOF {env_file} source ./hpl_env.sh; cat > ./HPL.dat << EOF {input_file}'
```
bgeltz: Is the 'f' at the start of the string required?  If not, please remove.

Please use ```'some {} string'.format(string_to_insert)``` for these sorts of things.  The matches the style here: 
https://github.com/geopm/geopm/blob/dev/integration/apps/nekbone/nekbone.py#L72

This whole line is a little hard to read, and would be easier to grok if it were broken up into multiple lines above like:
```
setup_commands = 'cat > hpl_env.sh << EOF {};'.format(env_file)
setup_commands += 'source ./hpl_env.sh;'
setup_commands += 'cat > ./HPL.dat << EOF {};'.format(input_file)
return setup_commands
```

Alternatively, you could just dump the contents of the env_file here, and avoid sourcing it entirely (unless it's needed somewhere else, but it looks like you just use it here).

integration/apps/hpl_cpu/hpl_cpu.py
```diff
@@ -0,0 +1,155 @@
...
+        return self._cpu_per_rank
+
+    def get_bash_exec_path(self):
+        return self.exe_path
+
+    def get_bash_exec_args(self):
+        return ''
+
+    def get_custom_geopm_args(self):
+        return ['--geopm-ctl=application',
```
bgeltz, line 140: I recall that this is required for some reason and process mode does not work.  Please capture this here and/or in the README so we do not lose this detail.

bgeltz (changes requested):
This is looking pretty good.  There are a few little things that need to be fixed, and a few items that I think need to be described in the README.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: I had the first two just in case, I guess I can remove them. I will keep MKL_NUM_THREADS. MKL_NUM_THREADS is used by MKL libraries, that is how we communicate hpl how many threads per MPI it is supposed to create, I was told to use this by Nitin. Looks like it is intricate how MKL_NUM_THREADS is used by NKL libraries vs OMP_NUM_THREADS: https://stackoverflow.com/questions/13783190/openmp-and-mkl-threading

fkeceli: I put a bunch of info in the README.md.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: I put a bunch of info in the README.md.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: I agree that I should split up the command, that is much nicer to read.

As for the f-string notation: I discovered formatted string literals a few months ago (https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals) and I kind of like it better that the previous string formatting. Basically what used to be:

   "some string {} more string".format(literal)

is replaced by:

   "some string {literal} more string"
 
I will use this notation unless it is against the code conventions.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: Wow this is news to me :) I thought with Python3 "print something" was replaced with "print(something)" (statement vs function). I could not find any reference to print function to be deprecated though. When did that happen?

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: ...and comments in code.

integration/apps/hpl_cpu/0003-Changed-the-definiton-of-the-TOPdir-in-Makefile.patch
fkeceli: Thank you :)

integration/apps/hpl_cpu/hpl_cpu.py
bgeltz: You're correct about the print statement vs function.  I misspoke a bit about it being deprecated.

We never use print (statement or function) in any of the Python code in the GEOPM repo.  We still have to support Python 2, and sys.stdout.write() works the same in Python 2 and 3.  My comment should have originally been about preserving backwards compatibility with Python 2, and matching the existing style of Python code in this repo.

integration/apps/hpl_cpu/hpl_cpu.py
bgeltz: It's against the code conventions.  Please match the existing Python style in the repo.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: No problem, I will change it. However I should also mention that I was not able to run the integration scripts with Python 2 (yaml packages missing, etc). Maybe it is because I don't have my Python 2 packages up to date?

integration/apps/hpl_cpu/0002-Changed-LAinc-lib-to-the-correct-paths-and-openmp-fl.patch
fkeceli: Done.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: Done.

integration/apps/hpl_cpu/hpl_cpu.py
```diff
@@ -0,0 +1,151 @@
...
+        self.NBs=384 # This is the recommended size for Intel Scalable Xeon family.
+        process_grid_ratios = {
+            1: {'P': 1, 'Q': 1},
+            2: {'P': 1, 'Q': 2},
+            4: {'P': 2, 'Q': 2},
+            8: {'P': 2, 'Q': 4},
+            16: {'P': 4, 'Q': 4}
+        }
+        if num_nodes not in process_grid_ratios:
+            raise RuntimeError(f"Number of nodes {num_nodes} is not defined for HPL.")
```
bgeltz: Please use ''.format()

bgeltz (changes requested):
Last nit on string usage then this is g2g.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: Addressed in the last commit.

bgeltz (approved):

PR #1369, opened 2020-10-16
Added support for MKL HPL-CPU.

fkeceli:
This PR follows the HPL-CPU enabling PR, adding support for the MKL version.

### DIFF
--- a/integration/apps/hpl_cpu/build.sh
+++ b/integration/apps/hpl_cpu/build.sh
@@ -30,23 +30,89 @@
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
-set -x
 set -e
 
-# Get helper functions
-source ../build_func.sh
+codebase="netlib"
 
-# Set variables for workload
-DIRNAME=hpl-2.3
-ARCHIVE=${DIRNAME}.tar.gz
-URL=https://www.netlib.org/benchmark/hpl/
+function usage {
+    set +x
+    echo "Usage: build.sh [--mkl] [--netlib] [-h |--help]"
+    echo "   Default is netlib if not specified."
+}
 
-# Run helper functions
-clean_source ${DIRNAME}
-get_archive ${ARCHIVE} ${URL}
-unpack_archive ${ARCHIVE}
-setup_source_git ${DIRNAME}
+if (( $# > 1 )); then
+    set +x
+    shift
+    usage
+    echo "Extra arguments: $@"
+    exit 1
+fi
 
-# Build application
-cd ${DIRNAME}
-make arch=Linux_Intel64
+while [ "$1" != "" ]; do
+    case $1 in
+        --netlib )    codebase="netlib"
+                      ;;
+        --mkl )       codebase="mkl"
+                      ;;
+        --help | -h ) usage
+                      exit
+                      ;;
+        * )           echo "Argument not known: $1"
+                      usage
+                      exit 1
+    esac
+    shift
+done
+
+if [[ "${codebase}" == "netlib" ]]; then
+    set -x
+    # Get helper functions
+    source ../build_func.sh
+
+    # Set variables for workload
+    DIRNAME=hpl-2.3
+    ARCHIVE=${DIRNAME}.tar.gz
+    URL=https://www.netlib.org/benchmark/hpl/
+
+    # Run helper functions
+    clean_source ${DIRNAME}
+    get_archive ${ARCHIVE} ${URL}
+    unpack_archive ${ARCHIVE}
+    setup_source_git ${DIRNAME}
+
+    # Build application
+    cd ${DIRNAME}
+    make arch=Linux_Intel64
+else
+    set +x
+    echo "This script will check if:"
+    echo "   1. MKL is installed in your system and environment is set (check $MKLROOT environment variable)."
+    echo "   2. Intel HPL binary is installed as a part of the MKL installation in the \$MKLROOT/benchmarks/mp_linpack directory."
+    echo "   3. Check MPI version."
+    echo "It will not perform any download/build. User is expected to install MKL manually."
+
+    echo "Check 1..."
+    if [[ "${MKLROOT}" == "" ]]; then
+        echo "ERROR: MKLROOT env variable is not found. Make sure that the MKL is installed."
+    fi
+
+    echo "Check 2..."
+    xhpl_executable=${MKLROOT}/benchmarks/mp_linpack/xhpl_intel64_dynamic
+    if ! [[ -f "${xhpl_executable}" ]]; then
+        echo "ERROR: HPL executable is not found in the MKL installation at ${xhpl_executable}."
+    fi
+
+    xhpl_executable_wrapper=${MKLROOT}/benchmarks/mp_linpack/runme_intel64_prv
+    if ! [[ -f "${xhpl_executable_wrapper}" ]]; then
+        echo "ERROR: HPL executable wrapper is not found in the MKL installation at ${xhpl_executable_wrapper}."
+    fi
+
+    echo "Check 3..."
+    mpilib=$(ldd ${xhpl_executable} |grep libmpi.so)
+    echo "MPI library: ${mpilib}"
+    if echo ${mpilib} |grep mvapich > /dev/null; then
+        echo "INFO: Detected MVAPICH MPI. Not recommended for MKL HPL runs."
+    fi
+
+    echo "MKL HPL seems to be installed on your system."
+fi
--- a/integration/apps/hpl_cpu/build.sh
+++ b/integration/apps/hpl_cpu/build.sh
@@ -32,39 +32,28 @@
 
 set -e
 
-codebase="netlib"
-
 function usage {
-    set +x
-    echo "Usage: build.sh [--mkl] [--netlib] [-h |--help]"
-    echo "   Default is netlib if not specified."
+    echo "Usage: build.sh [--mkl | --netlib | -h | --help]"
+    echo "   Default is mkl if not specified."
 }
 
-if (( $# > 1 )); then
-    set +x
-    shift
-    usage
-    echo "Extra arguments: $@"
-    exit 1
-fi
+# Command line argument processing.
+case $# in
+    0 ) argv="--mkl"
+        ;;
+    1 ) argv=$1
+        ;;
+    * ) shift
+        usage
+        echo "Extra arguments: $@"
+        exit 1
+esac
 
-while [ "$1" != "" ]; do
-    case $1 in
-        --netlib )    codebase="netlib"
-                      ;;
-        --mkl )       codebase="mkl"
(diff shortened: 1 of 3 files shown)


integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
...
+    echo "Usage: build.sh [--mkl] [--netlib] [-h |--help]"
+    echo "   Default is netlib if not specified."
+}
 
-# Run helper functions
-clean_source ${DIRNAME}
-get_archive ${ARCHIVE} ${URL}
-unpack_archive ${ARCHIVE}
-setup_source_git ${DIRNAME}
+if (( $# > 1 )); then
```
bgeltz: Please match the style in the existing bash scripts:
```
integration/apps/build_func.sh:100:    if [ $# == 2 ]; then
integration/apps/build_func.sh:126:        elif [ $# -eq 2 ]; then
integration/smoke/gen_smoke.sh:116:if [ $# -ne 1 ]; then
integration/smoke/run_smoke.sh:94:if [ $# -ne 1 ]; then
```
So here that would be either ```[ $# -gt 1 ]``` or ```[ $# -ne 1 ]```

integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
-set -x
```
bgeltz, line 33: You have a lot of ```set +x``` calls below.  Why not leave this one and remove the others?

integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
...
+    shift
+    usage
+    echo "Extra arguments: $@"
+    exit 1
+fi
 
-# Build application
-cd ${DIRNAME}
-make arch=Linux_Intel64
+while [ "$1" != "" ]; do
```
bgeltz: I see ```codebase="netlib"``` is the default at the top of the script.  Why not remove the while loop here, and just allow for an mkl override of the default?  I'm thinking it'd be just:
```
case $1 in
    --netlib )
        echo "Do Netlib things."
        ;;
    --mkl )
        echo "Do mkl things"
        ;;
esac
```

If you're worried about error checking $1, I'd remove the default at the top of the script, and force the user to specify $1.

integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
...
+                      usage
+                      exit 1
+    esac
+    shift
+done
+
+if [[ "${codebase}" == "netlib" ]]; then
+    set -x
+    # Get helper functions
+    source ../build_func.sh
```
bgeltz, line 59: Why was this moved from the top of the script?  I'm pretty sure we always want to do this.

integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
...
+else
+    set +x
+    echo "This script will check if:"
+    echo "   1. MKL is installed in your system and environment is set (check $MKLROOT environment variable)."
+    echo "   2. Intel HPL binary is installed as a part of the MKL installation in the \$MKLROOT/benchmarks/mp_linpack directory."
+    echo "   3. Check MPI version."
+    echo "It will not perform any download/build. User is expected to install MKL manually."
+
+    echo "Check 1..."
+    if [[ "${MKLROOT}" == "" ]]; then
```
bgeltz: Please use ```if [ -z "${MKLROOT}" ]; then```

integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
...
+        echo "ERROR: MKLROOT env variable is not found. Make sure that the MKL is installed."
+    fi
+
+    echo "Check 2..."
+    xhpl_executable=${MKLROOT}/benchmarks/mp_linpack/xhpl_intel64_dynamic
+    if ! [[ -f "${xhpl_executable}" ]]; then
```
bgeltz: Please use ```if [ ! -f "${xhpl_executable}" ]; then``` for these sorts of checks.

When in doubt, check these for references:
https://github.com/geopm/geopm/blob/dev/integration/config/build_env.sh
https://github.com/geopm/geopm/blob/dev/tutorial/admin/00_test_prereqs.sh

integration/apps/hpl_cpu/hpl_cpu.py
```diff
@@ -116,8 +127,23 @@ def get_bash_setup_commands(self):
...
-        setup_commands = 'export MKL_NUM_THREADS={cpu_per_rank}\n'.format(cpu_per_rank=self._cpu_per_rank)
-        setup_commands += 'cat > ./HPL.dat << EOF {input_file}\n'.format(input_file=input_file)
+        setup_commands = 'cat > ./HPL.dat << EOF {input_file}\n'.format(input_file=input_file)
+
+        # Below are only needed if the codebase is MKL.
+        if not self.mklroot is None:
+            setup_commands += 'export MKL_NUM_THREADS={cpu_per_rank}\n'.format(cpu_per_rank=self._cpu_per_rank)
+            setup_commands += textwrap.dedent('''
+            # For Mvapich
+            if [ -n "${MPIRUN_RANK}" ]
```
bgeltz: Please use consistent style for the if statements:
```if [ -n "${MPIRUN_RANK}" ]; then```

File: integration/apps/hpl_cpu/hpl_cpu.py [hunk 8]
```diff
@@ -116,8 +127,23 @@ def get_bash_setup_commands(self):
         EOF
         '''.format(N=self.N, NBs=self.NBs, P=self.P, Q=self.Q))
 
-        setup_commands = 'export MKL_NUM_THREADS={cpu_per_rank}\n'.format(cpu_per_rank=self._cpu_per_rank)
-        setup_commands += 'cat > ./HPL.dat << EOF {input_file}\n'.format(input_file=input_file)
+        setup_commands = 'cat > ./HPL.dat << EOF {input_file}\n'.format(input_file=input_file)
+
+        # Below are only needed if the codebase is MKL.
+        if not self.mklroot is None:
```
bgeltz: ```if self.mklroot is not None```

integration/apps/hpl_cpu/build.sh
```diff
@@ -30,23 +30,89 @@
...
+        echo "ERROR: HPL executable is not found in the MKL installation at ${xhpl_executable}."
+    fi
+
+    xhpl_executable_wrapper=${MKLROOT}/benchmarks/mp_linpack/runme_intel64_prv
+    if ! [[ -f "${xhpl_executable_wrapper}" ]]; then
+        echo "ERROR: HPL executable wrapper is not found in the MKL installation at ${xhpl_executable_wrapper}."
+    fi
+
+    echo "Check 3..."
+    mpilib=$(ldd ${xhpl_executable} |grep libmpi.so)
```
bgeltz, line 104: Please put a space after all your pipes.

integration/apps/hpl_cpu/hpl_cpu.py
bgeltz: This could just as easily be ```if codebase == "mkl"```.  Then you don'e need ```self.mklroot``` at all.

bgeltz (changes requested):

integration/experiment/monitor/run_monitor_hpl_cpu.py
```diff
@@ -49,10 +49,12 @@
     parser.add_argument('--perc-dram', dest='perc_dram_per_node',
                     action='store', type=float, default=0.9,
                     help='Ratio of the total node DRAM that should be used for the HPL matrix (assuming DP). Default is 0.9.')
+    parser.add_argument('--codebase', action='store', type=str, default="netlib", choices=["mkl", "netlib"],
+                    help='Choose mkl or netlib based on the version of HPL to run. Default is netlib.')
```
dianarg: indentation here and above

dianarg (commented):
My initial reaction is that is we have options for both the build and run, perhaps these two versions should be separate AppConf classes (with some shared code). Rather than have an option to the run script, there would be separate run scripts.

integration/apps/hpl_cpu/build.sh
fkeceli: Done.

integration/apps/hpl_cpu/build.sh
fkeceli: I sanitized the usage of -+x. I want it to be only shown for the netlib case and not for MKL case, which is checking a bunch of conditions for MKL to run and not a real build.

integration/apps/hpl_cpu/build.sh
fkeceli: Done.

integration/apps/hpl_cpu/build.sh
fkeceli: Build is only done for netlib version. MKL version just checks the existence of MKL and IMPI. For MKL IMPI module needs to be loaded which removes some of the compiler variables that build_func.sh complains about. Alternatives are (1) define these variables redundantly, (2) don't call build_func.sh. I chose (2).

integration/apps/hpl_cpu/build.sh
fkeceli: Done.

integration/apps/hpl_cpu/build.sh
fkeceli: Done.

integration/apps/hpl_cpu/build.sh
fkeceli: Done.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: Done.

integration/apps/hpl_cpu/hpl_cpu.py
fkeceli: Done. That was a left over from a previous version, thanks for catching it.

integration/experiment/monitor/run_monitor_hpl_cpu.py
fkeceli: > My initial reaction is that is we have options for both the build and run, perhaps these two versions should be separate AppConf classes (with some shared code). Rather than have an option to the run script, there would be separate run scripts.

The MKL and and netlib versions run with the same exact set of parameters but their executable path are different and there are slight differences in setup. One could argue that we could use OOP to extend the base class, it's a choice. Is it problematic to keep it this way?

fkeceli: Also netlib was the starting version however I can see that it performs very poorly compared to MKL so I made MKL default.

fkeceli:
We decided to change how we add HPL MKL: it will be a new directory.

PR #1373, opened 2020-10-17
Addition of perfmon as described in SDM vol 3 Section 18.2.3

lhlawson:

Addition of IA32_PMC0-4, PERFEVTSEL and other related controls as outlined in SDM vol 3 section 18.2.3.  Only 4 perfmon counters are provided in this PR, though this could be expanded to support 8 counters on systems that disable hyperthreading (see SDM vol 3 section 18 for relevant arch details and restrictions).  

The expectation is the user will set the perf_global_control en_pmc*, perfevetsel* and other associated fields as required to enable these register in an sbatch file or through another means prior to launching their job on various nodes. 

IA32_PMC*:PERFCTR have been added as read only, however they could be made writable for the purposes of zeroing the PMON metrics prior to runs.


cmcantalupo (approved):
Looks fine to me.  I'm a little unsure about how to deal with the extra counters when hyperthreading is disabled.   I feel like it might be better to add them, even if they are invalid some of the time.  We can also add that in a separate patch.

PR #1378, opened 2020-10-21
Added 0x1B0 - Energy Perf Bias MSR register

lhlawson:

Addition of Performance Energy Bias Hint IA32_ENERGY_PERF_BIAS MSR for testing.

This may be coupled with a check of CPUID.6H:EXC[3] = 1 to confirm the MSR is valid for a system if used on a mixture of systems that do and do not support EPB.  See SDM IA-32 Architectural MSRs documentation for more information.

### DIFF
--- a/src/msr_data_arch.cpp
+++ b/src/msr_data_arch.cpp
@@ -966,7 +966,7 @@ namespace geopm
                 }
             }
         },
-        "IA32_ENERGY_PERF_BIAS": {
+        "ENERGY_PERF_BIAS": {
             "offset": "0x1B0",
             "domain": "cpu",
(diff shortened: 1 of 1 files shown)


src/msr_data_arch.cpp
```diff
@@ -966,6 +966,20 @@ namespace geopm
                 }
             }
         },
+        "IA32_ENERGY_PERF_BIAS": {
```
dianarg: We usually leave the IA32_ part off the name of these

lhlawson: Removed in the newest patch.

PR #1382, opened 2020-10-22
Pass by const ref

### LINKED ISSUE #1348: Change complex parameters to pass by const reference
Some functions with large types as parameters are not using const reference, e.g. functions with `std::string` parameters should have a declaration like `void func(const std::string &param);`.  These can be found using grep within the geopm source directory.  Some examples are listed below.  Please submit separate commits for each class, as some of these are in public headers that may require a version change.

 Find std::string examples: `grep '.*(.*std::string [^&]*)' src/*.[ch]pp` 
- src/geopm_plugin.[ch]pp:    void plugin_load(std::string plugin_prefix);
- src/Helper.[ch]pp:    bool string_ends_with(std::string str, std::string key);
- src/Imbalancer.[ch]pp:        Imbalancer(const std::string config_path);
- src/Profile.[ch]pp:    uint64_t ProfileImp::region(const std::string region_name, long hint);
- src/Profile.[ch]pp:    void ProfileImp::print(const std::string file_name);
- src/ProfileSampler.[ch]pp:            ProfileRankSamplerImp(const std::string shm_key, size_t table_size);

Find std::vector examples: `grep '.*(.*std::vector<.*> [^&]*)' src/*.[ch]pp`
- src/DomainControl.[ch]pp:            DomainControl(std::vector<std::shared_ptr<Control> > controls);
- src/ModelApplication.cpp:    ModelApplication::ModelApplication(uint64_t repeat, std::vector<std::string> region_name, std::vector<double> big_o, int verbosity, int rank);
- src/ProfileSampler.[ch]pp:            void sample(std::vector<std::pair<uint64_t, struct geopm_prof_message_s> > &content, size_t &length, std::shared_ptr<Comm> comm) override;
- src/TreeComm.[ch]pp:            int num_level_controlled(std::vector<int> coords);

Hacktoberfest: to run the unit tests, follow the instructions in #1319.

pathway27:
Summary of change.

- Pass by const reference, commit per Class
- Fixes #1348 from github issues.

### DIFF
--- a/src/DomainControl.cpp
+++ b/src/DomainControl.cpp
@@ -38,7 +38,7 @@
 
 namespace geopm
 {
-    DomainControl::DomainControl(std::vector<std::shared_ptr<Control> > controls)
+    DomainControl::DomainControl(const std::vector<std::shared_ptr<Control> > &controls)
         : m_controls(controls)
         , m_is_batch_ready(false)
     {
--- a/src/DomainControl.hpp
+++ b/src/DomainControl.hpp
@@ -43,7 +43,7 @@ namespace geopm
     class DomainControl : public Control
     {
         public:
-            DomainControl(std::vector<std::shared_ptr<Control> > controls);
+            DomainControl(const std::vector<std::shared_ptr<Control> > &controls);
             DomainControl(const DomainControl &other) = delete;
             virtual ~DomainControl() = default;
             void setup_batch(void) override;
--- a/src/Imbalancer.cpp
+++ b/src/Imbalancer.cpp
@@ -49,7 +49,7 @@ class Imbalancer
 {
     public:
         Imbalancer();
-        Imbalancer(const std::string config_path);
+        Imbalancer(const std::string &config_path);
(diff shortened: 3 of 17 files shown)


src/Helper.cpp
```diff
@@ -182,7 +182,7 @@ namespace geopm
         return (str.find(key) == 0);
     }
 
-    bool string_ends_with(std::string str, std::string key)
+    bool string_ends_with(const std::string &str, const std::string &key)
```
dianarg: It looks like the implementation was relying on these being copied.  Perhaps it would be better to use std::string::rfind with the length of the key?

dianarg (changes requested):
Thanks for your contribution @pathway27!  Since Helper.hpp is in the public API (installed headers), it will require a major version increase.  I think the easiest solution is to remove the changes to Helper.hpp and and Helper.cpp from this PR.  You can also submit those changes in a different PR, if you feel like trying to fix the implementation.

dianarg (approved):

Issue #1374, opened 2020-10-19
As a user of the integration infrastructure, I would like to be able to specify rank affinity for applications


dianarg:
Done criteria
--------------
- Add a feature to AppConf allowing it to specify pinning instead of the launcher

Notes
-------
- May depend on option to launcher to disable automatic pinning #XX
- Use affinity class; already in launcher.py somewhere. 
- Method to return python dictionary with pinning info (agnostic to MPI family)

PR #1383, opened 2020-10-22
Baseline experiment

### LINKED ISSUE #1244: As a user of the integration infrastructure I would like to be able to evaluate applications without using GEOPM
- Enable use of the launcher with the controller disabled as part of the experiment.
- Evaluate the figure of merit for comparison with GEOPM runs
- Print total runtime

Done criteria
---------------
- Baseline experiment with the following runs: 1) all cores, no geopm  2) no geopm but with some cores idle (as if geopm controller were there); 3) run with geopm monitor on reserved; 4) run with monitor but sharing core with the app 
- Add option to AppConf: string configuration that tells how many cores to use -> for some apps, this determines number of ranks

dianarg:
Scripts for running experiments comparing runs with the monitor agent and without geopm.

- Resolves #1244

### DIFF
--- a/integration/apps/apps.py
+++ b/integration/apps/apps.py
@@ -142,7 +142,6 @@ class AppConf(object):
         '''
         return None
 
-
     def trial_setup(self, run_id, output_dir):
         '''Create files to be used during a launch of a step in an experiment.
            This method may create symbolic links, copy files into the
@@ -277,7 +276,7 @@ class AppConf(object):
                 log_path (str): path to the log file containing the
                                 standard output from the previous run.
 
-            Return:
+            Returns:
                 float or None: application figure of merit for the run.
         '''
         return None
@@ -293,7 +292,7 @@ def make_bash(app_conf, run_id, log_file):
             log_path (str): path to the log file containing the
                             standard output from the previous run.
 
-        Return:
+        Returns:
             str: path to the generated script
     '''
 
@@ -323,11 +322,22 @@ def make_bash(app_conf, run_id, log_file):
 
 
 def get_available_app_cores(mach, pin_config):
-    '''
-    pin_config: str
-    mach: Machine
+    '''Returns the number of cores per node available to be used by
+    an application for the given pinning configuration.  The pinning
+    configuration determines how many cores will be reserved for the
+    OS and the GEOPM controller, if any.
 
-    Returns number of cores available to be used by the application.
+    Args:
+        pin_config (str): One of "all_cores" (no cores reserved),
+                          "geopm_os_shared" (1 core reserved to be
+                          shared by OS and GEOPM), or
+                          "geopm_os_reserved" (2 cores reserved for
+                          the OS and GEOPM).
+        mach (Machine): the machine configuration for the compute
+                        nodes containing the total number of cores.
+
+    Returns:
+        int: number of cores available to be used by the application.
 
     '''
     app_cores = mach.num_core()
--- a/integration/apps/apps.py
+++ b/integration/apps/apps.py
@@ -320,3 +320,23 @@ def make_bash(app_conf, run_id, log_file):
     # read, write, execute for owner
     os.chmod(bash_file, stat.S_IRWXU)
     return bash_file
+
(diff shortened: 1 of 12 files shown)


integration/apps/minife/minife.py
```diff
@@ -56,12 +55,20 @@ def __init__(self, num_nodes):
             raise RuntimeError("No input size defined for minife on {} nodes".format(num_nodes))
         self.app_params = problem_sizes[num_nodes]
 
+        app_cores = apps.get_available_app_cores(mach, pin_config)
+        # TODO: test other thread configs
+        self.cpu_per_rank = 1
+        self.ranks_per_node = app_cores // self.cpu_per_rank
```
bgeltz: So before, ranks_per_node was 2.

Now, I think it's going to be a lot more.  Thinking about 44 cores on McFly:
```
(44 - 2) // 1 = 42
```

Do you want 42 ranks per node?

integration/experiment/launch_util.py
```diff
@@ -97,23 +114,29 @@ def launch_run(agent_conf, app_conf, run_id, output_dir, extra_cli_args,
...
+    end_time = timeit.default_timer()
 
     # Get app-reported figure of merit
     fom = app_conf.parse_fom(log_path)
+    # Get total runtime
+    total_runtime = datetime.timedelta(seconds=end_time-start_time).total_seconds()
     # Append to report
     with open(report_path, 'a') as report:
         report.write('\nFigure of Merit: {}\n'.format(fom))
+        report.write('Total Runtime: {}\n'.format(total_runtime))
```
bgeltz, line 128: Thank you for adding this.

bgeltz (changes requested):

integration/apps/apps.py
```diff
@@ -319,3 +319,23 @@ def make_bash(app_conf, run_id, log_file):
     # read, write, execute for owner
     os.chmod(bash_file, stat.S_IRWXU)
     return bash_file
+
+
+def get_available_app_cores(mach, pin_config):
+    '''
+    pin_config: str
```
dianarg: Better docstring

integration/apps/minife/minife.py
dianarg: I still need to test miniFE with different rank configurations.  I don't know why we thought 2 ranks was the best, but the documentation I could dig up was using 1 rank per core.  Given that this version has OMP threading, I doubt that MPI threads only is the right configuration.

integration/experiment/baseline/gen_summary.py
```diff
@@ -0,0 +1,101 @@
...
+    result['min_fom'] = group['FOM'].min()
+    result['max_fom'] = group['FOM'].max()
+    result['std_fom'] = group['FOM'].std()
+    if args.use_stdev:
+        sys.stdout.write('Using stdev for error percent\n')
+        result['error_pct_runtime'] = result['std_runtime'] / result['total_runtime']
+        result['error_pct_fom'] = result['std_fom'] / result['FOM']
+    else:
+        sys.stdout.write('Using min-max for error percent\n')
+        # TODO: what is this supposed to be for min-max
```
dianarg: Revisit

integration/experiment/baseline/baseline.py
```diff
@@ -0,0 +1,73 @@
...
+
+from experiment import launch_util
+from experiment import common_args
+
+
+def setup_run_args(parser):
+    common_args.setup_run_args(parser)
+
+
+def launch(app_conf_configs, args, experiment_cli_args):
```
dianarg, line 46: app_conf_configs could have a better name

integration/experiment/launch_util.py
dianarg: I got tired of forgetting to add it to my scripts.  What might be interesting for later analysis is that this number is longer than the application total runtime in the report because that's the time between MPI_Init and MPI_Finalize.

dianarg:
This requires #1374 to actually get good performance in the no-geopm case

PR #1385, opened 2020-10-22
Added HPL MKL application integration.

fkeceli:


copying_headers/MANIFEST.EXEMPT
```diff
@@ -31,11 +31,12 @@ geopm-theta.spec.in
 integration/README.md
 integration/apps/README.md
 integration/apps/amg/README.md
+integration/apps/hpl_mkl/README.md
+integration/apps/hpl_netlib/README.md
 integration/apps/minife/README.md
 integration/apps/nasft/nasft/COPYING-NAS
 integration/apps/nasft/nasft/README
 integration/apps/nasft/README.md
-integration/apps/hpl_netlib/README.md
```
fkeceli, line 38: Alphabetical order correction... Just like the ones below.

integration/apps/hpl_mkl/README.md
```diff
@@ -0,0 +1,55 @@
...
+    computers. It can thus be regarded as a portable as well as freely
+    available implementation of the High Performance Computing Linpack Benchmark.
+
+This AppConf runs the MKL version of HPL, which is optimized for Intel
+processors.
+
+Testing has been done with MKL 19.0.4.243 and impi 2019.4.243 (dynamically linked).
+IMPI is required for the best performance.
+
+The --perc-dram option passed to the run scripts that call this banchmark
```
dianarg: benchmark

integration/apps/hpl_mkl/hpl_mkl.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, num_nodes, mach, perc_dram_per_node=0.9, cores_per_node=None):
+        super().__init__(num_nodes, mach, perc_dram_per_node, cores_per_node)
+        self.mklroot = os.getenv('MKLROOT')
+        self.exec_path = os.path.join(self.mklroot, 'benchmarks/mp_linpack/xhpl_intel64_dynamic')
+
+    def get_bash_setup_commands(self):
+        benchmark_dir = os.path.dirname(os.path.abspath(__file__))
+
+        setup_commands += '{}\n'.format(os.path.join(benchmark_dir, 'check_env.sh'))
```
fkeceli: @dianarg @cmcantalupo Is this the right place to put this check? It is being executed by all MPI ranks, which is kind of hard to track in the output file.

integration/apps/hpl_mkl/README.md
dianarg: Is this value from 0.0 to 1.0, or 0 to 100?

integration/apps/hpl_mkl/check_env.sh
```diff
@@ -0,0 +1,68 @@
...
+xhpl_executable=${MKLROOT}/benchmarks/mp_linpack/xhpl_intel64_dynamic
+if [ ! -f "${xhpl_executable}" ]; then
+    echo "ERROR: HPL executable is not found in the MKL installation at ${xhpl_executable}."
+    exit_val=1
+else
+    echo "Passed."
+fi
+
+echo "Check 3..."
+mpilib=$(ldd ${xhpl_executable} |grep libmpi.so)
```
dianarg: Check in the other PR for some of Brad's comments on bash style.  Here there should be a space after all the pipes.

integration/apps/hpl_mkl/check_env.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "    MPI library: ${mpilib}"
+if echo ${mpilib} |grep mvapich > /dev/null; then
```
dianarg, line 62: Is there a way to do a positive check for IMPI specifically, in case of OpenMPI or MPICH?

integration/apps/hpl_mkl/hpl_mkl.py
```diff
@@ -0,0 +1,64 @@
...
+from apps.hpl_netlib import hpl_netlib
+
+
+class HplMklAppConf(hpl_netlib.HplNetlibAppConf):
+    @staticmethod
+    def name():
+        return 'hpl_mkl'
+
+    def __init__(self, num_nodes, mach, perc_dram_per_node=0.9, cores_per_node=None):
+        super().__init__(num_nodes, mach, perc_dram_per_node, cores_per_node)
```
dianarg: Might not work with python2.  Do we care?

integration/apps/hpl_mkl/hpl_mkl.py
dianarg: Yes, this would be a motivation for bringing back the experiment_setup() method.  It would still get executed multiple times (once per trial)

fkeceli: Well at least it will be executed on one node and will be easier to read. I will move it to experiment_setup() if we ever bring it back.

integration/apps/hpl_mkl/README.md
fkeceli: Thanks.

fkeceli: I clarified it and will push update.

integration/apps/hpl_mkl/check_env.sh
fkeceli: Fixed. I lost some fixes I made previously because of a force push:P

integration/apps/hpl_mkl/check_env.sh
fkeceli: Unfortunately I cannot find a way (and I discussed with Brad too). Both OpenMP and IMPI seem to be linked from libmpi and the only way to differentiate is to look at the address at the output of ldd. That seems to be a bit fragile though so I didn't use that method. @dianarg Please let me know if you can think of something else.

integration/apps/hpl_mkl/hpl_mkl.py
fkeceli: Good point, no need to risk it. I will change it to super(HplMklAppConf, self)

dianarg (approved):

PR #1387, opened 2020-10-23
Helper refactor

pathway27:
As mentioned in https://github.com/geopm/geopm/pull/1382, pulling out the `Helper` pass by const ref. changes here.
And refactoring `string_ends_with` to use `rfind`.

Related to #1348

### DIFF
--- a/src/Helper.cpp
+++ b/src/Helper.cpp
@@ -184,9 +184,7 @@ namespace geopm
 
     bool string_ends_with(const std::string &str, const std::string &key)
     {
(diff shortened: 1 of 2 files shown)


dianarg:
You should be able to debug the failing unit test more quickly by running the tests locally.  Let me know if you need any help getting that set up.

cmcantalupo:
C++20 adds the starts_with() and ends_with() methods for std::string.  We may want our functions to reflect the APIs defined in the standard.

PR #1393, opened 2020-10-29
Addition of the Weather Research and Forecasting (WRF) Application to integration

lhlawson:

Addition of the Weather Research and Forecasting model v3.9.1.1 application to the integration apps, along with patches for switching the build to use the Intel toolchain, scripts for sizing jobs for multinode runs, and setup with the conus 2.5km benchmark.

License information included in the integration/apps/wrf/README.md
Workload/application and benchmark input sourced from https://www2.mmm.ucar.edu/
Dependencies and their build instructions sourced from https://www2.mmm.ucar.edu/wrf/OnLineTutorial/compilation_tutorial.php   
                                                   
- Addition of geopm/integration/apps/wrf: README.md, init, wrf.py, build.sh, and patch files for workload and dependency build                                                                              - Addition of monitor and frequency sweep scripts for WRF
- Addition of mcfly configuration script in geopm/integration/config
- Inclusion of plotting scripts for figure of merit and runtime vs inverse frequency (as part of frequency sweeps)  

Note that the WRF build scripts use perl to change the configure file before build, so the generalized setup of compilers used for other apps in the GEOPM integration infrastructure will not work as expected.  As such the configure file has been modified in the patches to use ifort, mpiifort, and mpiicc.


integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,139 @@
...
+        # TODO: use self._machine_file to determine
+        return self._num_rank_per_node
+
+    def get_bash_setup_commands(self):
+        return_string =  "ulimit -s unlimited;\n"
+        return_string += "export OMP_PLACES=cores;\n"
+        return_string += "export OMP_PROC_BIND=close;\n"
+        return_string += "export OMP_STACKSIZE=512M;\n" #TODO: May change to 128M for low node count
+        return_string += "export WRFIO_NCD_LARGE_FILE_SUPPORT=1;\n"
+        return_string += "export WRF_NUM_TILES="+str(self._wrf_num_tiles)+";\n"
```
cmcantalupo: use string format method:
```
.
.
.
export WRF_NUM_TITLES={WRF_NUM_TITLES}
'''.format(WRF_NUM_TITLE=self._wrf_num_titles)
```

integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,139 @@
...
+    def get_cpu_per_rank(self):
+        return self._cpu_per_rank
+
+    #Sets ranks per node
+    def get_rank_per_node(self):
+        # TODO: use self._machine_file to determine
+        return self._num_rank_per_node
+
+    def get_bash_setup_commands(self):
+        return_string =  "ulimit -s unlimited;\n"
```
cmcantalupo: multiline string is easier to read:

```
        return_string = '''\
ulimit -s unlimited
export OMP_PLACES=cores
.
.
.
```

integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,139 @@
...
+        return ""
+
+    def experiment_setup(self, output_dir):
+        pass
+
+    def experiment_teardown(self, output_dir):
+        pass
+
+    def trial_setup(self, run_id, output_dir):
+        exec_str = "ln -sf " + self._Wrf_path + "* " + output_dir + "/."
```
cmcantalupo: please use string format() throughout.

integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,139 @@
...
+        with open(filename) as fid:
+            for line in fid.readlines():
+                line = line.rstrip()  # remove '\n' at end of line
+                if key in line:
+                    if idx > 0: #First value skews the FOM
+                        timestep_sum += float(line.split(':')[-1].strip().split()[0])
+                    idx += 1;
+
+        avg_time = timestep_sum/idx;
+        return avg_time;
```
cmcantalupo: Please invert for now.

integration/experiment/frequency_sweep/gen_plot_freqinv_fom.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
+#
```
cmcantalupo: Please move plotting scripts not related to wrf into separate PR.

integration/experiment/frequency_sweep/gen_plot_freqinv_runtime.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python
+#
```
cmcantalupo: Please move into separate PR

integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,139 @@
...
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+
+'''
+Describes the best known configuration for WRF.
+
+On one node of mcfly, one iteration with the monitor agent takes about
+???sec and produces a trace file of about ??MB.
```
cmcantalupo: ???

integration/apps/wrf/wrf.py
cmcantalupo: I'm not sure I understand why the * globbing is being done.  It looks like there is just one file in Wrf_path?  If so why don't we know the name of it?

integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,139 @@
...
+    def experiment_teardown(self, output_dir):
+        pass
+
+    def trial_setup(self, run_id, output_dir):
+        exec_str = "ln -sf " + self._Wrf_path + "* " + output_dir + "/."
+        subprocess.check_call(exec_str, stdout=subprocess.PIPE,
+                              stderr=subprocess.PIPE, shell=True)
+
+    def trial_teardown(self, run_id, output_dir):
+        exec_str = "mv " + output_dir + "/rsl.out.0000 " + output_dir + "/rsl.out." + run_id
```
cmcantalupo: format function please

integration/apps/wrf/0001-Whitespace-cleanup.patch
```diff
@@ -0,0 +1,1168 @@
+From 4c87740e2e59135c9f9b275e4b1d9c6cda147972 Mon Sep 17 00:00:00 2001
```
cmcantalupo: Why is whitespace clean up required?  Seems like we asking for a future conflict by having this patch that serves little value.  Please remove this patch

lhlawson: Whitespace cleanup was added following the examples from hpcg & nekbone
https://github.com/geopm/geopm/blob/dev/integration/apps/hpcg/0001-Fix-whitespace-issues.patch
https://github.com/geopm/geopm/blob/dev/integration/apps/nekbone/0001-Fix-whitespace-issues.patch

integration/apps/wrf/wrf.py
lhlawson: The workload expectation is that it will be run from the wrf/run path, with everything in that directory present (namelist, restart file, input file, etc).  We could trim this down to just the list of files needed, but it would likely include everything in the wrf/run/ dir.

integration/apps/wrf/wrf.py
```diff
@@ -0,0 +1,136 @@
...
+        with open(filename) as fid:
+            for line in fid.readlines():
+                line = line.rstrip()  # remove '\n' at end of line
+                if key in line:
+                    if idx > 0: #First value skews the FOM
+                        timestep_sum += float(line.split(':')[-1].strip().split()[0])
+                    idx += 1;
+
+        avg_time = 1/(timestep_sum/idx);
+        return avg_time;
```
cmcantalupo: either rename variable, or return the inverse:

```
avg_time = timestep_sum / idx
return 1 / avg_time
```

or probably easier to read:

```
avg_rate = idx / timestep_sum
return avg_rate
```

Also "idx" is really serving the purpose of a counter, perhaps naming "idx" to "count" would be clearer

integration/config/mcfly_env.sh
```diff
@@ -0,0 +1,59 @@
...
+# This script is intended to be sourced within an existing script or shell ONLY.
+# It is NOT intended to be ./ executed.
+
+# Default Intel Toolchain compiler overrides
+# The MPI compiler wrappers are supported by Intel, but are not Intel specific.
+
+module purge
+module load autotools intel impi
+
+export CC=icc
```
cmcantalupo: CC, CXX, FC, F77 AND F90 are all unchanged from the defaults, so these are not required.

@bgeltz What do you think about having the mcfly config default to the intel compiler wrappers?

integration/apps/wrf/README.md
```diff
@@ -0,0 +1,29 @@
...
+The Weather Research and Forecasting (WRF) Model is an open source
+weather prediction system that can operate across a broad range of
+geographic scales.  We provide a series of scripts that download,
+modify configuration, build, and run the workload.
+
+The WRF build scripts use perl to change the configure file before
+build, so the generalized setup of compilers used for other apps
+in the GEOPM integration infrastructure will not work as expected.
+The configure file has been modified in the patches to use ifort,
+mpiifort, and mpiicc.  Please modify as necessary.
```
cmcantalupo: Maybe more information about how to change the compiler?   Like "change references to ifort and mpiifort in file <FILE_PATH> to enable a different toolchain"

integration/config/mcfly_env.sh
bgeltz: What is motivating this change?

If the motivation is that typically Intel MPI is used on the other large systems, and that it's more the exception than the rule that sites are using MVAPICH2 or something else, then I'm fine with switching the defaults to be Intel.

integration/apps/wrf/wrf.py
lhlawson: Updated to use count and avg_rate as above.  I've left avg_time in the script as a convenience to users more familiar with WRF's FoM (avg_time) than proposed the GEOPM FoM.

integration/config/mcfly_env.sh
lhlawson: Removed the mcfly_env.sh from the WRF integration branch so that we can focus on the system defaults separately as part of an arch or system config discussion

integration/apps/wrf/README.md
lhlawson: Updated with path required for future/different compiler changes in the WRF model.

PR #1398, opened 2020-11-04
Add uncore frequency sweep

dianarg:


integration/experiment/uncore_frequency_sweep/uncore_frequency_sweep.py
```diff
@@ -0,0 +1,158 @@
...
+                        help='the maximum core frequency setting for the sweep')
+    parser.add_argument('--min-core-frequency', dest='min_core_frequency',
+                        action='store', type=float, default=None,
+                        help='the minimum core frequency setting for the sweep')
+    parser.add_argument('--step-core-frequency', dest='step_core_frequency',
+                        action='store', type=float, default=None,
+                        help='the step size in hertz between core frequency settings for the sweep')
+    parser.add_argument("--run-max-turbo", dest="run_max_turbo",
+                        action='store_true', default=False,
+                        help='add extra runs to the experiment at maximum turbo frequency')
```
bgeltz: This is identical to the one in frequency_sweep.py now.  Should this be moved to common_args.py?

integration/experiment/uncore_frequency_sweep/uncore_frequency_sweep.py
```diff
@@ -0,0 +1,158 @@
...
+    common_args.setup_run_args(parser)
+    parser.add_argument('--max-core-frequency', dest='max_core_frequency',
+                        action='store', type=float, default=None,
+                        help='the maximum core frequency setting for the sweep')
+    parser.add_argument('--min-core-frequency', dest='min_core_frequency',
+                        action='store', type=float, default=None,
+                        help='the minimum core frequency setting for the sweep')
+    parser.add_argument('--step-core-frequency', dest='step_core_frequency',
+                        action='store', type=float, default=None,
+                        help='the step size in hertz between core frequency settings for the sweep')
```
bgeltz: It would be great if these were integrated into the common args.  I see we have options for min/max/step frequency there already, but the common args options don't specify if it's core frequency or uncore.

I'm thinking we should make the options in the common args reflect that they're for core frequency, and then have the special options here for doing uncore bounds.  Thoughts?

bgeltz (changes requested):
A few ideas for refactoring.  If you'd rather have these in a future PR, that's fine too.  Let me know.

integration/experiment/uncore_frequency_sweep/uncore_frequency_sweep.py
dianarg: yes, I think it should

integration/experiment/uncore_frequency_sweep/uncore_frequency_sweep.py
dianarg: that makes sense to me if we also update the normal frequency sweep

bgeltz (approved):

PR #1412, opened 2020-11-13
Added the PPERF register for SKX

lhlawson:

Addition of the PPERF register to the SKX MSR list, and update to MSR Whitelist testing.


cmcantalupo (approved):
LGTM.  I assume you have tested this on a machine that does not support pperf and didn't have any issues with loading the MSRIOGroup.

PR #1415, opened 2020-11-13
Add more apps and uncore sweep to smoke tests

dianarg:
- Added HPL and pennant
- Updated smoke test to use run_env.sh


integration/smoke/run_smoke.sh
```diff
@@ -31,8 +31,22 @@
...
+if [ ! -z ${GEOPM_SYSTEM_ENV} ]; then
+    source ${GEOPM_SYSTEM_ENV}
+fi
+
+GEOPM_SOURCE=${GEOPM_SOURCE:?Please set GEOPM_SOURCE in your environment.}
+source ${GEOPM_SOURCE}/integration/config/run_env.sh
+EXP_DIR=${GEOPM_SOURCE}/integration/experiment
+
+APPLICATIONS="dgemm dgemm_tiny nekbone minife amg nasft hpcg hpl_mkl hpl_netlib pennant"
+
```
bgeltz: If the lines here are identical to gen_smoke.sh lines 34-48, I think it's a good candidate for an extract refactor.  I think it would be easy to get the files out-of-sync in these sections otherwise.

Perhaps it makes sense to put this all in new ```smoke_env.sh``` in this dir and source it here?

bgeltz (changes requested):
LGTM.  Refactor request if you feel it makes sense to do it now.  If you'd rather do it later, also fine w/me.

integration/smoke/gen_smoke.sh
```diff
@@ -31,8 +31,21 @@
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
-source geopm_env.sh
-EXP_DIR=$GEOPM_SRC/integration/experiment
+# TODO: fix this
```
dianarg: rm todo

integration/smoke/run_smoke.sh
dianarg: that makes sense to me.  That may have been what "fix this" was about but I really can't remember

bgeltz (approved):

PR #1419, opened 2020-11-16
Fixes to apps from smoke tests

dianarg:


integration/apps/nekbone/nekbone.py
```diff
@@ -76,7 +76,10 @@ def trial_setup(self, run_id, output_dir):
 
     def experiment_teardown(self, output_dir):
         file_name = os.path.join(output_dir, self._input_name)
-        os.unlink(file_name)
+        try:
+            os.unlink(file_name)
+        except FileNotFoundError:
```
bgeltz: I don't understand how it's possible that the data.rea file does not exist when experiment_teardown() is called.  It should be created (or just checked to make sure is still exists) with every trial_setup().  Is there another code path somewhere that removes it?

bgeltz (changes requested):
? on Nekbone changes.  Everything else looks fine.

integration/apps/nekbone/nekbone.py
dianarg: It's probably worth retesting this.  I thought that this function was getting called multiple times, and that's why the file sometimes didn't exist, but experiment_teardown implies that it only happens once per experiment.

dianarg: Ok, I just looked at this and it can get called multiple times.  For example, we might be running an experiment like the barrier_frequency_sweep where there are multiple configs of the same app (which will use the same input file name).

bgeltz: FileNotFoundError does not exist in Python 2.  You can use IOError instead for backwards compatability.

Sorry I didn't catch this on my first run through.  I was bitten by this exact issue yesterday, so now it's fresh in my mind. :)

bgeltz (changes requested):

bgeltz (approved):

PR #1432, opened 2020-11-24
Pennant configuration changes

fkeceli:
1. Added alternative builds with different outer loop iterations to GEOPM epoch ratios. Different builds are chosen for different datasets.
2. Added more information to Pennant README.
3. Changed number of physical cores per node reserved for to be automatically deduced from the node configuration.
4. Added new command line options for Pennant run control.
5. New pennant.create_appconf(mach, args) function: Called from the run scripts to create the AppConfig. This way, we can add/remove command line options without having to change all the run scripts in the future.
6. Cleaned up Pennant constructor.
7. Added integration/apps/pennant/pennant_1_0_1.tgz to .gitignore.

### DIFF
--- a/integration/apps/pennant/README.md
+++ b/integration/apps/pennant/README.md
@@ -1,5 +1,35 @@
+# What is Pennant?
+
 From https://asc.llnl.gov/coral-2-benchmarks:
 
     PENNANT is a mini-app for hydrodynamics on general unstructured
     meshes in 2D (arbitrary polygons). It makes heavy use of indirect
     addressing and irregular memory access patterns.
+
+A detailed description of the algorithm can be found in the Pennant
+repository (downloaded to integration/apps/pennant/PENNANT by the
+build.sh script) under doc/pennantdoc.pdf.
+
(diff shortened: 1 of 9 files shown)


integration/apps/pennant/README.md
```diff
@@ -1,5 +1,35 @@
...
+cores in the node. This way, at least one physical core is reserved for
+the OS and GEOPM.
+
+# GEOPM Epoch Selection
+
+Based on the problem size we choose a different build of pennant.
+The difference between these builds is how GEOPM epochs are marked:
+
+    epoch100: One epoch every 100 outer loops iterations.
+    epoch1: One epoch every 100 outer loops iterations.
```
dianarg: what's the difference between these?

fkeceli: That was a typo. Fixed it.
    epoch100: One epoch every 100 outer loop iterations.
    epoch1: One epoch every outer loop iteration.

dianarg (approved):

PR #1434, opened 2020-11-25
ApplicationRecordLog implementation

### LINKED ISSUE #1424: Define and implement ApplicationRecordLog helper class and unit tests
ApplicationRecordLog is a wrapper around shared memory for passing region entry, region exit, epoch, and short region events.  It is used by both the Profile (application side) and the ApplicationSampler (controller side).

dianarg:
Resolves #1424


src/ApplicationRecordLog.hpp
```diff
@@ -0,0 +1,108 @@
...
+namespace geopm
+{
+    class SharedMemory;
+    class SharedMemoryUser;
+    class SharedMemoryScopedLock;
+
+    class ApplicationRecordLog
+    {
+        public:
+            static std::unique_ptr<ApplicationRecordLog> make_unique(std::shared_ptr<SharedMemory> shmem);
```
cmcantalupo, line 88: Doxygen comments required.

src/ApplicationRecordLog.cpp
```diff
@@ -0,0 +1,257 @@
...
+        return sizeof(m_layout_s);
+    }
+
+    ApplicationRecordLogImp::ApplicationRecordLogImp(std::shared_ptr<SharedMemory> shmem)
+        : m_process(-1)
+        , m_shmem(shmem)
+        , m_time_zero({{0, 0}})
+        , m_is_setup(false)
+        , m_epoch_count(0)
+    {
```
cmcantalupo, line 69: Need to check that shmem.size() >= buffer_size() in constructor

Maybe we should have a version check for the layout structure?

cmcantalupo (commented):
Comments from discussion with Daniel

src/ApplicationRecordLog.hpp
```diff
@@ -0,0 +1,108 @@
...
+            static std::unique_ptr<ApplicationRecordLog> make_unique(std::shared_ptr<SharedMemoryUser> shmem);
```
cmcantalupo: Delete me

src/ApplicationRecordLog.hpp
```diff
@@ -0,0 +1,108 @@
...
+#include <vector>
+#include <memory>
+
+#include "geopm_time.h"
+#include "record.hpp"
+
+namespace geopm
+{
+    class SharedMemory;
+    class SharedMemoryUser;
```
cmcantalupo: delete me

src/ApplicationRecordLog.cpp
```diff
@@ -0,0 +1,257 @@
...
+            GEOPM_DEBUG_ASSERT(region_idx >= 0 && region_idx < layout.num_region,
+                               "Invalid region index");
+            auto &region = layout.region_table[region_idx];
+            ++(region.num_complete);
+            region.total_time += geopm_time_diff(&(enter_info.enter_time), &time);
+        }
+        m_entered_region_hash = GEOPM_REGION_HASH_INVALID;
+    }
+
+    void ApplicationRecordLogImp::epoch(const geopm_time_s &time)
```
dannosliwcd, line 185: Should we open an issue for the finalize latency that results from high-frequency epoch events?

src/ApplicationRecordLog.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    void ApplicationRecordLogImp::dump(std::vector<record_s> &records,
+                                       std::vector<short_region_s> &short_regions)
+    {
+        // this function should not do anything with m_hash_region_enter_map
+        std::unique_ptr<SharedMemoryScopedLock> lock = m_shmem->get_scoped_lock();
+        m_layout_s &layout = *((m_layout_s *)(m_shmem->pointer()));
+        records.resize(layout.num_record);
+        std::copy(layout.record_table, layout.record_table + layout.num_record, records.begin());
```
dannosliwcd: I think this calls the constructors of each entry twice (first default-constructed, then copy-constructed). The duplicate construction could be avoided with `records.assign(layout.record_table, layout.record_table + layout.num_record)`. I assume there is a pointer specialization with a built-in reserve call to avoid unnecessary reallocations, but you could precede it with `reserve` if necessary.

src/ApplicationRecordLog.cpp
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+
+    void ApplicationRecordLogImp::append_record(m_layout_s &layout, const record_s &record)
+    {
+        int record_idx = layout.num_record;
+        // Don't overrun the buffer
+        if (record_idx < M_MAX_RECORD) {
+            layout.record_table[record_idx] = record;
```
dannosliwcd, line 256: Since record is just being copied anyways, could it be a pass-by-value instead? Then we could do `layout.record_table[record_idx] = std::move(record);`. Compilers should be easily able to perform copy elision for both of the callers' cases, since record goes in and out of scope immediately before/after passing to this function at the callers. Maybe they can even optimize away the reference in the existing implementation, so maybe the net result is the same.

I understand that this code is in a hot path, but maybe this is too micro-optimizey. Might be better left unless we can measure any improvement.

dannosliwcd (commented):
Partially-complete review. Will pick up at test/ApplicationRecordLogTest.cpp

test/ApplicationRecordLogTest.cpp
```diff
@@ -0,0 +1,383 @@
...
+    EXPECT_EQ(geopm::EVENT_EPOCH_COUNT, records[0].event);
+    EXPECT_EQ(2ULL, records[0].signal);
+    ASSERT_EQ(0ULL, short_regions.size());
+
+    m_record_log->epoch({7, 0});
+    m_record_log->exit(hash, {8, 0});
+    m_record_log->dump(records, short_regions);
+    ASSERT_EQ(2ULL, records.size());
+    EXPECT_EQ(geopm::EVENT_EPOCH_COUNT, records[0].event);
+    EXPECT_EQ(geopm::EVENT_SHORT_REGION, records[1].event);
```
dannosliwcd, line 350: Your explanation last week helped me understand this. Maybe a similar comment here might help draw attention to the significance of this line. If I understand correctly, this is expected to be promoted to a short region because the finishing region in this dump was started as any instance after a full short-running-region cycle for that hash while still in the previous dump.

dannosliwcd (commented):
Done with an initial review pass

src/ApplicationRecordLog.hpp
```diff
@@ -45,15 +45,47 @@
...
+    /// but prior to the next dump() call will be aggregated into this
+    /// short region event.
+    ///
+    /// If a region hash was observed as a short region during a
+    /// control interval (it was both entered and exited), but the
+    /// dump() call occurs prior to an exit() call that would close
+    /// the region, the subsequent exit() call to this region will be
+    /// classified as a short region event and the intervening
+    /// interval of time will be recorded in this event created by the
+    /// exit() call.
```
dianarg, line 70: maybe add something about the implications of the short region compression: the order of events in the log will not exactly match the order the regions occurred in the app

src/ApplicationRecordLog.hpp
```diff
@@ -45,15 +45,47 @@
...
     class ApplicationRecordLog
     {
         public:
+            /// @brief Factory constructor
+            /// @param shmem [in] Shared memory object of at least the
```
dianarg: `@param [in] shmem`

src/ApplicationRecordLog.hpp
```diff
@@ -45,15 +45,47 @@
...
+            ///              size returned by buffer_size().
             static std::unique_ptr<ApplicationRecordLog> make_unique(std::shared_ptr<SharedMemory> shmem);
-            static std::unique_ptr<ApplicationRecordLog> make_unique(std::shared_ptr<SharedMemoryUser> shmem);
+            /// @brief Destructor for pure virtual base class.
             virtual ~ApplicationRecordLog() = default;
+            /// @brief Called by the Profile to set the process
+            ///        identifier that will be used to tag all control
+            ///        messages.
+            /// @param process [in] The process identifier, e.g. the
```
dianarg: `@param [in] process`

src/ApplicationRecordLog.cpp
```diff
@@ -0,0 +1,267 @@
...
+        m_time_zero = time;
+    }
+
+    void ApplicationRecordLogImp::enter(uint64_t hash, const geopm_time_s &time)
+    {
+        check_setup();
+        std::unique_ptr<SharedMemoryScopedLock> lock = m_shmem->get_scoped_lock();
+        m_layout_s &layout = *((m_layout_s *)(m_shmem->pointer()));
+        check_reset(layout);
+        /// @todo: Optimize with piecewise_construct
```
dianarg: can move this todo

src/ApplicationRecordLog.hpp
dianarg: more below

test/ApplicationRecordLogTest.cpp
```diff
@@ -0,0 +1,395 @@
...
+    EXPECT_EQ(1, short_regions[0].num_complete);
+    EXPECT_EQ(4.0, short_regions[0].total_time);
+}
+
+TEST_F(ApplicationRecordLogTest, overflow_record_table)
+{
+    std::vector<record_s> records;
+    std::vector<short_region_s> short_regions;
+    int proc_id = 123;
+    //uint64_t hash = 0xABCD;
```
dianarg: rm

src/ApplicationRecordLog.cpp
cmcantalupo: https://github.com/geopm/geopm/issues/1436

cmcantalupo (approved):

src/ApplicationRecordLog.hpp
```diff
@@ -0,0 +1,257 @@
...
+                int record_idx;
+                int region_idx;
+                geopm_time_s enter_time;
+            };
+            void check_setup(void);
+            void check_reset(m_layout_s &layout);
+            void append_record(m_layout_s &layout, const record_s &record);
+            int m_process;
+            std::shared_ptr<SharedMemory> m_shmem;
+            std::map<uint64_t, m_region_enter_s> m_hash_region_enter_map;
```
dannosliwcd, line 249: I think you previously mentioned considering unordered_map, but that maybe there were some memory size concerns about it. Is that to-be-evaluated, or did it work out in favor of map? Or was that for a different change set?

PR #1439, opened 2020-12-02
Pennant GEOPM region markup

fkeceli:
- Added GEOPM markup to Pennant with two flavors: High level functions and lower level functions. The default is lower level functions and we can switch to the other flavor by editing the build.sh script in the future if needed for studies.
- Fixed a bug in pennant.py that was affecting the executable choice based on dataset.

### DIFF
--- a/integration/apps/pennant/0003-Added-GEOPM-region-markup.patch
+++ b/integration/apps/pennant/0003-Added-GEOPM-region-markup.patch
@@ -1,4 +1,4 @@
-From 31d5ec851755fb6a948b41bae9a3bacb3fd6f2ab Mon Sep 17 00:00:00 2001
+From 49a76ed325466f32036ae9f6344988108f1465f6 Mon Sep 17 00:00:00 2001
 From: Fuat Keceli <fuat.keceli@intel.com>
 Date: Tue, 1 Dec 2020 17:44:13 -0800
 Subject: [PATCH 3/3] Added GEOPM region markup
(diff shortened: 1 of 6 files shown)


integration/apps/pennant/0003-Added-GEOPM-region-markup.patch
```diff
@@ -0,0 +1,404 @@
...
+--- a/Makefile
++++ b/Makefile
+@@ -40,6 +40,12 @@ CXXFLAGS := $(CXXFLAGS_OPT)
+ ifdef USEGEOPM
+ 	LDFLAGS += $(GEOPM_LDFLAGS) $(GEOPM_LDLIBS)
+ 	CXXFLAGS += $(GEOPM_CFLAGS) -DUSEGEOPM -DEPOCH_TO_OUTERLOOP=$(EPOCH_TO_OUTERLOOP)
++ifdef USEGEOPMMARKUP
++	CXXFLAGS += -DUSEGEOPMMARKUP
++endif
++ifdef USEGEOPMMARKUP_2
```
dianarg: do we need both sets of markup? Is one of these going to become the canonical version?

fkeceli: I removed the 2nd set of markup.

dianarg (dismissed):

dianarg (approved):

PR #1442, opened 2020-12-05
Use ApplicationRecordLog in ApplicationSampler

cmcantalupo:
- Modify test constructor to take filter name and avoid
      environment() call in implementation.
    - Update ApplicationSampler unit tests to use ApplicationRecordLog mock
    - Removed checks for logic about hints related to regions.
    - Added a ApplicationSampler test with two ranks.
    - Add ApplicationSampler test for short region functionality

### DIFF
--- a/src/ApplicationIO.cpp
+++ b/src/ApplicationIO.cpp
@@ -54,16 +54,14 @@ namespace geopm
 {
     constexpr size_t ApplicationIOImp::M_SHMEM_REGION_SIZE;
 
-    ApplicationIOImp::ApplicationIOImp(const std::string &shm_key)
-        : ApplicationIOImp(shm_key,
-                           ApplicationSampler::application_sampler(),
+    ApplicationIOImp::ApplicationIOImp()
+        : ApplicationIOImp(ApplicationSampler::application_sampler(),
                            platform_io(), platform_topo())
     {
 
     }
 
-    ApplicationIOImp::ApplicationIOImp(const std::string &shm_key,
-                                       ApplicationSampler &application_sampler,
+    ApplicationIOImp::ApplicationIOImp(ApplicationSampler &application_sampler,
                                        PlatformIO &platform_io,
                                        const PlatformTopo &platform_topo)
         : m_platform_io(platform_io)
--- a/src/ApplicationIO.hpp
+++ b/src/ApplicationIO.hpp
@@ -128,9 +128,8 @@ namespace geopm
     class ApplicationIOImp : public ApplicationIO
     {
         public:
-            ApplicationIOImp(const std::string &shm_key);
-            ApplicationIOImp(const std::string &shm_key,
-                             ApplicationSampler &application_sampler,
+            ApplicationIOImp();
+            ApplicationIOImp(ApplicationSampler &application_sampler,
                              PlatformIO &platform_io,
                              const PlatformTopo &platform_topo);
             virtual ~ApplicationIOImp();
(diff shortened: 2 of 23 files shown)


src/ApplicationIO.cpp
```diff
@@ -91,6 +94,7 @@ namespace geopm
 
             if (m_application_sampler.get_io_sample() == nullptr) {
                 auto rank_per_node = m_application_sampler.get_sampler()->rank_per_node();
+
```
cmcantalupo: delete change

src/ApplicationIO.cpp
```diff
@@ -99,6 +103,7 @@ namespace geopm
 
                 platform_io().get_profileio()->connect();
             }
+            m_application_sampler.connect(m_shm_key);
```
cmcantalupo: Maybe this call should just go in the controller next to where application_io->connect() is called.

src/ApplicationIO.hpp
```diff
@@ -168,6 +169,8 @@ namespace geopm
             std::vector<bool> m_is_epoch_changed;
             bool m_is_connected;
             ApplicationSampler &m_application_sampler;
+            std::string m_shm_key;
```
cmcantalupo: left over cruft

test/ApplicationSamplerTest.cpp
```diff
@@ -56,179 +65,174 @@ class ApplicationSamplerTest : public ::testing::Test
...
-    std::vector<struct geopm_prof_message_s> message_buffer {
-    //   rank     region_id     timestamp     progress
-        {0,       region_id,    {{10,0}},     0.0},
-        {0,       epoch_id,     {{11,0}},     0.0},
-        {0,       region_id,    {{12,0}},     1.0},
-        {0,       region_id,    {{13,0}},     0.0},
-        {0,       epoch_id,     {{14,0}},     0.0},
-        {0,       region_id,    {{15,0}},     1.0},
+    uint64_t region_hash = 0xabcdULL;
+    std::vector<record_s> message_buffer {
```
dianarg: this test could have multiple ranks too

dianarg (commented):
LGTM

src/ApplicationIO.cpp
```diff
@@ -54,16 +54,14 @@ namespace geopm
 {
     constexpr size_t ApplicationIOImp::M_SHMEM_REGION_SIZE;
 
-    ApplicationIOImp::ApplicationIOImp(const std::string &shm_key)
-        : ApplicationIOImp(shm_key,
```
bgeltz, line 58: Seems like the shm_key handling here was never used previously?  I'm trying to recall why we plumbed this all in.

src/ApplicationSampler.cpp
```diff
@@ -130,133 +132,57 @@ namespace geopm
...
+        // Dump the record log from each process, filter the results,
+        // and reindex the short region event signals.
 
-        std::vector<record_s> tmp_buffer;
-        for (auto &record_it : m_record_buffer) {
-            auto &proc = get_process(record_it.process);
+        // Clear the buffers that we will be building.
+        m_record_buffer.clear();
+        m_short_region_buffer.clear();
+        // Itereate over the record log for each process
```
bgeltz, line 143: *Iterate

bgeltz (changes requested):
Nits.

src/ApplicationIO.cpp
dianarg: Maybe we had intended to pass it through and forgot to fix this?  The other objects that use shared memory independently look up the environment variable for the shm_key so it should actually be the same in practice.

bgeltz (dismissed):

PR #1448, opened 2020-12-10
Add an SSTIO object to communicate with the isst_interface driver

dannosliwcd:
This change introduces an object that interacts with isst_interface ioctls. The object is meant to provide the interface for an SSTIOGroup to use the mailbox and mmio SST controls and signals. Supporting ioctl calls enable functionality to build a map of Linux CPUs to punit cores, and to check how many messages can be bundled in a single isst_interface ioctl argument.

Co-authored-by: Diana Guttman <diana.r.guttman@intel.com>
Co-authored-by: Asma H Al-Rawi <asma.h.al-rawi@intel.com>

### DIFF
--- a/src/SSTIO.cpp
+++ b/src/SSTIO.cpp
@@ -1,0 +1,537 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+
+#include "SSTIO.hpp"
+
+#include <fcntl.h>
+#include <sys/ioctl.h>
+
+#include <algorithm>
+#include <utility>
+
+#include "Exception.hpp"
+#include "SSTIOImp.hpp"
+
+#define GEOPM_IOC_SST_VERSION _IOR(0xfe, 0, struct geopm::SSTIOImp::sst_version_s *)
+#define GEOPM_IOC_SST_GET_CPU_ID _IOWR(0xfe, 1, struct geopm::SSTIOImp::sst_mbox_interface_batch_s *)
+#define GEOPM_IOC_SST_MMIO _IOW(0xfe, 2, struct geopm::SSTIOImp::sst_mmio_interface_batch_s *)
+#define GEOPM_IOC_SST_MBOX _IOWR(0xfe, 3, struct geopm::SSTIOImp::sst_mbox_interface_batch_s *)
+
+namespace geopm
+{
+    std::shared_ptr<SSTIO> SSTIO::make_shared(uint32_t max_cpus)
+    {
+        return std::make_shared<SSTIOImp>(max_cpus);
+    }
+
+    SSTIOImp::SSTIOImp(uint32_t max_cpus)
+        : m_path("/dev/isst_interface")
+        , m_fd(open(m_path.c_str(), O_RDWR))
+        , m_batch_command_limit(0)
+    {
+        if (m_fd < 0) {
+            throw Exception("SSTIOImp failed to open SST driver",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+
+        sst_version_s sst_version;
+        int err = ioctl(m_fd, GEOPM_IOC_SST_VERSION, &sst_version);
+        if (err == -1) {
+            throw Exception("SSTIOImp::SSTIOImp() failed to get the SST driver version information",
+                            errno, __FILE__, __LINE__);
+        }
+        if (!sst_version.is_mbox_supported) {
+            throw Exception("SSTIOImp::SSTIOImp() SST driver does not support MBOX messages",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        if (!sst_version.is_mmio_supported) {
+            throw Exception("SSTIOImp::SSTIOImp() SST driver does not support MMIO messages",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        if (sst_version.batch_command_limit == 0) {
+            throw Exception("SSTIOImp::SSTIOImp() SST driver reports 0-command batch size limit",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        m_batch_command_limit = sst_version.batch_command_limit;
+
+        std::vector<sst_cpu_map_interface_s> batch_read_data;
+        batch_read_data.reserve(max_cpus);
+        for (uint32_t i = 0; i < max_cpus; ++i) {
+            batch_read_data.emplace_back(sst_cpu_map_interface_s{ i, 0 });
+        }
+        auto batch_reads = ioctl_structs_from_vector<sst_cpu_map_interface_batch_s>(
+            batch_read_data);
+
+        for (auto &batch_read : batch_reads) {
+            err = ioctl(m_fd, GEOPM_IOC_SST_GET_CPU_ID, batch_read.get());
+            if (err == -1) {
+                throw Exception("SSTIOImp::SSTIOImp() failed to get CPU map",
+                                errno, __FILE__, __LINE__);
+            }
+
+            for (size_t i = 0; i < batch_read->num_entries; ++i) {
+                m_cpu_punit_core_map.emplace(batch_read->interfaces[i].cpu_index,
+                                             // LSB indicates which hyperthread
+                                             // is mapped. Right-shift once to
+                                             // drop the hyperthread bit
+                                             batch_read->interfaces[i].punit_cpu >> 1);
+            }
+        }
+    }
+
+    int SSTIOImp::add_mbox_read(uint32_t cpu_index, uint16_t command,
+                                uint16_t subcommand, uint32_t subcommand_arg,
+                                uint32_t interface_parameter)
+    {
+        // save the stuff in the list
+        struct sst_mbox_interface_s mbox {
+            .cpu_index = cpu_index,
+            .mbox_interface_param = interface_parameter,
+            .write_value = subcommand_arg,
+            .read_value = 0,
+            .command = command,
+            .subcommand = subcommand,
+            .reserved = 0
+        };
+
+        // Stage everything in a vector for now. It will be copied to the ioctl
+        // buffer later.
+        int mbox_idx = m_mbox_read_interfaces.size();
+        auto it = std::find_if(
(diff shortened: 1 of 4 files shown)


src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+{
+    std::shared_ptr<SSTIO> SSTIO::make_shared(int max_cpus)
+    {
+        return std::make_shared<SSTIOImp>(max_cpus);
+    }
+
+    SSTIOImp::SSTIOImp(int max_cpus)
+        : m_path("/dev/isst_interface")
+        , m_fd(open(m_path.c_str(), O_RDWR))
+        , m_batch_command_limit(0)
```
dianarg, line 61: these are all empty vectors; usually we don't list out default constructed members but either way is fine

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+        }
+
+        sst_version_s sst_version;
+        int err = ioctl(m_fd, GEOPM_IOC_SST_VERSION, &sst_version);
+        if (err == -1) {
+            throw Exception("SSTIOImp::SSTIOImp() failed to get the SST driver version information",
+                            errno, __FILE__, __LINE__);
+        }
+        if (!sst_version.is_mbox_supported)
+        {
```
dianarg: {  on same line as if ()

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+            throw Exception("SSTIOImp::SSTIOImp() SST driver reports 0-command batch size limit",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        m_batch_command_limit = sst_version.batch_command_limit;
+
+        std::vector<sst_cpu_map_interface_s> batch_read_data;
+        batch_read_data.reserve(max_cpus);
+        for (int i = 0; i < max_cpus; ++i) {
+            batch_read_data.emplace_back(sst_cpu_map_interface_s{static_cast<uint32_t>(i), 0});
```
dianarg: declare `i` as uint32_t

src/SSTIO.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include <vector>
+#include <memory>
+
+namespace geopm
+{
+    class SSTIO
+    {
+        public:
+            /// Interact with the mailbox on commands that are expected to return data
```
dianarg: Add `@brief` for doxygen comments and add comments for other methods.  I don't know whether it will really be useful to describe every parameter here since they match the struct members.  Maybe when we get to the IOGroup we can explain these commands in more detail.

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,217 @@
...
+            };
+
+            struct sst_mbox_interface_batch_s
+            {
+                uint32_t num_entries;
+                sst_mbox_interface_s interfaces[1];  // TODO: might need to be array instead
+            };
+
+            template<typename ListStruct, typename CommandStruct>
+            class CommandList
```
dianarg: Did this class get used?

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,217 @@
...
+            using InnerStruct =
+                typename std::remove_all_extents<decltype(OuterStruct::interfaces)>::type;
+
+            // Given a single vector of messages to send to an ioctl, split it
+            // into multiple structs to send to that ioctl. Each InnerStruct
+            // contains a single message. Each OuterStruct contains multiple
+            // messages, with size upper-bounded by m_batch_command_limit.
+            template<typename OuterStruct>
+            std::vector<std::unique_ptr<OuterStruct> >
+                ioctl_structs_from_vector(const std::vector<InnerStruct<OuterStruct> >& commands)
```
dianarg, line 158: Stared at this function for a while and it looks fine to me.  This should fail to compile if OuterStruct doesn't have members named `num_entries` and `interfaces`.

I guess the reason for the `new` and `reinterpret_cast` is so that we don't have to redefine the same constructor for all these similar outer struct types.  Is that right?

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+            {
+                std::vector<std::unique_ptr<OuterStruct> > outer_structs;
+
+                int handled_commands = 0;
+                while (handled_commands < commands.size())
+                {
+                    size_t batch_size = std::min(
+                        m_batch_command_limit,
+                        static_cast<int>(commands.size() - handled_commands));
```
dianarg: make handled_commands a size_t?

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+                    // The inner struct is embedded in the outer struct, and
+                    // the inner struct's size depends on how many entries it
+                    // can contain. That size is dynamically determined, so we
+                    // manually allocate the outer struct here.
+                    outer_structs.emplace_back(reinterpret_cast<OuterStruct *>(
+                        new char[sizeof(OuterStruct::num_entries) +
+                                 sizeof(InnerStruct<OuterStruct>) * commands.size()]));
+                    outer_structs.back()->num_entries = batch_size;
```
dianarg, line 177: in c++17 we can use the return value from emplace :)

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,217 @@
...
+                               uint32_t read_mask) override;
+            int add_mmio_read(uint32_t cpu_index, uint16_t register_offset,
+                              uint32_t register_value) override;
+            int add_mmio_write(uint32_t cpu_index, uint16_t register_offset,
+                               uint32_t register_value, uint32_t read_mask) override;
+            // call ioctl() for both mbox list and mmio list,
+            // unless we end up splitting this class
+            void read_batch(void) override;
+
+            // TODO: might need separate call for mbox and mmio
```
dianarg: is this TODO still relevant?

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+        for (auto &batch_read : batch_reads) {
+            err = ioctl(m_fd, GEOPM_IOC_SST_GET_CPU_ID, batch_read.get());
+            if (err == -1) {
+                throw Exception("SSTIOImp::SSTIOImp() failed to get CPU map",
+                                errno, __FILE__, __LINE__);
+            }
+
+            for (size_t i = 0; i < batch_read->num_entries; ++i) {
+                m_cpu_punit_core_map.emplace(batch_read->interfaces[i].cpu_index,
+                                             // Right-shift once to drop the
```
dianarg: Can you explain this more?  punit_cpu[31:1] is a field?

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+            .cpu_index = cpu_index,
+            .mbox_interface_param = interface_parameter,
+            .write_value = 0,
+            .read_value = 0,
+            .command = command,
+            .subcommand = subcommand,
+            .reserved = 0
+        };
+        int mbox_idx = m_mbox_write_interfaces.size();
+        auto it = std::find_if(
```
dianarg, line 186: note: linear search here (same as in MSRIOGroup).  shouldn't be a problem since we have a small number of controls, and push_control is not in the hot loop.  Does it make sense to add the same check for add_mbox_read() and add_mmio*?

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+                                    errno, __FILE__, __LINE__);
+                }
+            }
+        }
+    }
+
+    uint64_t SSTIOImp::sample(int batch_idx) const
+    {
+        const auto& interface = m_added_interfaces[batch_idx];
+        uint64_t sample_value;
```
dianarg: initialize to 0

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+        if (interface.first == MMIO) {
+            size_t mmio_batch_idx = interface.second / m_batch_command_limit;
+            size_t mmio_interface_idx = interface.second % m_batch_command_limit;
+            sample_value = m_mmio_read_batch[mmio_batch_idx]->interfaces[mmio_interface_idx].value;
+        }
+        else {
```
dianarg, line 314: logic error if interface.first is not MBOX

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+        }
+    }
+
+    uint64_t SSTIOImp::sample(int batch_idx) const
+    {
+        const auto& interface = m_added_interfaces[batch_idx];
+        uint64_t sample_value;
+        if (interface.first == MMIO) {
+            size_t mmio_batch_idx = interface.second / m_batch_command_limit;
+            size_t mmio_interface_idx = interface.second % m_batch_command_limit;
```
dianarg, line 311: I don't understand the trick here with / and %

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+                // the buffer that contains the mailbox write locations (which
+                // may be different from the read locations for some controls)
+                for (size_t i = 0; i < batch->num_entries; ++i) {
+                    // Mask the read so we only propagate the bits that we are
+                    // supposed to read. Mask the write so we only update the
+                    // adjusted bits.
+                    m_mbox_write_interfaces[i].write_value |=
+                        ~m_mbox_rmw_write_masks[i] &
+                        (batch->interfaces[i].read_value &
+                         m_mbox_rmw_read_masks[i]);
```
dianarg, line 350: is there a check somewhere that these two masks don't overlap?  what are the bits not covered by either mask?

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+        err = ioctl(m_fd, GEOPM_IOC_SST_MMIO, &batch);
+        if (err == -1) {
+            throw Exception("sstioimp::write_mmio_once() mmio write failed",
+                            errno, __FILE__, __LINE__);
+        }
+    }
+
+
+    void SSTIOImp::adjust(int index, uint64_t write_value, uint64_t write_mask)
```
dianarg: rename index -> idx or batch_idx

src/SSTIO.cpp
```diff
@@ -0,0 +1,514 @@
...
+    {
+        const auto& interface = m_added_interfaces[index];
+        auto &write_destination = interface.first == MMIO
+                                ? m_mmio_write_interfaces[interface.second].value
+                                : m_mbox_write_interfaces[interface.second].write_value;
+        write_destination &= ~write_mask;
+        write_destination |= write_value;
+        if (interface.first == MBOX) {
+            m_mbox_rmw_write_masks[interface.second] |= write_mask;
```
dianarg, line 526: why are these being updated?

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,217 @@
...
+#ifndef SSTIOIMP_HPP_INCLUDE
+#define SSTIOIMP_HPP_INCLUDE
+
+#include <algorithm>
+#include <map>
+#include <memory>
+#include <string>
+
+#include "SSTIO.hpp"
+#include <type_traits>
```
dianarg: move to cpp header group

src/SSTIO.cpp
dannosliwcd: I'll expand the comment. The entire 32 bits map to an individual hyperthread. Differences in only the last bit map to the same core, but different hyperthreads,

src/SSTIO.cpp
dannosliwcd: Yeah, that makes sense. I'll add the others.

src/SSTIO.cpp
dannosliwcd: I'll add a comment. This works because ioctl_structs_from_vector ensures that all m_mmio_read_batch[] entries are the same size (the maximum of the system), except for the last one. So the divide determines which of those entries to look in, and the modulo determines how far to look within that entry.

src/SSTIO.cpp
dannosliwcd: Overlap indicates bits that will be overwritten. 100% overlap could be an opportunity to avoid an unnecessary pre-read. I think that would be possible to check in adjust(), though I'll need to add another vector or map to indicate which reads map to which writes (since those vectors will then be different lengths). Do you see a more clean way to avoid that?

Bits not covered by either mask will be zeroed. So anything we want to retain needs to be specified in the rmw_read_masks by the user of this object. In the full solution, that is ultimately provided by the SSTIOGroup, (which knows all of the fields it cares to maintain in the register).

src/SSTIO.cpp
dannosliwcd: I'll add a comment. This is so we know which bits to use in the write phase of the ioctl RMW operations.

src/SSTIOImp.hpp
dannosliwcd: Nope. I will remove it.

src/SSTIOImp.hpp
dannosliwcd: Nope. I think that was an abandoned solution for the same problem ioctl_structs_from_vector() addresses. I'll remove it.

src/SSTIOImp.hpp
dannosliwcd: Yeah. It should be applicable to any data structure that has those two members. There's probably some trickery that could be applied to also make it fail if the struct is composed of anything more than just those two, but I'm not sure (static_assert on added sizeofs jumps to mind, but I think that wouldn't account for padding).

Yeah, `reinterpret_cast<struct>(new char[])` avoids having a separate struct definition for each size OuterStruct that we want to support. A fixed-max-size struct doesn't quite work either. The maximum count of interfaces in our test system is 64, but that is determined at run time, so we can't compile in with all sized at 64 entries.

src/SSTIOImp.hpp
dannosliwcd: Yeah, that would be better. I'll change it.

src/SSTIOImp.hpp
dannosliwcd: Things will just be too easy when we switch to C++17!

dannosliwcd (commented):
Responding to questions immediately, will continue with code changes.

src/SSTIOImp.hpp
```diff
@@ -0,0 +1,210 @@
...
+                    size_t batch_size = std::min(
+                        static_cast<size_t>(m_batch_command_limit),
+                        commands.size() - handled_commands);
+
+                    // The inner struct is embedded in the outer struct, and
+                    // the inner struct's size depends on how many entries it
+                    // can contain. That size is dynamically determined, so we
+                    // manually allocate the outer struct here.
+                    outer_structs.emplace_back(reinterpret_cast<OuterStruct *>(
+                        new char[sizeof(OuterStruct::num_entries) +
```
cmcantalupo, line 175: Seems like there should be a delete somewhere?

cmcantalupo (approved):
LGTM

PR #1449, opened 2020-12-10
Add thread profile API to Profile

dianarg:
- Adds Profile::thread_init() and Profile::thread_post()
- Removes geopm_tprof_init_loop()
- Removes Profile::tprof_table()
- Removes the existing Profile unit tests; they will be replaced with new tests in a later patch


src/Profile.hpp
```diff
@@ -167,9 +167,30 @@ namespace geopm
...
+            /// @brief Update the total work for one CPU and reset the
+            ///        work completed.  This method should be called
+            ///        by all threads in the same parallel region with
+            ///        the total expected to be completed by the
+            ///        entire group.
+            /// @param [in] cpu The Linux logical CPU obtained with
+            ///             get_cpu().
+            /// @param [in] num_work_unit The total work units for all
+            ///             threads in the same parallel region.
+            virtual void thread_init(int cpu, uint32_t num_work_unit) = 0;
```
bgeltz, line 179: Was the change in precision (from how it used to be with size_t's in init_prof_loop, to all uint32_t's) just made to align everything to uint32_t?

Is this enough precision?

bgeltz (changes requested):
LGTM

There are references to ```geopm_tprof_init_loop()``` in ```ronn/geopm_fortran.3.ronn``` and ```tutorial/tutorial_region_prof.c``` that need to be cleaned up.  Happy to approve after that.

src/Profile.hpp
cmcantalupo: This is to enable a lock free implementation, we need to stick to 32 bit integers which are atomic.  Note the hints are shifted right 32 bits so they fit.

cmcantalupo (approved):

bgeltz (approved):

PR #1450, opened 2020-12-11
Remove geopm_prof_progress()

dianarg:
- Updated examples and ModelRegion to use geopm_tprof_init() and
  geopm_tprof_post()
- Updated geopm_prof_c man page.
- Changes to integration tests will come after full runtime is enabled
  with the new code path.


cmcantalupo (approved):
LGTM.  I wonder if we it is a good idea to get rename geopm_tprof_* to just geopm_prof_* since there is no need to distinguish the different progress routines?

PR #1452, opened 2020-12-12
Modify ApplicationSampler constructor to use num cpus rather than topo

cmcantalupo:


src/ApplicationStatus.cpp
```diff
@@ -39,27 +39,25 @@
 #include "SharedMemory.hpp"
 #include "Helper.hpp"
 #include "Exception.hpp"
-#include "PlatformTopo.hpp"
```
dianarg, line 42: you can remove the forward declaration for this

dianarg (approved):

PR #1453, opened 2020-12-12
Added policy recommender.

alawibaba:
Added integration/experiment/power_sweep/gen_policy_recommendation.py, which ingests a power sweep and emits a policy recommendation for the power governor based on basic models it creates of total workload runtime and energy consumption.


integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+minimum energy use subject to a runtime degradation constraint.
+'''
+
+import os
+import sys
+import argparse
+import json
+
+import numpy as np
+from sklearn.preprocessing import PolynomialFeatures
```
bgeltz: Is sklearn a default Python module?

If it is not, it must be added to https://github.com/geopm/geopm/blob/dev/scripts/requirements.txt.  This must be done for all new/required Python modules.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+    parser.add_argument('--path', help='path containing reports and machine.json', required=True)
+    parser.add_argument('--region_filter', help='comma-separated list of regions to include, default to include all regions', default=None)
+    parser.add_argument('--dump_prefix', help='prefix to dump statistics to, empty to not dump stats')
+    parser.add_argument('--min_pl', help='minimum power limit, default behavior is to use machine.json', default=-1, type=float)
+    parser.add_argument('--max_pl', help='maximum power limit, default behavior is to use machine.json', default=-1, type=float)
+    parser.add_argument('--tdp', help='tdp (sticker power limit), default behavior is to use machine.json', default=-1, type=float)
+    parser.add_argument('--max_degradation', help='maximum allowed runtime degradation, default is 0.1 (i.e., 10%%)', default=0.1, type=float)
+    parser.add_argument('--min_energy', help='ignore max degradation, just give the minimum energy possible', action='store_true')
+
+    args = parser.parse_args(argv[1:])
```
bgeltz: Consider refactoring this so this is in the ```if __name__``` section below.  Look at https://github.com/geopm/geopm/blob/dev/integration/experiment/power_sweep/gen_plot_balancer_comparison.py as an example.

There may be some common args you could use here as well.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+import os
+import sys
+import argparse
+import json
+
+import numpy as np
+from sklearn.preprocessing import PolynomialFeatures
+from sklearn.linear_model import LinearRegression
+
+from geopmpy import io as geoio
```
bgeltz: Please don't refer to this as ```geoio```.  We use ```geopmpy.io``` in all code elseware.  It only has a single occurrence in this file.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+
+def extract_from_reports(path, region_filter):
+    simplified_data = []
+    
+    try:
+        df = geoio.RawReportCollection("*report", dir_name=path).get_df()
+    except RuntimeError:
+        sys.stderr.write("<geopm> Error: No report data found in " + path + \
+                "; run a power sweep before using this analysis.\n")
+        sys.exit(1)
```
bgeltz: Please refactor this so the initial report parsing happens in the ```if __name__``` block at the bottom.  Use  https://github.com/geopm/geopm/blob/dev/integration/experiment/power_sweep/gen_plot_balancer_comparison.py as a reference.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+    
+            simplified_data.append((power_limit, total_rt, total_en))
+    return simplified_data
+
+
+def dump_simplified_data(simplified_data, fname):
+    raw_out = open(fname, "w")
+    for pl, rt, en in sorted(simplified_data):
+        raw_out.write("%d %f %f\n" % (pl, rt, en))
+    raw_out.close()
```
bgeltz: This should be refactored to use a ```with``` statement.  See: https://github.com/geopm/geopm/blob/dev/integration/experiment/power_sweep/gen_plot_balancer_comparison.py#L259-L260

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+
+    for pl in count:
+        rtx[pl] /= count[pl]
+        rtxx[pl] /= count[pl]
+        enx[pl] /= count[pl]
+        enxx[pl] /= count[pl]
+
+    stats_out = open(fname, "w")
+    for pl in sorted(count.keys()):
+        stats_out.write("%d %f %f %f %f\n" % (pl, rtx[pl], (rtxx[pl] - rtx[pl]**2)**.5, enx[pl], (enxx[pl] - enx[pl]**2)**.5))
```
bgeltz: Please refactor all of your ```write()``` calls (and any other string manipulation that uses the % syntax) to use ```"".format()```.  That's the style we use for all of our string manipulation.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+
+    rtmodel = CubicPowerModel()
+    enmodel = CubicPowerModel()
+
+    analyze_extracted_data(simplified_data, rtmodel, enmodel)
+
+    min_pl, tdp, max_pl = None, None, None
+    if min(args.min_pl, args.tdp, args.max_pl) < 0:
+        try:
+            machine_info = json.load(open(os.path.join(args.path, "machine.json")))
```
bgeltz: We have a method for getting at the machine info.  Take a look at:
https://github.com/geopm/geopm/blob/dev/integration/experiment/monitor/gen_plot_achieved_power.py#L88

This can be refactored into the ```if __name__``` section.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,287 @@
...
+    if args.tdp > 0:
+        tdp = args.tdp
+    if args.max_pl > 0:
+        max_pl = args.max_pl
+
+    plrange = [min_pl + i for i in range(int(max_pl - min_pl))] + [max_pl]
+    bestpl, bestrt, besten = policy_min_energy(plrange, enmodel, rtmodel, tdp, max_degradation=(None if args.min_energy else args.max_degradation))
+    tdprt, tdpen = rtmodel.test(tdp), enmodel.test(tdp)
+
+    print("AT TDP = %dW, RUNTIME = %f s,  ENERGY = %f J" % (tdp, tdprt, tdpen))
```
bgeltz: Please don't use ```print```.  Use only ```sys.stdout.write``` or ```sys.stderr.write```.

Print is not portable between Python 2 and 3.  We're trying to keep ```print``` out of the codebase.

bgeltz (changes requested):
I see a few refactors here that should clean this up a bit.

I evaluated more on style and form than on logic since there's no docu.  I think I'd benefit from a live session on a logic walk through for how this algorithm works if you'd like me to evaluate this on correctness.

Also, there may be some ways to simplify the data flow from report parsing on down if I understand what all you need.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: D'oh, I meant to fix this. Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Ok, I moved a lot of the main function into the if __name__ section down below; I tried to keep the argument parsing and file input in that section, but the function calls and logic in the main function. I hope the separation makes sense.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: mea culpa

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+            'best': best_policy}
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    common_args.add_min_power(parser)
+    common_args.add_max_power(parser)
+    parser.add_argument('--path', required=True,
+                        help='path containing reports and machine.json')
+    parser.add_argument('--region_filter', default=None,
```
dianarg: these can be `--region-filter` (with hyphen instead of underscore) if you add a `dest` parameter

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+                        help='comma-separated list of regions to include, '
+                             'default to include all regions')
+    parser.add_argument('--dump_prefix',
+                        help='prefix to dump statistics to, empty to not dump '
+                             'stats')
+    parser.add_argument('--tdp', default=-1, type=float,
```
dianarg: can this just use None for default, and check for None on 304 below?  any other place we can get negative values?

integration/smoke/run_smoke.sh
```diff
@@ -58,8 +58,10 @@ function run_all_power_sweep {
     APPLICATIONS="dgemm dgemm_tiny nekbone minife amg nasft hpcg"
     EXP_SUBDIR=power_sweep
     EXP_TYPE=power_sweep
-    ARGS="--min-power=220 --max-power=230"
+    ARGS="--min-power=190 --max-power=230"
     run_all
+    python3 ${EXP_DIR}/power_sweep/gen_policy_recommendation.py \
```
dianarg: It would be better to put this in the gen_smoke script if possible.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+import geopmpy.io
+
+from experiment import common_args
+from experiment import machine
+
+class PowerLimitModel:
+    "Abstract class for predicting a statistic based on a power limit."
+    def __init__(self):
+        "Default constructor."
+        raise NotImplemented()
```
cmcantalupo: I think this should be
```
raise NotImplementedError
```

I tried:

```
>>> raise NotImplemented()
  File "<stdin>", line 1, in <module>
TypeError: 'NotImplementedType' object is not callable
```

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+        self._model = None
+
+    def train(self, df, key):
+        self._model = Polynomial.fit(
+                df.index,
+                df[key].values,
+                deg=self._degree).convert()
+
+    def test(self, PL):
+        return sum(self._model.coef * PL**np.arange(self._degree+1))
```
cmcantalupo: space between binary operators please:

```
sum(self._model.coef * PL ** np.arange(self._degree + 1))
```

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def train(self, df, key):
+        """
+        Train this model to fit the data of column key contained in the
+        dataframe (df), which is indexed by power limit. This modifies
+        the instance it is called on, preparing it for calls of the
+        methods test and batch_test."""
+        raise NotImplemented()
+
+    def test(self, PL):
```
cmcantalupo: Maybe a different name for the method?  I would reserve "test" for unit test related methods (it is a key for unittest).  Is it returning a time?  Pehaps the name should reflect what it is returning? 

```
def time_estimate(self, power_limit):

File: integration/experiment/power_sweep/gen_policy_recommendation.py [hunk 15]
```diff
@@ -0,0 +1,350 @@
...
+        """
+        Evaluate a trained model on a single power limit (PL). Returns
+        the model's prediction for this power limit."""
+        raise NotImplemented()
+
+    def batch_test(self, PL_list):
```
cmcantalupo: batch_time_estimate()

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+                deg=self._degree).convert()
+
+    def test(self, PL):
+        return sum(self._model.coef * PL**np.arange(self._degree+1))
+
+    def __str__(self):
+        if self._model is not None:
+            terms = ["", " PL"] + \
+                    [" PL ** {}".format(i)
+                            for i in range(2, self._degree+1)]
```
cmcantalupo: ws

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+        else:
+            best_runtime = None
+        return {'power': best_pl,
+                'runtime': best_runtime,
+                'energy': best_energy}
+    else:
+        rt_predictions = rtmodel.batch_test(plrange)
+        rt_at_tdp = rtmodel.test(pltdp)
+        constrained_values = [
+                (energy, pl, runtime)
```
cmcantalupo: I suppose there is no chance that energy will be equal for two evaluations, but if it were, I would order the tuple

(energy, runtime, pl)

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+    """
+    The main function. full_df is a report collection dataframe,
+    region_filter is a list of regions to include, dump_prefix
+    a filename prefix for debugging output (if specified), min_pl,
+    max_pl, tdp are the minimum, maximum, and reference power limits,
+    respectively, min_energy is a flag indicating whether or not
+    minimum energy is sought not subject to a constraint on runtime
+    (True if runtime should be ignored) and max_degradation specifies
+    what maximum runtime degradation is accepted."""
+    df = extract_columns(full_df, args.region_filter)
```
cmcantalupo: how is args defined? I think this should just be region_filter

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -0,0 +1,350 @@
...
+
+    runtime_model = CubicPowerModel()
+    energy_model = CubicPowerModel()
+
+    runtime_model.train(df, key='runtime')
+    energy_model.train(df, key='energy')
+
+    plrange = [min_pl + i for i in range(int(max_pl - min_pl))] + [max_pl]
+    best_policy = policy_min_energy(plrange, energy_model, runtime_model, tdp,
+            max_degradation=(None if args.min_energy else args.max_degradation))
```
cmcantalupo: would be more readable if max_degradation were a parameter to main(), and args is not defined in this context anyway.

cmcantalupo (commented):
Looking pretty good, let me know if you want to discuss any of the comments over the phone.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done throughout.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/smoke/run_smoke.sh
alawibaba: Did I do this right?

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Fixed. I think NotImplemented is a singleton that I misused here.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: I was thinking of the machine learning "train vs test" jargon, but what you're saying makes sense. I changed it to "evaluate" -- which is used in the docstring anyway. Sound ok?

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: I don't want to do "time" because it's not always time! I changed it to batch_evaluate.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: I like it, fixed.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Good catch.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Fixed... but that meant moving this line down into the `if __name__` block. Is that ok or did you have something else in mind?

alawibaba:
Hey guys, thanks for the comments. I think I addressed everything with a couple of questions. Please take a look when you have a chance!

integration/smoke/run_smoke.sh
dianarg: yes, it's in the right file now, thanks.

dianarg (dismissed):

dianarg (commented):
Smoke tests need to be rebased on dev; let's retry after merge

bgeltz (approved):

PR #1454, opened 2020-12-14
Updates to Profile to use ApplicationRecordLog and ApplicationStatus 

dianarg:
- Region and thread info uses the new code path.
- Handshake setup and teardown uses the old code path.

### DIFF
--- a/test/ProfileTest.cpp
+++ b/test/ProfileTest.cpp
@@ -1,0 +1,245 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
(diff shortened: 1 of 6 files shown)


test/ProfileTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+    EXPECT_CALL(*m_status, set_hint(3, 0ULL));
+    // progress is cleared when exiting top-level region
+    EXPECT_CALL(*m_status, set_total_work_units(2, 0));
+    EXPECT_CALL(*m_status, set_total_work_units(3, 0));
+
+    m_profile->exit(region_id);
+}
+
+// TODO: get rid of GEOPM_REGION_ID_MPI, epoch bit if still there
+// TODO: fix geopm_mpi_region_enter/exit to set hint instead and
```
cmcantalupo, line 151: Remove comment

cmcantalupo (approved):

PR #1464, opened 2020-12-18
Update SampleAggregator to use Accumlators

cmcantalupo:


dianarg (commented):
Leaving a comment for when we come back to this.  Where does the knowledge about which signals are summed and which are averaged come from?  We talked about maybe having a new IOGroup API.  The existing agg_function is about aggregating domains, this is about how we agg over time.

test/SampleAggregatorTest.cpp
```diff
@@ -250,31 +249,29 @@ TEST_F(SampleAggregatorTest, sample_total)
...
-        EXPECT_EQ(exp_energy[0][region], m_agg->sample_total(M_SIGNAL_ENERGY_0, region));
-        EXPECT_EQ(exp_energy[1][region], m_agg->sample_total(M_SIGNAL_ENERGY_1, region));
-        EXPECT_EQ(exp_cycles[0][region], m_agg->sample_total(M_SIGNAL_CYCLES_0, region));
-        EXPECT_EQ(exp_cycles[1][region], m_agg->sample_total(M_SIGNAL_CYCLES_1, region));
-        EXPECT_EQ(exp_cycles[2][region], m_agg->sample_total(M_SIGNAL_CYCLES_2, region));
-        EXPECT_EQ(exp_cycles[3][region], m_agg->sample_total(M_SIGNAL_CYCLES_3, region));
+        EXPECT_EQ(exp_time[0][region], m_agg->sample_region(M_SIGNAL_TIME, region));
+        EXPECT_EQ(exp_energy[0][region], m_agg->sample_region(M_SIGNAL_ENERGY_0, region));
+        EXPECT_EQ(exp_energy[1][region], m_agg->sample_region(M_SIGNAL_ENERGY_1, region));
+        EXPECT_EQ(exp_cycles[0][region], m_agg->sample_region(M_SIGNAL_CYCLES_0, region)) << region;
```
cmcantalupo: either make the region print verbose and for all errors, or delete.

src/SampleAggregator.cpp
```diff
@@ -61,108 +58,320 @@ namespace geopm
...
+                signal.app_accum->update(delta);
+                // If the epoch count has changed, call the exit/enter
+                if (epoch_count != signal.epoch_count_last) {
+                    if (signal.epoch_count_last != 0) {
+                        signal.epoch_accum->exit();
+                    }
+                    signal.epoch_accum->enter();
+                }
+                // If we have observed our first epoch, update epoch totals
+                if (signal.epoch_count_last != 0) {
```
dianarg, line 202: doesn't this skip updating for the first sample in the first epoch, since epoch is not updated until 206?

dianarg (commented):
Unit tests for averages would be nice; maybe we can come back to it.

dianarg (approved):

src/SampleAggregator.cpp
cmcantalupo: Same logic as for region entry/exit

PR #1465, opened 2020-12-29
Update Reporter to use new code path

dianarg:


dianarg:
Already included in #1468

PR #1471, opened 2021-01-04
Small fixes for new-profile

dianarg:


src/ProfileIOGroup.cpp
```diff
@@ -198,67 +215,68 @@ namespace geopm
...
-            case M_SIGNAL_REGION_HASH:
-                result = hash_to_signal(m_per_cpu_hash[cpu_idx]);
-                break;
-            case M_SIGNAL_REGION_HINT:
-                result = m_per_cpu_hint[cpu_idx];
-                break;
-            case M_SIGNAL_THREAD_PROGRESS:
-                result = m_per_cpu_progress[cpu_idx];
+        uint64_t result = GEOPM_REGION_HINT_UNSET;
+        switch (signal_type) {
```
cmcantalupo: Maybe we should use a map or vector instead of a switch?

cmcantalupo (commented):
Looks good to me.  One comment only.

PR #1480, opened 2021-01-08
Integration infrastructure fixes/enhancements

bgeltz:
- Backup cloned git repos and archives automatically
- Allow for args to be passed to the ```git clone``` command; for checking out out specific git branches
- Build optimization: only unpack archives the first time
- And env for SMNG.
- Bug fix for ```gen_slurm.sh```
- Add UNCORE status and ctl registers to default trace_signals for frequency_sweep and monitor
- Add "checkpoint/restart" to ```launch_util.py```
- Add trial retry logic to ```launch_util.py```

### DIFF
--- a/integration/apps/build_func.sh
+++ b/integration/apps/build_func.sh
@@ -138,7 +138,10 @@ unpack_archive() {
     if [ "${ARCHIVE##*.}" == zip ]; then
         unzip ${ARCHIVE}
     else
-        tar xvf ${ARCHIVE}
+        local DIRNAME=$(tar -tf ${ARCHIVE} | head -1 | cut -f1 -d"/")
+        if [ ! -d ${DIRNAME} ]; then
+            tar xvf ${ARCHIVE}
+        fi
     fi
 }
 
--- a/integration/apps/build_func.sh
+++ b/integration/apps/build_func.sh
@@ -73,9 +73,21 @@ clone_repo_git(){
         return 1
     fi
 
+    backup_archive ${ARCHIVE} ${DIRNAME}
+}
+
+backup_archive() {
+    local ARCHIVE=$1
+    local DIRNAME=$2
+
     # If the user set the cache dir, pack up the source for next time
     if [ -d ${GEOPM_APPS_SOURCES} ]; then
-        tar czvf ${GEOPM_APPS_SOURCES}/${ARCHIVE} ${DIRNAME}
+        if [ ! -f ${ARCHIVE} ]; then
+            tar czvf ${ARCHIVE} ${DIRNAME}
+        fi
+        if [ ! -f ${GEOPM_APPS_SOURCE}/${ARCHIVE} ]; then
+            cp ${ARCHIVE} ${GEOPM_APPS_SOURCES}
+        fi
     else
         echo "Warning: Please set GEOPM_APPS_SOURCES in your environment to enable the local source code cache."
     fi
@@ -114,9 +126,7 @@ get_archive() {
         elif [ $# -eq 2 ]; then
             local URL=$2
             wget ${URL}/${ARCHIVE}
-            if [ -d ${GEOPM_APPS_SOURCES} ]; then
-                cp ${ARCHIVE} ${GEOPM_APPS_SOURCES}
-            fi
+            backup_archive ${ARCHIVE}
         fi
     fi
 }
--- a/integration/apps/build_func.sh
+++ b/integration/apps/build_func.sh
@@ -57,9 +57,10 @@ clone_repo_git(){
     local REPOURL=$1
     local DIRNAME=$2
     local TOPHASH=$3
+    local CLONEARGS=$4
     local ARCHIVE=${DIRNAME}_${TOPHASH}.tgz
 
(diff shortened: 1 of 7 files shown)


integration/apps/build_func.sh
```diff
@@ -73,9 +74,21 @@ clone_repo_git(){
...
+}
+
+backup_archive() {
+    local ARCHIVE=$1
+    local DIRNAME=$2
+
     # If the user set the cache dir, pack up the source for next time
     if [ -d ${GEOPM_APPS_SOURCES} ]; then
-        tar czvf ${GEOPM_APPS_SOURCES}/${ARCHIVE} ${DIRNAME}
+        if [ ! -f ${ARCHIVE} ]; then
```
cmcantalupo, line 86: Can you add some comments about why these changes are required, it is not really clear to me what the effect of these changes is.

integration/config/smng_env.sh
```diff
@@ -0,0 +1,64 @@
...
+
+export MPIFORT=mpiifort
+export MPIFC=mpiifort
+export MPIF77=mpiifort
+export MPIF90=mpiifort
+
+export GEOPM_USER_ACCOUNT=h1906
+export GEOPM_SYSTEM_DEFAULT_QUEUE=test
+export GEOPM_SBATCH_EXTRA_LINES="
+#SBATCH --constraint=\"noscratch\"
```
cmcantalupo, line 57: On mose systems the env.sh file is simply sourced prior to submitting a job, I'm not sure what these #SBATCH lines are doing in that context.

integration/experiment/frequency_sweep/frequency_sweep.py
```diff
@@ -86,7 +86,8 @@ def report_signals():
 
 
 def trace_signals():
-    return []
+    return ['MSR::UNCORE_PERF_STATUS:FREQ@package', 'MSR::UNCORE_RATIO_LIMIT:MAX_RATIO@package',
```
cmcantalupo, line 89: I don't think traces are enabled for the frequency sweep run by default.

integration/experiment/uncore_frequency_sweep/uncore_frequency_sweep.py
```diff
@@ -95,7 +95,8 @@ def report_signals():
 
 
 def trace_signals():
-    return []
+    return ['MSR::UNCORE_PERF_STATUS:FREQ@package', 'MSR::UNCORE_RATIO_LIMIT:MAX_RATIO@package',
```
cmcantalupo, line 98: Are traces enabled?

integration/experiment/launch_util.py
```diff
@@ -93,17 +94,20 @@ def launch_run(agent_conf, app_conf, run_id, output_dir, extra_cli_args,
     num_ranks = app_conf.get_rank_per_node() * num_nodes
     cpu_per_rank = app_conf.get_cpu_per_rank()
 
-    launcher = geopmpy.launcher.Factory().create(argv,
-                                                 num_node=num_nodes,
-                                                 num_rank=num_ranks,
-                                                 cpu_per_rank=cpu_per_rank)
-    launcher.run()
+    # Attempt to run if there are no valid report files for the current configuration
+    if not (os.path.isfile(report_path) and os.path.getsize(report_path) > 0):
```
cmcantalupo, line 98: I thought this was managed by the creation of the machine file.  You cannot create a machine file if reports are present...

cmcantalupo (commented):
Some quick feedback.  Might have to sort out some of the script changes over slack.

integration/apps/build_func.sh
bgeltz: This one is easier if you use the split view instead of unified.  I refactored this code that was responsible for backing up downloads in GEOPM_APPS_SOURCES.  Previously there were 2 different backup paths, one in clone_repo_git() and one in get_archive().

They were doing almost the same operations, so I did an extract refactor so all the backup code is in the same place.

backup_archive() is just a refactor.  No functional change.

integration/config/smng_env.sh
bgeltz: These lines are used when SBATCH scripts are generated using gen_slurm.sh.  This is the only time they are used.

integration/experiment/frequency_sweep/frequency_sweep.py
bgeltz: They are not, that's true.  I was hoping to get this in since it's very commonly requested when we start digging into data.  When someone turns traces on, it would be nice to not have to remember to also add the UNCORE fields.

integration/experiment/launch_util.py
bgeltz: That has not been my experience with the runs I did in December.  This was required.

integration/experiment/uncore_frequency_sweep/uncore_frequency_sweep.py
bgeltz: No, not by default.  I was hoping to set the trace_signals() up for when they are enabled.

cmcantalupo (approved):

PR #1481, opened 2021-01-08
Add man page describing the reports

dianarg:

### DIFF
--- a/ronn/geopm_report.7.ronn
+++ b/ronn/geopm_report.7.ronn
@@ -1,0 +1,288 @@
+geopm_report(7) -- geopm summary report file
+============================================
+
+## DESCRIPTION
+
+The GEOPM report is a YAML file containing summarized information
+about runtimes, network usage, energy consumption, and other data
+about an application's behavior on every compute node where it ran.
+The report is broken into sections with totals for each compute node,
+and further divided into sections detailing each application region
+and epoch.  By default, the report will be saved into a file called
+"geopm.report", but this name can be customized with the GEOPM_REPORT
+environment variable; refer to the description for `--geopm-report` in
+**geopmlaunch(1)**.
+
+The application regions and epoch are defined by use of the
+**geopm_prof_c(3)** interface to mark up the user application, or
+through automatic inference of regions based on interposing on the MPI
+or OpenMP interfaces (interposing on OpenMP requires that OMPT is
+enabled at GEOPM compile time, and the `--geopm-ompt-disable` option
+is not provided to `geopmlaunch`). Alternatively, epochs may be
+inserted automatically when `--geopm-record-filter` is used, as
+described in **geopmlaunch(1)**.
+
+## NOTES ON SAMPLING
+
+Most data in the report is derived from PlatformIO signals (described
+in **geopm::PlatformIO(3)**) sampled at a rate determined by the
+Agent::wait() implementation for the active agent (see
+**geopm::Agent(3)**).  Every sample interval, all values of interest
+to the report are read, and the delta values between the last sample
+and current sample are calculated.  This delta is attributed to the
+current region hash for the corresponding domain.  By default,
+synchronous values in the report are aggregated to the board domain,
+so deltas will be attributed to a region only when all active CPUs on
+the compute node are in that region.  User extensions through
+GEOPM_REPORT_SIGNALS may be aggregated to smaller domains.  If any
+CPUs within the domain are in different regions at the time of the
+sample, the delta value will be attributed to the unmarked region.
+Only active CPUs assigned to the application are considered when
+determining the region to attribute values to.
+
+Values that are the result of sampling (most fields except 'runtime'
+and 'count') may include small contributions from adjacent regions,
+especially if regions are very short.  The amount of slop depends on
+the controller sample rate and the number of region entries.  Sampled
+values should not be compared with the per-process average runtime
+`runtime (s)`; use `sysn-runtime (s)` instead.
+
+Refer to the EXAMPLES section below.
+
+
+## REPORT FORMAT
+
+* `Header Fields`:
+
+  `GEOPM Version`: The version of the GEOPM library and tools used
+                   for the run.  Note that if building from source,
+                   ./autogen.sh must be rerun to update the version
+                   to the latest git SHA.
+
+  `Start Time`: The start time of the job recorded by the GEOPM
+                controller.
+
+  `Profile`: The profile string, obtained from the GEOPM_PROFILE
+             environment variable.
+
+  `Agent`: The name of the agent, obtained from the GEOPM_AGENT
+           environment variable.
+
+  `Policy`: A JSON string containing the policy values as set in
+            GEOPM_POLICY, or "DYNAMIC" if the policy was received
+            through the endpoint.
+
+
+* `Per-host Region, Epoch, and Application Totals`:
+
+Under the `Hosts` section of the report, each host in the job creates
+a section with its hostname as the key.  Each host's section contains
+`Application Totals`, the totals for each metric from the beginning to
+the end of the job; `Epoch Totals`, the totals from the first detected
+epoch to the end of the job; and `Regions`, the summarized data for
+each region.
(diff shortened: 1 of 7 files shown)


ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+sample, the delta value will be attributed to the unmarked region.
+Only active CPUs assigned to the application are considered when
+determining the region to attribute values to.
+
+Values that are the result of sampling (most fields except 'runtime'
+and 'count') may include small contributions from adjacent regions,
+especially if regions are very short.  The amount of slop depends on
+the controller sample rate and the number of region entries.  Sampled
+values should not be compared with the per-process average runtime
+`runtime (s)`; use `sysn-runtime (s)` instead.
```
fkeceli: Typo? (sysn --> sync)

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+current region hash for the corresponding domain.  By default,
+synchronous values in the report are aggregated to the board domain,
+so deltas will be attributed to a region only when all active CPUs on
+the compute node are in that region.  User extensions through
+GEOPM_REPORT_SIGNALS may be aggregated to smaller domains.  If any
+CPUs within the domain are in different regions at the time of the
+sample, the delta value will be attributed to the unmarked region.
+Only active CPUs assigned to the application are considered when
+determining the region to attribute values to.
+
```
fkeceli, line 46: Can you add a note for sync-runtime vs runtime? Sync runtime is sum(delta(time)) for all samples that all active cores were in the same region so it is sample based vs runtime is async wrt GEOPM sampling intervals.

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+                           region hint was GEOPM_REGION_HINT_UNKNOWN.
+                           The region hint is determined by the hint
+                           passes to `geopm_region()` for the region
+                           and any nested regions.
+
+  `time-hint-unset (s)`: The portion of sync-runtime where the region
+                         hint was GEOPM_REGION_HINT_UNSET.  The region
+                         hint is unset when exiting a region
+                         (i.e. when a CPU is in the unmarked region).
+
```
fkeceli, line 204: Maybe a bit more explicit here: I time-hint-unset only possible when geopm controller considers the app to be in the unmarked region? Why not call this something like time-hint-unmarkedregion then (less confusing)?

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+Agents can add keys to the report header, host section, or individual
+region sections using the `report_header()`, `report_host()` or
+`report_region()` methods respectively.  See **geopm::Agent(3)** for
+more information about the report extensions available to agents.
+
+
+## EXAMPLES
+
+Suppose we have a system with 2 sockets per node and 2 CPUs per packge
```
fkeceli: typo: package

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+(socket) (assume 1 CPU per core for simplicity).  The application
```
fkeceli: assume 1 CPU per core for simplicity, i.e. hyper-threading is turned off if applicable

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+epoch, and the current region hash for sampling.
+
+The following is an example of the regions being entered and exited on
+each CPU.  The sample rate is 0.002 s; the region hash for the
+REGION_HASH sample at larger domains is shown on the right side of the
+table; "-" indicates the unmarked region.  Note that in a real report,
+the entry and exit times used to calculate the "runtime" may not
+line up exactly with the sample boundaries, and samples may not be
+exactly the same throughout the run.
+
```
fkeceli, line 246: In a separate section, maybe explain how GEOPM sampling works: GEOPM might take longer than 2 msecs to sample and it will try to make up for it at the next sample so average is back to 2msecs.

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+0.012     finalize: report generated ||   -       -      -
+
+A subset of the report is shown below.  The "runtime" and "count"
+fields are averaged across the 4 CPUs.  The user extensions for
+"TIME@package" correspond to the sync runtime for each package.
+
+  Regions:
+    -
+    "name": "A",
+    "runtime": 0.00375,   // average of [0.002, 0.004, 0.004, 0.005]
```
fkeceli, line 270: For the uninitiated calculation of the runtime here might be misleading since it looks like we are taking the sample times. I agree that there is no easy way to do that here but at least have a footnote emphasizing that runtime is not calculated based on sample.

fkeceli (changes requested):

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+Under the `Hosts` section of the report, each host in the job creates
+a section with its hostname as the key.  Each host's section contains
+`Application Totals`, the totals for each metric from the beginning to
+the end of the job; `Epoch Totals`, the totals from the first detected
+epoch to the end of the job; and `Regions`, the summarized data for
+each region.
+
+The fields in each of these sections are described below:
+
+  `name`: Name of the region.  For user-defined regions, this is the
```
cmcantalupo, line 91: Should now be "region:"

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+          string passed to `geopm_region()`; for automatically
+          detected OpenMP and MPI regions this is the function name.
+          See **geopm_prof_c(3)** for more information.
+
+  `hash`: The CRC32 hash of the region name.  This value is used by
```
cmcantalupo, line 96: maybe refer to geopm_hash_crc32str() rather than the generic CRC32 hash (which is somehow diffferent than the output of the x86 intrinsic).

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+           The count has no meaning for the unmarked region and
+           application totals.
+
+  `sync-runtime (s)`: Total time for which the sampled region hash
+                      matched this region on all CPUs on the compute
+                      node.  For epoch and application totals, this
+                      value is the same as `runtime (s)`.  The
+                      sync-runtime for the unmarked region represents
+                      the total time for which the ranks on a compute
+                      node were in different regions (i.e. unaligned).
```
cmcantalupo: "different regions" -> "not in the same region"

this will cover the ambiguity of the case where one or more of the ranks is not in a region, but other ranks are in the same region.

ronn/geopm_report.7.ronn
```diff
@@ -0,0 +1,288 @@
...
+  `frequency (Hz)`: Achieved core frequency for the processor in
+                    hertz.  This frequency is calculated using the
+                    ratio of CYCLES_THREAD to CYCLES_REFERENCE times
+                    the sticker (base) frequency.
+
+  `time-hint-network (s)`: The portion of sync-runtime where the
+                           region hint was GEOPM_REGION_HINT_NETWORK.
+                           The region hint is determined by the hint
+                           passes to `geopm_region()` for the region
+                           and any nested regions.
```
cmcantalupo: maybe it could be clearer that the most nested region determines the hint value for any sample.  Also passes -> passed.

cmcantalupo (commented):
A few comments, but this is a big improvement as it is to our existing documentation

ronn/geopm_report.7.ronn
dianarg: It is also possible for a user to create a region with GEOPM_REGION_HINT_UNSET if they really want to.

ronn/geopm_report.7.ronn
dianarg: added this to notes on sampling

PR #1482, opened 2021-01-11
Fixes for smoke tests

dianarg:


integration/smoke/run_smoke_parallel.sh
```diff
@@ -0,0 +1,137 @@
...
+        else
+            echo "${RUN_SCRIPT} not found; skipping"
+            # TODO: danger!
+            eval ${SKIP_RESULT}
+        fi
+    done
+
+done
+
+################
```
dianarg: rm

integration/smoke/run_smoke_parallel.sh
```diff
@@ -0,0 +1,166 @@
...
+EOF
+        # Remove blank lines
+        uniq ${SBATCH_NAME} .${SBATCH_NAME}
+        mv .${SBATCH_NAME} ${SBATCH_NAME}
+    else
+        SBATCH_NAME=""
+    fi
+}
+
+for EXP_TYPE in $ALL_EXP; do
```
bgeltz, line 213: A comment here denoting that this is really the main() section would help.

integration/smoke/run_smoke_parallel.sh
```diff
@@ -0,0 +1,166 @@
...
+
+    NUM_NODES=2
+    for APP in ${APPS_2NODE}; do
+        gen_sbatch
+        if [ ! -z $SBATCH_NAME ]; then
+            sbatch ${SBATCH_NAME}
+            echo "${SBATCH_NAME} submitted"
+        else
+            echo "${RUN_SCRIPT} not found; skipping"
+            # TODO: danger!
```
bgeltz: What's dangerous here?

integration/smoke/run_smoke_parallel.sh
```diff
@@ -0,0 +1,166 @@
...
+            eval ${SKIP_RESULT}
+        fi
+    done
```
bgeltz: Seems like this block is repeated from above.  It's probably a good candidate for an extract refactor into a function.

bgeltz (approved):

bgeltz (changes requested):
Perhaps we should have a review session to explore how to do this "right".

I'd really like to rewrite these functions so that there are no global vars.  It's a little difficult to follow because of that.  I bet we could either sit for an hour and rewrite this with a slightly different structure (still in BASH) or ditch BASH for Python in the hope that would help clean some things up.

integration/smoke/run_smoke_parallel.sh
```diff
@@ -0,0 +1,213 @@
...
+    # Generates the name of the run script for a given combination of
+    # experiment type and application.
+    #
+    # Inputs:
+    #   $1: name of the application
+    #   $2: type of experiment
+    #   GEOPM_SOURCE (global): root directory of the GEOPM source code
+    #
+    # Returns: name of the run script in integration/experiments that
+    # corresponds
```
dianarg: finish this comment

integration/smoke/run_smoke_parallel.sh
```diff
@@ -0,0 +1,213 @@
...
+        EXP_DIR="energy_efficiency"
+    fi
+    echo "${GEOPM_SOURCE}/integration/experiment/${EXP_DIR}/run_${2}_${1}.py"
+}
+
+function gen_sbatch {
+    # Creates a slurm batch script to run the given smoke test
+    # experiment.
+    #
+    # TODO: This function has some copied code from
```
dianarg, line 112: that script has some extra slurm options like account that might be useful to move here

bgeltz: It would be nice if you could call that script here instead of copying a big chunk of the code.

bgeltz (approved):

PR #1492, opened 2021-01-14
Enable --geopm-report-signals to support averaged signals

cmcantalupo:


src/MSRIOGroup.cpp
```diff
@@ -732,6 +732,27 @@ namespace geopm
         return result;
     }
 
+    int MSRIOGroup::signal_behavior(const std::string &signal_name) const
+    {
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("MSRIOGroup::signal_behavior(): signal_name " + signal_name +
+                            " not valid for MSRIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
bgeltz, line 742: Does every MSR field have an associated behavior defined?  If yes, then this should read something like:
```
MSRIOGroup::signal_behavior(): signal_name " + signal_name +
" is missing behavior field.",
```

bgeltz (changes requested):
Code changes look fine AFAICT.  Exception text change requested.

Also, I was expecting so see some manpage updates here as well.  At a minimum these should be updated with the API changes:
```
GEOPM_CXX_MAN_PlatformIO.3.ronn
GEOPM_CXX_MAN_IOGroup.3.ronn
GEOPM_CXX_MAN_MSRIOGroup.3.ronn
```

Additionally, I went looking for some docu about the msr_data_* files and did not find anything.  Do we have anything in the manpages describing the structure of these files that I missed?

src/MSRIOGroup.cpp
dianarg: This is checked in the JSON parsing step for any MSRs/fields in the JSON files (see `check_msr_field()`).  This exception here should not really be possible to hit through PlatformIO because PlatformIO checks the available signal names before forwarding to the IOGroup.  This exception would only happen if a user managed to create an MSRIOGroup standalone and pass it an invalid name.  There is a similar exception for all the other methods that take signal names.

cmcantalupo: Opened #1495

PR #1500, opened 2021-01-15
Allow MSRIOGroup to be used on systems with restricted allow lists

### LINKED ISSUE #1395: Inclusion of programmable counter MSRs causes a crash on systems where the offsets are not in whitelist
b186b1c73234aac5c3ea02e233df47610bb204ef
```
Error: <geopm> Could not read from MSR device: MSRIOImp::read_msr(): pread() failed at offset 0x189 system error: Permission denied: at geopm/src/MSRIO.cpp:124
Error: <geopm> Could not read from MSR device: MSRIOImp::read_msr(): pread() failed at offset 0x189 system error: Permission denied: at geopm/src/MSRIO.cpp:124
  what():  <geopm> Could not write to MSR device: MSRIOImp::write_msr(): pwrite() failed at offset 0x186 system error: Permission denied: at geopm/src/MSRIO.cpp:149
  what():  <geopm> Could not write to MSR device: MSRIOImp::write_msr(): pwrite() failed at offset 0x186 system error: Permission denied: at geopm/src/MSRIO.cpp:149
```

cmcantalupo:
- If a control is not in the allow list, throw only when
  the user tries to write the control, not when the save
  happens.

- fixes #1395

### DIFF
--- a/src/MSRIOGroup.cpp
+++ b/src/MSRIOGroup.cpp
@@ -522,10 +522,19 @@ namespace geopm
 
     void MSRIOGroup::save_control(void)
     {
+        std::vector<std::string> unallowed_controls;
         for (auto &ctl : m_control_available) {
-            for (auto &dom_ctl : ctl.second.controls) {
-                dom_ctl->save();
+            try {
+                for (auto &dom_ctl : ctl.second.controls) {
+                    dom_ctl->save();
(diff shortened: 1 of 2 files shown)


cmcantalupo:
Duplicate

PR #1501, opened 2021-01-16
Allow MSRIOGroup to be used on systems with restricted allow lists

### LINKED ISSUE #1395: Inclusion of programmable counter MSRs causes a crash on systems where the offsets are not in whitelist

cmcantalupo:
- If a control is not in the allow list, throw only when
  the user tries to write the control, not when the save
  happens.
- Fixes #1395

### DIFF
--- a/src/MSRIOGroup.cpp
+++ b/src/MSRIOGroup.cpp
@@ -522,10 +522,19 @@ namespace geopm
 
     void MSRIOGroup::save_control(void)
     {
+        std::vector<std::string> unallowed_controls;
         for (auto &ctl : m_control_available) {
-            for (auto &dom_ctl : ctl.second.controls) {
-                dom_ctl->save();
+            try {
+                for (auto &dom_ctl : ctl.second.controls) {
+                    dom_ctl->save();
+                }
             }
+            catch (const Exception &) {
+                unallowed_controls.push_back(ctl.first);
+            }
+        }
+        for (auto &ctl : unallowed_controls) {
+            m_control_available.erase(ctl);
         }
     }
 
--- a/src/PlatformIO.cpp
+++ b/src/PlatformIO.cpp
@@ -299,6 +299,10 @@ namespace geopm
             throw Exception("PlatformIOImp::push_control(): pushing controls after read_batch() or adjust().",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
+        if (!m_do_restore) {
+            throw Exception("PlatformIOImp::push_control(): pushing controls before calling save_control().",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
         if (domain_type < 0 || domain_type >= GEOPM_NUM_DOMAIN) {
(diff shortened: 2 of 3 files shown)


src/MSRIOGroup.cpp
```diff
@@ -522,10 +522,19 @@ namespace geopm
...
+        std::vector<std::string> unallowed_controls;
         for (auto &ctl : m_control_available) {
-            for (auto &dom_ctl : ctl.second.controls) {
-                dom_ctl->save();
+            try {
+                for (auto &dom_ctl : ctl.second.controls) {
+                    dom_ctl->save();
+                }
             }
+            catch (const Exception &) {
```
bgeltz, line 532: Any exception that occurs in the ```dom_ctl->save();``` call path will be stopped here.  For MSR based controls, I don't see how any other exceptions could be generated here, so this is fine.  If we need to use this pattern anywhere else, trapping ```Exception``` might be too heavy handed and we should instead use one of the custom GEOPM types.

src/MSRIOGroup.cpp
```diff
@@ -522,10 +522,19 @@ namespace geopm
...
+                unallowed_controls.push_back(ctl.first);
+            }
+        }
+        for (auto &ctl : unallowed_controls) {
+            m_control_available.erase(ctl);
```
bgeltz, line 537: Perhaps this would be a good place to add:
```
#ifdef GEOPM_DEBUG
std::cerr << "<geopm> Warning: MSR supported in GEOPM but not in msr-safe allowlist: " << ctl ...;
#endif
```

src/PlatformIO.cpp
```diff
@@ -299,6 +299,10 @@ namespace geopm
             throw Exception("PlatformIOImp::push_control(): pushing controls after read_batch() or adjust().",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
+        if (!m_do_restore) {
+            throw Exception("PlatformIOImp::push_control(): pushing controls before calling save_control().",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
```
bgeltz, line 305: ```GEOPM_CXX_MAN_PlatformIO.3.ronn``` needs an update to reflect this I think?

bgeltz (changes requested):

PR #1513, opened 2021-01-21
Remove Regions section from report when application has no regions

dianarg:


dianarg:
Might need to update io.py to handle this

PR #1516, opened 2021-01-21
Fix geopmwrite command line tool

cmcantalupo:
- Needs to use write_control() not write_batch().
- Implementation used write_batch() for historical reasons.


bgeltz (approved):
Code looks fine.  Awaiting results of manual testing.

PR #1520, opened 2021-01-21
Added better error reporting to ValidateRecord

fkeceli:
Added better error reporting to ValidateRecord

### DIFF
--- a/src/ValidateRecord.cpp
+++ b/src/ValidateRecord.cpp
@@ -56,7 +56,7 @@ namespace geopm
     {
         if (hash == GEOPM_REGION_HASH_INVALID ||
             hash > UINT32_MAX) {
-            throw Exception("ValidateRecord::filter(): Region hash out of bounds",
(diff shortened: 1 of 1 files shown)


src/ValidateRecord.cpp
```diff
@@ -56,7 +56,7 @@ namespace geopm
     {
         if (hash == GEOPM_REGION_HASH_INVALID ||
             hash > UINT32_MAX) {
-            throw Exception("ValidateRecord::filter(): Region hash out of bounds",
+            throw Exception("ValidateRecord::filter(): Region hash out of bounds: " + std::to_string(hash),
```
dianarg: It would be nice to print this in hex.  You can use string_format_hex() from Helper.hpp.  Same for other hashes below.

fkeceli: Done.

dianarg (approved):

cmcantalupo (approved):

PR #1521, opened 2021-01-21
New PENNANT GEOPM markup

fkeceli:
Signed-off-by: Fuat Keceli <fuat.keceli@intel.com>

Summary of change.

- Added high level GEOPM markup regions to Pennant.
- Added better hints and ignored regions.


dianarg (approved):
I'm surprised to see enter and exit for a region in different files.  If you've confirmed this markup is correct it looks fine to go in.

PR #1524, opened 2021-01-22
Integration tests of geopmread and geopmwrite

dianarg:

### DIFF
--- a/integration/test/test_geopmio.py
+++ b/integration/test/test_geopmio.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python
+#
+#  Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#      * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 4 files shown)


integration/test/test_geopmio.py
```diff
@@ -0,0 +1,277 @@
...
+        '''
+        def read_current_freq(domain, signal='FREQUENCY'):
+            return geopm_test_launcher.geopmread('{} {} {}'.format(signal, domain, '0'))
+
+        def read_min_sticker_freq():
+            return (geopm_test_launcher.geopmread('{} {} {}'.format('CPUINFO::FREQ_MIN', 'board', '0')),
+                    geopm_test_launcher.geopmread('{} {} {}'.format('CPUINFO::FREQ_STICKER', 'board', '0')))
+
+        def load_cpu_start():
+            self.load_pid = subprocess.Popen('stress-ng --cpu=$(lscpu | grep -e "^CPU(" | cut -d: -f2 | tr -d " ")', shell=True)
```
cmcantalupo, line 235: Is stress-ng a common enough package that we want to use this here?

cmcantalupo (commented):
Wow, that looks like it actually covers a lot of the platformio functionality as well.  Just wanted to note about the stress-ng application requirement.

PR #1530, opened 2021-01-25
Update environment integration test to be in its own python file

cmcantalupo:
Mostly just moving the code into a separate file, but also modified the test to use the frequency map agent instead of the ee agent.  This PR also disables the EE agent from the Agent factory.


dannosliwcd (approved):
These changes look good to me.

PR #1531, opened 2021-01-25
Add power governor integration test to new pattern

cmcantalupo:
Separate out power governor test from geopm_test_integration

### DIFF
--- a/integration/test/test_ompt.py
+++ b/integration/test/test_ompt.py
@@ -53,7 +53,7 @@ class TestIntegration_ompt(unittest.TestCase):
                          '.' + cls.__name__ + ') ...')
         test_name = 'test_ompt'
         cls._report_path = '{}.report'.format(test_name)
-        cls._agent_conf_path = 'test_' + test_name + '-agent-config.json'
+        cls._agent_conf_path = test_name + '-agent-config.json'
         # Clear out exception record for python 2 support
         geopmpy.error.exc_clear()
         # Set the job size parameters
--- a/integration/test/test_power_governor.py
+++ b/integration/test/test_power_governor.py
@@ -34,73 +34,68 @@
 import sys
 import unittest
 import os
+import pandas
 
 import geopm_context
 import geopmpy.io
 
 import util
-
-@util.skip_unless_do_launch()
-class TestIntegration(unittest.TestCase):
-    def setUp(self):
-        self.longMessage = True
-        self._agent = 'power_governor'
-        self._options = {'power_budget': 150}
-        self._tmp_files = []
-        self._output = None
-        self._power_limit = geopm_test_launcher.geopmread("MSR::PKG_POWER_LIMIT:PL1_POWER_LIMIT board 0")
-        self._frequency = geopm_test_launcher.geopmread("MSR::PERF_CTL:FREQ board 0")
-        self._original_freq_map_env = os.environ.get('GEOPM_FREQUENCY_MAP')
-
-    def tearDown(self):
-        geopm_test_launcher.geopmwrite("MSR::PKG_POWER_LIMIT:PL1_POWER_LIMIT board 0 " + str(self._power_limit))
-        geopm_test_launcher.geopmwrite("MSR::PERF_CTL:FREQ board 0 " + str(self._frequency))
-        if self._original_freq_map_env is None:
-            if 'GEOPM_FREQUENCY_MAP' in os.environ:
-                os.environ.pop('GEOPM_FREQUENCY_MAP')
-        else:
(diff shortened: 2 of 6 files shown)


integration/test/test_ompt.py
```diff
@@ -53,7 +53,7 @@ def setUpClass(cls):
                          '.' + cls.__name__ + ') ...')
         test_name = 'test_ompt'
         cls._report_path = '{}.report'.format(test_name)
-        cls._agent_conf_path = 'test_' + test_name + '-agent-config.json'
+        cls._agent_conf_path = test_name + '-agent-config.json'
```
bgeltz, line 56: Was this change to the OMPT code intentionally included?

integration/test/test_power_governor.py
```diff
@@ -34,73 +34,68 @@
...
+        cls._trace_path = '{}.trace'.format(test_name)
+        cls._agent_conf_path = test_name + '-agent-config.json'
+        # Clear out exception record for python 2 support
+        geopmpy.error.exc_clear()
+        # Set the job size parameters
+        cls._num_node = 4
         num_rank = 16
         loop_count = 500
-        app_conf = geopmpy.io.BenchConf(name + '_app.config')
-        self._tmp_files.append(app_conf.get_path())
```
bgeltz: I see that tmp_files tracking has been removed.  Does this mean these tests will be littering files all over the place?

integration/test/test_power_governor.py
```diff
@@ -114,11 +109,8 @@ def test_power_consumption(self):
             pkg_energy_cols = [s for s in power_data.keys() if 'ENERGY_PACKAGE' in s]
             dram_energy_cols = [s for s in power_data.keys() if 'ENERGY_DRAM' in s]
             power_data['SOCKET_POWER'] = power_data[pkg_energy_cols].sum(axis=1) / power_data['ELAPSED_TIME']
-            power_data['DRAM_POWER'] = power_data[dram_energy_cols].sum(axis=1) / power_data['ELAPSED_TIME']
-            power_data['COMBINED_POWER'] = power_data['SOCKET_POWER'] + power_data['DRAM_POWER']
 
             pandas.set_option('display.width', 100)
-            launcher.write_log(name, 'Power stats from {} :\n{}'.format(nn, power_data.describe()))
```
bgeltz: I see you've removed all the write_log calls.  This data was useful to track power stats from these tests over time.  Is this being captured somewhere else now?

bgeltz (changes requested):

integration/test/test_ompt.py
cmcantalupo: It is a copy paste error that I found when copy pasting again. This issue may be in other files too...I can check.

integration/test/test_power_governor.py
cmcantalupo: Yes. We had discussed using git to deal with extra files.  

There should also be zero requirement that the tests be run in the git repo now as long as the integration run_config.sh file is sourced.  We can double check that this is true before merging new-profile to dev.

integration/test/test_power_governor.py
cmcantalupo: I'm not sure how to support this feature and also support skip launch.  The log file is created when the launch happens.  These calls to "write_log" are happening in the the assertion part of the code.  We could just write this to standard output?

integration/test/test_power_governor.py
bgeltz: I don't think I was involved in the discussion about using git for those files.

The problem I see there is that we had been using tmp_files to conditionally remove the files in the test if the test passed.  If it failed, the files were saved for debug purposes.  That was the reason that this line exists up in setUpClass() for Python2: ```geopmpy.error.exc_clear()```.

If the idea was to invoke ```git clean``` if a test passed, that will remove all the files from any previously failed tests.  I don't see how git can understand what passed vs. failed.  I'm open to discussion if you've already figured this out though.

integration/test/test_power_governor.py
bgeltz: I'm fine with either a new log file that's specific to this test (so you can just create/open/write to it here, and avoid ```launcher.write_log``` or blasting it all in to stdout.  If it's a separate file, it'll be easier to sift through later.

integration/test/test_power_governor.py
dianarg: One problem that I don't think we ever resolved is that the check in teardown for `sys.exc_info()` is always `(None, None, None)` so it incorrectly cleans up files in python3 even when the test fails.

bgeltz (approved):

PR #1538, opened 2021-02-10
New profile hint time test

cmcantalupo:
Add an integration test that checks that nested regions change the hint time.

Test is failing on the new-profile branch.  The hint time seems to be off by a factor of two:

```
    -
      region: "network-memory"
      hash: 0x1921d14e
      runtime (s): 3.00014
      count: 1
      sync-runtime (s): 2.9963
      package-energy (J): 362.087
      dram-energy (J): 30.1939
      power (W): 120.845
      frequency (%): 154.192
      frequency (Hz): 3.23803e+09
      time-hint-network (s): 0.974966
      time-hint-ignore (s): 0
      time-hint-compute (s): 0
      time-hint-memory (s): 0.486259
      time-hint-io (s): 0
      time-hint-serial (s): 0
      time-hint-parallel (s): 0
      time-hint-unknown (s): 0
      time-hint-unset (s): 0.00244356
```

### DIFF
--- a/integration/test/test_hint_time.py
+++ b/integration/test/test_hint_time.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python
+#
+#  Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#      * Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#
+#      * Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in
+#        the documentation and/or other materials provided with the
+#        distribution.
+#
+#      * Neither the name of Intel Corporation nor the names of its
+#        contributors may be used to endorse or promote products derived
+#        from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+#  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+#  A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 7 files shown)


integration/test/test_hint_time.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        pass
+
+    def get_exec_path(self):
+        """Path to benchmark filled in by template automatically.
+
+        """
+        bin_name = 'test_hint_time'
+        # Look for in place build
+        script_dir = os.path.dirname(os.path.realpath(__file__))
```
cmcantalupo, line 74: Note this logic is different than it is in the template (which assumes an in place build).  This logic should be added to a function in the util module so it can be easily shared with the other exiting app configs in the test directory (including the template).

integration/test/test_hint_time.py
```diff
@@ -0,0 +1,182 @@
...
+        sys.stdout.write('(' + os.path.basename(__file__).split('.')[0] +
+                         '.' + cls.__name__ + ') ...')
+        cls._test_name = 'hint_time'
+        cls._report_path = 'test_{}.report'.format(cls._test_name)
+        cls._trace_path = 'test_{}.trace'.format(cls._test_name)
+        cls._image_path = 'test_{}.png'.format(cls._test_name)
+        cls._skip_launch = not util.do_launch()
+        cls._agent_conf_path = 'test_' + cls._test_name + '-agent-config.json'
+        # Clear out exception record for python 2 support
+        geopmpy.error.exc_clear()
```
bgeltz, line 112: I think we can remove all of this now, including from the template.

src/ApplicationSampler.cpp
```diff
@@ -281,19 +283,23 @@ namespace geopm
 
     double ApplicationSamplerImp::cpu_hint_time(int cpu_idx, uint64_t hint) const
     {
+        double result = NAN;
```
bgeltz, line 284: Shouldn't this just be 0?  I think that would fix the problem with the totals.

bgeltz (changes requested):

integration/test/test_hint_time.py
bgeltz: There's a few other bits re: Python 2.  I suppose we should just save all of this for a future clean-up task.

src/ApplicationSampler.cpp
cmcantalupo: No, the NaN is used to mark the signal to be ignored by the aggregation function.

bgeltz (approved):

PR #1540, opened 2021-02-12
Add callbacks for ompt progress and add test that shows it works

cmcantalupo:
- The ompt callback "dispatch" does not seem to be active.
- The ompt work callback does seem to be active
- [x] TODO: remove dispatch callback until it can be tested
- [x] TODO: validate that the work callback is doing exactly what is expected
- [x] TODO: Add loose assertions about total runtime of the test_progress run
- [x] TODO: Add assertions that the progress signal goes up twice
- [x] TODO: Add assertions that the progress signal is monotone within each OMPT detected region
- [x] TODO: Clean up history

### DIFF
--- a/integration/test/test_progress.py
+++ b/integration/test/test_progress.py
@@ -105,13 +105,15 @@ class TestIntegration_progress(unittest.TestCase):
 
         app_conf = AppConf()
 
-        trace_signals = 'REGION_PROGRESS@cpu'
+        trace_signals = 'REGION_PROGRESS@cpu,REGION_HASH@cpu'
+        report_signals = 'TIME@cpu'
         agent_conf = geopmpy.io.AgentConf(cls._test_name + '_agent.config')
 
         # Create the test launcher with the above configuration
         launcher = geopm_test_launcher.TestLauncher(app_conf=app_conf,
                                                     agent_conf=agent_conf,
                                                     report_path=cls._report_path,
+                                                    report_signals=report_signals,
                                                     trace_path=cls._trace_path,
                                                     trace_signals=trace_signals)
         launcher.set_num_node(cls._num_node)
@@ -183,25 +185,6 @@ class TestIntegration_progress(unittest.TestCase):
         error = (poly_out[1][0] / len(progress)) ** 0.5 / rr
         self.assertLess(error, 0.01, msg)
 
-    def sum_progress(self, cpu_wp, df):
-        """Sum of all progress values from a set of cpus
-
-        Args:
-            cpu_wp: A container of CPU index values
-
-            df: Data frame to select columns from
-
-        Returns:
-            The sum of all matching REGION_PROGRESS columns from the
-            data frame
-
-        """
-        result = numpy.zeros(len(df))
-        for cpu in cpu_wp:
-            name = 'REGION_PROGRESS-cpu-{}'.format(cpu)
-            result += df[name]
-        return result
-
     def test_num_host(self):
         """Check the number of hosts in the report
 
@@ -244,22 +227,15 @@ class TestIntegration_progress(unittest.TestCase):
         is well behaved for all CPUs and also collectively.
 
         """
-        df = self._trace.get_trace_df()
-        grouped_df = df.groupby('REGION_HASH')
-        cpu_with_progress = self.get_cpu_with_progress()
-        max_progress = 1.0 / self._thread_per_rank
-
         triad_post_hash = self.get_hash('triad_with_post')
-        triad_post_df = grouped_df.get_group(triad_post_hash)
-        total_progress = self.sum_progress(cpu_with_progress, triad_post_df)
-        err_msg = 'Bad fit for triad total proggress'
-        self.check_progress(triad_post_df['TIME'],
-                            total_progress,
-                            self._num_rank,
-                            err_msg)
-        for cpu in cpu_with_progress:
+        df = self._trace.get_trace_df()
+        max_progress = 1.0 / self._thread_per_rank
+        for cpu in self.get_cpu_with_progress():
+            group_name = 'REGION_HASH-cpu-{}'.format(cpu)
+            grouped_df = df.groupby(group_name)
+            triad_post_df = grouped_df.get_group(triad_post_hash)
             name = 'REGION_PROGRESS-cpu-{}'.format(cpu)
-            err_msg = 'Bad fit for triad cpu {} proggress'.format(cpu)
+            err_msg = 'Bad fit for triad cpu {} progress'.format(cpu)
             self.check_monotone(triad_post_df[name], err_msg)
             self.check_progress(triad_post_df['TIME'],
                                 triad_post_df[name],
@@ -271,22 +247,15 @@ class TestIntegration_progress(unittest.TestCase):
         is well behaved for all CPUs and also collectively.
 
         """
-        df = self._trace.get_trace_df()
-        grouped_df = df.groupby('REGION_HASH')
-        cpu_with_progress = self.get_cpu_with_progress()
-        max_progress = 1.0 / self._thread_per_rank
-
         dgemm_post_hash = self.get_hash('dgemm_with_post')
-        dgemm_post_df = grouped_df.get_group(dgemm_post_hash)
-        total_progress = self.sum_progress(cpu_with_progress, dgemm_post_df)
-        err_msg = 'Bad fit for dgemm total proggress'
-        self.check_progress(dgemm_post_df['TIME'],
-                            total_progress,
-                            self._num_rank,
-                            err_msg)
-        for cpu in cpu_with_progress:
+        df = self._trace.get_trace_df()
+        max_progress = 1.0 / self._thread_per_rank
+        for cpu in self.get_cpu_with_progress():
+            group_name = 'REGION_HASH-cpu-{}'.format(cpu)
+            grouped_df = df.groupby(group_name)
+            dgemm_post_df = grouped_df.get_group(dgemm_post_hash)
             name = 'REGION_PROGRESS-cpu-{}'.format(cpu)
-            err_msg = 'Bad fit for dgemm cpu {} proggress'.format(cpu)
+            err_msg = 'Bad fit for dgemm cpu {} progress'.format(cpu)
             self.check_monotone(dgemm_post_df[name], err_msg)
             self.check_progress(dgemm_post_df['TIME'],
                                 dgemm_post_df[name],
@@ -300,7 +269,6 @@ class TestIntegration_progress(unittest.TestCase):
         """
         df = self._trace.get_trace_df()
         grouped_df = df.groupby('REGION_HASH')
-        cpu_with_progress = self.get_cpu_with_progress()
 
         triad_post_hash = self.get_hash('triad_with_post')
         dgemm_post_hash = self.get_hash('dgemm_with_post')
@@ -311,7 +279,7 @@ class TestIntegration_progress(unittest.TestCase):
                                       unmarked_hash,
                                       'NAN')]
         for gg in other_groups:
-           for cpu in cpu_with_progress:
+           for cpu in self.get_cpu_with_progress():
                name = 'REGION_PROGRESS-cpu-{}'.format(cpu)
                uu = grouped_df.get_group(gg)[name].dropna().unique()
                if len(uu) != 0:
--- a/integration/test/test_progress.py
+++ b/integration/test/test_progress.py
@@ -163,7 +163,9 @@ class TestIntegration_progress(unittest.TestCase):
             msg: Error message to include with any failures
 
         """
-        good_idx = numpy.isfinite(progress)
+        # Get rid of nan values and values
+        good_idx = numpy.isfinite(progress) & ~numpy.isin(progress, expected_max)
+        self.assertLess(len(progress) * 0.6, sum(good_idx))
         progress = progress[good_idx]
         time = time[good_idx]
         poly_out = numpy.polyfit(time,
--- a/integration/test/test_progress.py
+++ b/integration/test/test_progress.py
@@ -161,6 +161,9 @@ class TestIntegration_progress(unittest.TestCase):
             msg: Error message to include with any failures
 
         """
+        good_idx = numpy.isfinite(progress)
+        progress = progress[good_idx]
+        time = time[good_idx]
         poly_out = numpy.polyfit(time,
                                  progress,
                                  1, full=True)
@@ -227,13 +230,14 @@ class TestIntegration_progress(unittest.TestCase):
         self.assertEqual(self._num_active_cpu, len(cpu_with_progress),
                          msg=err_msg)
 
-    def check_monotone(self, progress):
+    def check_monotone(self, progress, err_msg):
         """Check that progress does not decrease over time.
 
         """
-        self.assertEqual(progress[0], min(progress))
-        self.assertEqual(progress[-1], max(progress))
-        self.assertLessEqual(0, min(progress.diff()[1:]))
+        progress = progress.dropna()
+        self.assertEqual(progress[0], min(progress), err_msg)
+        self.assertEqual(progress[-1], max(progress), err_msg)
+        self.assertLessEqual(0, min(progress.diff()[1:]), err_msg)
 
     def test_linear_progress_triad(self):
         """Test that the triad region with with the post reports progress that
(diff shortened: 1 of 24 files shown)


integration/test/test_progress.py
```diff
@@ -0,0 +1,286 @@
...
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+import sys
+import unittest
+import os
+import pandas
+import numpy
+import socket
+import matplotlib.pyplot as plt
```
cmcantalupo: unused import

integration/test/test_progress.py
```diff
@@ -0,0 +1,379 @@
...
+        # Clear out exception record for python 2 support
+        geopmpy.error.exc_clear()
+
+        # Create machine
+        cls._machine = machine.Machine()
+        try:
+            cls._machine.load()
+            sys.stderr.write('Warning: {}: using existing file "machine.json", delete if invalid\n'.format(cls._test_name))
+        except RuntimeError:
+            cls._machine.save()
```
dianarg: could be a helper, if this is going to be a pattern in the tests.  I seem to remember we didn't want people to call the Machine constructor directly.

integration/test/test_progress.py
```diff
@@ -0,0 +1,379 @@
...
+        """Check that progress does not decrease over time.
+
+        """
+        progress = progress.dropna()
+        self.assertEqual(progress[0], min(progress), err_msg)
+        self.assertEqual(progress[-1], max(progress), err_msg)
+        self.assertLessEqual(0, min(progress.diff()[1:]), err_msg)
+
+    def test_linear_progress_triad(self):
+        """Test that the triad region with with the post reports progress that
```
dianarg: extra with

dianarg: post -> post()

integration/test/test_progress.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        df = self._trace.get_trace_df()
+        grouped_df = df.groupby('REGION_HASH')
+        cpu_with_progress = self.get_cpu_with_progress()
+        max_progress = 1.0 / self._thread_per_rank
+
+        triad_post_hash = self.get_hash('triad_with_post')
+        triad_post_df = grouped_df.get_group(triad_post_hash)
+        total_progress = self.sum_progress(cpu_with_progress, triad_post_df)
+        err_msg = 'Bad fit for triad total proggress'
```
dianarg: proggress -> progress, more below

integration/test/test_progress.py
```diff
@@ -0,0 +1,379 @@
...
+        self.check_progress(triad_post_df['TIME'],
```
dianarg: check_monotone() should also be valid for total progress

integration/test/util.py
```diff
@@ -142,6 +142,24 @@ def get_config_value(key):
...
+    """
+    path_list = []
+    int_test_dir = os.path.dirname(os.path.realpath(__file__))
+    path_list.append(os.path.join(int_test_dir, '.libs'))
+    int_dir = os.path.dirname(int_test_dir)
+    # Look for out of place build from using apps/build_func.sh
+    path_list.append(os.path.join(int_dir, 'build/integration/test/.libs'))
+    bin_list = [os.path.join(pp, bin_name) for pp in path_list]
+    for bb in bin_list:
+        if os.path.exists(bb):
```
dianarg: maybe also check isfile(), os.access() here if the errors from Popen are not as useful

src/DefaultProfile.cpp
```diff
@@ -183,15 +183,18 @@ extern "C"
...
-                geopm::Profile::default_profile().thread_init(cpu, num_work_unit);
+        try {
+            auto &prof = geopm::Profile::default_profile();
+            if (g_pmpi_tprof_enabled) {
+                prof.thread_init(geopm::Profile::get_cpu());
             }
-            catch (...) {
-                err = geopm::exception_handler(std::current_exception());
+            if (g_pmpi_prof_enabled) {
+                prof.thread_work(num_work_unit);
```
dianarg: Don't the non-lead threads pass 0 here?

cmcantalupo: Only the lead thread has "g_pmpi_prof_enabled", all the threads have "g_pmpi_tprof_enabled".  I settled on this solution so that the lead thread is responsible for setting the number of work units, but each thread is responsible for zeroing its own counter.  This avoids an error where the lead thread would zero the counter after the non-lead thread had already started incrementing it.  There should be a comment here explaining that!

integration/test/test_progress.py
cmcantalupo: I will move it into a machine method, "Machine.try_load_or_save()" or something like that.  I'm not sure I recall any issues with calling machine.Machine(), but I do think the Machine interface could probably use an overhaul.  It is a little more surprising than I'd like.

integration/test/test_progress.py
cmcantalupo: Thanks!

integration/test/test_progress.py
cmcantalupo: Thank you!

integration/test/test_progress.py
cmcantalupo: I will try to bring that back.

integration/test/util.py
cmcantalupo: Great, I'll look into that.

integration/test/test_progress.py
```diff
@@ -0,0 +1,379 @@
...
+        try:
+            cls._machine.load()
+            sys.stderr.write('Warning: {}: using existing file "machine.json", delete if invalid\n'.format(cls._test_name))
+        except RuntimeError:
+            cls._machine.save()
+
+        # Set the job size parameters
+        cls._num_node = 1
+        cls._num_rank = 10
+        cls._num_active_cpu = cls._machine.num_core() - 2 * cls._machine.num_package()
```
bgeltz, line 98: Speaking of the Machine object, it would be great if we could ask it for the number of active CPUs to avoid sprinkling the magic ```2``` around all the tests.  This will be nice if/when we change where and how the Controller runs (application mode).

bgeltz (changes requested):
I'm seeing a unit test failure with this PR when I build with debug enabled (release too):
```
FAIL: test/gtest_links/ApplicationSamplerTest.hint_time
```

The log shows:
```
Note: Google Test filter = ApplicationSamplerTest.hint_time
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from ApplicationSamplerTest
[ RUN      ] ApplicationSamplerTest.hint_time
test/ApplicationSamplerTest.cpp:387: Failure
Expected equality of these values:
  0.0
    Which is: 0
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:389: Failure
Expected equality of these values:
  0.0
    Which is: 0
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:391: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan
test/ApplicationSamplerTest.cpp:393: Failure
Expected equality of these values:
  0.0
    Which is: 0
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:395: Failure
Expected equality of these values:
  0.0
    Which is: 0
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:397: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: check_sample_end()
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
test/ApplicationSamplerTest.cpp:411: Failure
Expected equality of these values:
  0.0
    Which is: 0
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:413: Failure
Expected equality of these values:
  0.0
    Which is: 0
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:415: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan
test/ApplicationSamplerTest.cpp:417: Failure
Expected equality of these values:
  0.0
    Which is: 0
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:419: Failure
Expected equality of these values:
  0.0
    Which is: 0
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:421: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: check_sample_end()
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
test/ApplicationSamplerTest.cpp:436: Failure
Expected equality of these values:
  0.0
    Which is: 0
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:438: Failure
Expected equality of these values:
  1.0
    Which is: 1
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:440: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan
test/ApplicationSamplerTest.cpp:442: Failure
Expected equality of these values:
  1.0
    Which is: 1
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:444: Failure
Expected equality of these values:
  0.0
    Which is: 0
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:446: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: check_sample_end()
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
test/ApplicationSamplerTest.cpp:461: Failure
Expected equality of these values:
  0.0
    Which is: 0
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:463: Failure
Expected equality of these values:
  3.0
    Which is: 3
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:465: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan
test/ApplicationSamplerTest.cpp:467: Failure
Expected equality of these values:
  1.0
    Which is: 1
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:469: Failure
Expected equality of these values:
  0.0
    Which is: 0
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:471: Failure
Expected equality of these values:
  2.0
    Which is: 2
  memory_time
    Which is: nan

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: check_sample_end()
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
test/ApplicationSamplerTest.cpp:486: Failure
Expected equality of these values:
  3.0
    Which is: 3
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:488: Failure
Expected equality of these values:
  3.0
    Which is: 3
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:490: Failure
Expected equality of these values:
  0.0
    Which is: 0
  memory_time
    Which is: nan
test/ApplicationSamplerTest.cpp:492: Failure
Expected equality of these values:
  1.0
    Which is: 1
  compute_time
    Which is: nan
test/ApplicationSamplerTest.cpp:494: Failure
Expected equality of these values:
  3.0
    Which is: 3
  network_time
    Which is: nan
test/ApplicationSamplerTest.cpp:496: Failure
Expected equality of these values:
  2.0
    Which is: 2
  memory_time
    Which is: nan
[  FAILED  ] ApplicationSamplerTest.hint_time (2 ms)
[----------] 1 test from ApplicationSamplerTest (2 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (3 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ApplicationSamplerTest.hint_time

 1 FAILED TEST
```

integration/test/test_progress.py
```diff
@@ -0,0 +1,379 @@
...
+    def test_overhead_dgemm(self):
+        """Test that the overhead is small for dgemm which does many
+        operations for each work chunk.
+
+        """
+        self.check_overhead('dgemm', 0.05)
+
+    def test_overhead_triad(self):
+        """Test that the overhead not huge for triad which does only 3
+        operations per work chunk (FMA,add,store).
```
bgeltz:
```suggestion
        """Test that the overhead is not huge for triad which does only 3
        operations per work chunk (FMA, add, and store).
```

src/ApplicationStatus.cpp
```diff
@@ -76,7 +76,6 @@ namespace geopm
         // initialize shmem if all zero is not appropriate
         for (int cpu = 0; cpu < m_num_cpu; ++cpu) {
             set_process({cpu}, -1);
-            set_total_work_units(cpu, 0);
```
bgeltz, line 79: This seems important.  Why is it not necessary?

src/Profile.cpp
```diff
@@ -498,7 +495,8 @@ namespace geopm
                 // progress from decreasing at the end of a region.
                 // The thread progress value is not valid outside of a
                 // region.
-                m_app_status->set_total_work_units(cpu, 0);
+                m_app_status->set_hash(cpu, m_current_hash);
+                m_app_status->reset_work_units(cpu);
```
bgeltz, line 499: I expected to see someone setting the work units to 0 at the first region entry, but I'm not seeing it.

integration/test/test_progress.py
cmcantalupo: Your memory is correct, I added a helper function "Machine.try_machine()", in other cases we use "Machine.get_machine()".

integration/test/test_progress.py
cmcantalupo: The problem with the total progress analysis is that I had to select based on the region hash for the entire board (i.e. the synchronous part).  This meant that the end of the lagging ranks progress was cut off by the group_by() selection.  This caused issues with the helper check assertsions.  The sum of progress is a little contrived, and checking each of them is a finer grained check (now selecting on per-cpu region hash).

integration/test/test_progress.py
cmcantalupo: The unit test has been failing since we started deriving the hint from the ApplicationStatus lock free solution and not based on messages in the ApplicationRecordLog.  This unit test needs to be updated to mock the ApplicationStatus object, not the ApplicationRecordLog object.

In terms of the magic number 2, this is based on the geopmpy.launch behavior (reserve one core for the OS and one core for GEOPM, which is what this number two reprsents).  The machine file does not contain information about the launch command was issued, or what the application cpu affinities are.

integration/test/test_progress.py
```diff
@@ -0,0 +1,349 @@
...
+        max_prog = max(progress[good_idx])
+        util.assertNear(self, expected_max, max_prog, epsilon=epsilon,
+                        msg='{}: maximum value'.format(msg))
+        # Get rid of values at expected_max to avoid bad fit error
+        # when completed threads stop increasing progress
+        good_idx = good_idx & ~numpy.isin(progress, expected_max)
+        self.assertLess(len(progress) * 0.6, sum(good_idx))
+        progress = progress[good_idx]
+        time = time[good_idx]
+        poly_out = numpy.polyfit(time,
```
alawibaba, line 174: This is fine, but if we have a performance issue in the future there is repeated work in this call that we could eliminate if this is slow.

alawibaba:
Hey LGTM. I had one little comment about one of the tests but it's just a comment.

integration/test/test_progress.py
cmcantalupo: The test runs very quickly, so for now this is not a problem.  If we start using this pattern in tests that use more data, we may want to look into performance improvements.

bgeltz (approved):

PR #1541, opened 2021-02-16
Initial confidence (cross validation) policy recommender.

alawibaba:
Creates a policy recommender that uses cross validation to estimate fit confidence.

### DIFF
--- a/integration/experiment/power_sweep/gen_policy_recommendation.py
+++ b/integration/experiment/power_sweep/gen_policy_recommendation.py
@@ -268,9 +268,10 @@ def rootssq(ll):
 
 def normal_comparison(dist1, dist2):
     """Returns the probability that a sample drawn from the normal distribution
-    described by dist1 (a tuple of mu and sigma) will be larger than a sample
-    drawn from a normal distribution drawn from the normal distribution
-    described by dist2."""
+    described by dist1 (a dictionary with keys 'mean' and 'dev' and values the
+    mean and standard deviation of the distribution, respectively) will be
+    larger than a sample drawn from a normal distribution drawn from the normal
+    distribution described by dist2."""
     mu1, sigma1 = dist1['mean'], dist1['dev']
     mu2, sigma2 = dist2['mean'], dist2['dev']
     sigma1 = rootssq(sigma1)
--- a/integration/experiment/power_sweep/gen_policy_recommendation.py
+++ b/integration/experiment/power_sweep/gen_policy_recommendation.py
@@ -37,8 +37,9 @@ consumption. Offers power-governor policy recommendations for
 minimum energy use subject to a runtime degradation constraint.
 '''
 
-import sys
 import argparse
+import math
+import sys
 
 import pandas
 import numpy as np
@@ -49,7 +50,7 @@ import geopmpy.io
 from experiment import common_args
 from experiment import machine
 
-class PowerLimitModel:
+class PowerLimitModel(object):
     "Abstract class for predicting a statistic based on a power limit."
     def __init__(self):
         "Default constructor."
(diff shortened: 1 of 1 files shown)


integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -266,6 +391,9 @@ def main(full_df, region_filter, dump_prefix, min_pl, max_pl, tdp, max_degradati
     parser.add_argument('--min_energy', action='store_true',
                         help='ignore max degradation, just give the minimum '
                              'energy possible')
+    parser.add_argument('--no-confidence', dest='confidence',
```
cmcantalupo: its is a little long, but i think i'd prefer "--disable-confidence" to "--no-confidence" just for consistency with other GEOPM CLI's.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -214,7 +257,75 @@ def policy_min_energy(plrange, enmodel, rtmodel = None, pltdp = None,
...
+    if max_degradation is None:
+        # we don't need the runtime model in this case
+        best_energy, best_pl = min(zip(en_predictions, plrange))
+        if rtmodel:
+            best_runtime = rtmodel.evaluate(best_pl, True)
+        else:
+            best_runtime = None
+        return {'power': best_pl,
+                'runtime': best_runtime[0],
+                'runtimedev': rootssq(best_runtime[1]),
```
cmcantalupo: rootssq != standard deviation, you need to remove the mean.  Has the mean been removed from the best_runtime?

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -214,7 +257,75 @@ def policy_min_energy(plrange, enmodel, rtmodel = None, pltdp = None,
                 'energy': best_energy}
 
 
-def main(full_df, region_filter, dump_prefix, min_pl, max_pl, tdp, max_degradation):
+def rootssq(ll):
+    "Return the root of the sum of the squares of the list of values."
+    try:
+        return sum([ss**2  for ss in ll])**0.5
+    except TypeError:
```
cmcantalupo, line 265: why is this exception being ignored, can you add a comment please?

integration/experiment/power_sweep/gen_policy_recommendation.py
cmcantalupo: To make this clearer, when the evaluate() method is called, each of the three returned values should get a different variable name.  This will avoid the confusion I had when thinking that "best_runtime[1]" is a container with runtime values in it.

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -258,20 +258,21 @@ def policy_min_energy(plrange, enmodel, rtmodel = None, pltdp = None,
...
     try:
         return sum([ss**2  for ss in ll])**0.5
     except TypeError:
-        pass
+        pass  # ll is not iterable
     return ll
 
 def normal_comparison(dist1, dist2):
     """Returns the probability that a sample drawn from the normal distribution
     described by dist1 (a tuple of mu and sigma) will be larger than a sample
```
cmcantalupo: dist1 is not a tuple of mu and sigma anymore

integration/experiment/power_sweep/gen_policy_recommendation.py
```diff
@@ -176,7 +176,7 @@ def evaluate(self, PL, full=False):
         dev = (sum([val**2 for val in vals])/len(vals) - mean**2)**0.5
         res = (sum([mm.residual() for mm in self._model])/len(self._model))
         if full:
-            return mean, [dev, res]
+            return {'mean': mean, 'dev': [dev, res]}
```
cmcantalupo, line 179: why not:
```
return {'mean': mean, 'sigma': rootssq([dev, res])}
```
it seems like sigma is used throughout the rest of the code?  Perhaps there will be another use for each of the terms in the future, or this is for compatibility with another python interface?

cmcantalupo: Nevermind, I see where the residual is used independently below.

cmcantalupo (commented):
Just a missing documentation update required with the change to the type of the distribution objects to a dictionary.

integration/experiment/power_sweep/gen_policy_recommendation.py
alawibaba: Done.

alawibaba:
Done! PTAL.

cmcantalupo (approved):
LGTM

PR #1551, opened 2021-02-26
Update integration tests on new-profile

bgeltz:
Some changes to the tests need to be made to address changes on this branch compared to dev.  The following high level changes are made:
1. Updates to AppOutput usage to remove report parsing.  Trace files are the only form of data still handled.
2. Updates to field names from the report.
...

These updates allow the tests to run.  Presently there are test failures being exposed with these changes:
```
ERROR: test_regions_absent (test_omp_outer_loop.TestIntegrationOMPOuterLoop)
Test that the first run's report does NOT contain
  File "./test_omp_outer_loop.py", line 128, in test_regions_absent
    if observed_region['time-hint-network'] != 0:
KeyError: 'time-hint-network'

ERROR: setUpClass (test_tutorial_base.TestIntegration_tutorial_base)
  File "./test_tutorial_base.py", line 85, in setUpClass
    subprocess.check_call(build_script, shell=True)
  File "/usr/lib64/python3.6/subprocess.py", line 311, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '/home/bgeltz/git/geopm/integration/test/test_tutorial_base.sh' returned non-zero exit status 2.

FAIL: test_agent_frequency_map (test_frequency_map.TestIntegration_frequency_map)
  File "./test_frequency_map.py", line 165, in test_agent_frequency_map
    util.assertNear(self, expect, actual, msg=msg)
  File "./util.py", line 285, in assertNear
    self.fail('The fractional difference between {a} and {b} is greater than {epsilon}.  {msg}'.format(a=a, b=b, epsilon=epsilon, msg=msg))
AssertionError: The fractional difference between 1100000000.0 and 1208120000.0 is greater than 0.05.  all2all: frequency should be near assigned map frequency

FAIL: test_power_balancer_geopmbench (test_power_balancer.TestIntegration_power_balancer)
  File "./test_power_balancer.py", line 227, in test_power_balancer_geopmbench
    self.balancer_test_helper('geopmbench')
  File "./test_power_balancer.py", line 224, in balancer_test_helper
    agent_runtime['power_governor'], agent_runtime['power_balancer'], margin))
AssertionError: 62.889675000000004 not greater than 69.4377 : governor runtime: 68.998, balancer runtime: 69.4377, margin: 6.108325000000001

Ran 42 tests in 931.807s

FAILED (failures=2, errors=2, skipped=9, expected failures=1)
```


integration/test/test_frequency_hint_usage.py
```diff
@@ -140,15 +140,15 @@ def setUpClass(cls):
...
-            # EE run
-            launcher = geopm_test_launcher.TestLauncher(app_conf,
-                                                        ee_agent_conf,
-                                                        cls._ee_report_path,
-                                                        cls._ee_trace_path,
-                                                        time_limit=time_limit)
-            launcher.set_num_node(num_node)
-            launcher.set_num_rank(num_rank)
-            launcher.run(test_name)
+            # EE run - Not available; Removed from Agent factory.
```
cmcantalupo, line 143: On the fence:

One might argue that we should delete all tests in place for an agent since it was removed from the factory.  

On the other hand, we have left the EE agent code in the repository even if we have removed it from the factory.  The agent compiles, but does not execute without failure due to signals not provided by platformio.  Leaving the code in place but disabling the agent with the factory and disabling the tests for the agent with skip decorators is self consistent.  

By this logic we should leave the tests here in a disabled state until we (a) remove the EE agent source and then delete the tests or (b) re-implement the EE agent and modify the tests as needed once they are enabled again.

I don't love commented code, is the setUpClass still executed when the test is skipped?

cmcantalupo (approved):
Soon we should just delete all references to the ee agent throughout the code.

PR #1576, opened 2021-03-11
Add an interface between generic GEOPM signals and the interfaces from SSTIO

dannosliwcd:
- Create a new SSTSignal class for signals from SSTIO
- Create a new SSTControl class for controls from SSTIO
- Create a new SSTIO mock for testing the new signal and control interfaces

### DIFF
--- a/src/SSTControl.cpp
+++ b/src/SSTControl.cpp
@@ -63,7 +63,7 @@ namespace geopm
 
     void SSTControl::setup_batch(void)
     {
-        if (m_control_type == MMIO) {
+        if (m_control_type == M_MMIO) {
             m_adjust_idx = m_sstio->add_mmio_write(
                 m_cpu_idx, m_interface_parameter, m_write_value, m_rmw_read_mask);
         }
@@ -83,7 +83,7 @@ namespace geopm
 
     void SSTControl::write(double value)
     {
-        if (m_control_type == MMIO) {
+        if (m_control_type == M_MMIO) {
             m_sstio->write_mmio_once(
                 m_cpu_idx, m_interface_parameter, m_write_value, m_rmw_read_mask,
                 static_cast<uint64_t>(value * m_multiplier) << m_shift, m_mask);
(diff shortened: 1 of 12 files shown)


src/SSTControl.cpp
```diff
@@ -0,0 +1,108 @@
...
+        }
+        else {
+            m_sstio->write_mbox_once(
+                m_cpu_idx, m_command, m_subcommand, m_interface_parameter,
+                m_rmw_subcommand, m_rmw_interface_parameter, m_rmw_read_mask,
+                static_cast<uint64_t>(value * m_multiplier) << m_shift, m_mask);
+        }
+    }
+
+    void SSTControl::save(void)
```
cmcantalupo, line 99: Why isn't there a save/restore implementation?

src/SSTControl.hpp
```diff
@@ -0,0 +1,107 @@
...
+    class SSTIO;
+
+    /// This is the abstraction layer that exposes GEOPM controls from the
+    /// SSTIO object.
+    class SSTControl : public geopm::Control
+    {
+        public:
+            enum ControlType
+            {
+                MBOX, // SST Mailbox interface
```
cmcantalupo: It is rare for us to have a public enum used in this way.  A different solution would be to have two classes derived from Control:  SSTMBoxControl and SSTMMIOControl.  This would avoid the if () switch in the methods that use it, but could produce code duplication for the methods that do not.  We can discuss pros and cons.  The most similar example i could find to this pattern (probably where it was copied from is here:

https://github.com/geopm/geopm/blob/dev/src/MSR.hpp#L47-L50

In this case we used the convention of prefixing the values with M_, e.g. M_MBOX and M_MMIO.  The logic of distinguishing class members from globals is questionable for public enums, but maybe, do add the M_ prefix if the switch remains.

cmcantalupo (commented):
Just the nit about M_ in the enum names and add save/restore logic (with tests)

cmcantalupo (approved):
LGTM

PR #1578, opened 2021-03-11
Addition of PCNT Scalability based region hints

lhlawson:
Addition of 3 PCNT scalability based region hint signals:
1. MSR::PPERF:PCNT_REGION_HINT_COMPUTE - PCNT is > 0.5
2. MSR::PPERF:PCNT_REGION_HINT_MEMORY - PCNT is < 0.5 but greater than 0.05
3. MSR::PPERF:PCNT_REGION_HINT_IGNORE - PCNT is < 0.05

These region hints are based upon the addition of the following signals:
- MSR::PPERF:PCNT_RATE
- MSR::APERF:ACNT_RATE
- MSR::PPERF:CPU_SCALABILITY

Two new signal types have been added:
- DivisionSignal, used for basic division of two provided signals with division by zero checking.  Needed for PCNT_RATE/ACNT_RATE calculation
- ScalabilityRegionSignal, used for incrementing the REGION_HINT_ signals when CPU_SCALABILITY is within the specified range.

Changes to the tracking of MSR format_function have also been made in the MSRIOGroup to handle signals with unit type of "None" that do not use an integer format.  

Unit and integration tests have been added to test the functionality described above.  The test_pcnt_region_hints.py unit test contains two tests of interest:
1. test_runtime - a basic test that the three new region signals cover the entire runtime of a test (i.e. application total time == sum(compute, memory, ignore))
2. test_runtime_composition - a test that checks DGEMM is primarily compute regions, STREAM is primarily memory/ignore regions, etc.  It should be noted that this test will **FAIL** on systems that do not have Scalability enabled, as all regions will appear compute bound.


src/ScalabilityRegionSignal.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "Helper.hpp"
+#include "geopm_debug.hpp"
+
+#include <math.h>
+#include <unistd.h>
+
+namespace geopm
+{
+    ScalabilityRegionSignal::ScalabilityRegionSignal(std::shared_ptr<Signal> scalability_sig, std::shared_ptr<Signal> time_sig,
+                                         double range_upper, double range_lower,
```
cmcantalupo: WS

src/ScalabilityRegionSignal.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if (!m_is_batch_ready) {
+            m_scalability->setup_batch();
+            m_time->setup_batch();
+            m_is_batch_ready = true;
+        }
+    }
+
+    double ScalabilityRegionSignal::compute_region_time(double scalability,
+                               double curr_time,
```
cmcantalupo: WS

integration/test/test_pcnt_region_hints.py
```diff
@@ -0,0 +1,188 @@
...
+        """Create launcher, execute benchmark and set up class variables.
+
+        """
+        sys.stdout.write('(' + os.path.basename(__file__).split('.')[0] +
+                         '.' + cls.__name__ + ') ...')
+        test_name = 'pcnt_region_hints'
+        cls._report_path = 'test_{}.report'.format(test_name)
+        cls._trace_path = 'test_{}.trace'.format(test_name)
+        cls._image_path = 'test_{}.png'.format(test_name)
+        cls._report_signals = "MSR::PPERF:PCNT_REGION_HINT_COMPUTE,MSR::PPERF:PCNT_REGION_HINT_MEMORY,MSR::PPERF:PCNT_REGION_HINT_IGNORE"
```
cmcantalupo: `MSR::PPERF:PCNT_REGION_HINT_COMPUTE` -> `MSR::TIME_HINT_COMPUTE` etc

src/ScalabilityRegionSignal.cpp
```diff
@@ -0,0 +1,120 @@
...
+                                             curr_time,
+                                             m_prev_time,
+                                             m_range_upper,
+                                             m_range_lower);
+        m_prev_time = curr_time;
+
+        return m_region_time;
+    }
+
+    double ScalabilityRegionSignal::read(void) const
```
cmcantalupo: I'm not sure this interface is useful except for testing a single step of sampling.

src/ScalabilityRegionSignal.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if (!m_is_batch_ready) {
+            throw Exception("setup_batch() must be called before sample().",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+
+        double scalability = m_scalability->sample();
+        double curr_time = m_time->sample();
+
+        m_region_time += compute_region_time(scalability,
```
cmcantalupo: I don't think the "region" name should be used in this class.  Maybe ScalabilityHintTimeSignal?

src/DivisionSignal.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace geopm
+{
+    /// A composite signal used by an IOGroup to produce a signal as
+    /// the Division of two signals.
+    class DivisionSignal : public Signal
+    {
+        public:
+            DivisionSignal(std::shared_ptr<Signal> numerator,
+                             std::shared_ptr<Signal> denominator);
```
bgeltz:
```suggestion
                           std::shared_ptr<Signal> denominator);
```

bgeltz: Small spacing issue.

src/DivisionSignal.hpp
```diff
@@ -0,0 +1,61 @@
...
+#ifndef DIVISIONSIGNAL_HPP_INCLUDE
+#define DIVISIONSIGNAL_HPP_INCLUDE
+
+#include <memory>
+
+#include "Signal.hpp"
+
+namespace geopm
+{
+    /// A composite signal used by an IOGroup to produce a signal as
```
bgeltz:
```suggestion
    /// @brief A composite signal used by an IOGroup to produce a signal as
```

You need to use ```@brief``` if you want this do be a docstring.  Otherwise the triple slash is not required.  See for an example:  https://github.com/geopm/geopm/blob/dev/src/CSV.hpp#L45

src/MSRIOGroup.cpp
```diff
@@ -280,11 +284,160 @@ namespace geopm
...
+        // energy signal.
+        struct cnt_data
+        {
+            std::string cnt_name;
+            std::string description;
+            std::string msr_name;
+        };
+        std::vector<cnt_data> cnt_signals {
+            {"MSR::PPERF:PCNT_RATE",
+                    "Average cpu pcnt value over 20 ms or 4 control loop iterations",
```
bgeltz: How can you make any assertions about the control loop cadence here?  Isn't that agent specific?

bgeltz (changes requested):

src/MSRIOGroup.cpp
```diff
@@ -280,11 +284,160 @@ namespace geopm
...
+                                                   cnt_domain,
+                                                   IOGroup::M_UNITS_HERTZ,
+                                                   agg_function(msr_name),
+                                                   ps.description + "\n    alias_for: " + ps.msr_name + " rate of change",
+                                                   IOGroup::M_SIGNAL_BEHAVIOR_VARIABLE,
+                                                   string_format_double};
+            }
+        }
+
+        std::string signal_name = "MSR::PPERF:CPU_SCALABILITY";
```
cmcantalupo: MSR::PPERF:CPU_SCALABILITY -> MSR::CPU_SCALABILTY_RATIO

src/MSRIOGroup.cpp
```diff
@@ -280,11 +284,160 @@ namespace geopm
...
+        std::string msr_name = "MSR::PPERF:PCNT_RATE";
```
cmcantalupo: In general, remove PPERF: from these names which are not sub-fields.

src/MSRIOGroup.cpp
cmcantalupo: fix comment (sample times are user controlled)

test/ScalabilityRegionSignalTest.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    m_time_sig = std::make_shared<MockSignal>();
+    m_scal_sig = std::make_shared<MockSignal>();
+    for (const auto &range : m_range) {
+        m_sig.push_back(geopm::make_unique<ScalabilityRegionSignal>(m_scal_sig, m_time_sig,
+                                                     range.first, range.second, m_sleep_time));
+    }
+}
+
+TEST_F(ScalabilityRegionSignalTest, read)
```
cmcantalupo: Can drop since read() should always return NAN.

PR #1579, opened 2021-03-15
Add interface for periodic aggregation

cmcantalupo:
- This feature will enable the power balancer to balance
  over periods of time rather than epochs.

### DIFF
--- a/src/SampleAggregator.cpp
+++ b/src/SampleAggregator.cpp
@@ -121,6 +121,7 @@ namespace geopm
                 0,
                 SumAccumulator::make_unique(),
                 SumAccumulator::make_unique(),
+                SumAccumulator::make_unique(),
                 {},
                 {},
            };
@@ -150,6 +151,7 @@ namespace geopm
                 GEOPM_REGION_HASH_INVALID,
                 m_platform_io.push_signal("EPOCH_COUNT", domain_type, domain_idx),
                 0,
+                AvgAccumulator::make_unique(),
                 AvgAccumulator::make_unique(),
                 AvgAccumulator::make_unique(),
                 {},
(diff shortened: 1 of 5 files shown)


test/SampleAggregatorTest.cpp
```diff
@@ -114,6 +114,7 @@ void SampleAggregatorTest::SetUp(void)
     EXPECT_CALL(m_platio, push_signal("TIME", GEOPM_DOMAIN_BOARD, 0))
         .WillOnce(Return(M_SIGNAL_TIME));
     m_agg = geopm::make_unique<SampleAggregatorImp>(m_platio);
+    m_agg->period_duration(1.0);
```
cmcantalupo, line 118: Add a check that m_agg returns NAN prior to setting period.

src/SampleAggregator.cpp
```diff
@@ -419,4 +458,25 @@ namespace geopm
...
+    double SampleAggregatorImp::sample_period_last(int signal_idx)
+    {
+        double result = NAN;
+        if (m_period_duration != 0.0) {
+            auto sum_it = m_sum_signal.find(signal_idx);
+            if (sum_it != m_sum_signal.end()) {
+                result = sum_it->second.period_accum->interval_total();
+            }
+            else {
+                auto avg_it = m_avg_signal.find(signal_idx);
```
cmcantalupo: no test for this branch of the code

src/SampleAggregator.cpp
```diff
@@ -218,8 +229,24 @@ namespace geopm
         return result;
     }
 
+    void SampleAggregatorImp::period_duration(double duration)
+    {
```
cmcantalupo, line 233: Make sure this is only called prior to any update calls and not afterward.

src/SampleAggregator.cpp
cmcantalupo: This is a general problem for this unit test, not particular to this feature.  I opened an issue here:

https://github.com/geopm/geopm/issues/1590

dannosliwcd (approved):
LGTM!

This is cool! So is the workflow that an agent would hold an instance of this, register some signals to either sum or average, set a period (or omit that step if epochs are used instead), then sample this at any time the sum or average is needed (without needing to manually compute or track time deltas)?

PR #1582, opened 2021-03-16
Fix irregularities in hash/hint reporting

cmcantalupo:
This PR contains two changes.  One writes the hash and hint on the same pass through the CPU list when entering or exiting an outer most region. The other fix is for an off-by-one inconsistency between how the region aggregation is done and how the hint time aggregation is done.

### DIFF
--- a/src/ApplicationSampler.cpp
+++ b/src/ApplicationSampler.cpp
@@ -149,6 +149,7 @@ namespace geopm
         , m_is_cpu_active(is_cpu_active)
         , m_update_time({{0, 0}})
         , m_is_first_update(true)
+        , m_hint_last(m_num_cpu, GEOPM_REGION_HINT_UNSET)
     {
         if (m_is_cpu_active.empty()) {
             m_is_cpu_active.resize(m_num_cpu, false);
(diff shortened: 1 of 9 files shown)


src/ApplicationStatus.hpp
```diff
@@ -68,6 +68,7 @@ namespace geopm
             /// @brief Set the hash of the region currently running on
             ///        a CPU.
             virtual void set_hash(int cpu_idx, uint64_t hash) = 0;
```
bgeltz: Is there any reason to keep the old ```set_hash``` API?  Is it called anywhere else?

src/ApplicationSampler.cpp
```diff
@@ -183,11 +184,16 @@ namespace geopm
         m_sampler->check_sample_end();
 
         m_status->update_cache();
-        if (!m_is_first_update) {
+        if (m_is_first_update) {
+            for (int cpu_idx = 0; cpu_idx != m_num_cpu; ++cpu_idx) {
+                m_hint_last[cpu_idx] = m_status->get_hint(cpu_idx);
```
bgeltz, line 189: Does it make sense to set ```m_hint_time``` on the first update as well?

cmcantalupo: The current solution is to make m_hint_last a const map with a special initializer object:

https://github.com/geopm/geopm/pull/1582/files/f55967855a5a7b192bf0b768bb740b85eeed5377#diff-f3b7daf82cbdf6c5548dbd738a8faa5b015fb5477299011efc5672d89e6cf6feR122-R133

This has the benefit of telling the compiler that the number of keys in the map will not increase.

src/ApplicationStatus.hpp
cmcantalupo: Not really.  Only if we think there will be a case were we want set the hash but not the hint int the future.  The ApplicationStatus is intended to be an abstraction of the shared memory structure it manipulates.  I'm okay with dropping the method, it will increase our code coverage fraction, since this is no longer covered by the unit test.

cmcantalupo: Top of the stack removes this method

bgeltz (approved):

PR #1584, opened 2021-03-18
Fix DerivativeSignal so that the sample() method can be called multiple times per read_batch()

cmcantalupo:


lhlawson (dismissed):
Tested locally with the PCNT Hint Time derivative signals, no issues encountered.

src/DerivativeSignal.cpp
```diff
@@ -110,8 +111,15 @@ namespace geopm
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
         double time = m_time_sig->sample();
-        double signal = m_y_sig->sample();
-        return compute_next(m_history, m_derivative_num_fit, time, signal);
+        size_t history_size = m_history.size();
+        // Check if this is the first call to sample() since the last
+        // call to read_batch()
```
cmcantalupo:
```suggestion
        // Check if this is the first call ever to sample() (history_size == 0)
        // Or check if this is the first call to sample() since the last call to read_batch() (last element of history buffer does not match the sampled time)
```

bgeltz (approved):

PR #1586, opened 2021-03-23
Add an IOGroup for the isst_interface driver

dannosliwcd:
This change creates the SSTIOGroup class, and registers it as an IOGroup in GEOPM. 

The new IOGroup maps GEOPM signal and control names to specific inputs for the SST ioctl interface. The name-to-ioctl mappings are encoded into the SSTIOGroup class via the following variables:
* `sst_signal_mbox_info`: Signals that depend on SST mailbox operations
* `sst_control_mbox_info`: Controls (and their equivalent signals) that depend on SST mailbox operations
* `sst_control_mmio_info`: Controls (and their equivalent signals) that depend on SST mmio operations

There is no `sst_signal_mmio_info` because the only MMIO-type signals that exist in this iogroup are those that are derived from their controls.

The SST driver has more interfaces than just those two (e.g., an msr interface). The above ones are the only ones we use at this time.

Interactions with the ioctl interface are managed through SSTControl and SSTSignal objects. Those objects are created from the above name-to-ioctl mappings when the IOGroup is constructed.

### DIFF
--- a/src/SSTIOGroup.cpp
+++ b/src/SSTIOGroup.cpp
@@ -1,0 +1,834 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, 2020, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+
+#include "SSTIOGroup.hpp"
+
+#include <algorithm>
+#include <cinttypes>
+#include <iomanip>
+#include <iterator>
+#include <sstream>
+
+#include "Agg.hpp"
+#include "Exception.hpp"
+#include "Helper.hpp"
+#include "MSR.hpp"
+#include "MSRFieldSignal.hpp"
+#include "PlatformTopo.hpp"
+#include "SSTControl.hpp"
+#include "SSTIO.hpp"
+#include "SSTSignal.hpp"
+#include "geopm_debug.hpp"
+#include "geopm_topo.h"
+
+namespace geopm
+{
+    const std::map<std::string, SSTIOGroup::sst_signal_mailbox_raw_s> SSTIOGroup::sst_signal_mbox_info = {
+        { "SST::CONFIG_LEVEL",
+          { SSTIOGroup::SSTMailboxCommand::M_TURBO_FREQUENCY, 0x00,
+            {{ "LEVEL", { 0x00, 16, 23, 1.0, M_UNITS_NONE,
+                            "SST configuration level", M_SIGNAL_BEHAVIOR_CONSTANT } }}
+          } },
+        { "SST::TURBOFREQ_SUPPORT",
+          { SSTIOGroup::SSTMailboxCommand::M_TURBO_FREQUENCY, 0x01,
+            {{ "SUPPORTED", { 0x00, 0, 0, 1.0, M_UNITS_NONE,
+                              "SST-TF is supported",
+                              M_SIGNAL_BEHAVIOR_CONSTANT } }}
+          } },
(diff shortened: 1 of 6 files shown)


src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+          { SSTIOGroup::SSTMailboxCommand::TURBO_FREQUENCY, 0x01,
+            {{ "SUPPORTED", { 0x00, 0, 0, 1.0, M_UNITS_NONE,
+                              "SST-TF is supported",
+                              M_SIGNAL_BEHAVIOR_CONSTANT } }}
+          } },
+        { "SST::HIGHPRIORITY_NCORES",
+          { SSTIOGroup::SSTMailboxCommand::TURBO_FREQUENCY, 0x10,
+            {{ "0", { 0x0000, 0, 7, 1.0, M_UNITS_NONE,
+                      "Count of high-priority turbo frequency cores in bucket 0",
+                      M_SIGNAL_BEHAVIOR_CONSTANT } },
```
cmcantalupo, line 73: All of these signals are labled as constant.  This implies that they will never change during runtime, is that correct here?  I think these are hardware values that cannot change, but I just wanted to check.

src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+    SSTIOGroup::SSTIOGroup(const PlatformTopo &topo, std::shared_ptr<SSTIO> sstio)
+        : m_topo(topo)
+        , m_sstio(sstio)
+        , m_is_read(false)
+        , m_signal_available()
+        , m_control_available()
+        , m_signal_pushed()
+        , m_control_pushed()
+    {
+        if (m_sstio == nullptr) {
```
cmcantalupo, line 257: Nit:  I'd prefer to use the bool operator here.  
```
if (!m_sstio) {
...
```
https://en.cppreference.com/w/cpp/memory/shared_ptr/operator_bool

I would think you would have to use m_sstio.get() to compare with nullptr, but i guess the compiler effectively casts nullptr to a std::shared_ptr()?

src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+            m_sstio = SSTIO::make_shared(topo.num_domain(GEOPM_DOMAIN_CPU));
```
cmcantalupo: Nit: topo -> m_topo

src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+        std::set<std::string> result;
+        for (const auto &kv : m_signal_available) {
+            result.insert(kv.first);
+        }
+        return result;
+    }
+
+    std::set<std::string> SSTIOGroup::control_names(void) const
+    {
+        std::set<std::string> s;
```
cmcantalupo: Nit: s -> result?

src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+        }
+        return result;
+    }
+
+    int SSTIOGroup::push_signal(const std::string &signal_name, int domain_type, int domain_idx)
+    {
+        int result = -1;
+        auto it = m_signal_available.find(signal_name);
+        if (it != m_signal_available.end()) {
+
```
cmcantalupo: Extra ws?

src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+            if (domain_type != it->second.domain) {
+                throw Exception("wrong domain type", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
cmcantalupo: I know it is a bit verbose considering the use of `__FILE__` and `__LINE__`, but please prefix exceptions with the class name and method name:

```
throw Exception("SSTIOGroup::push_signal(): Wrong domain type", ...
```

some folks like to use `__func__` macro to avoid copy paste errors, but you still need to type out the class name and there may be some casting issues if you want to use the + operator, so I usually just type it out.

There are other examples below of exceptions without reference to calling function.

src/SSTIOGroup.cpp
```diff
@@ -0,0 +1,824 @@
...
+                    .description = description,
+                    .behavior = behavior
+                };
+            }
+
+            std::string field_signal_name = raw_name + ":" + field_name;
+            auto raw_sst_list = m_signal_available.at(raw_signal_name).signals;
+            std::vector<std::shared_ptr<Signal> > signals;
+            for (int domain_idx = 0; domain_idx < num_domain; ++domain_idx) {
+                auto field_signal = std::make_shared<MSRFieldSignal>(
```
cmcantalupo, line 775: Will this signal be visible to users with a name  prefixed by "MSR::", but provided by the "SST" IOGroup?  Would be nice if we could avoid that.

src/SSTIOGroup.hpp
```diff
@@ -0,0 +1,285 @@
...
+            std::function<double(const std::vector<double> &)> agg_function(const std::string &signal_name) const override;
+            std::function<std::string(double)> format_function(const std::string &signal_name) const override;
+            std::string signal_description(const std::string &signal_name) const override;
+            std::string control_description(const std::string &control_name) const override;
+            int signal_behavior(const std::string &signal_name) const override;
+            static std::string plugin_name(void);
+            static std::unique_ptr<IOGroup> make_plugin(void);
+
+            enum class SSTMailboxCommand : uint16_t {
+                TURBO_FREQUENCY = 0x7f,
```
cmcantalupo: Prefix enums with M_ please.

src/SSTIOGroup.hpp
```diff
@@ -0,0 +1,285 @@
...
+                int units;
+                std::string description;
+                m_signal_behavior_e behavior;
+            };
+            struct sst_signal_mailbox_raw_s {
+                //! @param command Which type of mailbox command
+                //! @param subcommand Subtype of the given command
+                //! @param fields Subfields of the mailbox
+                sst_signal_mailbox_raw_s(SSTIOGroup::SSTMailboxCommand command, uint16_t subcommand,
+                                         std::map<std::string, sst_signal_mailbox_field_s> fields)
```
cmcantalupo: should this be a `const std::map<std::string, sst_signal_mailbox_field_s> &`?  There are struct constructors below that follow the same pattern of pass by value for std::map.

cmcantalupo (commented):
Looks pretty good.  Mostly a bunch of nits.

src/SSTIOGroup.cpp
dannosliwcd: That's right. This is just a table that specifies the sst ioctl inputs associated with each signal. The variants are excluded from this table.

Any of the ioctl variants that exist within one type of signal/control have been given different names, like `SST::COREPRIORITY:0 ... SST::COREPRIORITY:3`. So then they're constant with respect to GEOPM signal/control name.

src/SSTIOGroup.cpp
dannosliwcd: The current implementation uses `operator==(const shared_ptr&, nullptr_t)` rather than casting. I'll change it to use bool.

src/SSTIOGroup.cpp
dannosliwcd: This won't include MSR in the name. This is just reusing the code that breaks a bitmasked field down to multiple signals. I'll add a comment to indicate that in the code.

src/SSTIOGroup.cpp
cmcantalupo: Oh if @dianarg were actively reviewing, I'm sure she'd disagree with my recommendation. Given that there is also an == operator specifically for nullptr_t, I'm not sure my recommendation is *clearer to the reader*.

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #1588, opened 2021-03-24
Initialize m_adjust_idx and m_saved_value in the SSTControl constructor

dannosliwcd:
Static analysis indicates that these variables are not initialized in the constructor. This change adds them to the initializer list.


cmcantalupo (approved):
LGTM

PR #1593, opened 2021-03-27
Add a PowerBalancerAgent policy parameter for enabling periodic balancing

cmcantalupo:

### DIFF
--- a/integration/experiment/power_sweep/power_sweep.py
+++ b/integration/experiment/power_sweep/power_sweep.py
@@ -78,7 +78,8 @@ def setup_power_bounds(mach, min_power, max_power, step_power):
 
 def report_signals():
     return ["CYCLES_THREAD@package", "CYCLES_REFERENCE@package",
-            "TIME@package", "ENERGY_PACKAGE@package"]
+            "TIME@package", "ENERGY_PACKAGE@package",
+            "MSR::PKG_POWER_LIMIT:PL1_POWER_LIMIT@package"]
 
 
 def trace_signals():
--- a/integration/experiment/power_sweep/power_sweep.py
+++ b/integration/experiment/power_sweep/power_sweep.py
@@ -100,7 +100,11 @@ def launch_configs(output_dir, app_conf, agent_types, min_power, max_power, step
     for power_cap in range(max_power, min_power-1, -step_power):
         for agent in agent_types:
             name = '{}'.format(power_cap)
-            options = {'power_budget': power_cap}
+            options = {'POWER_PACKAGE_LIMIT_TOTAL': power_cap}
+            if agent.startswith('power_balancer-'):
+                period = float(agent.split('power_balancer-')[1])
+                agent = 'power_balancer'
+                options['PERIOD_DURATION'] = period
             config_file = os.path.join(output_dir, name + '_agent.config')
             agent_conf = geopmpy.io.AgentConf(path=config_file,
                                               agent=agent,
--- a/src/PowerBalancerAgent.cpp
+++ b/src/PowerBalancerAgent.cpp
@@ -578,6 +578,7 @@ namespace geopm
                     package.runtime = balancer->runtime_sample();
                 }
             }
+            role.m_last_period_count = period_count;
         }
     }
 
--- a/src/PowerBalancerAgent.cpp
+++ b/src/PowerBalancerAgent.cpp
@@ -152,14 +152,14 @@ namespace geopm
         , m_platform_io(platform_io)
         , m_platform_topo(platform_topo)
         , m_sample_agg(sample_agg)
-        , m_num_domain(m_platform_topo.num_domain(GEOPM_DOMAIN_PACKAGE))
+        , m_num_package(m_platform_topo.num_domain(GEOPM_DOMAIN_PACKAGE))
         , m_num_core(m_platform_topo.num_domain(GEOPM_DOMAIN_CORE))
         , m_time_agg_idx(-1)
         , m_network_agg_idx(m_num_core, -1)
         , m_ignore_agg_idx(m_num_core, -1)
         , m_power_balancer(power_balancer)
         , M_STABILITY_FACTOR(3.0)
-        , m_package(m_num_domain, m_package_s {0.0, NAN, 0.0, 0.0, false, true, -1})
+        , m_package(m_num_package, m_package_s {0.0, NAN, 0.0, 0.0, false, true, -1})
         , M_MIN_PKG_POWER_SETTING(min_power)
         , M_MAX_PKG_POWER_SETTING(max_power)
         , m_is_single_node(is_single_node)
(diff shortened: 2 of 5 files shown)


integration/experiment/power_sweep/power_sweep.py
```diff
@@ -100,7 +100,11 @@ def launch_configs(output_dir, app_conf, agent_types, min_power, max_power, step
     for power_cap in range(max_power, min_power-1, -step_power):
         for agent in agent_types:
             name = '{}'.format(power_cap)
-            options = {'power_budget': power_cap}
+            options = {'POWER_PACKAGE_LIMIT_TOTAL': power_cap}
+            if agent.startswith('power_balancer-'):
```
bgeltz, line 105: Where is this funky agent string coming from?

src/PowerBalancerAgent.cpp
```diff
@@ -215,12 +222,21 @@ namespace geopm
...
                 m_policy = in_policy;
+                double duration = m_policy[M_POLICY_PERIOD_DURATION];
+                m_sample_agg->period_duration(duration);
+                double warm_time = M_WARM_PERIODS_MIN * duration;
+                if (warm_time >= M_WARM_TIME_MIN) {
+                    m_warm_periods = M_WARM_PERIODS_MIN;
+                }
+                else {
+                    m_warm_periods = static_cast<int>(M_WARM_TIME_MIN / duration) + 1;
+                }
```
bgeltz: Why isn't the else{} case sufficient for all usages?

In the case of periods longer than 1 sec, you're forcing M_WARM_PERIODS_MIN (10) when that many periods might not be necessary?

src/PowerBalancerAgent.cpp
```diff
@@ -541,26 +557,37 @@ namespace geopm
...
-                double balanced_epoch_runtime =  total - network - ignore;
+        int period_count = role.m_sample_agg->get_period();
+        // Wait m_warm_periods before starting algorithm
+        if (period_count > role.m_warm_periods &&
+            period_count != role.m_last_period_count) {
+            for (int pkg_idx = 0; pkg_idx < role.m_num_domain; ++pkg_idx) {
+                double sum_net_ig = 0.0;
+                std::set<int> nested_core = role.m_platform_topo.domain_nested(GEOPM_DOMAIN_CORE, GEOPM_DOMAIN_PACKAGE, pkg_idx);
+                std::set<int> active_nested_core;
+                std::set_intersection(nested_core.begin(), nested_core.end(),
```
bgeltz: This might be an area for performance optimization:
https://stackoverflow.com/questions/54763112/how-to-improve-stdset-intersection-performance-in-c

src/PowerBalancerAgent.cpp
```diff
@@ -143,23 +146,29 @@ namespace geopm
...
                                            double time_window,
                                            bool is_single_node,
-                                           int num_node)
+                                           int num_node,
+                                           const std::set<int> &active_core)
         : Role(num_node)
         , m_platform_io(platform_io)
         , m_platform_topo(platform_topo)
         , m_sample_agg(sample_agg)
         , m_num_domain(m_platform_topo.num_domain(GEOPM_DOMAIN_PACKAGE))
```
bgeltz: I realize it wasn't introduced in this PR, but ```m_num_domain``` is an awful variable name

src/PowerBalancerAgent.cpp
```diff
@@ -541,26 +557,37 @@ namespace geopm
...
+                for (auto &core_idx : active_nested_core) {
+                    /// We wish to measure runtime that is a function of node
+                    /// local optimizations only, and therefore uncorrelated
+                    /// between compute nodes.
+                    sum_net_ig += role.m_sample_agg->sample_period_last(role.m_network_agg_idx[core_idx]);
+                    sum_net_ig += role.m_sample_agg->sample_period_last(role.m_ignore_agg_idx[core_idx]);
+                }
+                double balanced_period_runtime = role.m_sample_agg->sample_period_last(role.m_time_agg_idx);
+                int num_selected = active_nested_core.size();
+                if (num_selected != 0) {
```
bgeltz: Seems like this check on ```active_nested_core.size()``` should be moved above line 571.  Then that for loop over active_nested_core could be bypassed if the size is 0.

The initial calculation of ```balanced_period_runtime``` on 578 could then move up above this check too?

bgeltz (changes requested):

integration/experiment/power_sweep/power_sweep.py
cmcantalupo: The user can specify it on the command line.

src/PowerBalancerAgent.cpp
cmcantalupo: The idea is that it will take about 200 periods to converge and show a benefit (at least that is how it has worked in the past), so we know that waiting 10 periods will be at most 5% of the runtime required for a benefit.

src/PowerBalancerAgent.cpp
cmcantalupo: Okay, I doubt this will be a performance bottle neck, as it is only done once per period and the set sizes will have only about 40 elements (number of cores on a socket)

src/PowerBalancerAgent.cpp
cmcantalupo: In principle the power control domain could be something other than the package, but the name of the control were are pushing has "PACKAGE" in the name, so introducing that ambiguity doesn't really make much sense.

src/PowerBalancerAgent.cpp
cmcantalupo: I mean the loop will be a noop if the size is zero.  I was just trying to avoid too much nesting

integration/experiment/power_sweep/power_sweep.py
bgeltz: I think this should be one of the argparse options for invoking this py file like the other policy params are derived.  Mashing the period into the agent name seems hacky.

src/PowerBalancerAgent.cpp
bgeltz: It's confusing when you consider the next line ```m_num_core```, is all.  I'd rather we were consistent when we setup for loops over domains so that you don't have to do sleuthing to ensure the loop is iterating over what you expect it to, especially now since we're considering multiple domains (PACKAGE and CORE).

To your point, this could have been ```m_num_default_power_domain``` or similar to make it more clear.

src/PowerBalancerAgent.cpp
bgeltz: Alrighty.

bgeltz (changes requested):

integration/experiment/power_sweep/power_sweep.py
cmcantalupo: Yeah that makes a lot of sense.  Especially because, as it is written, a user might try to specify "power_balancer,power_balancer-0.5,power_governor", but if they do that the output of the second balancer run overwrites the first one.  I know that by experience.

src/PowerBalancerAgent.cpp
cmcantalupo: Right on, I'll rename the member.

PR #1595, opened 2021-03-30
Address GitHub Actions CI issues

bgeltz:
- Properly use clang and gcc.
- Use the latest available versions of the above compilers.
- Test debug and release builds.
- Install missing dependencies (libomp).
- Address compile time issues arising from clang build.

These changes bring us back to parity with what we had running on TravisCI.


cmcantalupo (approved):
LGTM

PR #1596, opened 2021-03-31
Update BSD-3 header to include 2021

bgeltz:
Happy New Year.


cmcantalupo (dismissed):
at some point we prolly should to switch to

2015 - 2021

cmcantalupo (approved):

PR #1605, opened 2021-04-08
Introduce AVX comparison plot

bgeltz:
WIP

Based on presence of 'avx2' or 'avx512' in the Profile string in reports.

### DIFF
--- a/integration/experiment/launch_util.py
+++ b/integration/experiment/launch_util.py
@@ -46,7 +46,7 @@ from apps import apps
 
 
 def launch_run(agent_conf, app_conf, run_id, output_dir, extra_cli_args,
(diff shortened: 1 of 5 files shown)


integration/experiment/launch_util.py
```diff
@@ -65,6 +65,8 @@ def launch_run(agent_conf, app_conf, run_id, output_dir, extra_cli_args,
     app_name = app_conf.name()
 
     uid = '{}_{}_{}'.format(app_name.lower(), agent_name, run_id)
+    if label is not None:
+        uid += '_{}'.format(label)
```
bgeltz, line 69: This breaks parsing of the iteration # that's on the end of this string.  This needs to be revisited.

PR #1606, opened 2021-04-08
Feature branch for GEOPM service

cmcantalupo:
Creates a new directory with separate build infrastructure that supports the geopm systemd service.  This service will provide the PlatformIO features through a daemon interface with access control capabilities and also save/restore capabilities.


service/bin/geopmd
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+
+from dasbus.loop import EventLoop
+from dasbus.connection import SessionMessageBus
+import geopmpy.pio
+
+class GEOPMService(object):
+    __dbus_xml__ = """
```
dannosliwcd: Looks like dasbus can also generate the xml for us if we use python3 type annotations and decorate the class with `@dasbus.dbus_interface`. Not sure if there are advantages/disadvantages to that approach though.

service/bin/geopmd
```diff
@@ -0,0 +1,103 @@
...
+        </interface>
+    </node>
+    """
+    def __init__(self, pio=geopmpy.pio):
+        self._pio = pio
+
+    def AccessList(self, group):
+        return self._pio.signal_names(), self._pio.control_names()
+
+    def AccessControl(self, group,
```
dannosliwcd: Maybe `SetAccessList`

service/bin/geopmd
```diff
@@ -0,0 +1,103 @@
...
+                 <arg direction="in" name="query" type="as" />
+                 <arg direction="out" name="result" type="as" />
+            </method>
+        </interface>
+    </node>
+    """
+    def __init__(self, pio=geopmpy.pio):
+        self._pio = pio
+
+    def AccessList(self, group):
```
dannosliwcd: Maybe `GetAccessList`?

service/bin/geopmd
```diff
@@ -0,0 +1,103 @@
...
+
+    def AccessControl(self, group,
+                      allowed_signals, denied_signals,
+                      allowed_controls, denied_controls):
+        raise NotImplementedError
+
+    def AccessDelete(self, group):
+        raise NotImplementedError
+
+    def AccessAllowed(self):
```
dannosliwcd: Do we need both `AccessAllowed` and `AccessList`? Looks like they are the same except this also gives domain context.

service/bin/geopmd
```diff
@@ -0,0 +1,103 @@
...
+        names = self._pio.signal_names()
+        signals = [(nn, self._pio.signal_domain_type(nn))
+                   for nn in self._pio.signal_names()]
+        controls = [(nn, self._pio.control_domain_type(nn))
+                    for nn in self._pio.control_names()]
+        return signals, controls
+
+    def Loop(self, signals, controls, interval):
```
dannosliwcd: Maybe `StartXYZLoop`? I guess `XYZ` could be something like `PlatformIO`, `Sampling`, `Control`, `GeopmControl`, etc. Or maybe `StartGeopmController`

service/dbus-1/interfaces/io.github.geopm.xml
```diff
@@ -0,0 +1,136 @@
...
+      <arg name="query" direction="in" type="as">
+        <doc:doc><doc:summary>Array of control names</doc:summary></doc:doc>
+      </arg>
+      <arg name="result" direction="out" type="as">
+        <doc:doc><doc:summary>Array of structures each contains the control description</doc:summary></doc:doc>
+      </arg>
+      <doc:doc>
+        <doc:description>
+          <doc:para>
+            Enables a client to query information about an array of signals.
```
dannosliwcd: Array of `signals`->`controls`

service/system/systemd-geopm.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=User space interface to hardware features with fine grained access control
+
+[Service]
+Environment=PYTHONUNBUFFERED=true
```
dannosliwcd: This reminds me, do we need to do anything special for output, like using a systemd logger or something? Or is print-based logging sufficient?

dannosliwcd:
Should one of these methods (or an additional method) export a mapping of (signal/control)->(offset within shmem buffer)?

service/bin/geopmd
cmcantalupo: My idea here was that AccessAllowed() is for clients to find out what they might want to add to the Loop call (hence no group input parameter, and the domain), where as the AccessList is for an administrator to get the access list for a specified group.  If we can (unclear to me) restrict the AccessList to be root only, it will make more sense.

service/bin/geopmd
cmcantalupo: We will probably need to have access to the XML to write a C++ client, that was my thinking behind creating the interface/io.github.geopm.xml file.

service/README.md
```diff
@@ -0,0 +1,95 @@
...
+had when the session began which is reported as zero.
+
+
+Signals and Controls
+--------------------
+
+Each geopm signal and control has an associated name and a hardware
+domain.  The name is a unique string identifier defined by the IOGroup
+that provides the signal or control.  The hardware domain that
+provides the interface is represented by one of the enumerations as
```
dannosliwcd, line 34: Specifically, this is the integer value of those enumerations, right? (vs. a stringified enum identifier)

dannosliwcd: Also, domain _index_ is never used at this scope of the interface, right? (No need to explicitly state that here, just checking in case I'm assuming wrong)

service/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Access to signals and controls is determined by Unix group
+associations.  A default set of access is configured for all users on
+the system and this list can be augmented by lists associated with
+Unix groups.  A user will have access to all of the signals and
+controls provided in the default access list as well as any access
+lists that are associated with groups that they belong to.  The default
+lists are stored in:
+
+    /etc/geopm-service/0.DEFAULT_ACCESS/allowed_signals
```
dannosliwcd, line 55: Does the zero have any meaning specific to us, or is that just a D-Bus or systemd convention?

service/bin/geopmd
cmcantalupo: TBD in feature branch...This seems like a good feature of dasbus to use.

service/bin/geopmd
cmcantalupo: Decided on a concept of opening and closing a "session".   LMKWYT

service/system/systemd-geopm.service
cmcantalupo: Good question, I've certainly seen special logging it in some examples for a variety of reasons.  We can look into it.   Messages are turning up in the syslog, but we may want our own log file.

service/README.md
cmcantalupo: I just needed a number, because groups cannot begin with numbers.

service/README.md
cmcantalupo: correct

cmcantalupo:
This PR was merged into the geopm-service branch of the main repository to facilitate collaboration.  There are still many issues both mentioned in the PR and also unmentioned, that must be addressed before the geopm-service branch will be merged to the dev branch.

PR #1618, opened 2021-04-22
Address recent user feedback

bgeltz:
- Move README section about Intel compiler overrides.
- Allow acpi-cpufreq/userspace to be a valid combination for the frequency driver and governor.
- Move warning about the Controller oversubscribing CPUs.

### DIFF
--- a/src/ApplicationSampler.cpp
+++ b/src/ApplicationSampler.cpp
@@ -395,6 +395,7 @@ namespace geopm
         int result = m_num_cpu - 1;
         int num_core = m_topo.num_domain(GEOPM_DOMAIN_CORE);
         bool found_inactive_core = false;
+        bool found_inactive_cpu = false;
         std::vector<bool> is_core_active(num_core, false);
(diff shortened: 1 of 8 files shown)


src/ApplicationSampler.cpp
```diff
@@ -424,6 +424,13 @@ namespace geopm
 #ifdef GEOPM_DEBUG
         std::cout << "Info: <geopm> ApplicationSampler::sampler_cpu(): The Controller will run on logical CPU " << result << std::endl;
 #endif
+
+        if (result == m_num_cpu - 1) {
```
cmcantalupo: This should be triggered based on

    if (!found_inactive_core)  

Potentially we could add another bool and print a different message if we are sharing a core with the application, but on an unused hyperthread (found_inactive_thread set after line 419)?  Even more information could be printed if we are running on core 0 (sharing a core with the OS).

cmcantalupo (commented):
Don't think the warning message test is quite right.

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #1619, opened 2021-04-22
Update geopm service to provide support for an IOGroup

cmcantalupo:
- Simple mechanism for enabling support for signals and controls provided by the service.
- Put the ServiceIOGroup at the bottom of the stack, so it is given least preference for signals and controls
- Faster mechanisms provided by other IOGroups are used where possible.
- Use save/restore feature of the service only when a user requests a control through the service


service/README.md
```diff
@@ -8,22 +8,47 @@ access management is configurable by system administrators with the
...
+controls that cannot be provided by any of the IOGroups loaded by the
+unprivileged user process.
+
+The geopm service also provides a fail-safe save/restore mechanism for
+any platform controls that are exposed by PlatformIO.  This is done by
+initiating a session with the service when the ServiceIOGroup is first
+created by a libgeopm or libgeopmpolicy user.  All controls are saved
+by the GEOPM service prior to opening the session with the end-user.
+When the session is closed, either by an explicit D-Bus call by the
+client, or when the process that initiated the client session ends,
```
dannosliwcd: I think this is mentioned a few times in these docs, but I don't think the docs specify what mechanism is used to detect a terminated process. This failsafe discussion (or somewhere near it) is probably a good place to note that, since our failsafe hinges on that.

service/README.md
```diff
@@ -8,22 +8,47 @@ access management is configurable by system administrators with the
...
+counters) are reported with respect to the value they had when the
+session began, which is reported as zero.
+
+When a client calls read_signal() or write_control() on their
+PlatformIO object and the only IOGroup that provides the signal or
+control is the ServiceIOGroup, then each request goes through the slow
+D-Bus interface.  A fast batch mechanism is available to support the
+ServiceIOGroup::read_batch() and ServiceIOGroup::write_batch()
+interfaces.  When a client process uses the ServiceIOGroup for batch
+operations a batch session through the D-Bus interface in opened with
```
dannosliwcd: in opened -> is opened

service/README.md
```diff
@@ -8,22 +8,47 @@ access management is configurable by system administrators with the
...
+a request for access to a set of signals and/or controls.  Once this
+session is established, after validating all requested signals and
+controls, a protocol between the server and client is enabled that
+uses inter-process shared memory and sending Unix signals for fast
+access.  In this documentation we will call these "Unix signals" to
+differenciate from the GEOPM signal concept which is unrelated to the
```
dannosliwcd: differentiate

service/README.md
```diff
@@ -38,21 +63,17 @@ controls available on a system can be discovered with the
...
 administrators that will help them understand what is enabled for an
-end-user when access is granted to a signal or control.  The platform
-topology information about a system can also be shown through
-`geopmaccess`.
+end-user when access is granted to a signal or control.
 
 
-Access Control
---------------
+Access Managment
```
dannosliwcd: Management

service/README.md
```diff
@@ -38,21 +63,17 @@ controls available on a system can be discovered with the
...
-Access to signals and controls is determined by Unix group
-associations.  A default set of access is configured for all users on
-the system and this list can be augmented by lists associated with
-Unix groups.  A user will have access to all of the signals and
-controls provided in the default access list as well as any access
-lists that are associated with groups that they belong to.  The
-default lists are stored in:
+Access to signals and controls through the GEOPM service is configured
+by the system administrator.  The administrator controls a default
+access list applies to all users of the system and this list can be
```
dannosliwcd: applies -> applied

service/README.md
```diff
@@ -74,24 +95,74 @@ command line tool.
...
+PlatformIO object is created with libgeopm or libgeopmpolicy while the
+GEOPM systemd service is active.  This session is initially opened in
+read only mode.  If any calls are made to the D-Bus write APIs:
+
+    io.github.geopm.PlatformWriteControl
+    io.github.geopm.PlatformPushControl
+
+then the session is converted into a read/write mode session.  Only
+one read/write mode session is allowed at any time, and if a client
+attempts to begin a read/write session while another client has one
```
dannosliwcd: Is there a way for a privileged third party (e.g., admin) to stop an existing session? Or is there another way to handle denial of service that exploits this rule? I guess the failsafe disconnector + single-tenant assumption might be sufficient, as long as we know it is guaranteed.

dannosliwcd: e.g., I think we wouldn't be able to count on the destructor alone since that may not execute if the program does not terminate cleanly.

PR #1622, opened 2021-04-28
Update to allow for testing NVML accelerator code without nvml libraries

### LINKED ISSUE #1295: NVML code is not covered by CI
All NVML code in src/ and test/ is heavily ifdef'd out using ifdef GEOPM_ENABLE_NVML to reduce potential impact on users that do not use NVML devcies.  A side effect of this is that this code is not covered by Travis-CI as currently configured.

An update may be required to provide more extensive mocking and testing of the NVML components to achieve better coverage.

### LINKED ISSUE #1490: Unit tests for NVMLIOGroup require libnvml
There is a libnvml wrapper class, NVMLDevicePool which is intended to provide a mockable interface to enable unit testing of all dependency objects.  Unfortunately, unit testing of any NVML related class is fully disabled when libnvml is not available.

lhlawson:
GEOPM Accelerator code updates to allow for testing of NVML libraries without nvml.h.  
Tested on systems with and without NVML accelerators.

Summary of change.
- Updates to AcceleratorTopo that break out default functionality to AcceleratorTopoNull
- Addition of NVMLDevicePoolThrow to provide nvml_device_pool in absence of a full implementation
- Changed AcceleratorTopoTest to AcceleratorTopoNullTest   
- Fixes #1295 from github issues.
- Fixes #1490 from github issues.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -446,6 +446,8 @@ libgeopmpolicy_la_SOURCES = contrib/json11/json11.cpp \
                             contrib/json11/json11.hpp \
                             src/AcceleratorTopo.cpp \
                             src/AcceleratorTopo.hpp \
+                            src/AcceleratorTopoNull.cpp \
+                            src/AcceleratorTopoNull.hpp \
                             src/Accumulator.cpp \
                             src/Accumulator.hpp \
                             src/Admin.cpp \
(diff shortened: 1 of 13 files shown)


lhlawson:
As part of this commit additional small changes were made to the NVMLDevicePool, NVMLAcceleratorTopo, and NVMLIOGroup with respect to data types, casting, etc to fix type mismatch errors when using gcc-9 and clang-9.

bgeltz (approved):

PR #1624, opened 2021-05-05
Get ServiceIOGroup::read_signal functioning

cmcantalupo:

### DIFF
--- a/service/src/geopm_pio.h
+++ b/service/src/geopm_pio.h
@@ -103,11 +103,11 @@ int geopm_pio_control_description(const char *control_name,
 
 /// @brief C interface to get enums associated with a signal name.
 ///
-/// This interfacesupports DBus PlatformGetSignalInfo method.  This C
+/// This interface supports DBus PlatformGetSignalInfo method.  This C
 /// interface is implemented using several PlatformIO methods unlike
 /// the other wrappers in this header.
 ///
-/// @param [in] signal_name Name of signal to query
+/// @param [in] signal_name Name of signal to query.
 ///
 /// @param [out] aggregation_type One of the Agg::m_type_e enum values
 ///        describing the way the signal is aggregated.
@@ -120,7 +120,7 @@ int geopm_pio_control_description(const char *control_name,
 ///        IOGroup::m_signal_behavior_e enum values that decribes
 ///        the signals behavior over time.
 ///
-/// @return Zero on success, error value on failure
+/// @return Zero on success, error value on failure.
 int geopm_pio_signal_info(const char *signal_name,
                           int *aggregation_type,
                           int *format_type,
--- a/service/src/geopm_pio.h
+++ b/service/src/geopm_pio.h
@@ -100,6 +100,27 @@ int geopm_pio_control_description(const char *control_name,
                                   size_t description_max,
                                   char *description);
 
+
+/// @brief C interface to get enums associated with a signal name.
+///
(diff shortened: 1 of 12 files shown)


src/ServiceProxy.cpp
```diff
@@ -81,20 +81,186 @@ namespace geopm
...
+            error_message << "ServiceProxy::platform_get_signal_info(): Failed to call sd-bus function sd_bus_call(), error:"
+                          << err << " name: " << bus_error.name << ": " << bus_error.message;
+            throw Exception(error_message.str(), GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        err = sd_bus_message_enter_container(bus_reply, SD_BUS_TYPE_ARRAY, "(ssiiii)");
+        if (err < 0) {
+            throw Exception("ServiceProxy::platform_get_signal_info(): Failed to enter \"a(ssiiii)\" container",
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        for (bool is_done = false; !is_done;) {
```
bgeltz: Shouldn't this be a while loop?  This looks weird as a for loop.

service/src/geopm_pio.h
```diff
@@ -100,6 +100,11 @@ int geopm_pio_control_description(const char *control_name,
                                   size_t description_max,
                                   char *description);
 
+int geopm_pio_signal_info(const char *signal_name,
```
bgeltz, line 124: Docu?

I see by inspecting the implementation that everything past the first argument is supposed to be output.  If we're moving away from having all these APIs described in the ronn files, maybe we should put some notes here?

bgeltz (changes requested):
A few minor comments.

Largely this looks pretty good.

service/src/geopm_pio.h
```diff
@@ -101,6 +101,32 @@ int geopm_pio_control_description(const char *control_name,
                                   char *description);
 
 
+/// @brief C interface to get enums associated with a signal name.
+///
+/// This interfacesupports DBus PlatformGetSignalInfo method.  This C
```
bgeltz:
```suggestion
/// This interface supports DBus PlatformGetSignalInfo method.  This C
```

service/src/geopm_pio.h
```diff
@@ -101,6 +101,32 @@ int geopm_pio_control_description(const char *control_name,
...
+/// interface is implemented using several PlatformIO methods unlike
+/// the other wrappers in this header.
+///
+/// @param [in] signal_name Name of signal to query
```
bgeltz:
```suggestion
/// @param [in] signal_name Name of signal to query.
```

service/src/geopm_pio.h
```diff
@@ -101,6 +101,32 @@ int geopm_pio_control_description(const char *control_name,
...
+///
+/// @param [out] format_type One of the geopm::string_format_e
+///        enums defined in Helper.hpp that defines how to format
+///        the signal as a string.
+///
+/// @param [out] behavior_type One of the
+///        IOGroup::m_signal_behavior_e enum values that decribes
+///        the signals behavior over time.
+///
+/// @return Zero on success, error value on failure
```
bgeltz:
```suggestion
/// @return Zero on success, error value on failure.
```

bgeltz (changes requested):

bgeltz (approved):

PR #1626, opened 2021-05-06
Get automatic save restore wired up with GObj

cmcantalupo:
- Also, fix service directory build broken in recent rebase

### DIFF
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -39,6 +39,10 @@ import pwd
 import grp
 import json
 import shutil
+import psutil
+import gi
+from gi.repository import GLib
+
 from . import pio
 from . import topo
 from dasbus.connection import SystemMessageBus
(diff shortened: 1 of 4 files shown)


service/geopmdpy/service.py
```diff
@@ -219,6 +231,14 @@ def _write_mode(self, client_pid):
             # daemon restart
             self._pio.save_control()
 
+    def _watch_client(self, client_pid):
+        return GLib.timeout_add(self._WATCH_INTERVAL_MSEC, self._check_client, client_pid)
+
+    def _check_client(self, client_pid):
+        if client_pid in self._sessions and not psutil.pid_exists(client_pid):
+            self.close_session(client_pid)
```
dannosliwcd, line 239: The `return False` causes the timeout to be destroyed, and close_session also removes the timeout source. Is it okay for both to happen?

cmcantalupo: I was worried about this too.  If you call GObj.source_remove() and the handle you pass is already removed, a warning is printed.  If, however, you call GObj.source_remove(), and then the callback returns False, there is no warning or error emitted.

PR #1632, opened 2021-05-10
Fix use of default access list and enable architectural MSRs on all platforms

cmcantalupo:


service/src/MSRIOGroup.cpp
```diff
@@ -127,31 +147,11 @@ namespace geopm
...
-        register_power_signals();
-
-        register_control_alias("POWER_PACKAGE_LIMIT", "MSR::PKG_POWER_LIMIT:PL1_POWER_LIMIT");
-        register_control_alias("FREQUENCY", "MSR::PERF_CTL:FREQ"); // TODO: Remove @ v2.0
-        register_control_alias("CPU_FREQUENCY_CONTROL", "MSR::PERF_CTL:FREQ");
-        register_control_alias("POWER_PACKAGE_TIME_WINDOW", "MSR::PKG_POWER_LIMIT:PL1_TIME_WINDOW");
+        if (max_turbo_name != "") {
+            register_signal_alias("CPU_FREQUENCY_MAX", max_turbo_name);
+            set_signal_description("CPU_FREQUENCY_MAX", "Maximum processor frequency.");
+        }
```
bgeltz, line 158: I'm a little surprised by the alias changes here (i.e. adding "CPU_..." where appropriate).  I tried to introduce this change back in Q4-2020 through some work Diana was doing, and at the time the decision was made that these aliases (e.g. "FREQUENCY_MAX")  had already been communicated to our customers, and therefore should not be changed until our next major release.

Has this requirement been relaxed?  If not, the following legacy aliases are missing:
FREQUENCY
FREQUENCY_MAX

service/src/MSRIOGroup.cpp
```diff
@@ -100,17 +100,37 @@ namespace geopm
     {
         // Load available signals and controls from files
         parse_json_msrs(arch_msr_json());
-        parse_json_msrs(platform_data(m_cpuid));
+        try {
+            // Try to extend list of MSRs if CPUID is recognized
+            parse_json_msrs(platform_data(m_cpuid));
+        }
+        catch (const Exception &ex) {
+            // Only load architectural MSRs
```
bgeltz, line 108: This is cool.

bgeltz (changes requested):

PR #1636, opened 2021-05-13
Add integration test for geopmaccess and streamline error reporting in service integration test

cmcantalupo:


service/test/test_su_give_access.sh
```diff
@@ -0,0 +1,103 @@
...
+TEST_SCRIPT=${TEST_DIR}/test_write_session.sh
+SAVE_SIGNALS=test_su_give_access_save_signals.txt
+SAVE_CONTROLS=test_su_give_access_save_controls.txt
+TEST_SIGNALS=test_su_give_access_test_signals.txt
+TEST_CONTROLS=test_su_give_access_test_controls.txt
+
+test_error() {
+    echo "Error: $1" 1>&2
+    exit -1
+}
```
bgeltz, line 65: If we end up writing a whole suite of tests in bash, we'll probably want to pull out helper functions like this one into their own file.

bgeltz (approved):

PR #1637, opened 2021-05-13
First unit test for geopmdpy

cmcantalupo:


service/geopmdpy_test/test_service.py
```diff
@@ -0,0 +1,98 @@
...
+        self.assertEqual([], controls)
+
+    def test_get_group_access_default(self):
+        signals_expect = ['geopm', 'signals', 'default', 'energy']
+        controls_expect = ['geopm', 'controls', 'default', 'power']
+        default_dir = os.path.join(self._CONFIG_PATH.name, '0.DEFAULT_ACCESS')
+        os.makedirs(default_dir)
+        signal_file = os.path.join(default_dir, 'allowed_signals')
+        lines = list(signals_expect)
+        lines.insert(2, '# Comment line')
```
dannosliwcd: I saw there was some whitespace stripping code. If a line starts like whitespace before the `#`, should that also be a comment, or is that not allowed?

cmcantalupo: Good point, I'll strip whitespace before the hash check

bgeltz (approved):

PR #1639, opened 2021-05-17
Add a PAPI IOGroup

dannosliwcd:
Draft of the PAPI IOGroup for use in GEOPM. Currently determines which events to register from a whitespace-delimited GEOPM_PAPI_EVENTS environment variable. From a recent design discussion, an alternative solution may be to make the IOGroup dynamically register new signals by a control-based interface. E.g., writing 1 to PAPI_L1_DCM would make this IOGroup:
1. Determine whether that is a valid PAPI event. If not, simply ignore the request (it may be intended for a different IOGroup, so don't error out)
2. Add the event to the registered event set
3. Add a new signal with that event's name

Other things that still need to happen are to add unit tests and a man page. If you want to try this out before the man page is written, refer to the README from the plugin-based implementation: https://github.com/dannosliwcd/geopm_papi_iogroup/blob/master/README.md

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -544,6 +544,8 @@ libgeopmpolicy_la_SOURCES = contrib/json11/json11.cpp \
                             src/MonitorAgent.hpp \
                             src/OptionParser.cpp \
                             src/OptionParser.hpp \
+                            src/PAPIIOGroup.cpp \
+                            src/PAPIIOGroup.hpp \
                             src/PlatformIO.cpp \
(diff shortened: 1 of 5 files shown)


cmcantalupo (commented):
LGTM, be nice to have a PAPI class that is a thin wrapper around the PAPI API's that threw exceptions based on return values.  Also kind of want to append "_ENABLE" to the signal name to make it an enabling control, but that is for future work anyway.

PR #1646, opened 2021-05-21
Added Support for Intel Cache Monitoring Technology and Memory Bandwidth Monitoring

lhlawson:
Added MSRs and derived signals related to the CMT and MBM features to allow users to configure and query cache and bandwidth metrics as described in section 17.18.1 of Vol. 3B of the SDM.  This includes CPUID checks for feature support, adding per cpu RMID registers, and added package scoped QM event selection and counter registers

- Update to SKX MSR list: QM_EVTSEL, QM_CTR, PQR_ASSOC
- Update to arch MSR list: added CPUID Max Limit field to MISC_ENABLE register
- Added register_rdt function to derive a scaled QM_CTR and QM_CTR rate signal


src/MSRIOGroup.cpp
```diff
@@ -582,6 +661,38 @@ namespace geopm
         return ((family << 8) + model);
     }
 
+    MSRIOGroup::rdt_info MSRIOGroup::rdt(void) {
```
lhlawson: Move curly bracket to next line

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+
+        auto read_it = m_signal_available.find(msr_name);
+        if (read_it != m_signal_available.end()) {
+            auto readings = read_it->second.signals;
+            int ctr_domain = read_it->second.domain;
+
+            // Begin bit, end bit, and function are currently hardcoded as there's no accessor
+            // for MSRFieldSignal values that doesn't involve reparsing the JSON file
+            MSRIOGroup::add_msr_field_signal("QM_CTR",
+                                         signal_name,
```
cmcantalupo: ws

src/msr_data_arch.cpp
```diff
@@ -242,6 +242,15 @@ namespace geopm
                     "behavior":  "variable",
                     "writeable": false
                 },
+                "LIMIT_CPUID_MAXVAL": {
+                    "begin_bit": 22,
+                    "end_bit":   22,
+                    "function":  "scale",
+                    "units":     "none",
+                    "scalar":    1.0,
+                    "behavior":  "variable",
```
cmcantalupo, line 251: I see these fields are all marked as "variable".  Are these values not constant?

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+        for (int dom_idx = 0; dom_idx < num_domain; ++dom_idx) {
+            disable += read_signal("MSR::MISC_ENABLE:LIMIT_CPUID_MAXVAL", domain, dom_idx);
+        }
+        if ((disable != 0.0) || (!m_rdt.rdt_support)) {
+            return;
+        }
+
+        std::string signal_name = "QM_CTR_SCALED";
+        std::string msr_name = "MSR::QM_CTR:RM_DATA";
+        std::string description = "Resource Monitor Data scaled to number of bytes by the conversion factor";
```
cmcantalupo: Maybe this description could be improved? I'm not sure how this is used.

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+
+            // Begin bit, end bit, and function are currently hardcoded as there's no accessor
+            // for MSRFieldSignal values that doesn't involve reparsing the JSON file
+            MSRIOGroup::add_msr_field_signal("QM_CTR",
+                                         signal_name,
+                                         ctr_domain,
+                                         0, 31,
+                                         MSR::string_to_function("scale"),
+                                         (double)m_rdt.mbm_scalar,
+                                         IOGroup::M_UNITS_NONE,
```
cmcantalupo, line 321: Why don't we know the units?

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+
+
+        // register time signal; domain board
+        std::string time_name = "MSR::TIME";
+        std::shared_ptr<Signal> time_sig = std::make_shared<TimeSignal>(m_time_zero, m_time_batch);
+        m_signal_available[time_name] = {std::vector<std::shared_ptr<Signal> >({time_sig}),
+                                         GEOPM_DOMAIN_BOARD,
+                                         IOGroup::M_UNITS_SECONDS,
+                                         Agg::select_first,
+                                         "Time in seconds used to calculate power"};
```
cmcantalupo: power?

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
         }
     }
 
+    void MSRIOGroup::register_rdt_signals(void)
+    {
+        // It may be necessary to wrap the initial rdt() call in this check
+        int domain = signal_domain_type("MSR::MISC_ENABLE:LIMIT_CPUID_MAXVAL");
+        int num_domain = m_platform_topo.num_domain(domain);
+        double disable = 0.0;
```
cmcantalupo: bool?

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+        for (int dom_idx = 0; dom_idx < num_domain; ++dom_idx) {
+            disable += read_signal("MSR::MISC_ENABLE:LIMIT_CPUID_MAXVAL", domain, dom_idx);
```
cmcantalupo: If we use agg function of and logical_and this can be done with a single read_signal call.

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+        // register time signal; domain board
+        std::string time_name = "MSR::TIME";
+        std::shared_ptr<Signal> time_sig = std::make_shared<TimeSignal>(m_time_zero, m_time_batch);
+        m_signal_available[time_name] = {std::vector<std::shared_ptr<Signal> >({time_sig}),
+                                         GEOPM_DOMAIN_BOARD,
+                                         IOGroup::M_UNITS_SECONDS,
+                                         Agg::select_first,
+                                         "Time in seconds used to calculate power"};
+
+        int derivative_window = 8;
```
cmcantalupo: put these constants in the constructor maybe?

cmcantalupo (commented):
Some comments inline.

src/msr_data_arch.cpp
lhlawson: The LIMIT_CPUID_MAXVAL is listed as R/W in the SDM Vol4 Table 2-2. IA-32 Architectural MSRs.  Others appear to be the same.

src/MSRIOGroup.cpp
lhlawson: Updated.  It's intended as Bytes/Second.

src/MSRIOGroup.hpp
```diff
@@ -209,6 +212,17 @@ namespace geopm
             std::shared_ptr<geopm_time_s> m_time_zero;
             std::shared_ptr<double> m_time_batch;
 
+            struct rdt_info
+            {
+                bool rdt_support;
+                uint32_t rmid_max;
```
lhlawson: Currently rmid_max is not used, instead per cpu family bit ranges for RMID related registers are defined in the appropriate msr json file.

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+        if (read_it != m_signal_available.end()) {
+            auto readings = read_it->second.signals;
+            int ctr_domain = read_it->second.domain;
+
+            // Begin bit, end bit, and function are currently hardcoded as there's no accessor
+            // for MSRFieldSignal values that doesn't involve reparsing the JSON file
+            MSRIOGroup::add_msr_field_signal("QM_CTR",
+                                             signal_name,
+                                             ctr_domain,
+                                             0, 31,
```
lhlawson: We can't easily get begin and end bit for a signal once it's created.  Hard-coding range here because of that.

src/MSRIOGroup.cpp
```diff
@@ -1176,6 +1288,15 @@ namespace geopm
...
 
+                if (msr_field_name == "QM_EVT_SEL:RMID" || msr_field_name == "PQR_ASSOC:RMID") {
+                    if((end_bit - begin_bit) != m_rdt.rmid_bit_width) {
+                        throw Exception("MSRIOGroup::" + std::string(__func__) + "(): RMID bit width " +
+                                        std::to_string(end_bit - begin_bit) + "  does not match CPUID value: " +
+                                        std::to_string(m_rdt.rmid_bit_width), GEOPM_ERROR_INVALID, __FILE__,
+                                        __LINE__);
+                    }
+                }
+
```
lhlawson: Special casing in JSON parse because we can't easily access the signal bit fields later to check if they're valid.  This also is the earliest point at which we can catch a RMID max val vs bit field mismatch.

src/MSRIOGroup.cpp
lhlawson: These match the POWER derivative window (and reuse the time signal), is it reasonable to just consolidate this from register_rdt_signals and register_power_signals?

test/legacy_allowlist.out
```diff
@@ -63,6 +63,10 @@
 0x00000618 0x0000000000ffffff # DRAM_POWER_LIMIT
 0x00000620 0x0000000000007f7f # UNCORE_RATIO_LIMIT
 0x00000621 0x0000000000000000 # UNCORE_RATIO_STATUS
+0x00000c8d 0x000003FF000000FF # IA32_QM_EVTSEL
+0x00000c8e 0x0000000000000000 # IA32_QM_CTR
+0x00000c8f 0xFFFFFFFF000001FF # IA32_PQR_ASSOC
+0x00000e00 0x00000000ffffffff # C0_MSR_PMON_BOX_CTL
```
lhlawson, line 69: Additional testing may be needed on these bit fields, the msr_data_skx json file and the MSRIOGroup.cpp checks of bit width.  
According to the SDM the general support is:
image

but E5 v4 shows:
image

While a read of CPUID on a Broadwell system from shows:
Maximum range of RMID = 175

ceil(log2(175)+1) --> 9.   It appears N+31:32 in this case should be 40:32, not 41:32.

src/MSRIOGroup.cpp
cmcantalupo: I'm suggesting that there are two new constant members m_derivative_window and m_sleep_time, and that these be set in the constructor.  If they can be reused elsewhere, all the better

src/MSRIOGroup.cpp
```diff
@@ -285,6 +288,82 @@ namespace geopm
...
+    void MSRIOGroup::register_rdt_signals(void)
+    {
+        // It may be necessary to wrap the initial rdt() call in this check
+        int domain = signal_domain_type("MSR::MISC_ENABLE:LIMIT_CPUID_MAXVAL");
+        int num_domain = m_platform_topo.num_domain(domain);
+        bool disable = false;
+        for (int dom_idx = 0; dom_idx < num_domain; ++dom_idx) {
+            disable |= (bool)read_signal("MSR::MISC_ENABLE:LIMIT_CPUID_MAXVAL", domain, dom_idx);
+        }
+        if (disable || (!m_rdt.rdt_support)) {
```
cmcantalupo:
```suggestion
        if (disable || !m_rdt.rdt_support) {
```

src/MSRIOGroup.cpp
cmcantalupo: Okay

src/MSRIOGroup.cpp
```diff
@@ -1176,6 +1288,15 @@ namespace geopm
                     description = field_data["description"].string_value();
                 }
 
+                if (msr_field_name == "QM_EVT_SEL:RMID" || msr_field_name == "PQR_ASSOC:RMID") {
+                    if((end_bit - begin_bit) != m_rdt.rmid_bit_width) {
+                        throw Exception("MSRIOGroup::" + std::string(__func__) + "(): RMID bit width " +
```
cmcantalupo: I think this is fine, but I'd prefer using ostringstream rather than string conversion with adds to generate the error message.

src/MSRIOGroup.hpp
```diff
@@ -209,6 +212,17 @@ namespace geopm
...
+            struct rdt_info
+            {
+                bool rdt_support;
+                uint32_t rmid_bit_width;
+                uint32_t mbm_scalar;
+            };
+            rdt_info m_rdt;
+            /// @brief Return the Intel Resource Director Technology
+            ///        support information
+            static rdt_info rdt(void);
```
cmcantalupo: rdt -> get_rdt_info

src/MSRIOGroup.cpp
```diff
@@ -98,6 +98,7 @@ namespace geopm
         , m_is_fixed_enabled(false)
         , m_time_zero(std::make_shared<geopm_time_s>(time_zero()))
         , m_time_batch(std::make_shared<double>(NAN))
+        , m_rdt(rdt())
```
cmcantalupo: m_rdt -> m_rdt_info

src/MSRIOGroup.cpp
lhlawson: Update MSRIOGroup to provide a new signal exposing RMID scalar (MSR::RMID_QM_SCALAR?), and a combination signal that multiplies RMID Scalar Value and QM_CTR raw.

cmcantalupo (commented):
A couple of outstanding rename issues:

m_rdt -> m_rdt_info
rdt() -> get_rdt_info()

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #1654, opened 2021-06-10
GEOPM service sst integration test

cmcantalupo:


service/requirements.txt
```diff
@@ -0,0 +1,4 @@
+cffi==1.13.2
```
cmcantalupo, line 1: This file needs to be accounted for in the MANIFEST_EXEMPT and the two Makefile.am EXTRA_DIST.

service/test/test_sst_priority.sh
```diff
@@ -0,0 +1,106 @@
...
+echo 'CPU_FREQUENCY_STATUS core 0
+CPU_FREQUENCY_STATUS core 1
+CPU_FREQUENCY_STATUS core 2
+CPU_FREQUENCY_STATUS core 3
+CPU_FREQUENCY_STATUS core 4
+CPU_FREQUENCY_STATUS core 5
+CPU_FREQUENCY_STATUS core 6
+CPU_FREQUENCY_STATUS core 7' > ${READ_REQUEST}
+
+echo 'SST::COREPRIORITY_ENABLE:ENABLE board 0 1
```
dannosliwcd, line 66: Sorry, I omitted some steps in the runs I did earlier, since I had already configured them earlier in that session. You'll also want to set each of the weights and the frequency limits.

For buckets 0 through 3, something like:
```
SST::COREPRIORITY:<bucket>:WEIGHT board 0 <weight>
SST::COREPRIORITY:<bucket>:FREQUENCY_MIN board 0 <the starting frequency>
SST::COREPRIORITY:<bucket>:FREQUENCY_MAX board 0 <e.g. the system's max frequency>
```
Note that the buckets assigned here are defining the behavior that you map to cores with `SST::COREPRIORITY:ASSOCIATION`. That is, the `<bucket>` here is selected for a core by using that number as the input to `SST::COREPRIORITY:ASSOCIATION`.

For one configuration that would work for this test, you could set the min frequency to sticker for all except bucket 0. Maybe set bucket 0 to start somewhere inside the turbo range. Set a increasing weight as the bucket increases. You might consider bucket 0->0 weight, bucket 1->5 weight, bucket 2->10 weight, bucket 3->15 weight.

service/test/test_sst_priority.sh
```diff
@@ -0,0 +1,105 @@
...
+    --------------------------
+
+    Runs a CPU stress test binary (stress-ng) on all cores of the
+    platform.  While the all of the CPUs are loaded the GEOPM service
+    is used to configure the Intel SST feature to give high priority
+    to two cores (cores 0 and 5).  The CPU frequencies of the first 8
+    cores on the system are printed before and after configuring SST
+    core priority.
+
+    This test is a work in progress.
```
cmcantalupo: Remove this comment.

PR #1656, opened 2021-06-11
Make CI use clang-12 and ubuntu-20.04

dannosliwcd:
- Previously used clang-9 and ubuntu-latest
- ubuntu-latest is currently ubuntu-20.04, which is an LTS release
- clang-9 is not available in ubuntu-20.04, and clang-12 is the newest
  one available in that image


cmcantalupo (approved):
LGTM

bgeltz (approved):

PR #1659, opened 2021-06-14
Use new-style MOCK_METHOD calls in gmock

### LINKED ISSUE #858: As a GEOPM developer, I want to be able to detect inconsistencies between test mock classes and the library classes being tested
A way to achieve this would be to upgrade googletest to the latest version (https://github.com/google/googletest/releases/tag/release-1.10.0) which has support for marking methods with the `override` keyword.

Done criteria:
- (Already supported) adding a new virtual method in base class causes compiler error if not added in mock class.
- Removing a virtual method from the base class causes a compiler error if still present in mock.
- Changing the signature of a method in the base classes causes a compiler error if mock doesn't match.

dannosliwcd:
This set of changes updates gmock to use the new-style MOCK_METHOD calls that have been available to us since #1643. 

The second patch in the stack applies the new `MOCK_METHOD` call format, and fixes any formatting inconsistencies by running clang-format. The first patch applies `clang-format` without any other changes, with the intent of making it easier to see non-formatting differences in the second patch.

Migration was performed by the following script:
```
for migration_file in "$@"
do
        perl -0p \
                -e 's/MOCK_METHOD\d\(\s*([^,]+),\s*([^(]+)\s*([^;]*)\);/MOCK_METHOD\($2,$1,$3, (override)\);/gms; \
                    s/MOCK_CONST_METHOD\d\(\s*([^,]+),\s*([^(]+)\s*([^;]*)\);/MOCK_METHOD\($2,$1,$3, (const, override)\);/gms; \
                    s/MOCK_METHOD\(([^;]*)\(void\)([^;]*;)/MOCK_METHOD\($1\(\)$2/gms; ' \
                "$migration_file" \
                | clang-format --style=file \
                > "${migration_file}.tmp"
        mv "${migration_file}.tmp" "$migration_file"
done
```

The script does not handle all cases of gmock macro migration, but it does handle most of the ones that apply for our code. The main case it does not handle is nested types that contain commas (they need to be wrapped by a set of parentheses). The third patch manually addresses those few cases where it impacts our build.

As a result of using the new format, the build can now warn us about inconsistencies between mock code and the interface it implements. The fourth patch fixes some now-reported inconsistencies that started generating build failures. This changes fixes #858.

The last patch removes an ignore option for a warning that we no longer need to ignore since we consistently use `override` qualifiers in our mock code now.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -530,20 +530,6 @@ AC_COMPILE_IFELSE([
     AM_CONDITIONAL([GEOPM_ENABLE_MPI3], [false])
 ])
 
-[CFLAGS="$CFLAGS_SAVE"]
-[CFLAGS="$CFLAGS -Wall -Werror -Wno-inconsistent-missing-override"]
-AC_MSG_CHECKING([for support for -Wno-inconsistent-missing-override])
-AC_COMPILE_IFELSE([
-    AC_LANG_PROGRAM([[#include <stdio.h>]],
-          [[printf("Hello World.\n");]])
-],[
-    AC_MSG_RESULT(yes)
-    AM_CONDITIONAL([GEOPM_DISABLE_INCONSISTENT_OVERRIDE], [true])
-],[
-    AC_MSG_RESULT(no)
-    AM_CONDITIONAL([GEOPM_DISABLE_INCONSISTENT_OVERRIDE], [false])
-])
-
 [LDFLAGS="$LDFLAGS_SAVE"]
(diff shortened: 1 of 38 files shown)


cmcantalupo (approved):
Thanks for the work and the extensive documentation in the PR itself.  It will make modifying outstanding branches (e.g. the geopm-service branch) easier if any conflicts arise.  🤞

PR #1661, opened 2021-06-16
Modify integration configuration to explicitly use intel MPI and intel compiler

cmcantalupo:


integration/config/mcfly_env.sh
```diff
@@ -0,0 +1,52 @@
...
+export MPICC=mpiicc
+export MPICXX=mpiicpc
+export FC=ifort
+export F77=ifort
+export F90=ifort
+export MPIFORT=mpiifort
+export MPIFC=mpiifort
+export MPIF77=mpiifort
+export MPIF90=mpiifort
+export FI_PROVIDER=psm2
```
bgeltz:
```suggestion
```

Why is this necessary?

integration/config/mcfly_env.sh
```diff
@@ -0,0 +1,52 @@
...
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+# MCFLY BUILD ENVIRONMENT
+#
+# This script is intended to be sourced within an existing script or shell ONLY.
+# It is NOT intended to be ./ executed.
+
+module swap mvapich2 impi >& /dev/null || true
```
bgeltz:
```suggestion
```

I'm going to change the cluster wide defaults if this is necessary.

bgeltz (changes requested):
I thought these changes were going to go into build_env.sh.

When we wrote build_env.sh originally, it was written for the default McFly environment.

integration/config/mcfly_env.sh
cmcantalupo: Not the best solution as written anyway (no check for impi being loaded in the end)

dannosliwcd (approved):

bgeltz (approved):

PR #1683, opened 2021-06-17
Level Zero IOGroup, Device Pool, and Accelerator Topo

lhlawson:
This PR is intended to provide basic support of the [Intel Level Zero Specification v1.1.2](https://spec.oneapi.com/level-zero/1.1.2/sysman/PROG.html) for monitoring and controlling Intel GPUs.  It follows the architecture implemented for NVIDIA support provided in PR#1229 and PR#1246.  

This PR Provides:
- The ability to discover and enumerate Level Zero devices through the Sysman interface 
- Reporting of devices and a rudimentary static CPU to GPU mapping to PlatformTopo
- A device pool to control and monitor Level Zero devices
- An IO Group that uses the above to provide users with the ability to monitor and control Level Zero devices via geopmread/write and agents.
- Testing of the above features, with the exception of agent capabilities

Summary of Change: 
- Addition of LevelZeroAcceleratorTopo implementation of AcceleratorTopo.
- Addition of LevelZeroDevicePool to act as an interface to the LevelZero API.  See Known issues and future work 2 & 3.
- Addition of LevelZeroIOGroup to provide interface to the LevelZeroDevicePool.  
- Addition of LevelZeroSignal to allow leveraging existing the DerivativeSignal class.
- Modifications to IOGroup and AcceleratorTopo to include Level Zero code when needed.
- Modifications to the configure.ac to provide --enable-levelzero and --with-liblevelzero options for enabling functionality.
- Addition of Unit tests and mock object to test basic Device Pool, Accelerator Topo, and IO Group Functions.
- Addition of Integration tests to test IOGroup and Device Pool functionality on supported systems.

Known Issues & Future work:
1. Addition of signals: frequency throttle reasons, frequency tdp, frequency efficient, as describe [here](https://spec.oneapi.com/level-zero/latest/sysman/api.html#_CPPv416zes_freq_state_t)
2. Refactor of LevelZeroDevicePool to enable testability and provide an actual 'thin' wrapper.  Right now it's functional but too complex.
3. Addition of per device domain tracking to reduce requirement to iterate over number of domains in LevelZeroDevicePool

The devicepool is already slated for a refactor, so focusing on the IOGroup, signals names, and Topo is likely higher priority.  Regardless any feedback is appreciated.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -302,8 +302,8 @@ AC_ARG_WITH([liblevelzero], [AS_HELP_STRING([--with-liblevelzero=PATH],
             [specify directory for installed liblevelzero package.])])
 if test "x$with_liblevelzero" != x; then
   EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -I$with_liblevelzero/include/level_zero"
-  LD_LIBRARY_PATH="$with_liblevelzero/lib:$LD_LIBRARY_PATH"
-  EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_liblevelzero/lib64" #Chris, is this safe?
+  LD_LIBRARY_PATH="$with_liblevelzero/lib64:$LD_LIBRARY_PATH"
+  EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_liblevelzero/lib64"
 fi
 
 AC_ARG_WITH([libelf], [AS_HELP_STRING([--with-libelf=PATH],
--- a/configure.ac
+++ b/configure.ac
@@ -117,6 +117,19 @@ fi
 [enable_procfs="1"]
 )
 
+AC_ARG_ENABLE([levelzero],
+  [AS_HELP_STRING([--enable-levelzero], [Enables use of the levelzero library to support levelzero board accelerators])],
+[if test "x$enable_levelzero" = "xno" ; then
+  enable_levelzero="0"
+else
+  enable_levelzero="1"
+fi
+],
+[enable_levelzero="0"]
+)
+AC_SUBST([enable_levelzero])
+AM_CONDITIONAL([ENABLE_LEVELZERO], [test "x$enable_levelzero" = "x1"])
+
 AC_ARG_ENABLE([nvml],
   [AS_HELP_STRING([--enable-nvml], [Enables use of the NVML library to support NVML board accelerators])],
 [if test "x$enable_nvml" = "xno" ; then
@@ -166,6 +179,13 @@ if test "x$enable_msrsafe_ioctl_write" = "x1" ; then
   AC_DEFINE([GEOPM_MSRSAFE_IOCTL_WRITE], [ ], [Enables use of msr-safe ioctl feature for writing (broken as of msr-safe version 1.2.0)])
 fi
 AC_SUBST([enable_msrsafe_ioctl_write])
+
+if test "x$enable_levelzero" = "x1" ; then
+  ENABLE_LEVELZERO=True
+  AC_DEFINE([GEOPM_ENABLE_LEVELZERO], [ ], [Enables use of the levelzero library to support levelzero board accelerators])
+else
+  ENABLE_levelzero=False
+fi
 
 if test "x$enable_nvml" = "x1" ; then
   ENABLE_NVML=True
@@ -276,6 +296,14 @@ if test "x$with_libnvml" != x; then
   EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -I$with_libnvml/include"
   LD_LIBRARY_PATH="$with_libnvml/lib:$LD_LIBRARY_PATH"
   EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_libnvml/lib"
+fi
+
+AC_ARG_WITH([liblevelzero], [AS_HELP_STRING([--with-liblevelzero=PATH],
+            [specify directory for installed liblevelzero package.])])
+if test "x$with_liblevelzero" != x; then
+  EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -I$with_liblevelzero/include/level_zero"
+  LD_LIBRARY_PATH="$with_liblevelzero/lib:$LD_LIBRARY_PATH"
+  EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_liblevelzero/lib64" #Chris, is this safe?
 fi
 
 AC_ARG_WITH([libelf], [AS_HELP_STRING([--with-libelf=PATH],
@@ -481,6 +509,16 @@ if test "x$enable_nvml" = "x1" ; then
         exit -1])
 fi
 
+if test "x$enable_levelzero" = "x1" ; then
+    AC_SEARCH_LIBS([zeInit], [ze_loader])
+    AC_CHECK_LIB([ze_loader], [zeInit], [], [
+        echo "missing libze_loader: LevelZero library is required, use --with-liblevelzero to specify location"
+        exit -1])
+    AC_CHECK_HEADER([ze_api.h], [], [
+        echo "missing ze_api.h: LevelZero header is required, use --with-liblevelzero to specify location"
+        exit -1])
+fi
+
 # Checks for header files.
 AC_CHECK_HEADERS([arpa/inet.h fcntl.h float.h inttypes.h malloc.h netdb.h stddef.h stdint.h stdlib.h string.h sys/socket.h unistd.h])
 
@@ -651,4 +689,5 @@ AC_MSG_RESULT([ompt               : ${enable_ompt}])
 AC_MSG_RESULT([beta               : ${enable_beta}])
 AC_MSG_RESULT([bloat              : ${enable_bloat}])
 AC_MSG_RESULT([nvml               : ${enable_nvml}])
+AC_MSG_RESULT([levelzero          : ${enable_levelzero}])
 AC_MSG_RESULT([===============================================================================])
--- a/integration/test/util.py
+++ b/integration/test/util.py
@@ -109,6 +109,16 @@ def skip_unless_do_launch():
         return unittest.skip("Most tests in this suite require launch; do not set --skip-launch.")
     return lambda func: func
 
+def skip_unless_levelzero_power():
+    #try/catch read signal for levelzero
+    try:
+        power = geopm_test_launcher.geopmread("LEVELZERO::POWER board_accelerator 0")
+    except subprocess.CalledProcessError:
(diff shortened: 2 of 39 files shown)


configure.ac
```diff
@@ -265,6 +285,14 @@ if test "x$with_libnvml" != x; then
   EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_libnvml/lib"
 fi
 
+AC_ARG_WITH([liblevelzero], [AS_HELP_STRING([--with-liblevelzero=PATH],
+            [specify directory for installed liblevelzero package.])])
+if test "x$with_liblevelzero" != x; then
+  EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -I$with_liblevelzero/include/level_zero"
+  LD_LIBRARY_PATH="$with_liblevelzero/lib:$LD_LIBRARY_PATH"
+  EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_liblevelzero/lib64" #Chris, is this safe?
```
lhlawson: This deviates from the other --with-lib* in that is points to lib64 specifically

integration/test/util.py
```diff
@@ -109,6 +109,16 @@ def skip_unless_do_launch():
         return unittest.skip("Most tests in this suite require launch; do not set --skip-launch.")
     return lambda func: func
 
+def skip_unless_levelzero_power():
+    #try/catch read signal for levelzero
+    try:
+        power = geopm_test_launcher.geopmread("LEVELZERO::POWER board_accelerator 0")
```
lhlawson, line 115: This will result in an error on systems without LEVELZERO support or without GPUs, and catches the error to indicate we should skip the test.  
We may want to find a way to suppress the original error print to avoid confusion

src/LevelZeroAcceleratorTopo.cpp
```diff
@@ -0,0 +1,103 @@
...
+        if (m_num_accelerator == 0) {
+            std::cerr << "Warning: <geopm> LevelZeroAcceleratorTopo: No levelZero accelerators detected.\n";
+        }
+        else {
+            m_cpu_affinity_ideal.resize(m_num_accelerator);
+            unsigned int num_cpu_per_accelerator = num_cpu / m_num_accelerator;
+
+            //TODO: Add ideal cpu to accelerator affinitization that isn't a simple split.  This may come from
+            //      a call to oneAPI
+            for (unsigned int accel_idx = 0; accel_idx <  m_num_accelerator; ++accel_idx) {
```
lhlawson, line 65: No ideal affinitization appears to be available from oneAPI at this time (spec 1.1.2) from the core, tools, or sysman interface.  This TODO warrants review to see if this functionality exists elsewhere, otherwise it may be removed

src/LevelZeroAcceleratorTopo.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+            if ((num_cpu % m_num_accelerator) != 0) {
+                unsigned int accel_idx = 0;
+                for (int cpu_idx = num_cpu_per_accelerator*m_num_accelerator; cpu_idx < num_cpu; ++cpu_idx) {
+                    m_cpu_affinity_ideal.at(accel_idx%m_num_accelerator).insert(cpu_idx);
+                    ++accel_idx;
+                }
+            }
+
+            //if ((num_cpu % m_num_accelerator) != 0) {
```
lhlawson: Delete or replace with more advanced check that all CPUs have been affinitized.

src/LevelZeroDevicePool.hpp
```diff
@@ -0,0 +1,122 @@
...
+            /// @brief Number of accelerators on the platform.
+            /// @return Number of LevelZero accelerators.
+            virtual int num_accelerator(void) const = 0;
+            /// @brief Get the LevelZero device core clock rate
+            ///        in MHz.
+            /// @param [in] accel_idx The index indicating a particular
+            ///        accelerator.
+            /// @return Accelerator device core clock rate in MHz.
+            virtual double frequency_status_gpu(unsigned int accel_idx) const = 0;
+            virtual double frequency_status_mem(unsigned int accel_idx) const = 0;
```
lhlawson: Header descriptions required.  A large number of these functions will be moved into another helper class as part of the planned refactor, so bundling those changes together.

src/LevelZeroDevicePoolImp.hpp
```diff
@@ -0,0 +1,143 @@
...
+            virtual double temperature_memory(unsigned int accel_idx) const override;
+            virtual uint64_t active_time(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_compute(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_copy(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_media_decode(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_timestamp(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_timestamp_compute(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_timestamp_copy(unsigned int accel_idx) const override;
+            virtual uint64_t active_time_timestamp_media_decode(unsigned int accel_idx) const override;
+            /// @brief Get the LevelZero device power in ???.
```
lhlawson: Remove

src/LevelZeroDevicePoolImp.hpp
```diff
@@ -0,0 +1,143 @@
...
+            uint32_t m_num_mca;
+
+            std::vector<ze_driver_handle_t> m_levelzero_driver;
+            // TODO: This will need to be a map or vector of vectors if we're supporting multiple
+            //       types of sysman devices such as <BOARD GPU devices, PACKAGE GPU device, FPGA device>
+            std::vector<zes_device_handle_t> m_sysman_device;
+
+            // TODO: These will need to be something like a map of vector of vectors if we're supporting multiple
+            //       types of sysman devices such as <Temperature<device<Board GPU, Package GPU, FPGA>>>
+            std::vector<std::vector<zes_freq_handle_t> > m_freq_domain;
```
lhlawson: Consider moving tracking to a device level struct, such as:
`device_struct{
    device_handles;
    domain_handles[power,freq,engine,...]
    device_type;
    `

src/LevelZeroIOGroup.hpp
```diff
@@ -0,0 +1,122 @@
...
+            void init(void);
+
+            void register_derivative_signals(void);
+            void register_signal_alias(const std::string &alias_name, const std::string &signal_name);
+            void register_control_alias(const std::string &alias_name, const std::string &control_name);
+
+            const PlatformTopo &m_platform_topo;
+            const LevelZeroDevicePool &m_levelzero_device_pool;
+            bool m_is_batch_read;
+            std::vector<uint64_t> m_initial_power_limit;
```
lhlawson: Currently unused.  Remove until needed for save/restore.

src/LevelZeroIOGroup.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+            struct control_s
```
lhlawson, line 98: We're currently using GEOPM Signals for the signal in order to get support for the DerivativeSignal functionality.  We can move to a LevelZeroControl for controls, however there's not key functionality we're missing by not using it at this time.  

Adding a LevelZeroControl seems preferable overall to keep things consistent.

src/LevelZeroIOGroup.hpp
lhlawson: Expanded to include signals we expect to need for the restore portion of save restore.

configure.ac
```diff
@@ -265,6 +285,14 @@ if test "x$with_libnvml" != x; then
   EXTRA_LDFLAGS="$EXTRA_LDFLAGS -L$with_libnvml/lib"
 fi
 
+AC_ARG_WITH([liblevelzero], [AS_HELP_STRING([--with-liblevelzero=PATH],
+            [specify directory for installed liblevelzero package.])])
+if test "x$with_liblevelzero" != x; then
+  EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -I$with_liblevelzero/include/level_zero"
+  LD_LIBRARY_PATH="$with_liblevelzero/lib:$LD_LIBRARY_PATH"
```
cmcantalupo: Why would the path be different for -L than for LD_LIBRARY_PATH?

We should be using whatever path is "standard" when you install levelzero for both LD_LIBRARY_PATH and LDFLAGS.  In the current implementation they do not match.

integration/test/util.py
cmcantalupo: The error message should be immediately followed by a message saying that a test will be skipped.  I think this should be clear for a user, and the error message should not be suppressed.

configure.ac
lhlawson: Agree, this was a miss.  Fixed in latest.

lhlawson:
After initial feedback a refactor has been performed, focusing on the LevelZeroDevicePool.  The DevicePool has been broken into the DevicePool and shim, with the shim handling all API calls and the DevicePool being responsible for range checking & device to subdevice mapping.  

In addition to the refactor subdevice support has been added.  Due to the implementation, non-subdevice LevelZero GPUs are no longer supported.  This may impact usage of the ZE_AFFINITY_MASK environment variable.  Checking is done at run time to catch cases of devices without sub-devices.  This change was included after the refactor since it reduced Device Pool and shim complexity.  

The refactor and changes have been tested on live systems, but as part of the refactor existing tests have **NOT** been updated and as such this PR should not be merged.  If the refactor is considered appropriate the tests will be updated to mock the shim class (instead of the previous mocking of the device pool), and additional signals that were removed will be re-added.

lhlawson:
> 
> The refactor and changes have been tested on live systems, but as part of the refactor existing tests have **NOT** been updated and as such this PR should not be merged. If the refactor is considered appropriate the tests will be updated to mock the shim class (instead of the previous mocking of the device pool), and additional signals that were removed will be re-added.

The unit tests have now been updated to reflect the changes to the refactor and should cover the LevelZeroDevicePool, IOGroup, and AcceleratorTopo.

In addition the integration test has been updated to remove signals that were removed from the PR as part of the refactor (ex: power_limit_sustained).  These signals will be re-added in a later PR, and the test updated at that time.

lhlawson:
Currently the device and sub-device handles are cached at LevelZeroShim initialization time, as it was expected that these values would not change during a run.  It is, however, possible for the user to change the ZES_AFFINITY_MASK at any time, which may cause devices or sub-devices to appear and disappear.  This may be an issue for a later PR.  A simple fix is to add a basic check in the device pool on every shim call that checks if the ZES_AFFINITY_MASK has changed since instantiation.

lhlawson:
Closing PR to break up into sub PRs.  #1781 is the first of these.

PR #1723, opened 2021-06-24
Remove regex reserved characters from version in spec

bgeltz:
The __requires_exclude_from RPM macro was not functioning properly because of a non-release version string, ```geopm-1.1.0+dev1123g3558ff21```.  There is no way to escape the string (specifically the '+') in such a way that it is passed properly to the exclude macro, so I removed it entirely.

Checking the requirements from the RPM now properly omits all Python module required deps.

Before (when things were not functioning properly):
```
$ rpm -qpR geopm-1.1.0+dev1123g3558ff21-1.x86_64.rpm | grep fort
libgfortran-2e0d59d6.so.5.0.0()(64bit)
libgfortran-2e0d59d6.so.5.0.0(GFORTRAN_8)(64bit)
libmpifort.so.12()(64bit)
```

After the changes in this PR:
```
$ rpm -qpR geopm-1.1.0+dev1122g11cfee1b-1.x86_64.rpm | grep fort
libmpifort.so.12()(64bit)
```

More info:
https://docs.fedoraproject.org/en-US/packaging-guidelines/AutoProvidesAndRequiresFiltering/

### DIFF
--- a/specs/geopm-dudley.spec.in
+++ b/specs/geopm-dudley.spec.in
@@ -4,10 +4,11 @@
 %define python_family python3
 %define pip_family pip3
 %define install_prefix /opt/ohpc/pub/libs/%{pname}/%{pname}-%{version}
+%define install_prefix_short /opt/ohpc/pub/libs/%{pname}/%{pname}-
 %define module_prefix /opt/ohpc/pub/modulefiles/%{pname}
 %define module_name %{version}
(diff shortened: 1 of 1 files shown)


cmcantalupo (dismissed):
@bgeltz, if you are still working on this without a resolution, lets just go with this solution.  Sorry for being a pain.

cmcantalupo (approved):

PR #1725, opened 2021-06-25
Arithmetic Intensity benchmark command line fixes

fkeceli:
- Fixed command line arguments to arithmetic intensity benchmark and made some improvement to them.
- Made cosmetic changes to make the linter happy.

### DIFF
--- a/integration/apps/arithmetic_intensity/arithmetic_intensity.py
+++ b/integration/apps/arithmetic_intensity/arithmetic_intensity.py
@@ -34,7 +34,6 @@ AppConf class for Arithmetic Intensity benchmark.
 '''
 
 import os
-import re
 
 from apps import apps
 
@@ -42,12 +41,15 @@ from apps import apps
 def setup_run_args(parser):
     """ Add common arguments for all run scripts.
     """
-    parser.add_argument('--run-type', dest='run_type', choices=['sse', 'avx2', 'avx512'], default='sse',
+    parser.add_argument('--run-type', dest='run_type',
+        choices=['sse', 'avx2', 'avx512'], default='sse',
         help='Choose a vectorization type for the run.')
     parser.add_argument('--slowdown', type=float, default=1,
-        help='When imbalance is present, this specifies the amount of work for slow ranks to perform, as a factor of the amount of work the fast ranks perform.')
+        help='When imbalance is present, this specifies the amount of work for slow ranks' +
+             ' to perform, as a factor of the amount of work the fast ranks perform.')
     parser.add_argument('--base-internal-iterations', type=int, default=1,
-        help='How many iterations to perform in the inner loop, for the fast set of ranks (all ranks if there is no imbalance).')
+        help='How many iterations to perform in the inner loop, for the fast set of ranks ' +
+             '(all ranks if there is no imbalance).')
     parser.add_argument('--slow-ranks', type=int, default=0,
         help='The number of ranks to run with extra work for an imbalanced load.')
     parser.add_argument('--floats', type=int, default=67108864,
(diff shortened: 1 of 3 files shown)


integration/apps/arithmetic_intensity/arithmetic_intensity.py
```diff
@@ -78,17 +83,23 @@ def create_appconf(mach, args):
...
-            if type(values) == list:
+            if isinstance(values, list):
+                app_args.append(arg)
                 app_args += [str(ii) for ii in values]
+            elif isinstance(values, bool):
+                if values is True:
+                    app_args.append(arg)
             else:
+                app_args.append(arg)
                 app_args.append(str(values))
```
dannosliwcd: This problem is handled differently by geopmlaunch: https://github.com/geopm/geopm/blob/dev/scripts/geopmpy/launcher.py#L156

In that case, they only call add_argument for geopm-targeted args and not for any other wrapped args. I guess that would mean registering only --run-type and --ranks-per-node here. The rest are handled by forwarding the second return  value from parser.parse_known_args().

I think the current approach is also fine if that's preferred, particularly if you're interested in keeping all of the help text visible from the python wrapper.

dannosliwcd (dismissed):

integration/apps/arithmetic_intensity/arithmetic_intensity.py
fkeceli: You are absolutely right. The only reason I did this, is to have the help text. Since I have added the run-type argument, to enable the help text from python I had to add all the program arguments to the python wrapper. Alternatively we can remove all that and add a '--app-help' option which we can pass to the benchmark command line as -h. Would you prefer that?

fkeceli: We can even get fancier and override the argparse help to call the app help in addition to the help it provides.

fkeceli: Actually, I realized this is something we need to do. The new options you added to the benchmark break if we use the tagged version.

dannosliwcd: The new options add but don't change. Basically, they should only break in the sense that they don't exist if you have the old version, but you can still completely use all features of the old version even if you're aware of the features of the new version. The same applies whether you manually write the help text or delegate it to the installed version of the workload.

I think either approach can work (unless by your last comment, you mean something you tried to use isn't working). So I'd say go with whichever approach provides the least resistance to your experimentation workflow.

integration/apps/arithmetic_intensity/arithmetic_intensity.py
```diff
@@ -33,83 +33,82 @@
...
+                        help='Choose a vectorization type for the run (default: sse).')
     parser.add_argument('--ranks-per-node', dest='ranks_per_node',
                         action='store', type=int,
                         help='Number of physical cores to reserve for the app. ' +
                              'If not defined, all nodes but one will be reserved (leaving one ' +
                              'node for GEOPM).')
     parser.add_argument('--distribute-slow-ranks', action='store_true',
                         help='Distribute slow ranks across nodes and packages. Otherwise, slow '
-                             'ranks are assigned to fill nodes and packages.')
+                             'ranks are assigned to fill nodes and packages (default: True).')
```
dannosliwcd: Isn't the default False for store_true actions?

fkeceli: You are absolutely right. That was a copy/paste artifact :P

dannosliwcd (approved):

PR #1730, opened 2021-07-04
Changed the order that CLI arguments are passed to the launch function.

fkeceli:
GEOPM allows user to pass the same command line arguments multiple times with different values and the last value will be the effective one (due to how Python argparse works). We use this property to override the default trace/report signals defined in integration experiments from experiment command line. However the experiment command line should be added to the end of the internal command line definition that will be passed to GEOPM from inside the integration test script. This is currently the case for all the experiments except for uncore/core frequency sweeps. This PR fixes that inconsistency.


cmcantalupo (approved):
LGTM

PR #1734, opened 2021-07-07
Launch util fix

fkeceli:
Fixed a behavior in launch_util that wouldn't respond to CTRL-C and go into a loop if there is a bug in the integration script.


fkeceli:
This was accidentally merged with the arithmetic intensity integration PR. I will re-open.

PR #1735, opened 2021-07-07
Added geopmbench integration with command-line defined config files.

fkeceli:
As explained by the title.

### DIFF
--- a/integration/apps/geopmbench/geopmbench.py
+++ b/integration/apps/geopmbench/geopmbench.py
@@ -88,3 +88,39 @@ class TinyAppConf(apps.AppConf):
 
     def get_bash_exec_args(self):
         return self._bench_conf.get_path()
+
+class GeopmbenchAppConf(apps.AppConf):
+    ''' A version that user can pass config files.'''
+    @staticmethod
+    def name():
+        return 'geopmbench'
+
+    def __init__(self, bench_conf_path, ranks_per_node):
+        self._bench_conf_path = bench_conf_path
+        self._ranks_per_node = ranks_per_node
+
+    def get_rank_per_node(self):
+        return self._ranks_per_node
+
+    def get_bash_exec_path(self):
+        # TODO: may need to find local version if not installed
+        return 'geopmbench'
+
+    def get_bash_exec_args(self):
+        return self._bench_conf_path
+
+def setup_geopmbench_run_args(parser):
+    """ Add common arguments for all run scripts.
+    """
+    parser.add_argument('config_file', action='store', type=str,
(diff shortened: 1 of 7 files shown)


integration/apps/geopmbench/geopmbench.py
```diff
@@ -88,3 +88,39 @@ def get_bash_exec_path(self):
...
+        # TODO: may need to find local version if not installed
+        return 'geopmbench'
+
+    def get_bash_exec_args(self):
+        return self._bench_conf_path
+
+def setup_geopmbench_run_args(parser):
+    """ Add common arguments for all run scripts.
+    """
+    parser.add_argument('config_file', action='store', type=str,
```
cmcantalupo, line 116: I think `config_file` should be a required argument.

```
required=True
```

Otherwise get_bash_exec_args() will return None and the string "None" will be added to the bash command line.

cmcantalupo (commented):
Minor change required.

integration/apps/geopmbench/geopmbench.py
fkeceli: I believe ‘required’ is for optional arguments not for positional arguments: https://docs.python.org/3/library/argparse.html#required (edited) 

config_file does not start with ‘-’ or ‘--’ so it is a positional argument.

cmcantalupo: You are correct.

cmcantalupo (dismissed):

integration/apps/geopmbench/geopmbench.py
```diff
@@ -88,3 +88,39 @@ def get_bash_exec_path(self):
...
+                        help='Path to the geopmbench config json file.')
+    parser.add_argument('--ranks-per-node', dest='ranks_per_node',
+                        action='store', type=int,
+                        help='Number of MPI ranks physical cores to reserve for the app.')
```
bgeltz:
```suggestion
                        help='Number of MPI ranks to reserve for the app.')
```

The bit about "physical cores" doesn't make sense to me.

bgeltz (changes requested):

integration/apps/geopmbench/geopmbench.py
fkeceli: You are right that is a typo and fixed now.

integration/apps/geopmbench/geopmbench.py
fkeceli: I added a comment to the function to make it clear.

bgeltz (approved):

PR #1736, opened 2021-07-07
Launch util fix (retry)

fkeceli:
Fixed a behavior in launch_util that wouldn't respond to CTRL-C and go into a loop if there is a bug in the integration script.

### DIFF
--- a/integration/experiment/launch_util.py
+++ b/integration/experiment/launch_util.py
@@ -168,6 +168,9 @@ def launch_all_runs(targets, num_nodes, iterations, extra_cli_args, output_dir,
                     # Hit if e.g. the app calls MPI_ABORT
                     sys.stderr.write('Warning: <geopm> Execption encountered during run {}'.format(e))
                     sys.stderr.write('Retrying previous trial...')
+                    # Without sleep, the failed run tries to start over and over again and becomes
+                    # unresponsive to CTRL-C.
+                    time.sleep(5)
                 finally:
                     app_conf.trial_teardown(run_id, output_dir)
 


bgeltz (dismissed):

integration/experiment/launch_util.py
```diff
@@ -168,6 +168,9 @@ def launch_all_runs(targets, num_nodes, iterations, extra_cli_args, output_dir,
                     # Hit if e.g. the app calls MPI_ABORT
                     sys.stderr.write('Warning: <geopm> Execption encountered during run {}'.format(e))
                     sys.stderr.write('Retrying previous trial...')
+                    # Without sleep, the failed run tries to start over and over again and becomes
```
bgeltz, line 171: Why is the run failing?  Ah, I see this is about bugs in the integration script.

Why not just run a single iteration until you're sure the script works?

bgeltz (changes requested):

integration/experiment/launch_util.py
fkeceli: It is not because of iterations but retries independent of iteration counts. It seems that the purpose here was if an experiment failed it will be retried. We might also want to put a limit on how many times it is retried.

fkeceli: Because of this: https://github.com/geopm/geopm/blob/d0e4506c924832b946f20c8c376e9c0457c6af05/integration/experiment/launch_util.py#L159

Even with 1 iteration, it never reaches the completed state. I think we should also put a limit on this.

bgeltz: Ah, I see.  Thanks for the explanation.
Yeah, I wrote this for when GADGET does a bad walk of the simulation and aborts.  In that case, I wanted to keep retrying until it ran successfully.
I agree that the number of attempted retries should be capped (~10 or something).
This is fine for now.

bgeltz (approved):

PR #1738, opened 2021-07-07
Fix issue with ACNT and MCNT msr aggregation

### LINKED ISSUE #1737: ACNT, MCNT and PCNT aggregation in report files is broken
When ACNT, MCNT and PCNT are dumped in report files at both core level and package level, the package level values seem to be picking up the first value instead of sum or averages of values in a package.

cmcantalupo:
- Aggregation method was not specified in json description.
- Default is "select_first", which is not correct in this case.
- Set aggration to sum which is the type for all counters.
- Fixes #1737 from github issues.

### DIFF
--- a/src/msr_data_arch.cpp
+++ b/src/msr_data_arch.cpp
@@ -68,7 +68,7 @@ namespace geopm
                     "units":     "none",
                     "scalar":    1.0,
                     "behavior":  "monotone",
-                    "writeable": false
+                    "writeable": false,
                     "aggregation": "sum"
                 }
             }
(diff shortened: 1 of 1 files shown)


src/msr_data_arch.cpp
```diff
@@ -69,6 +69,7 @@ namespace geopm
                     "scalar":    1.0,
                     "behavior":  "monotone",
                     "writeable": false
+                    "aggregation": "sum"
```
fkeceli, line 72: Is this missing a comma at the previous line?

src/msr_data_arch.cpp
```diff
@@ -84,6 +85,7 @@ namespace geopm
...
```
fkeceli, line 88: Is this missing a comma at the previous line?

fkeceli (approved):

Issue #1702, opened 2021-06-22
Remove all Python 2 support


bgeltz:
> estimate 1

Estimate is for remaining work.

This is the "epic" issue around removing all things related to Python 2 support.  Not limited to but including:

- [x] Python 2 specifics in integration/test: e.g. exception stack clearing logic.
- [x] Cleanup sys.path.append lines in e.g. test_hint_time.py and template, geopm_context. Relates to #2055 
- [x] Python dependency on mock which was a backwards compatability wrapper for py3 mocking to use in py2.  Imports will need to be updated to: `from unittest import mock` instead of `import mock`. Resolved by #1742 
- [ ] `__future__` and `builtin` imports and the `future` package dependency. https://python-future.org/compatible_idioms.html documents idiomatic translations between 2-only, 3-only, and the cross-version ones we have here.
- [x] RPM specfile dependencies on python 2. Resolved in #1753

PR #1742, opened 2021-07-08
Fix mock import

cmcantalupo:


dannosliwcd (approved):
LGTM. Looks like this finishes one of the tasks in #1702!

PR #1743, opened 2021-07-09
Fixed PCNT aggregation method

### LINKED ISSUE #1737: ACNT, MCNT and PCNT aggregation in report files is broken

fkeceli:
Add the missing PCNT metric fix to #1738 PR.
- Aggregation method was not specified in json description.
- Default is "select_first", which is not correct in this case.
- Set aggregation to sum which is the type for all counters.
- Fixes #1737 from github issues.

### DIFF
--- a/src/msr_data_skx.cpp
+++ b/src/msr_data_skx.cpp
@@ -620,7 +620,8 @@ namespace geopm
                     "units": "none",
                     "scalar": 1.0,
                     "behavior":  "monotone",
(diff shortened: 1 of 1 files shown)


cmcantalupo (approved):
LGTM thanks

PR #1744, opened 2021-07-09
Add infrastructure for make check target in service directory

### LINKED ISSUE #1733: Add build infrastructure to run unit tests in service directory
Parent story: https://github.com/geopm/geopm/issues/1672

Working branch: https://github.com/cmcantalupo/geopm/tree/platform-service-utest

cmcantalupo:
- Fixes #1733 from github issues.


service/geopmdpy_test/geopmdpy_test.sh
```diff
@@ -0,0 +1,48 @@
...
+
+test_class=$(echo $test_name | awk -F. '{print $1}')
+service_dir=$top_dir/service
+lib_dir=$obj_dir/.libs
+export PYTHONPATH=$service_dir:$PYTHONPATH
+export LD_LIBRARY_PATH=$lib_dir:$LD_LIBRARY_PATH
+
+# PYTHON is expected to be set when running make check. However, the default
+# value may be used when invoking this script directly from the command line.
+"${PYTHON-python3}" $service_dir/geopmdpy_test/$test_class.py --verbose $test_name >& $test_dir/$test_name.log
```
bgeltz:
```suggestion
"${PYTHON:-python3}" $service_dir/geopmdpy_test/$test_class.py --verbose $test_name >& $test_dir/$test_name.log
```

If ```PYTHON``` is set to the empty string, ```""```, then ```${PYTHON-python3}``` evaluates to the empty string ```""``` and may lead to undesired behavior.  IMO this should be as I suggested here to avoid this problem.

bgeltz (changes requested):

service/geopmdpy_test/geopmdpy_test.sh
bgeltz:
```suggestion
"${PYTHON-python3}" ${service_dir}/geopmdpy_test/${test_class}.py --verbose ${test_name} >& ${test_dir}/${test_name}.log
```

Adding ```{}``` where appripriate.  This should be done everywhere.  I'm not a fan of expansions like: ```$test_class.py``` where ```test_class``` is the BASH variable.  This can/will lead to subtle expansion problems later and should be taken care of now.

bgeltz (changes requested):

service/geopmdpy_test/geopmdpy_test.sh
cmcantalupo: This is a pattern that appears in scripts/test/geopmpy_test.sh as well, so I pushed a separate patch.

scripts/test/geopmpy_test.sh
```diff
@@ -45,4 +45,4 @@ export LD_LIBRARY_PATH=$lib_dir:$LD_LIBRARY_PATH
 
 # PYTHON is expected to be set when running make check. However, the default
 # value may be used when invoking this script directly from the command line.
-"${PYTHON-python3}" $scripts_dir/test/$test_class.py --verbose $test_name >& $test_dir/$test_name.log
+"${PYTHON:-python3}" $scripts_dir/test/$test_class.py --verbose $test_name >& $test_dir/$test_name.log
```
bgeltz: Should also use ```{}``` consistently here for the BASH vars to be expanded.

bgeltz (approved):

PR #1754, opened 2021-07-15
Fix various integration test issues

bgeltz:
Uncovered after the switchover to SLES from CentOS and with the latest release of OpenHPC.

### DIFF
--- a/scripts/geopmpy/launcher.py
+++ b/scripts/geopmpy/launcher.py
@@ -1078,24 +1078,26 @@ class SrunLauncher(Launcher):
             result = ['--reservation', self.reservation]
         return result
 
-    current_governor = None
-    desired_governor = None
+    static_performance_governor_option = None
     def performance_governor_option(self):
-        governor_file = '/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor'
-        if SrunLauncher.current_governor is None:
+        if SrunLauncher.static_performance_governor_option is None:
+            governor_file = '/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor'
             pid = subprocess.Popen(['srun', 'cat', '{}'.format(governor_file)], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             output, err = pid.communicate()
-            SrunLauncher.current_governor = output.decode().strip('\n')
-
-        if SrunLauncher.current_governor is not None and \
-           SrunLauncher.desired_governor is None:
-            if SrunLauncher.current_governor not in ['performance', 'userspace']:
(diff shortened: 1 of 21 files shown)


scripts/geopmpy/launcher.py
```diff
@@ -1078,8 +1078,20 @@ def reservation_option(self):
             result = ['--reservation', self.reservation]
         return result
 
+    do_warning_once = True
     def performance_governor_option(self):
-        return ['--cpu-freq=Performance']
+        result = []
+        governor_file = '/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor'
+        with open(governor_file) as f:
```
cmcantalupo: This is opening the governor file on the node where geopmlaunch is being executed.  I think we need to do something to run on the compute node, like when we create the topo, a launcher is used to do that.  Could be more of a circular dependency in this case, since you need the use launcher to figure out how the launcher should be configured.

bgeltz: doh

scripts/geopmpy/launcher.py
```diff
@@ -1078,8 +1078,24 @@ def reservation_option(self):
...
     def performance_governor_option(self):
-        return ['--cpu-freq=Performance']
+        governor_file = '/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor'
+        if SrunLauncher.current_governor is None:
+            pid = subprocess.Popen(['srun', 'cat', '{}'.format(governor_file)], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            output, err = pid.communicate()
+            SrunLauncher.current_governor = output.decode().strip('\n')
+
+        if SrunLauncher.current_governor is not None and \
+           SrunLauncher.desired_governor is None:
```
cmcantalupo:
```
        if (SrunLauncher.current_governor is not None and
            SrunLauncher.desired_governor is None):
```
?

I think the lines above ensure that SrunLauncher.current_governor will be a string (and therefore not None) right?

bgeltz:
I'm cherry-picking this into the nightlies until it is merged.

scripts/geopmpy/launcher.py
```diff
@@ -1078,8 +1078,28 @@ def reservation_option(self):
             result = ['--reservation', self.reservation]
         return result
 
+    _static_performance_governor_option = None
     def performance_governor_option(self):
-        return ['--cpu-freq=Performance']
+        if SrunLauncher._static_performance_governor_option is None:
+            governor_file = '/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor'
+            pid = subprocess.Popen(['srun', 'cat', '{}'.format(governor_file)], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
cmcantalupo, line 1085: For init_topo() we had been careful to always carry over the user provided options:

```
                                  host_file=self.host_file,
                                  node_list=self.node_list,
                                  exclude_list=self.exclude_list,
                                  reservation=self.reservation)
```
Hopefully, the specific nodes we land on are not as important for the governor setting.

cmcantalupo (approved):
See comment.  I think it is GTG

PR #1757, opened 2021-07-16
Add four dgemm executions to big_o() method to do warmup

cmcantalupo:


fkeceli:
I think it would be better to have a configurable amount to epochs. I have seen that the time to get to a stable place depends on the big-o. Also we can generalize this to the other types as well.

fkeceli (approved):
As a result of discussion on 7/19, we decided it is not probably not necessary to have this a configurable value and it is beyond the use cases of geopmbench.

PR #1761, opened 2021-07-21
Address missing options for running commands on computes

bgeltz:
- Refactor the code to remove repeated sections.

### DIFF
--- a/scripts/geopmpy/launcher.py
+++ b/scripts/geopmpy/launcher.py
@@ -571,6 +571,29 @@ class Launcher(object):
         if is_geopmctl and geopm_pid.returncode:
             raise subprocess.CalledProcessError(geopm_pid.returncode, geopm_argv)
 
+    def run_compute_cmd(self, argv, num_node=None):
+        """
+        Run a command on the compute nodes.
+        """
+        if num_node is None:
+            num_node = self.num_node # Run on all the nodes in the job by default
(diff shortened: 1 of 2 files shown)


cmcantalupo (dismissed):
LGTM.  Be nice to use run_compute_cmd to run geopmread and geopmwrite rather than:

https://github.com/geopm/geopm/blob/95f1b9959eefe2000dbba6576c0d254c60f00c5f/integration/experiment/util.py#L96-L132

cmcantalupo (commented):
Looks like some unit test clean up is required:

https://github.com/geopm/geopm/pull/1761/checks?check_run_id=3127380108

bgeltz:
> LGTM. Be nice to use run_compute_cmd to run geopmread and geopmwrite rather than:
> 
> https://github.com/geopm/geopm/blob/95f1b9959eefe2000dbba6576c0d254c60f00c5f/integration/experiment/util.py#L96-L132

Yeah, I'm still looking, but there's no easy way to do this.

```run_compute_cmd()``` requires a launcher to be created before it can be called.  Switching those util functions for read/write around to use ```run_compute_cmd()``` is still going to involve some hoop jumping like what's in ```allocation_node_test()``` to create the right launcher (SLURM vs. APRUN...) via injecting things into the argv that's passed along.

The only line I see changing after all of that is the ```launcher.run(...)``` line (changing to ```run_compute_cmd()```) and that doesn't really buy us anything.

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #1762, opened 2021-07-22
Incremental update to service admin doc

cmcantalupo:
- Remove redundant "Example:" from headers


service/docs/geninfo.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
bgeltz, line 1: Missing license header

bgeltz: Also, should this source ~/.geopmrc?

There are assumptions baked in below that ```geopmread|write``` are in your ```PATH```.

bgeltz (changes requested):
LGTM beyond the comments.

service/docs/geninfo.sh
cmcantalupo: I'm not thrilled that this script depends on files outside of the service directory.  I had worked on trying to replicate the `geopmread --info` and `geopmwrite --info` output with `geopmaccess`, but some information printed by the other command line tools has not been exposed through DBus.  In the future we should try to enable this.  I will open an issue.

Since this is for developers generating documentation rather than for end users, I figure we can let it ride for now.

cmcantalupo: https://github.com/geopm/geopm/issues/1764

bgeltz (approved):

PR #1763, opened 2021-07-22
Remove an extra underscore and fix an off-by-one issue with QM_EVTSEL

dannosliwcd:


cmcantalupo (dismissed):
LGTM

lhlawson:
The _ fix looks good, great catch!  

On the +1 addition, there are a few additional issues under all this, which I called out partially in this [comment](https://github.com/geopm/geopm/pull/1646#discussion_r645820645) but didn't fully explore.

The general form of how we use the RMID value is supposed to be ceiling(log2( CPUID value + 1)), as discussed here:
image
and here
image

Currently we are calculating it incorrectly as:
`.rmid_bit_width = (uint32_t)ceil(log2((double)max) + 1)-1,`
which based on the SDM examples above should instead be (moved the parentheses and dropped the -1):
`.rmid_bit_width = (uint32_t)ceil(log2((double)max+1)),` 

RMID Bit width described in the SDM for Haswell, Broadwell, and Skylake is 10 bits.  The current msr_data_skx.cpp is incorrect in the end bit for both QM_EVTSEL.RMID and PQR_ASSOC.RMID.  
Haswell: 
image
Broadwell:
image
Skylake:
image

For the Skylake systems I have available we get values such as 159.  
SDM way: ceil(log2(159+1)) --> 8
Assuming SDM meant +1 outside the log2 call: ceil(log2(159)+1) --> 9

Neither of which add up to the 10 bit width described later in the SDM for these parts.  Until we have clarity on this issue we may want to consider disabling the MSR bit width vs CPUID value check.  While this is not ideal, there is still some protection in the form of a general protection fault as described here::
image

bgeltz:
> Until we have clarity on this issue we may want to consider disabling the MSR bit width vs CPUID value check.

SGTM.  We'll have to choose one as being "correct" while we chase down the other.
I think that change should either be in a new PR with the extra underscore fix, or rework this PR to just disable the check for now and not modify the off-by-one maths.

dannosliwcd:
Some more context from the instruction reference's side:
https://software.intel.com/content/www/us/en/develop/download/intel-architecture-instruction-set-extensions-programming-reference.html

image

So maybe we could less-strictly check that the MSRs are _at least_ that wide

lhlawson (dismissed):
After joint fixups addressing comments made this LGTM.

bgeltz (dismissed):

bgeltz (dismissed):

bgeltz (approved):

PR #1768, opened 2021-07-29
Align dudley patch to new Python requirements

bgeltz:
Needed because the mock module was removed from the geopmpy requirements.txt.


lhlawson (approved):
Looks good.  I have not performed an independent system test with it.

PR #1776, opened 2021-08-07
Refactor geopmdpy/session.py and add doc strings

### LINKED ISSUE #1711: Add doc strings to geopmdpy/session.py
> estimate 1

Parent issue: #1663

cmcantalupo:
- Fixes #1711 from github issues.

### DIFF
--- a/service/docs/source/reference.rst
+++ b/service/docs/source/reference.rst
@@ -2,6 +2,64 @@
 Reference Manual
 ================
 
+geopmaccess
+-----------
+
+Access management for the geopm service.
+
+Command line tool for reading and writing the access management lists
+for the geopm service signals and controls.
+
+
+geopmsession
+------------
+
+Command line interface for the geopm service read/write features.
+
+This command can be used to read signals and write controls by opening
+a session with the geopm service.  The user specifies which signals to
+read or which controls to write with standard input. When no command
+line options are provided, a single read of the signals requested
+though standard input is made and the results are printed to the
+screen.
+
+
+The input to the command line tool has one request per line.  A
+request for reading is made of up three strings separated by white
+space.  The first string is the signal name, the second string is the
+domain name, and the third string is the domain index.  An example
+where the entire THERM_STATUS model specific register is read from
+core zero:
+
+.. code-block:: bash
+
+    $ geopmsession
+    MSR::THERM_STATUS# core 0
+
+    0x0000000088430800
+
+
+A request for writing is made up of four strings separated by white
+space.  The first string is the control name, the second string is the
+domain name, and the third string is the domain index.  An example
+where the frequency limit of the PERF_CTL model specific register is
+set to 2.0 GHz for 10 seconds:
+
+
+.. code-block:: bash
+
+    $ geopmsession -w -t 10
+    MSR::PERF_CTL:FREQ core 0 2.0e9
+
+
+The time specified is the duration after the write has completed until
+the process of restoring all written values begins.  The total run
+time for the command will be longer than the time specified because of
+the overhead time spent reading and writing all controls for the
+save/restore feature of the service in addition to the time spent
+writing the requested controls.
+
+
 geopmdpy.service
 ----------------
 
--- a/service/geopmdpy/session.py
+++ b/service/geopmdpy/session.py
@@ -270,9 +270,10 @@ class TimedLoop:
 
     Example:
 
-        >>> time_zero = time.time()
-        >>> for loop_idx in TimedLoop(10, 0.1):
-        ...     print('{}: {}'.format(loop_idx, time.time() - time_zero))
+        >>> from time import time
+        >>> time_0 = time()
+        >>> for index in TimedLoop(10, 0.1):
+        ...     print(f'{index}: {time() - time_0}')
         ...
         0: 0.0008680820465087891
         1: 0.10126090049743652
--- a/service/geopmdpy/session.py
+++ b/service/geopmdpy/session.py
@@ -150,7 +150,7 @@ class Session:
             period (float): The user specified period between samples
                             in units of seconds.
 
-            out_stream (file): Object with .write() method where output
+            out_stream (file): Object with write() method where output
                                will be printed (typically sys.stdout).
 
         """
@@ -190,7 +190,8 @@ class Session:
         """Check that the run time and period are valid for a read session
 
         Args:
-            run_time (float): Time duration of the session in seconds.
+            run_time (float): Time duration of the session open in
+                              seconds.
 
             period (float): The user specified period between samples
                             in units of seconds.
@@ -563,6 +564,7 @@ class WriteRequestQueue(RequestQueue):
 
 def main():
     """Command line interface for the geopm service read/write features.
+
     This command can be used to read signals and write controls by
     opening a session with the geopm service.
 
--- a/service/geopmdpy/session.py
+++ b/service/geopmdpy/session.py
@@ -319,7 +319,6 @@ class TimedLoop:
         Iteration method for timed loop.  This iterator can be used in
         a for statement to execute the loop periodically.
 
-
         """
         self._loop_idx = 0
         self._target_time = time.time()
@@ -373,7 +372,7 @@ class Requests:
         return [rr[0] for rr in self._requests]
 
     def iterate_stream(self, request_stream):
-        """Iterated over a stream of requests
+        """Iterate over a stream of requests
 
         This is a generator function that will filter out comment
         lines and trailing white space from the input stream.  It can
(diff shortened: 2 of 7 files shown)


service/geopmdpy/session.py
```diff
@@ -32,90 +32,237 @@
...
+        for line in self.iterate_stream(request_stream):
+            words = line.split()
+            if len(words) != 3:
+                raise RuntimeError('Invalid command for reading: "{}"'.format(line))
+            signal_name = words[0]
+            domain_type = topo.domain_type(words[1])
+            domain_idx = int(words[2])
+            self._requests.append((signal_name, domain_type, domain_idx))
+
+    def init_signal_format(self):
```
bgeltz: Some docstrings would help explain what's going on here.

Why is the 0th request (of potentially multiple?) being paried with the 4th value back from SignalInfo?  I assume this is to extract the format function?

What if there are multiple requests?

bgeltz (changes requested):
Mostly LGTM.  1 question.

service/geopmdpy/session.py
cmcantalupo: each request is a tuple (signal_name, domain_type, domain_idx).  Later we use the signal_name (the 0th element of the request) to get back an info tuple.  Element 4 of the info tuple is the format enum.

service/src/PlatformIO.cpp
```diff
@@ -967,4 +967,28 @@ extern "C" {
         return err;
 #endif
     }
+
+    int geopm_pio_format_signal(double signal,
+                                int format_type,
+                                size_t result_max,
+                                char *result)
```
bgeltz, line 974: result needs to be checked for NULL.

https://en.cppreference.com/w/c/string/byte/strncpy
The behavior is undefined ... if either dest or src is not a pointer to a character array (including if dest or src is a null pointer)

service/src/PlatformIO.cpp
```diff
@@ -967,4 +967,28 @@ extern "C" {
...
+    {
+        int err = 0;
+        try {
+            auto format_func = geopm::string_format_type_to_function(format_type);
+            std::string result_cxx = format_func(signal);
+            result[result_max - 1] = '\0';
+            strncpy(result, result_cxx.c_str(), result_max);
```
bgeltz, line 981: We need to open an issue about changing all strncpy's to strncpy_s's.

service/geopmdpy/session.py
```diff
@@ -31,91 +31,563 @@
...
+        Raises:
+            RuntimeError: The geopm systemd service is not running
+
+        """
+        try:
+            geopm_proxy.PlatformOpenSession
+        except DBusError as ee:
+            if 'io.github.geopm was not provided' in str(ee):
+                err_msg = """The geopm systemd service is not enabled.
+    Install geopm service and run 'systemctl start geopm'"""
```
bgeltz, line 79:
```suggestion
                err_msg = """The geopm systemd service is not enabled.
                             Install geopm service and run 'systemctl start geopm'"""
```

service/geopmdpy/session.py
```diff
@@ -31,91 +31,563 @@
...
+            run_time (float): Time duration of the session in seconds.
+
+            period (float): The user specified period between samples
+                            in units of seconds.
+
+        Raises:
+            RuntimeError: The period is non-zero or total time is negative.
+
+        """
+        if run_time <= 0.0:
```
bgeltz, line 222: I'm wondering if we ought to have some sort of private member, call it self._min_write_time or similar, that's the realistic minimum time for which a control can be applied.

This is going to be IO group control latency dependent, DBus latency dependent, etc.  I know it's greater than 0, but beyond that I'm not sure how to put a bound on it.

There should be docu that describes all these latencies and how they relate to this minimum run_time.

service/geopmdpy/session.py
```diff
@@ -31,91 +31,563 @@
...
+        #     total_time == num_loop * period
+        # because we do not delay the start iteration
+        self._num_loop = num_period + 1
+
+    def __iter__(self):
+        """Set up a timed loop
+
+        Iteration method for timed loop.  This iterator can be used in
+        a for statement to execute the loop periodically.
+
```
bgeltz, line 322:
```suggestion
```

service/geopmdpy/session.py
```diff
@@ -31,91 +31,563 @@
...
+        Returns:
+            list(string): The name of the signal or control associated
+                           with each user request.
+
+        """
+        return [rr[0] for rr in self._requests]
+
+    def iterate_stream(self, request_stream):
+        """Iterated over a stream of requests
```
bgeltz:
```suggestion
        """Iterate over a stream of requests
```

service/geopmdpy/session.py
```diff
@@ -31,91 +31,563 @@
...
+        """
+        for line in request_stream:
+            line = line.strip()
+            if line == '':
+                break
+            if not line.startswith('#'):
+                yield line
+
+
+class ReadRequests(Requests):
```
bgeltz: We're doing plurals now?

service/geopmdpy/session.py
```diff
@@ -31,91 +31,563 @@
...
+
+        """
+
+        try:
+            result = [info[4] for info in
+                      geopm_proxy.PlatformGetSignalInfo(signal_names)]
+        except DBusError as ee:
+            if 'io.github.geopm was not provided' in str(ee):
+                err_msg = """The geopm systemd service is not enabled.
+    Install geopm service and run 'systemctl start geopm'"""
```
bgeltz, line 488:
```suggestion
                err_msg = """The geopm systemd service is not enabled.
                             Install geopm service and run 'systemctl start geopm'"""
```

bgeltz (changes requested):
Looks pretty good overall.

service/geopmdpy/session.py
cmcantalupo: The main purpose is to iterate over the object, so plural seemed appropriate.  We could call it a ReadRequestQueue?

service/geopmdpy/session.py
cmcantalupo: We don't want that much white space in the printed message, I thought a 4 space indent looks okay.

service/geopmdpy/session.py
cmcantalupo: We don't want that much white space in the printed message, I thought a 4 space indent looks okay.

service/geopmdpy/session.py
cmcantalupo: I'm not sure what someone would be trying to achieve with a small, non-zero value, but it seems like user error if they expect that using a command line tool written in python is going to respond in microseconds.  

If the run_time is very small, we will set the controls, sleep for that period of time and then revert the values.  The latencies you are talking about are captured in the calls to geopm, not in the call to sleep.  We can add some documentation about how the run_time is the duration of time after all controls have been set until just before the process of reverting the values will begin.

service/geopmdpy/session.py
bgeltz: I really don't like the indentation.  Do we have this pattern any where else?

What about this instead?
```suggestion
                err_msg = "The geopm systemd service is not enabled. " \
                          "Install geopm service and run 'systemctl start geopm'"
```

service/geopmdpy/session.py
bgeltz: > We can add some documentation about how the run_time is the duration of time after all controls have been set until just before the process of reverting the values will begin.

This would clear it all up.  I'm just trying to avoid a situation where the time spent to apply the control is significantly greater than the sleep time, and the user being surprised they're not getting a period they ask for.

service/geopmdpy/session.py
bgeltz: Fine w/either.  I just want to be consistent with all our code and I know we've been back and forth on having plural names in the codebase in the past.

service/geopmdpy/session.py
bgeltz: What about this instead?
```suggestion
                err_msg = "The geopm systemd service is not enabled. " \
                          "Install geopm service and run 'systemctl start geopm'"
```

bgeltz (changes requested):

service/docs/source/reference.rst
```diff
@@ -2,6 +2,64 @@
...
+set to 2.0 GHz for 10 seconds:
+
+
+.. code-block:: bash
+
+    $ geopmsession -w -t 10
+    MSR::PERF_CTL:FREQ core 0 2.0e9
+
+
+The time specified is the duration after the write has completed until
```
cmcantalupo, line 55: @bgeltz Doc for latency is here.

service/src/PlatformIO.cpp
cmcantalupo: https://github.com/geopm/geopm/issues/1778

bgeltz (approved):

PR #1781, opened 2021-08-13
Addition of levelzero device pool and topo support.

lhlawson:
The first of three PRs spun out of #1683 to ease of review and integration.  A second PR is planned to provide IOGroup Support, and a third to provide additional signals (added to the IOGroup, Device Pool, and Shim).

This PR is intended to provide basic support of the Intel Level Zero Specification v1.1.2 for monitoring and controlling Intel GPUs. It follows the architecture implemented for NVIDIA support provided in PR#1229 and PR#1246.

This PR Provides:
- The ability to discover and enumerate Level Zero devices through the Sysman interface
- Reporting of devices and a rudimentary static CPU to GPU mapping to PlatformTopo
- A device pool to control and monitor Level Zero devices
- Testing of the above features

Summary of Change:

- Addition of LevelZeroAcceleratorTopo implementation of AcceleratorTopo.
- Addition of LevelZeroDevicePool and LevelZeroShim to act as an interface to the LevelZero API. 
- Addition of LevelZeroSignal to allow leveraging existing the DerivativeSignal class.
- Modifications to AcceleratorTopo to include Level Zero code when needed.
- Modifications to the configure.ac to provide --enable-levelzero and --with-liblevelzero options for enabling functionality.
- Addition of Unit tests and mock object to test basic Device Pool and Accelerator TopoFunctions.
- Modifications to Platform Topo to support Accelerator subdevices
- Modifications to base accelerator topos and NVML accelerator topos to support subdevices

PR #1787, opened 2021-08-18
Addition of levelzero IO Group and integration tests

lhlawson:
A follow on PR to #1781 based on a breakdown of #1683.  

This PR Provides:
- An IO Group that uses the above to provide users with the ability to monitor and control Level Zero devices via geopmread/write and agents.
- Testing of the above features

Summary of Change:
- Addition of LevelZeroIOGroup to provide interface to the LevelZeroDevicePool.
- Addition of LevelZeroSignal to allow leveraging existing the DerivativeSignal class.
- Modifications to IOGroup to include Level Zero code when needed.
- Addition of Unit tests and mock object to test basic IO Group Functions.
- Addition of Integration tests to test IOGroup and Device Pool functionality on supported systems.

Known Issues & Future Work:
- Currently signals required a timestamp (energy, active time and its variants) first call for the value, then the timestamp.  This may lead to inaccurate measurements over time.  New _pair function variants have been provided by the DevicePool to address this issue, but the Signal updates have not been completed.
- Addition of signals: frequency throttle reasons, frequency tdp, frequency efficient, etc

### DIFF
--- a/src/LevelZeroDevicePool.cpp
+++ b/src/LevelZeroDevicePool.cpp
@@ -175,7 +175,7 @@ namespace geopm
     {
         if (domain != GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
             throw Exception("LevelZeroDevicePool::" + std::string(__func__) +
-                             ": domain " + std::to_string(domain) +
+                            ": domain " + std::to_string(domain) +
                             " is not supported for the frequency domain.",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
--- a/src/LevelZeroDevicePool.cpp
+++ b/src/LevelZeroDevicePool.cpp
@@ -84,7 +84,8 @@ namespace geopm
     {
         if (size == 0) {
             throw Exception("LevelZeroDevicePool::" + std::string(func) +
-                            ":Not supported on this hardware",
+                            ": Not supported on this hardware for the specified "
+                            "LevelZero domain",
                             GEOPM_ERROR_INVALID, __FILE__, line);
         }
     }
@@ -166,6 +167,26 @@ namespace geopm
 
         return m_levelzero.frequency_max(dev_subdev_idx_pair.first, l0_domain,
                                          dev_subdev_idx_pair.second);
+    }
+
+    std::pair<double, double> LevelZeroDevicePoolImp::frequency_range(int domain,
+                                                                      unsigned int domain_idx,
+                                                                      int l0_domain) const
+    {
+        if (domain != GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
+            throw Exception("LevelZeroDevicePool::" + std::string(__func__) +
+                             ": domain " + std::to_string(domain) +
+                            " is not supported for the frequency domain.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        std::pair<unsigned int, unsigned int> dev_subdev_idx_pair;
+        dev_subdev_idx_pair = subdevice_device_conversion(domain_idx);
+        check_domain_exists(m_levelzero.frequency_domain_count(
+                                        dev_subdev_idx_pair.first, l0_domain), __func__, __LINE__);
+
+        return m_levelzero.frequency_range(dev_subdev_idx_pair.first, l0_domain,
+                                           dev_subdev_idx_pair.second);
+
     }
 
     std::pair<uint64_t, uint64_t> LevelZeroDevicePoolImp::active_time_pair(int domain,
@@ -315,7 +336,8 @@ namespace geopm
     }
 
     void LevelZeroDevicePoolImp::frequency_control(int domain, unsigned int domain_idx,
-                                                   int l0_domain, double setting) const
+                                                   int l0_domain, double range_min,
+                                                   double range_max) const
     {
         if (domain != GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
             throw Exception("LevelZeroDevicePool::" + std::string(__func__) +
@@ -330,6 +352,7 @@ namespace geopm
                                         __func__, __LINE__);
 
         m_levelzero.frequency_control(dev_subdev_idx_pair.first, l0_domain,
-                                      dev_subdev_idx_pair.second, setting);
+                                      dev_subdev_idx_pair.second, range_min,
+                                      range_max);
     }
 }
--- a/src/LevelZeroIOGroup.cpp
+++ b/src/LevelZeroIOGroup.cpp
@@ -40,6 +40,7 @@
 #include <sched.h>
 #include <errno.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "IOGroup.hpp"
 #include "Signal.hpp"
@@ -765,7 +766,7 @@ namespace geopm
             control_name == "FREQUENCY_ACCELERATOR_CONTROL") {
             m_levelzero_device_pool.frequency_control(domain_type, domain_idx,
                                                       geopm::LevelZero::M_DOMAIN_COMPUTE,
-                                                      setting / 1e6);
+                                                      setting / 1e6, setting / 1e6);
         }
         else {
     #ifdef GEOPM_DEBUG
@@ -781,9 +782,22 @@ namespace geopm
     void LevelZeroIOGroup::save_control(void)
     {
         for (int domain_idx = 0;
-             domain_idx < m_platform_topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR);
+             domain_idx < m_platform_topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP);
              ++domain_idx) {
-            // TODO: Read ALL LEVELZERO Power Limits, frequency settings, etc
+
+            try {
+                // Currently only the levelzero compute domain control is supported.
+                // As new controls are added they should be included
+                m_frequency_range.push_back(m_levelzero_device_pool.frequency_range(
+                                            GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP, domain_idx,
+                                            geopm::LevelZero::M_DOMAIN_COMPUTE));
+            }
+            catch (const geopm::Exception &ex) {
+                throw Exception("LevelZeroIOGroup::" + std::string(__func__) + ": "
+                                + " Failed to fetch frequency control range for "
+                                " BOARD_ACCELERATOR_CHIP domain " + std::to_string(domain_idx),
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
         }
     }
 
@@ -791,10 +805,27 @@ namespace geopm
     // platform settings
     void LevelZeroIOGroup::restore_control(void)
     {
-        /// @todo: Usage of the LEVELZERO API for setting frequency, power, etc
-        ///        requires root privileges.  As such several unit tests will
-        ///        fail when calling restore_control.  Once a non-privileged
-        ///        solution is available this may be used
+        for (int domain_idx = 0;
+             domain_idx < m_platform_topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP);
+             ++domain_idx) {
+            try {
+                // Currently only the levelzero compute domain control is supported.
+                // As new controls are added they should be included
+                m_levelzero_device_pool.frequency_control(GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP,
+                                        domain_idx, geopm::LevelZero::M_DOMAIN_COMPUTE,
+                                        m_frequency_range.at(domain_idx).first,
+                                        m_frequency_range.at(domain_idx).second);
+            }
+            catch (const geopm::Exception &ex) {
+#ifdef GEOPM_DEBUG
+                std::cerr << "Warning: <geopm> LevelZeroIOGroup: Failed to "
+                             "restore frequency control settings for "
+                             "BOARD_ACCELERATOR_CHIP domain " << std::to_string(domain_idx)
+                             << ".  Exception: " << ex.what()
+                             << std::endl;
+#endif
+            }
+        }
     }
 
     // Hint to Agent about how to aggregate signals from this IOGroup
--- a/src/LevelZeroIOGroup.cpp
+++ b/src/LevelZeroIOGroup.cpp
@@ -453,7 +453,6 @@ namespace geopm
                                                    };
             }
         }
-
     }
 
     // Extract the set of all signal names from the index map
@@ -893,7 +892,7 @@ namespace geopm
     }
 
     void LevelZeroIOGroup::register_signal_alias(const std::string &alias_name,
-                                            const std::string &signal_name)
+                                                 const std::string &signal_name)
     {
         if (m_signal_available.find(alias_name) != m_signal_available.end()) {
             throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
@@ -909,12 +908,12 @@ namespace geopm
         // copy signal info but append to description
         m_signal_available[alias_name] = it->second;
         m_signal_available[alias_name].m_description =
-            m_signal_available[signal_name].m_description + '\n' +
-                                            "    alias_for: " + signal_name;
+            m_signal_available[signal_name].m_description +
+            "\n    alias_for: " + signal_name;
     }
 
     void LevelZeroIOGroup::register_control_alias(const std::string &alias_name,
-                                           const std::string &control_name)
+                                                  const std::string &control_name)
     {
         if (m_control_available.find(alias_name) != m_control_available.end()) {
             throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
@@ -930,7 +929,7 @@ namespace geopm
         // copy control info but append to description
         m_control_available[alias_name] = it->second;
         m_control_available[alias_name].m_description =
-       m_control_available[control_name].m_description + '\n' +
-                                         "    alias_for: " + control_name;
+            m_control_available[control_name].m_description +
+            "\n    alias_for: " + control_name;
     }
 }
--- a/src/LevelZeroIOGroup.cpp
+++ b/src/LevelZeroIOGroup.cpp
@@ -641,8 +641,7 @@ namespace geopm
     {
         m_is_batch_read = true;
         for (size_t ii = 0; ii < m_signal_pushed.size(); ++ii) {
-            std::dynamic_pointer_cast<LevelZeroSignal>(m_signal_pushed[ii])->
-                set_sample(m_signal_pushed[ii]->read());
+            m_signal_pushed[ii]->set_sample(m_signal_pushed[ii]->read());
         }
     }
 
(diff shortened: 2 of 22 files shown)


src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,862 @@
...
+                                  {},
+                                  [this](unsigned int domain_idx) -> double
+                                  {
+                                      return this->m_levelzero_device_pool.energy_timestamp(
+                                                   GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                                   domain_idx,
+                                                   geopm::LevelZero::M_DOMAIN_ALL);
+                                  },
+                                  1 / 1e6
+                                  }},
```
bgeltz: These are going to be collapsed eventually.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,862 @@
...
+            // If not pushed, add to pushed control
+            result = m_control_pushed.size();
+            m_control_pushed.push_back(control);
+        }
+
+        return result;
+    }
+
+    // Parse and update saved values for signals
+    void LevelZeroIOGroup::read_batch(void)
```
bgeltz, line 639: Do the reading and storing here.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,862 @@
...
+        }
+        //Not strictly necessary, but keeping to enforce general
+        //flow of read_batch followed by sample.
+        if (!m_is_batch_read) {
+            throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
+                            ": signal has not been read.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        return m_signal_pushed[batch_idx]->sample();
```
bgeltz, line 679: This should return the cached value.  The value should only change after a call to read_batch().

src/LevelZeroSignal.hpp
```diff
@@ -0,0 +1,69 @@
...
+    class LevelZeroSignal : public Signal
+    {
+        public:
+            virtual ~LevelZeroSignal() = default;
+            LevelZeroSignal(std::function<double (unsigned int)> devpool_func,
+                         unsigned int accelerator, double scalar);
+            LevelZeroSignal(const LevelZeroSignal &other) = delete;
+            void setup_batch(void) override;
+            double sample(void) override;
+            double read(void) const override;
```
bgeltz, line 55: new method here:
read_batch_element

bgeltz (changes requested):

src/LevelZeroIOGroup.cpp
lhlawson: read & store handled in the LevelZeroSignal read_batch_element function, however doing so required a dynamic_pointer_cast from Signal to LevelZeroSignal here, which doesn't seem ideal.

src/LevelZeroIOGroup.cpp
lhlawson: These are not collapsed, but are intrinsically linked with TIMESTAMP being a cached value based upon the ENERGY read.  This is now reflected in the signal description.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+                                  [this](unsigned int domain_idx) -> double
+                                  {
+                                      return this->m_levelzero_device_pool.energy(
+                                                   GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                                   domain_idx,
+                                                   geopm::LevelZero::M_DOMAIN_ALL);
+                                  },
+                                  1 / 1e6
+                                  }},
+                              {"LEVELZERO::ENERGY_TIMESTAMP", {
```
lhlawson, line 136: There are currently 4 timestamp signals.
ENERGY_TIMESTAMP
ACTIVE_TIME_TIMESTAMP
ACTIVE_TIME_COMPUTE_TIMESTAMP
ACTIVE_TIME_COPY_TIMESTAMP

If preferred these can be changed to the form of TIMESTAMP_, but this will requires updates to code in the push_signal function.

cmcantalupo: I think the order here is fine since the timestamp is a supporting qualifier to the primary subject which comes first?

src/LevelZeroIOGroup.hpp
```diff
@@ -0,0 +1,133 @@
...
+            void init(void);
+
+            void register_derivative_signals(void);
+            void register_signal_alias(const std::string &alias_name,
+                                       const std::string &signal_name);
+            void register_control_alias(const std::string &alias_name,
+                                        const std::string &control_name);
+
+            struct control_s
+            {
```
cmcantalupo:
```suggestion
            struct control_s {
```

src/LevelZeroIOGroup.hpp
```diff
@@ -0,0 +1,133 @@
...
+                std::function<std::string(double)> m_format_function;
+            };
+
+            const PlatformTopo &m_platform_topo;
+            const LevelZeroDevicePool &m_levelzero_device_pool;
+            bool m_is_batch_read;
+
+            std::map<std::string, signal_info> m_signal_available;
+            std::map<std::string, control_info> m_control_available;
+            std::vector<std::shared_ptr<Signal> > m_signal_pushed;
```
cmcantalupo, line 127: can this be a `std::vector<std::shared_ptr<LevelZeroSignal> >` ?

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+
+        return result;
+    }
+
+    // Parse and update saved values for signals
+    void LevelZeroIOGroup::read_batch(void)
+    {
+        m_is_batch_read = true;
+        for (size_t ii = 0; ii < m_signal_pushed.size(); ++ii) {
+            std::dynamic_pointer_cast<LevelZeroSignal>(m_signal_pushed[ii])->
```
cmcantalupo: Be nice if we could avoid this cast here.  Not sure why m_signal_pushed needs to be a Signal pointer and not a LevelZeroSignal pointer.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+                throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
+                                "Handling not defined for " + control_name,
+                                GEOPM_ERROR_LOGIC, __FILE__, __LINE__);
+    #endif
+        }
+    }
+
+    // Implemented to allow an IOGroup to save platform settings before starting
+    // to adjust them
+    void LevelZeroIOGroup::save_control(void)
```
cmcantalupo, line 780: This is a serious gap in the responsibilities of the IOGroup.  Ideally, all controls that the user has permissions to modify should be saved and restored.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+             ++domain_idx) {
+            // TODO: Read ALL LEVELZERO Power Limits, frequency settings, etc
+        }
+    }
+
+    // Implemented to allow an IOGroup to restore previously saved
+    // platform settings
+    void LevelZeroIOGroup::restore_control(void)
+    {
+        /// @todo: Usage of the LEVELZERO API for setting frequency, power, etc
```
cmcantalupo: Now that we have a LevelZero abstraction for interfacing with the library, unit testing should not be a problem.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+    }
+
+    // Function used by the factory to create objects of this type
+    std::unique_ptr<IOGroup> LevelZeroIOGroup::make_plugin(void)
+    {
+        return geopm::make_unique<LevelZeroIOGroup>();
+    }
+
+    void LevelZeroIOGroup::register_signal_alias(const std::string &alias_name,
+                                            const std::string &signal_name)
```
cmcantalupo:
```suggestion
    void LevelZeroIOGroup::register_signal_alias(const std::string &alias_name,
                                                 const std::string &signal_name)
```

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+        }
+        auto it = m_control_available.find(control_name);
+        if (it == m_control_available.end()) {
+            // skip adding an alias if underlying control is not found
+            return;
+        }
+        // copy control info but append to description
+        m_control_available[alias_name] = it->second;
+        m_control_available[alias_name].m_description =
+       m_control_available[control_name].m_description + '\n' +
```
cmcantalupo, line 903:
```suggestion
        m_control_available[alias_name].m_description =
            m_control_available[control_name].m_description + '\n' +
```

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+                                         "    alias_for: " + control_name;
```
cmcantalupo:
```suggestion
        m_control_available[alias_name].m_description =
            m_control_available[control_name].m_description +
            "\n    alias_for: " + control_name;
```

cmcantalupo (commented):
I'd like to understand if we can avoid casting the Signal pointer to a LevelZeroSignal pointer.  I'm also pretty concerned about the lack of save/restore functionality.  We should be able to save expressly the values that can be modified, and not the others.

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,906 @@
...
+                                                   domain,
+                                                   agg_function(ds.base_name),
+                                                   format_function(ds.base_name),
+                                                   result,
+                                                   nullptr,
+                                                   1
+                                                   };
+            }
+        }
+
```
lhlawson:
```suggestion
```

src/LevelZeroIOGroup.cpp
lhlawson: Resolved in latest commit.  It is specific to the single control we currently support and will need to be expanded as additional controls are added.

src/LevelZeroIOGroup.cpp
lhlawson: m_signal_pushed contains pointers to both Derivative and LevelZeroSignals, hence the Signal type being used.  
Added a set_sample function to the Signal base class, which helps avoid the cast being needed.

src/LevelZeroIOGroup.hpp
lhlawson: The pushed signals could be LevelZeroSignals or DerivativeSignals, so Signal is used.

lhlawson:
> 
> 
> I'd like to understand if we can avoid casting the Signal pointer to a LevelZeroSignal pointer. I'm also pretty concerned about the lack of save/restore functionality. We should be able to save expressly the values that can be modified, and not the others.

Save & Restore has been added, and the casting has been avoided through modifying the base Signal class to provide the set_sample(double) function.  Please see replies to individual comments for more info.

src/LevelZeroDevicePool.cpp
```diff
@@ -168,6 +169,26 @@ namespace geopm
...
     }
 
+    std::pair<double, double> LevelZeroDevicePoolImp::frequency_range(int domain,
+                                                                      unsigned int domain_idx,
+                                                                      int l0_domain) const
+    {
+        if (domain != GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
+            throw Exception("LevelZeroDevicePool::" + std::string(__func__) +
+                             ": domain " + std::to_string(domain) +
+                            " is not supported for the frequency domain.",
```
cmcantalupo:
```suggestion
            throw Exception("LevelZeroDevicePool::" + std::string(__func__) +
                            ": domain " + std::to_string(domain) +
                            " is not supported for the frequency domain.",
```

cmcantalupo (dismissed):
LGTM!!!

src/LevelZeroSignal.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "config.h"
+
+#include "LevelZeroSignal.hpp"
+
+#include "geopm_hash.h"
+#include "geopm_debug.hpp"
+#include "MSRIO.hpp"
+#include "Exception.hpp"
+#include "Helper.hpp"
```
bgeltz:
```suggestion
#include "Exception.hpp"
```

src/LevelZeroSignal.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#ifndef LEVELZEROSIGNAL_HPP_INCLUDE
+#define LEVELZEROSIGNAL_HPP_INCLUDE
+
+#include <cstdint>
+#include <cmath>
+
+#include <string>
+#include <memory>
+#include <functional>
```
bgeltz:
```suggestion
#include <functional>
```

src/LevelZeroIOGroup.hpp
```diff
@@ -0,0 +1,135 @@
...
+                                        const std::string &control_name);
+
+            struct control_s {
+                double m_setting;
+                bool m_is_adjusted;
+            };
+
+            struct signal_info {
+                std::string m_description;
+                int domain_type;
```
bgeltz:
```suggestion
                int m_domain_type;
```

src/LevelZeroIOGroup.hpp
```diff
@@ -0,0 +1,135 @@
...
+                std::function<std::string(double)> m_format_function;
+                std::vector<std::shared_ptr<Signal> > m_signals;
+                std::function<double (unsigned int)> m_devpool_func;
+                double scalar;
+            };
+
+            struct control_info {
+                std::string m_description;
+                std::vector<std::shared_ptr<control_s> > controls;
+                int domain_type;
```
bgeltz:
```suggestion
                std::vector<std::shared_ptr<control_s> > m_controls;
                int m_domain_type;
```

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,935 @@
...
+        };
+
+        std::vector<signal_data> derivative_signals {
+            {"LEVELZERO::POWER",
+                    "Average accelerator power over 40 ms or 8 control loop iterations",
+                    "LEVELZERO::ENERGY",
+                    "LEVELZERO::ENERGY_TIMESTAMP"},
+            {"LEVELZERO::UTILIZATION",
+                    "GPU utilization"
+                        "n  Level Zero logical engines may may to the same hardware"
```
bgeltz:
```suggestion
                        "\n  Level Zero logical engines may map to the same hardware"
```

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,935 @@
...
+                        "\n  resulting in a reduced signal range (i.e. not 0 to 1)",
+                    "LEVELZERO::ACTIVE_TIME",
+                    "LEVELZERO::ACTIVE_TIME_TIMESTAMP"},
+            {"LEVELZERO::UTILIZATION_COMPUTE",
+                    "Compute engine utilization"
+                        "n  Level Zero logical engines may may to the same hardware"
```
bgeltz:
```suggestion
                        "\n  Level Zero logical engines may map to the same hardware"
```

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,935 @@
...
+                        "\n  resulting in a reduced signal range (i.e. not 0 to 1)",
+                    "LEVELZERO::ACTIVE_TIME_COMPUTE",
+                    "LEVELZERO::ACTIVE_TIME_COMPUTE_TIMESTAMP"},
+            {"LEVELZERO::UTILIZATION_COPY",
+                    "Copy engine utilization"
+                        "n  Level Zero logical engines may may to the same hardware"
```
bgeltz:
```suggestion
                        "\n  Level Zero logical engines may map to the same hardware"
```

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,935 @@
...
+            throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
+                            ": domain_idx out of range.", GEOPM_ERROR_INVALID,
+                            __FILE__, __LINE__);
+        }
+
+        if (control_name == "LEVELZERO::FREQUENCY_GPU_CONTROL" ||
+            control_name == "FREQUENCY_ACCELERATOR_CONTROL") {
+            m_levelzero_device_pool.frequency_control(domain_type, domain_idx,
+                                                      geopm::LevelZero::M_DOMAIN_COMPUTE,
+                                                      setting / 1e6, setting / 1e6);
```
bgeltz, line 767: Does it make sense to move the ```1e6``` into a scalar field of the control_s like what you did with signal_s?
The ```1e6``` is pretty buried in the code here.

bgeltz: Also, why aren't you using ```m_control_available``` here?

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,935 @@
...
+    {
+        if (m_signal_available.find(alias_name) != m_signal_available.end()) {
+            throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
+                            ": signal_name " + alias_name +
+                            " was previously registered.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        auto it = m_signal_available.find(signal_name);
+        if (it == m_signal_available.end()) {
+            // skip adding an alias if underlying signal is not found
```
bgeltz, line 905: Throw in this case because the code is messed up?

Or is this about certain signals only being available on certain HW flavors?

src/LevelZeroIOGroup.cpp
```diff
@@ -0,0 +1,935 @@
...
+    {
+        if (m_control_available.find(alias_name) != m_control_available.end()) {
+            throw Exception("LevelZeroIOGroup::" + std::string(__func__) +
+                            ": contro1_name " + alias_name +
+                            " was previously registered.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        auto it = m_control_available.find(control_name);
+        if (it == m_control_available.end()) {
+            // skip adding an alias if underlying control is not found
```
bgeltz, line 926: Same thing here.  Throw or is this a HW support issue?

bgeltz (changes requested):
Structure looks pretty good.  A few docu changes, and a review of the includes should make this g2g.

Is there a test_levelzero_controls.py on the way?

src/LevelZeroIOGroup.cpp
lhlawson: The LevelZero controls do not use an underlying Control object, unlike the signals (which do use the Signal object), so they don't have a LevelZeroDevicePool function associated with them.  Instead they act more like the NVMLIOGroup controls and signals, so a simple if/else was used with hard coded function calls and scalars.

Moving to a full LevelZeroControl object is likely the best course of action long term.

src/LevelZeroIOGroup.cpp
lhlawson: This followed the pattern used in the CNLIOGroup::register_signal_aliases.  There's no LevelZero specific reason for this implementation, unless there's an overarching need for supporting incorrect aliases in GEOPM going forward.  The same applies to the NVMLIOGroup usage of this approach.

src/LevelZeroIOGroup.cpp
lhlawson: Same as above, no LevelZero specific need for this approach vs throw.

bgeltz (approved):

PR #1807, opened 2021-08-24
Avoid a dangling pointer in geopmagent

dannosliwcd:
- Some build types on some systems result in failure to get agent policy
  information. The root cause is a dangling pointer in the CLI parser.


cmcantalupo (approved):
Thanks!  LGTM

PR #1810, opened 2021-08-24
Add unit tests for runtime.py

### LINKED ISSUE #1803: Add unit tests for geopmdpy/runtime.py
> estimate 1

Parent issue: #1672

Done criteria: Add unit tests that cover runtime.py implementation to at least 80% statement coverage and have these tests merged into the geopm-service branch.

Note that this issue was created 3 days after the sprint started as this class did not exist at the start of the sprint.

bgeltz:
- Resolves #1803.

Signed-off-by: Brad Geltz <brad.geltz@intel.com>

### DIFF
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -275,5 +275,5 @@ class Controller:
             raise
         finally:
             pio.restore_control()
-            self._returncode = pid.returncode
+        self._returncode = pid.returncode
         return self._agent.get_report()
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -212,6 +212,8 @@ class Controller:
                              process) if timeout == 0.
 
         """
+        if not isinstance(agent, Agent):
+            raise ValueError('agent must be a subclass of Agent.')
         self._agent = agent
         self._signals = agent.get_signals()
         self._controls = agent.get_controls()
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -86,7 +86,9 @@ class TimedLoop:
                             units of seconds.
 
             num_period (int): Number of time periods spanned by the
-                              loop + 1.
+                              loop.  The total loop time is num_periods * period,
+                              but since there is no delay in the first loop, there
+                              will be num_period + 1 loop iterations.
 
         """
 
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -214,6 +214,8 @@ class Controller:
         """
         if not isinstance(agent, Agent):
             raise ValueError('agent must be a subclass of Agent.')
+        if timeout < 0:
+            raise ValueError('timeout must be >= 0')
         self._agent = agent
         self._signals = agent.get_signals()
         self._controls = agent.get_controls()
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -38,6 +38,7 @@
 import math
 import time
 import subprocess
+import sys
 from . import pio
 
 
@@ -226,6 +227,7 @@ class Controller:
         if timeout != 0:
             self._num_update = math.ceil(timeout / self._update_period)
         self._argv = argv
+        self._returncode = None
 
     def read_all_signals(self):
         """Sample for all signals pushed with pio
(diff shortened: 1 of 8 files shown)


service/geopmdpy/runtime.py
```diff
@@ -85,10 +85,17 @@ def __init__(self, period, num_period=None):
                             units of seconds.
 
             num_period (int): Number of time periods spanned by the
-                              loop.
+                              loop + 1.
```
cmcantalupo: We should clarify in the documentation.  The total loop time is num_periods * period, but since there is no delay in the first loop, there will be num_period + 1 loop iterations.

service/geopmdpy/runtime.py
```diff
@@ -238,13 +247,16 @@ def run(self):
             pid = subprocess.Popen(self._argv)
             for loop_idx in TimedLoop(self._update_period, self._num_update):
                 if pid.poll() is not None:
+                    # Nice process has died error, and RC
```
cmcantalupo: Totally, we should be tracking the return code of the subprocess and add an accessor method to the Controller object so it can be returned from main (should update the test_controller.py accordingly).

service/geopmdpy/runtime.py
```diff
@@ -238,13 +247,16 @@ def run(self):
...
                     break
                 pio.read_batch()
                 signals = self.read_all_signals()
                 controls = self._agent.update(signals)
                 for control_idx in self._controls_idx:
-                    pio.adjust(control_idx, control[control_idx])
```
cmcantalupo, line 246: Nice!

service/geopmdpy_test/TestRuntime.py
```diff
@@ -0,0 +1,237 @@
...
+    def get_period(self):
+        return self._period
+
+    def update(self, signals):
+        return [self._adjust_value] * len(signals)
+
+    def get_report(self):
+        return self._report_data
+
+class TestRuntime(unittest.TestCase):
```
cmcantalupo: This works, but I'd rather have the test classes map to the classes they are testing.  It means we have to have multiple test files for runtime.py, TestTimedLoop.py and TestController.py, and the tests for TimedLoop and Controller would be independent.

service/geopmdpy_test/TestRuntime.py
```diff
@@ -0,0 +1,237 @@
...
+            if ii == 2:
+                elapsed = time() - t1
+                self.assertAlmostEqual(period, elapsed, delta = 0.001)
+            t1 = time()
+            self.assertEqual(ii, next(it))
+
+        with self.assertRaisesRegex(StopIteration, ''):
+            next(it)
+
+    def test_agent(self):
```
cmcantalupo: Could probably leave this class in the TestController test class in TestController.py.  Seems a little overkill to have a separate file to test a class without an implementation.  May be worth checking that all methods have a __doc__ string that is not empty.

service/geopmdpy_test/TestRuntime.py
```diff
@@ -0,0 +1,237 @@
...
+            self.assertEqual(argv, con._argv)
+
+            calls = [mock.call(*cc) for cc in signals]
+            pps.assert_has_calls(calls)
+            calls = [mock.call(*cc) for cc in controls]
+            ppc.assert_has_calls(calls)
+
+        # Again, with timeout, same expectations otherwise
+        timeout = 10
+        with mock.patch('geopmdpy.pio.push_signal', side_effect = itertools.count()) as pps, \
```
cmcantalupo: what is the itertools.count() doing here?

bgeltz: https://docs.python.org/3/library/itertools.html#itertools.count

I'm using it so that every time you call the mocked push_signal, it'll return an ever increasing integer, starting at 0 in this case.  Here's an example:

```
>>> import itertools
>>> from unittest import mock
>>> m = mock.MagicMock()
>>> m()
<MagicMock name='mock()' id='140437089806936'>
>>> m.side_effect = itertools.count()
>>> m()
0
>>> m()
1
>>> m()
2
```

service/geopmdpy/runtime.py
cmcantalupo:
```suggestion
                              loop.  The total loop time is num_periods * period,
                              but since there is no delay in the first loop, there
                              will be num_period + 1 loop iterations.
```

cmcantalupo (approved):

PR #1817, opened 2021-08-27
LevelZero Update to handle devices without sub-devices

### LINKED ISSUE #1815: Current levelzero implementation requires subdevices to function
A limitation of the current sub-device support implementation is that we do not support devices that do not enumerate sub-devices.  Theoretically this is ANY level zero gpu with ZES_AFFINITY_MASK set in a particular configuration.  

This differs from our implementation of core vs. cpu when hyper-threading is enabled/disabled (1 cpu is always enumerated per core).

lhlawson:
An update to force devices without sub-devices to act as if a single sub-device is present, keeping the accelerator & accelerator_chip relationship in line with core & cpu.  

Tested on internal systems with & without enumerated sub-devices.

- Fixes #1815 from github issues.

### DIFF
--- a/src/LevelZero.cpp
+++ b/src/LevelZero.cpp
@@ -109,26 +109,31 @@ namespace geopm
                 check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
                                 "LevelZero::" + std::string(__func__) +
                                 ": LevelZero Sub-Device acquisition failed.", __LINE__);
-                // A limitation of the current subdevice support implementation is that we do NOT support devices
-                // without subdevices.  Theoretically this is ANY Level Zero GPU, depending on how the user sets
-                // the ZE_AFFINITY_MASK environment variable.
+#ifdef GEOPM_DEBUG
                 if (num_subdevice == 0) {
-                    throw Exception("LevelZero::" + std::string(__func__) +
-                                    ": GEOPM Requires subdevices to be enumerated." +
-                                    " Please check ZE_AFFINITY_MASK enviroment variable settings",
-                                    GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                    std::cerr << "LevelZero::" << std::string(__func__)  <<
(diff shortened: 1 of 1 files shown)


src/LevelZero.cpp
```diff
@@ -109,26 +109,31 @@ namespace geopm
...
+                            // device as a subdevice.
+                            m_num_board_gpu_subdevice += 1;
+                        }
 
                         //NOTE: We're only supporting Board GPUs to start with
                         m_devices.push_back({
                             device_handle.at(device_idx),
                             property,
-                            num_subdevice,
+                            num_subdevice, //if there are no subdevices, leave this a 0
```
lhlawson:
```suggestion
                            num_subdevice, //if there are no subdevices, leave this as 0
```

It may be better to change this to a boolean such as m_has_fake_subdevice

src/LevelZero.cpp
```diff
@@ -109,26 +109,31 @@ namespace geopm
...
                         ++m_num_board_gpu;
-                        m_num_board_gpu_subdevice += num_subdevice;
+                        if (num_subdevice != 0) {
+                            m_num_board_gpu_subdevice += num_subdevice;
+                        }
+                        else {
+                            // If there are no subdevices we are going to treat the
+                            // device as a subdevice.
+                            m_num_board_gpu_subdevice += 1;
+                        }
```
cmcantalupo:
```suggestion
                        m_num_board_gpu_subdevice += num_subdevice;
                        if (num_subdevice == 0) {
                            // If there are no subdevices we are going to treat the
                            // device as a subdevice.
                            m_num_board_gpu_subdevice += 1;
                        }
```

cmcantalupo (commented):
Just a suggestion for readability, but otherwise looks good to me.  If you want to merge my suggestion, I'd be happy to approve the change.  Please make sure I didn't miss something in the suggested change.

cmcantalupo (approved):

PR #1836, opened 2021-09-03
Fixed the python controller so we can set controls

alawibaba:
Hey folks, this fixes some minor issues with the python controller. Also fixed another python test that was failing.

### DIFF
--- a/service/test/test_controller.py
+++ b/service/test/test_controller.py
@@ -45,21 +45,24 @@ class LocalAgent(geopmdpy.runtime.Agent):
     """
     def __init__(self):
         self._loop_idx = 0
+        self._cpu_freq = None
 
     def get_signals(self):
         return [("TIME", geopmdpy.topo.DOMAIN_BOARD, 0),
                 ("INSTRUCTIONS_RETIRED", geopmdpy.topo.DOMAIN_PACKAGE, 0)]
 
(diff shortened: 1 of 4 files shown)


service/test/test_controller.py
```diff
@@ -45,21 +45,24 @@ class LocalAgent(geopmdpy.runtime.Agent):
...
         return [("TIME", geopmdpy.topo.DOMAIN_BOARD, 0),
                 ("INSTRUCTIONS_RETIRED", geopmdpy.topo.DOMAIN_PACKAGE, 0)]
 
     def get_controls(self):
-        return []
+        return [("FREQUENCY", geopmdpy.topo.DOMAIN_PACKAGE, 0)]
 
     def update(self, signals):
+        if self._cpu_freq is None:
+            self._cpu_freq = geopmdpy.pio.read_signal('CPU_FREQUENCY_MAX', 'board', 0)
```
cmcantalupo: put this call in __init__()?

cmcantalupo (commented):
Thanks Ali!

service/test/test_controller.py
alawibaba: Done!

alawibaba:
Thank you!

Issue #1694, opened 2021-06-18
As a GEOPM developer, I would like to maintain only one copy of the CFFI binding code


cmcantalupo:
> estimate 2

The pio.py, topo.py error.py and version.py were copied from geopmpy to geopmdpy.  These are bindings to C interfaces and in the case of geopmdpy these bind to libgeopmd and for geopmpy these bind to libgeopmpolicy.

This implies that installing the geopmpy module will require the geopmdpy module as a dependency.  Consider renaming the geopmdpy module?

geopmdpy -> geopmpy
geopmpy -> geopm????

Parent story: #1695

PR #1837, opened 2021-09-08
Completely separate service directory source from base directory build

### LINKED ISSUE #1695: As a user of the geopm libraries, I would prefer a clear hierachy of library dependecies rather than having symbols defined in multiple libraries
The service directory is constructed as an independent package, perhaps it would be better if libgeopmpolicy.so, and libgeopm.so depended on libgeopmd.so.  Might want to consider changing some .so names in the process, and we would have to have a different object name for the service side PlatformIO class.  

We currently depend on configure time options to compile different PlatformIO classes into the two libraries, but this may be better handled in C++ rather than the C preprocessor.

### LINKED ISSUE #1825: Modify Makefile.am in base repo to not reference .cpp or .hpp files in service directory
> estimate 3

Modify the build so that all of the objects that are compiled into libgeopmd.so are linked through that library when satisfying symbol dependencies in libgeopm.so and libgeopmpolicy.so.  Use `geopm/service/src` as an include directory for the base build.

Done criteria (partial):
If a user does an in place build of the service directory and then does an in place build of the base directory (or vice-versa) without doing a make clean in between they will end up with the wrong objects because of how configure time macros are being used to differentiate the two builds.

Parent issue: #1695

### LINKED ISSUE #1830: Move service related unit tests into the service directory
> estimate 1

0. Change existing geopm/service/test directory to geopm/service/test_integration
1. Unit test files in geopm/test should be moved to geopm/service/test as apprioriate.
2. Modify ```make check``` in service directory to run those tests

cmcantalupo:
- Fixes #1695 
- Fixes #1825 
- Fixes #1830
- Related to #1694


service/src/IOGroup.cpp
```diff
@@ -34,20 +34,17 @@
...
+#include "geopm/IOGroup.hpp"
 
 #include "geopm_plugin.hpp"
-#include "MSRIOGroup.hpp"
-#include "CpuinfoIOGroup.hpp"
-#include "TimeIOGroup.hpp"
+#include "geopm/MSRIOGroup.hpp"
+#include "geopm/CpuinfoIOGroup.hpp"
+#include "geopm/TimeIOGroup.hpp"
 #include "SSTIOGroup.hpp"
```
bgeltz, line 43: We need some docu somewhere that spells out why certain IOGroups are in certain spots.

service/src/geopm_hash.c
```diff
@@ -43,7 +47,19 @@ extern "C"
 
 uint64_t geopm_crc32_u64(uint64_t begin, uint64_t key)
```
bgeltz, line 48: Why does the Agg class have a dependency on this thing?  If this is a feature of the runtime, it should not be in the service directory.

service/test/Makefile.mk
```diff
@@ -0,0 +1,373 @@
...
+
+PHONY_TARGETS += gtest-checkprogs
+
+$(GTEST_TESTS): test/gtest_links/%:
+	mkdir -p test/gtest_links
+	rm -f $@
+	ln -s $(abs_srcdir)/test/geopm_test.sh $@
+
+coverage: check
+	lcov --no-external --capture --directory src --output-file coverage-base.info
```
bgeltz: Rename coverage-base.info.  And coverage-base below.

src/FrequencyMapAgent.cpp
```diff
@@ -39,17 +39,18 @@
...
-#include "contrib/json11/json11.hpp"
+#include "geopm/json11.hpp"
 
 #include "geopm.h"
 #include "geopm_hash.h"
 
 #include "Environment.hpp"
-#include "PlatformIO.hpp"
-#include "PlatformTopo.hpp"
+#include "PlatformIOProf.hpp"
```
bgeltz, line 48: For agents that load as a plugin, this header should be installed?

src/PlatformIOProf.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+namespace geopm
+{
+    PlatformIO &PlatformIOProf::platform_io(void)
+    {
+        static PlatformIOProf instance;
+        return instance.m_platform_io;
+    }
+
+    PlatformIOProf::PlatformIOProf()
```
bgeltz, line 55: Every line can be covered with a negative test here, so a test constructor is worthwhile.

test/Makefile.mk
```diff
@@ -823,5 +531,3 @@ clean-local-gtest-script-links:
 	rm -f test/gtest_links/*
 
 CLEAN_LOCAL_TARGETS += clean-local-gtest-script-links
-
-include test/googletest.mk
```
bgeltz: This should not have been moved.

Makefile.am
```diff
@@ -991,6 +756,7 @@ PHONY_TARGETS = clean-local \
                 # end
 
 include test/Makefile.mk
+include test/googletest.mk
```
bgeltz: Move this back to test/Makefile.am.

configure.ac
```diff
@@ -507,6 +441,14 @@ AC_SUBST([enable_ompt])
 AM_CONDITIONAL([ENABLE_OMPT], [test "x$enable_ompt" = "x1"])
 AM_CONDITIONAL([HAS_OMPT], [test "x$has_ompt" = "x1"])
 
+AC_SEARCH_LIBS([geopm_pio_read_signal], [geopmd])
+AC_CHECK_LIB([geopmd], [geopm_pio_read_signal], [], [
+    echo "missing libgeopmd: Install the geopm-service and use --with-geopmd or --with-geopmd-lib to specify location"
+    exit -1])
+AC_CHECK_HEADER([geopm.h], [], [
```
bgeltz: geopm_pio.h

Does geopm.h have to be moved?  Can it be moved back?

bgeltz (changes requested):

bgeltz:
We still have to figure out how to deal with building the service code with GNU, and the base dir runtime with intel.
Build script for both dirs?
Documentation spelling out exactly what to do and why: In the basedir README and in the (TODO/NEW) sphinx dev docu.
Configure checks/overrides for making sure the service dir is built with GNU
etc., etc.

service/src/IOGroup.cpp
cmcantalupo: I have removed all but MSRIOGroup.hpp from the installed headers.  geopmadmin does a lot of runtime related tasks in /etc/geopm in addition to the MSR allow list generation.

src/FrequencyMapAgent.cpp
cmcantalupo: I am now installing it, but we can still refer to it without the geopm/ since it is located in the same directory as this file.

src/PlatformIOProf.cpp
cmcantalupo: Currently the only constructor is private.  Perhaps we can figure out how to test this in another patch?

service/src/geopm_hash.c
cmcantalupo: It doesn't, but many of the tests in the service/test/*.cpp include the geopm_hash.h header.

configure.ac
```diff
@@ -507,6 +441,14 @@ AC_SUBST([enable_ompt])
 AM_CONDITIONAL([ENABLE_OMPT], [test "x$enable_ompt" = "x1"])
 AM_CONDITIONAL([HAS_OMPT], [test "x$has_ompt" = "x1"])
 
+AC_SEARCH_LIBS([geopm_pio_read_signal], [geopmd])
+AC_CHECK_LIB([geopmd], [geopm_pio_read_signal], [], [
+    echo "missing libgeopmd: Install the geopm-service and use --with-geopmd or --with-geopmd-lib to specify location"
+    exit -1])
+AC_CHECK_HEADER([geopm_pio.h], [], [
+    echo "missing geopm.h: Intall the geopm-service and use --with-geopmd or --with-geopmd-include to specify location"
```
bgeltz:
```suggestion
    echo "missing geopm_pio.h: Intall the geopm-service and use --with-geopmd or --with-geopmd-include to specify location"
```

service/test/Makefile.mk
```diff
@@ -0,0 +1,373 @@
...
+PHONY_TARGETS += gtest-checkprogs
+
+$(GTEST_TESTS): test/gtest_links/%:
+	mkdir -p test/gtest_links
+	rm -f $@
+	ln -s $(abs_srcdir)/test/geopm_test.sh $@
+
+coverage: check
+	lcov --no-external --capture --directory src --output-file coverage-service.info
+	genhtml coverage-service.info --output-directory coverage-base
```
bgeltz:
```suggestion
	genhtml coverage-service.info --output-directory coverage-service
```

bgeltz: The coverage target needs to be updated in the base dir/test Makefile.mk as well.  It still has refs to the service directory.

bgeltz (changes requested):

bgeltz (approved):

PR #1840, opened 2021-09-09
WIP: Refresh of Variorum plugin for GEOPM

amarathe84:
This is a WIP PR for the Variorum plugin for GEOPM. The Variorum PlatformIO plugin is intended to enable integration of Variorum with GEOPM.

### DIFF
--- a/src/VariorumControl.cpp
+++ b/src/VariorumControl.cpp
@@ -1,0 +1,128 @@
+/*
+ */
+
+#include "config.h"
+
+#include "VariorumFieldControl.hpp"
+#include "VariorumIO.hpp"
+#include "Variorum.hpp"
+#include "Exception.hpp"
+#include "geopm_debug.hpp"
+
+namespace geopm
+{
+    VariorumFieldControl::VariorumFieldControl(std::shared_ptr<MSRIO> msrio,
+                                     int cpu,
+                                     uint64_t offset,
+                                     int begin_bit,
+                                     int end_bit,
+                                     int function,
+                                     double scalar)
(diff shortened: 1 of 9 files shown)


cmcantalupo:
Thanks for the submission @amarathe84!  I will look over soon.

cmcantalupo:
@amarathe84 Can we set up a phone meeting with you and anyone else who is going to be working on this port (maybe @slabasan)?  

I think it would be good to start this work from the bottom up, and it would be better if I understood the Variorum interface better, so I could advise how to do that.  

Usually we start by defining a C++ abstraction around the external interface (in this case the Variorum library interface).  Once this interface is defined, the logic for the classes that support the IOGroup can be tested while mocking out the external dependency.  

I suggest that we start by determining the lowest level abstraction for the Variorum interface.  It may be as simple as a class with methods that map exactly to Variorum function calls.  We just need to select which of the Variorum functions will be required to implement the IOGroup.  

Some examples of last level abstraction classes:

https://github.com/geopm/geopm/blob/dev/src/MSRIO.hpp
https://github.com/geopm/geopm/blob/dev/src/SSTIoctl.hpp
https://github.com/geopm/geopm/blob/dev/src/NVMLDevicePool.hpp
https://github.com/geopm/geopm/blob/dev/src/LevelZero.hpp
https://github.com/geopm/geopm/blob/dev/src/MPIComm.hpp

These each provide a thin wrapper around an external dependency, and enable us to test the rest of the implementation without the external requirements by using googlemock.  

Thanks!

PR #1847, opened 2021-09-10
Ci fixes

bgeltz:
A collection of changes to get our CI pipeline working with the service branch.  The build process and dependency resolution has changed a bit, and there were many changes needed to adhere to the architecture on the service branch.


.github/workflows/build.yml
```diff
@@ -68,37 +67,51 @@ jobs:
...
     - name: install system dependencies
-      run: sudo apt-get update && sudo apt-get install libelf-dev mpich libmpich-dev libomp-9-dev libsystemd-dev
-    - name: install python dependencies
-      uses: py-actions/py-dependency-install@v2
-      with:
-        path: scripts/requirements.txt
+      run: sudo apt-get update && sudo apt-get install libelf-dev mpich libmpich-dev libomp-9-dev libsystemd-dev libgirepository1.0-dev
+    - name: install base and service dir python dependencies
+      run: |
+           python3 -m pip install --upgrade pip setuptools wheel pep517
```
cmcantalupo, line 73: I guess we could go back to the standard github script now that the requirements.txt no longer reference geopmdpy?

service/geopmdpy/__init__.py
```diff
@@ -34,6 +34,7 @@
 
 """
 
+from __future__ import absolute_import
```
cmcantalupo, line 37: This is only for supporting python 2.x, we really want to get rid of all of these abolute_import lines, not add them back

service/geopmdpy/runtime.py
```diff
@@ -35,6 +35,8 @@
 
 """
 
+from __future__ import absolute_import
+from __future__ import division
```
cmcantalupo, line 39: Similar issue with division, we are in the __future__ :)

cmcantalupo (changes requested):
Just a couple comments

.github/workflows/build.yml
bgeltz: No, we can't. The PyGobject requirement is problematic, and requires pip517 in order for pip to be able to successfully complete the source build within the GH actions VM.  There's no option with the standard github script thing to use pep517.

Example error:
https://github.com/bgeltz/geopm/runs/3591856424?check_suite_focus=true

You can disable pep517 outright:
https://github.com/bgeltz/geopm/runs/3591889886?check_suite_focus=true

But it seems better to try and use it since it's the new standard.

service/geopmdpy/__init__.py
bgeltz: Should I do this a part of this PR, or a new one?

service/geopmdpy/runtime.py
bgeltz: Sure, just trying to satisfy the test we have that checks all this stuff.  I can rip it all out in this PR too if that's the desire.

cmcantalupo (approved):

PR #1849, opened 2021-09-13
Fix test wrapper scripts in the service directory

cmcantalupo:


build.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Set this variable to append configure options for base build
+# GEOPM_BASE_CONFIG_OPTIONS=
+
+# Set this variable to append configure options for service build
+# GEOPM_SERVICE_CONFIG_OPTIONS=
+
+
+if [ ! -z "$NUM_THREAD" ]; then
+    NUM_THREAD=20
```
bgeltz: 20 is way too may IMO.  I typically use 9, but perhaps we should run some tests.

There used to be old advice that make doesn't parallelize well above 9.

build.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Simple script that will build and test both the base and the service
+
+set -e
+
+# Set this variable to append configure options for base build
+# GEOPM_BASE_CONFIG_OPTIONS=
+
+# Set this variable to append configure options for service build
+# GEOPM_SERVICE_CONFIG_OPTIONS=
```
bgeltz: Maybe some examples?  Like ```--enable-debug```

build.sh
```diff
@@ -0,0 +1,71 @@
...
+cd ..
+./autogen.sh
+mkdir -p build
+cd build
+../configure --with-geopmd-lib=../service/build/.libs \
+             --with-geopmd-include=../service/src \
+             ${GEOPM_BASE_CONFIG_OPTIONS}
+make -j${NUM_THREAD}
+make -j${NUM_THREAD} checkprogs
+make check
```
bgeltz, line 71: I'm surprised you didn't have to do any more hacking to your LD_LIBRARY_PATH or PYTHONPATH to make this work.

bgeltz (changes requested):

build.sh
bgeltz: Ah, I see now that I'm looking at the full PR that the changes in geopm_test.sh.in provide this functionality.  Nice.

PR #1850, opened 2021-09-14
Fix copying_headers/test-dist so it uses both distribution tarballs

cmcantalupo:


cmcantalupo:
This work was merged with https://github.com/geopm/geopm/pull/1847

PR #1866, opened 2021-09-16
Document systemd-devel package requirement

### LINKED ISSUE #1865: The geopm-service.spec.in file does not specify BuildRequires for libsystemd.so and headers
The ServiceIOGroup requires libsystemd.so and sdbus.h.  Need to update the spec file to reflect this requirement for both RHEL and SLES flavors of package names.

cmcantalupo:
- Fixes #1865 from github issues.

### DIFF
--- a/service/Makefile.am
+++ b/service/Makefile.am
@@ -180,6 +180,8 @@ libgeopmd_la_SOURCES = $(include_HEADERS) \
                        src/PlatformTopoImp.hpp \
                        src/RawMSRSignal.cpp \
                        src/RawMSRSignal.hpp \
+                       src/SDBus.hpp \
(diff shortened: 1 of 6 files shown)


fkeceli:
Looks good.

fkeceli (dismissed):
Looks good.

bgeltz (changes requested):
Configure time check to make sure systemd-devel is present if ENABLE_SYSTEMD was specified?

cmcantalupo:
@bgeltz 

https://github.com/geopm/geopm/blob/geopm-service/service/configure.ac#L216-L223

bgeltz (approved):

PR #1868, opened 2021-09-16
Purge unused CI workflows and tools

bgeltz:
- TravisCI has been abandoned in favor of GitHub CI.
- The pylint package is no longer necessary as we no longer support Python 2.
- Cancel in-progress CI runs if a new commit is pushed to the branch that is currently being run.  This works for both branches on forks that will be triggered by CI (currently "dev" and "geopm-service") and for PRs that are in-progress.

Example rendering of new repo badges and links:
https://github.com/bgeltz/geopm/tree/geopm-service

Example of a cancelled CI run:
https://github.com/geopm/geopm/actions/runs/1242822888
And the real run:
https://github.com/geopm/geopm/actions/runs/1242825756


cmcantalupo (approved):
Looks good.  Might be good to have some pylint feature in the future, but we can add them in as a separate commit.

PR #1872, opened 2021-09-16
Update Agent interface and improve documentation

### LINKED ISSUE #1862: As a user of the python controller/agent infrastructure I would like to be able to call Controller.run() multiple times
> estimate 1

At least one additional method for the Agent class must be defined to tell the agent object that a new run is beginning.  May also want another method that defines the end of a run for the agent.  The Agent.get_report() method could be used for this purpose, but seems to be overloading the interface.

cmcantalupo:
- Enables an Agent object to be used with a single controller to
  execute multiple run() calls.
- Fixes #1862 from github issues

### DIFF
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -149,6 +149,9 @@ class Agent:
     def get_signals(self):
         """Get list of read requests
 
+        The returned signals will be sampled from the platform by the
+        Controller and passed into Agent.update().
+
         Returns:
             list((str, int, int)): List of request tuples. Each
                                    request comprises a signal name,
@@ -160,6 +163,9 @@ class Agent:
     def get_controls(self):
         """Get list of control requests
 
+        The returned controls will be set in the platform by the Controller
+        based on the return value from Agent.update().
+
         Returns:
             list((str, int, int)): List of request tuples. Each
                                    request comprises a control name,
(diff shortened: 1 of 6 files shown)


service/geopmdpy/runtime.py
```diff
@@ -265,9 +317,10 @@ def run(self):
         """
         try:
             pid = subprocess.Popen(self._argv)
+            self._agent.run_begin(policy)
             for loop_idx in TimedLoop(self._update_period, self._num_update):
                 if pid.poll() is not None:
-                    sys.stderr.write('Ths app process has ended.  return code = {}'.format(pid.returncode))
+                    sys.stderr.write('The app process has ended.  return code = {}'.format(pid.returncode))
```
fkeceli: Can you put a \n at the end of the string?

fkeceli (commented):
Few things:

- What about moving argv to run() method as I brought up before?
- In the Controller __init__ (...timeout) documentation it is mentioned that "the agent will run indefinitely (until killed by another process) if timeout == 0. I know that run will end if the app finishes naturally (not killed). I think that doc can be clearer.

cmcantalupo:
Thanks for the feedback @fkeceli.  I think the last two patches address your comments.

fkeceli (dismissed):
Good changes. Thanks Chris. I actually hit "Do not throw on calls to ServiceIOGroup::read_batch()/write_batch()" bug in Saruman.Thanks for making the change.

service/geopmdpy/runtime.py
```diff
@@ -181,7 +227,7 @@ def update(self, signals):
         raise NotImplementedError('Agent is an abstract base class')
 
     def get_period(self):
-        """Get the target time interal for the control loop
+        """Get the target time internal for the control loop
```
bgeltz:
```suggestion
        """Get the internal target time for the control loop
```
or just
```suggestion
        """Get the target time for the control loop
```

It's a getter.  It's assumed it's returning internal state.

service/integration/test/test_controller.py
```diff
@@ -69,24 +83,29 @@ def get_report(self):
...
         sys.stderr.write(help_msg)
         return err
+    cpu_frequency_max = float(sys.argv[1])
     agent = LocalAgent()
-    controller = geopmdpy.runtime.Controller(agent, sys.argv[1:])
-    print(controller.run())
-    return err
+    controller = geopmdpy.runtime.Controller(agent)
+    print(controller.run(sys.argv[2:], cpu_frequency_max))
+    return controller.returncode()
```
bgeltz, line 108: Nice.

service/src/ServiceIOGroup.cpp
```diff
@@ -171,14 +171,12 @@ namespace geopm
 
     void ServiceIOGroup::read_batch(void)
     {
-        throw Exception("ServiceIOGroup::read_batch()",
-                        GEOPM_ERROR_NOT_IMPLEMENTED, __FILE__, __LINE__);
+
```
bgeltz: Maybe at least a comment here saying these are still unimplemented, and not intentionally empty?

bgeltz (changes requested):

service/geopmdpy/runtime.py
cmcantalupo: I think that was supposed to be *interval* not internal

bgeltz (approved):

PR #1873, opened 2021-09-16
Collection of changes to enable nightly CI

### LINKED ISSUE #1832: Run nightly unit/integration tests with the service branch
> estimate 2

-1. Integrate the new build process (service subdir first, then base dir) into the nightly scripting.
0. Run the unit tests with the new structure.

Then:

1. Run the integration tests on the service branch similar to how we always have (where the user has direct access to the IOGroups).
2. ~~Prototype a second run where the user does not have direct access to the IOGroups, but the GEOPM service is installed (so the root user has access).  Then the expectation would be the service provides access for the desired signals if the batch interface was implemented.  Since the batch interface is not implemented, it is expected that these tests will fail.~~ This is captured in issue #1861 
3. Generate code coverage numbers with lcov for the service directory.

bgeltz:
- This resolves multiple issues with running the integration tests.
- Fixes #1832.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,7 +32,7 @@
 # GLOBAL SETTINGS
 ACLOCAL_AMFLAGS = -I m4
 AM_CPPFLAGS += -I$(top_srcdir)/src \
-               -DGEOPM_DEFAULT_PLUGIN_PATH=\"$(pkglibdir)\" \
+               -DGEOPM_DEFAULT_PLUGIN_PATH=\"$(libdir)/geopm\" \
                -D_POSIX_C_SOURCE=200809L \
                -DOMPI_SKIP_MPICXX
 
--- a/service/src/geopm_plugin.cpp
+++ b/service/src/geopm_plugin.cpp
@@ -31,10 +31,12 @@
  */
 
 #include <dlfcn.h>
+#include <stdlib.h>
 
 #include <vector>
 #include <algorithm>
 #include <iostream>
+#include <string>
 
 #include "geopm_error.h"
 #include "geopm_plugin.hpp"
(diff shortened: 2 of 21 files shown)


Makefile.am
```diff
@@ -32,7 +32,7 @@
 # GLOBAL SETTINGS
 ACLOCAL_AMFLAGS = -I m4
 AM_CPPFLAGS += -I$(top_srcdir)/src \
-               -DGEOPM_DEFAULT_PLUGIN_PATH=\"$(pkglibdir)\" \
```
cmcantalupo, line 35: Since the base package name is geopm, I think this change should resolve to the same string?  Perhaps it is clearer with the change?

service/src/geopm_plugin.cpp
```diff
@@ -48,10 +50,17 @@ namespace geopm
 {
     void plugin_load(const std::string &plugin_prefix)
     {
+        std::string env_plugin_path_str(geopm::get_env("GEOPM_PLUGIN_PATH"));
         std::vector<std::string> plugin_paths {GEOPM_DEFAULT_PLUGIN_PATH};
         std::string so_suffix = ".so." GEOPM_ABI_VERSION;
         std::replace(so_suffix.begin(), so_suffix.end(), ':', '.');
 
+        if (!env_plugin_path_str.empty()) {
```
cmcantalupo, line 58: This looks correct, but I'm surprised we didn't already have this feature...

cmcantalupo (approved):
Couple of comments, but LGTM

service/src/geopm_plugin.cpp
bgeltz: You had removed to back in April to get the build working.  It's in dev.

Makefile.am
bgeltz: Yes, it does resolve to the same string.  This was done so that it's obvious where this is pointing when you compare this against the same line in ```service/Makefile.am```.  Previously this was broken in the service Makefile.

cmcantalupo (approved):

PR #1893, opened 2021-09-28
First draft of updated build and coverage instructions

bgeltz:
- Many TODOs added.

PR #1898, opened 2021-09-29
Pare down README so it applies to both the service and base source

### LINKED ISSUE #1666: As a GEOPM developer I would like for the top level README to explain the what the service directory is for
> estimate 1

Parent story: #1811 

Top level geopm/README.md needs to be updated to describe the service.

cmcantalupo:
- Fixes #1666 from github issues.

### DIFF
--- a/README
+++ b/README
@@ -8,21 +8,26 @@ GEOPM - Global Extensible Open Power Manager
 [![License](https://img.shields.io/badge/License-BSD%203--Clause-blue.svg)](https://opensource.org/licenses/BSD-3-Clause)
 
 
-WEB PAGES
+Web Pages
 ---------
 https://geopm.github.io <br>
+https://geopm.github.io/geopmdpy/ <br>
 https://geopm.github.io/man/geopm.7.html <br>
 https://geopm.slack.com
 
-SUMMARY
+
+Summary
 -------
+
 The Global Extensible Open Power Manager (GEOPM) is a framework for
-exploring power and energy optimizations targeting high performance
-computing.
+exploring power and energy optimizations on heterogeneous platforms.
 
 With GEOPM you can:
 
-- Interact with hardware settings using a platform-agnostic interface
+- Interact with hardware settings and sensors using a
+  platform-agnostic interface
+- Restore changes to hardware settings when configuring process
+  terminates
 - Profile applications to study their power and energy behavior
 - Automatically detect MPI and OpenMP phases in an application
 - Optimize MPI applications to improve energy efficiency or reduce the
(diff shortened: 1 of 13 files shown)


RUNTIME.md
```diff
@@ -0,0 +1,536 @@
+
+
+
+
+
```
bgeltz:
```suggestion
```

README
```diff
@@ -598,16 +138,19 @@ GEOPM with the `--enable-beta` configure flag.  The features currently
...
+
 SEE COPYING FILE FOR LICENSE INFORMATION.
 
-LAST UPDATE
+Last Update
 -----------
-2020 September 25
+2021 September 24
 
 Christopher Cantalupo <christopher.m.cantalupo@intel.com> <br>
```
bgeltz, line 155: ```Brad Geltz <brad.geltz@intel.com> <br>```

README
```diff
@@ -32,553 +37,88 @@ With GEOPM you can:
...
-To build all targets and install it in a "build/geopm" subdirectory of your
-home directory run the following commands:
+Guide for GEOPM Developers
+--------------------------
 
-    ./autogen.sh
-    ./configure --prefix=$HOME/build/geopm
-    make
-    make install
+GEOPM is an open development project and we use github to plan, review
```
bgeltz:
```suggestion
GEOPM is an open development project and we use GitHub to plan, review
```

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+control system can optimize for various objective functions including
+maximizing global application performance within a power bound or
+minimizing energy consumption with marginal degradation of application
+performance.  The root of the control hierarchy tree can communicate
+with the system resource manager to extend the hierarchy above the
+individual MPI application and enable the management of system power
+resources for multiple MPI jobs and multiple users by the system
+resource manager.
+
+The GEOPM package provides two libraries: libgeopm for use with MPI
```
bgeltz: What about libgeopmd?  There's no discussion of the service here.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+users.
+
+PYTHON INSTALL
+--------------
+The GEOPM python tools are packaged in the RPMs described above, but
+the latest release is also available from PyPI as the `geopmpy`
+package.  For example, to install the geopmpy package into your home
+directory, run the following command:
+
+    python -m pip install --user geopmpy
```
bgeltz: Also geopmdpy

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+with MPI and OpenMP through their profiling decorators, and the direct
+calls are made through the geopm_prof_c(3) or geopm_fortran(3)
+interfaces.  Marking up a compute application with profiling
+information through these interfaces can enable better integration of
+the GEOPM runtime with the compute application and more precise
+control.
+
+INSTALL
+-------
+The OpenHPC project provides the most robust way to install GEOPM.
```
bgeltz: The version in OpenHPC does not have any of the service bits.  Not sure this statement is true anymore.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+    python -m pip install --user "git+https://github.com/geopm/geopm.git@dev#egg=geopmpy&subdirectory=scripts"
+
+Note these commands only install the GEOPM Python tools and do not
+install the binaries required by the GEOPM runtime.  Without these binary files
+*only* the following modules will load successfully:
+
+    geopmpy.io
+    geopmpy.launcher
+    geopmpy.plotter
+    geopmpy.update_report
```
bgeltz: We should check that this is still correct.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+
+All other modules rely on the output of the GEOPM build (namely
+libgeopmpolicy.so) being present and available in LD_LIBRARY_PATH or
```
bgeltz: No discussion of libgeopmd.so dependency.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+    yum install openmpi-devel elfutils-libelf-devel ruby-devel rubygems
+
+SUSE:
+
+    zypper install openmpi-devel elfutils-libelf-devel ruby-devel rubygems
+
+UBUNTU (as of 18.04.3 LTS):
+
+    apt install libtool automake libopenmpi-dev build-essential gfortran \
+        libelf-dev ruby ruby-dev python libsqlite3-dev
```
bgeltz: Missing reqs for service (e.g. python-gobject).

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+for details on how to use non-standard install locations for build
+requirements through the
+
+    ./configure --with-<feature>
+
+options.
+
+
+BUILD INSTRUCTIONS
+------------------
```
bgeltz: This whole section needs to be reworked to discuss the build script and/or the environment files in integration/config that setup CC, CXX, etc. for a given system.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+home directory run the following commands:
+
+    ./autogen.sh
+    ./configure --prefix=$HOME/build/geopm
+    make
+    make install
+
+An RPM can be created on a RHEL or SUSE system with the
+
+    make rpm
```
bgeltz: We have a whole other guide on how to build RPMs now.  We should just point to that.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+
+    ./configure --disable-mpi
+
+option can be used to build only targets which do not require MPI.  By default
+MPI targets are built.
+
+
+RUN REQUIREMENTS
+----------------
+We are targeting SLES12 and RHEL7 distributions for functional runtime
```
bgeltz: Old OS versions discussed.  This isn't accurate today.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+If the script output contains:
+    WARNING: The lock bit for the PKG_POWER_LIMIT MSR is set.  The power_balancer
+             and power_governor agents will not function properly until this is cleared.
+
+Please enable RAPL in your BIOS and if such an option doesn't exist please
+contact your BIOS vendor to obtain a RAPL supported BIOS.
+
+For additional information, please contact the GEOPM team.
+
+### USER ENVIRONMENT
```
bgeltz: This should be reworked to discuss the ```integration/config/run_env.sh``` script.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+processes in all configurations.  Please refer to your site
+documentation about CPU affinity for the best solution on the system
+you are using and consider extending the geopmlaunch wrapper to
+support your system configuration (please see the CONTRIBUTING.md file
+for information about how to share these implementation with the
+community).
+
+TESTING
+-------
+From within the source code directory, unit tests can be executed with
```
bgeltz: Which source code directory?  The base or service?

We should be explicit about their being two different spots that you need to do this process in.  Examples would be nice too.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+test cyle is executed with the following commands:
+
+    make -j
+    make checkprogs -j
+    make check
+
+The unit tests can be executed on any development system, including
+VMs and containers, that meets the BUILD REQUIREMENTS section above.
+
+The integration tests are located in the "integration/test" directory.
```
bgeltz: This should be under a sub heading for "Integration Tests"

RUNTIME.md
bgeltz: This should be under a sub heading for "Unit Tests"

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+    python3 $GEOPM_SOURCE/integration/test
+
+These integration tests are based on pyunit and leverage the geopmpy
+python package to validate the runtime.  Please report failures of
+these tests as issues.
+
+Note that if the `GEOPM_SKIP_COMPILER_CHECK` environment variable is
+set when `${GEOPM_SOURCE}/build.sh` is called, then the check to
+verify that the Intel compilers are used will be bypassed.  This can
+be useful for performing builds with the GNU toolchain.
```
bgeltz: This should be moved to the build section.  Not sure why it's here.  Not relevant to integration tests.

RUNTIME.md
```diff
@@ -0,0 +1,536 @@
...
+
+RESOURCE MANAGER INTEGRATION
```
bgeltz: Are there implications of the service and libgeopmd.so that we have not captured in this section?

e.g. do we need libgeopmd.so to make these calls to save_control, etc.?

service/docs/source/devel.rst
```diff
@@ -3,7 +3,11 @@ Guide for GEOPM Developers
 ==========================
 
 These are instructions for a developer that would like to modify the
-source code in the GEOPM repository.
+source code in the GEOPM repository.  The GEOPM repository contains
+two independent build systems that are used to compile, test and
+install software components written in C++ and Python.  The `service`
+subdirectory of the geopm repository contains a build system for the
```
bgeltz:
```suggestion
subdirectory of the GEOPM repository contains a build system for the
```

bgeltz (changes requested):

service/docs/source/devel.rst
```diff
@@ -2,5 +2,123 @@
...
+
+   ./configure --disable-mpi
+   make
+   make check
+
+
+The integration tests require a system with MPI and multiple compute nodes.
+
+Coverage Instructions
+---------------------
```
bgeltz: Check out what I started here:
https://github.com/geopm/geopm/pull/1893/files

build.sh
```diff
@@ -72,19 +72,25 @@ if [ ! -f "${BUILD_ENV}" ]; then
...
-# Clear out old builds
-if [ -d "${GEOPM_INSTALL}" ]; then
+# Check that GEOPM_INSTALL is defined unless we are skipping install
+if [ -z "${GEOPM_INSTALL}" ] && [ -z "${GEOPM_SKIP_INSTALL+x}" ];
+    echo "Please define GEOPM_INSTALL prior to using this script" 1>&2
+    exit 1
+fi
+
+# Clear out old builds unless we are skipping install
+if [ -d "${GEOPM_INSTALL}" ] && [ -z "${GEOPM_SKIP_INSTALL+x}" ]; then
```
bgeltz, line 86: This seems risky.  I have absolutely debugged issues before where a stale build was being picked up because it was in the LD_LIBRARY_PATH o and/or PATH, and the current build was never installed.  I'd really prefer to just remove any old build that's laying around to avoid problems like this.

build.sh
```diff
@@ -72,19 +72,25 @@ if [ ! -f "${BUILD_ENV}" ]; then
 fi
 source ${BUILD_ENV}
 
-# Clear out old builds
-if [ -d "${GEOPM_INSTALL}" ]; then
+# Check that GEOPM_INSTALL is defined unless we are skipping install
+if [ -z "${GEOPM_INSTALL}" ] && [ -z "${GEOPM_SKIP_INSTALL+x}" ];
+    echo "Please define GEOPM_INSTALL prior to using this script" 1>&2
+    exit 1
+fi
```
bgeltz: If they're using the recommendations from the docs, then they're using ~/.geopmrc and GEOPM_INSTALL is set.  Even if we aren't installing, having GEOPM_INSTALL set is benign.  Why bother?

Additionally, I don't think this is even reachable if GEOPM_INSTALL is not set.  build_env.sh was sourced on line 73, and enforces that GEOPM_INSTALL is set:
https://github.com/geopm/geopm/blob/dev/integration/config/build_env.sh#L49

build.sh
```diff
@@ -72,19 +72,25 @@ if [ ! -f "${BUILD_ENV}" ]; then
...
     echo "Removing old build located at ${GEOPM_INSTALL}"
     rm -rf ${GEOPM_INSTALL}
 fi
 
-# e.g. To do a debug build, add --enable-debug to this env var
-GEOPM_GLOBAL_CONFIG_OPTIONS="${GEOPM_GLOBAL_CONFIG_OPTIONS} --prefix=${GEOPM_INSTALL}"
+# Append prefix to global config options unless skipping install
+if [ -z "${GEOPM_SKIP_INSTALL+x}" ]; then
+    GEOPM_GLOBAL_CONFIG_OPTIONS="${GEOPM_GLOBAL_CONFIG_OPTIONS} --prefix=${GEOPM_INSTALL}"
+fi
```
bgeltz, line 94: Having the prefix set when we are not installing should not affect anything, right?  Why bother?

bgeltz (changes requested):

build.sh
cmcantalupo: It seems really strange to error because GEOPM_INSTALL is not defined but GEOPM_SKIP_INSTALL is defined. If the user has chosen to "skip install" why would we want call rm -rf on whatever happens to by living in the environment variable "GEOPM_INSTALL"?  Perhaps we should add a prompt asking the user before force removing an entire directory?  I don't like it when rm -rf applies to a variable that I'm not using.

build.sh
cmcantalupo: It goes in the rpath of the compiled binaries.

service/docs/source/devel.rst
```diff
@@ -3,59 +3,236 @@ Guide for GEOPM Developers
...
+There are many options that may be passed to each of the two configure
+scripts that are part of the GEOPM repository build system.  Two
+scripts called ``autogen.sh`` are provided, one in the base of the
+GEOPM repository and the other in the service directory.  Each of
+these scripts manage the GEOPM version that is imbedded in the build
+artifacts, and create the two ``configure`` scripts using the
+autotools package.
+
+Running the configure scripts generate a number of output files,
+including the ``Makefile`` that is use for the rest of the build
```
cmcantalupo: use -> used

service/docs/source/devel.rst
```diff
@@ -3,59 +3,236 @@ Guide for GEOPM Developers
...
+steps.  The ``configure`` scripts accept a large number of command
+line options, and environment variables that affect the behavior.
+Each of configure script will provide user documentation through the
+``./configure --help`` command.  Some important options and
+environment variables are listed below.
+
+Both configure scripts
+^^^^^^^^^^^^^^^^^^^^^^
+
+* ``--enable-debug``
```
cmcantalupo, line 135: document --prefix

RUNTIME.md
cmcantalupo: I've been trying to refer to the "GEOPM package" and the "GEOPM Service package" perhaps we should call the base build package the "GEOPM HPC Runtime package" throughout?

integration/config/build.sh
```diff
@@ -74,19 +74,25 @@ if [ ! -f "${BUILD_ENV}" ]; then
 fi
 source ${BUILD_ENV}
 
-# Clear out old builds
-if [ -d "${GEOPM_INSTALL}" ]; then
+# Check that GEOPM_INSTALL is defined unless we are skipping install
+if [ -z "${GEOPM_INSTALL}" ] && [ -z "${GEOPM_SKIP_INSTALL+x}" ];
```
bgeltz: Continuing from: https://github.com/geopm/geopm/pull/1898#discussion_r719458258
It's impossible to get here if ```${GEOPM_INSTALL}``` is not defined.  This is enforced by ```set -e```  and the sourcing of ```build_env.sh``` above.  If the user has chosen to skip the install, we want to be absolutely certain that we remove anything that's in ```GEOPM_INSTALL``` so it is not picked up.  We have a number of different scripts that will add to a user's ```PATH``` and ```LD_LIBRARY_PATH```, so in order to avoid debugging problems with stale builds it seems to be the most safe to just remove an old build.

I'm fine to add a check around the ```rm -fr``` to be sure they want it removed.  But if they don't want it removed, then I want to ```exit 1``` here and stop.  The user should be forced to move the old build from ```$GEOPM_INSTALL``` to some other backup directory.   Having ```$GEOPM_INSTALL``` contain an old build that's not currently under test is a recipe for debugging headaches.

integration/config/build.sh
```diff
@@ -74,19 +74,25 @@ if [ ! -f "${BUILD_ENV}" ]; then
...
     echo "Removing old build located at ${GEOPM_INSTALL}"
     rm -rf ${GEOPM_INSTALL}
 fi
 
-# e.g. To do a debug build, add --enable-debug to this env var
-GEOPM_GLOBAL_CONFIG_OPTIONS="${GEOPM_GLOBAL_CONFIG_OPTIONS} --prefix=${GEOPM_INSTALL}"
+# Append prefix to global config options unless skipping install
+if [ -z "${GEOPM_SKIP_INSTALL+x}" ]; then
+    GEOPM_GLOBAL_CONFIG_OPTIONS="${GEOPM_GLOBAL_CONFIG_OPTIONS} --prefix=${GEOPM_INSTALL}"
+fi
```
bgeltz, line 94: Continuing from: https://github.com/geopm/geopm/pull/1898#discussion_r719460267
Is that a problem?  If we enforce that anything that was in ```$GEOPM_INSTALL``` was removed above, is there another issue here?

service/docs/source/devel.rst
```diff
@@ -3,59 +3,213 @@ Guide for GEOPM Developers
...
 
-To launch the GEOPM unit tests, run the following command in the geopm
-build directory after running configure:
+.. code-block:: bash
 
-.. code-block::
+    # Build the geopm-service package
+    cd service/
+    ./autogen.sh
+    ./configure
```
bgeltz, line 33: If systemd is default on still (I think it is), this will be broken on systems that do not have the systemd dependencies from the package manager.

Since most users won't have access to install these requirements on large systems (e.g. Theta) I think we should default to systemd disabled.

service/docs/source/devel.rst
```diff
@@ -3,59 +3,213 @@ Guide for GEOPM Developers
...
+
+* ``export CXX=``
+  Set the C++ compiler with environment variable
 
-.. code-block::
 
-   ./configure --disable-mpi
-   make
-   make check
+Service configure script
```
bgeltz, line 121: The requirement to build with GNU is not captured.

service/docs/source/devel.rst
```diff
@@ -3,59 +3,213 @@ Guide for GEOPM Developers
...
+  Set the MPI C compiler wrapper with environment variable
+
+* ``export MPICXX=``
+  Set the MPI C++ compiler wrapper with environment variable
+
+* ``export MPIFC=``
+  Set the Fortran compiler wrapper with environment variable
+
+* ``export MPIF77=``
+  Set the Fortran 77 compiler wrapper with environment variable
```
bgeltz, line 190: Also ~MPIF90, MPIFORT~, F90, F77, FC...

bgeltz (changes requested):

integration/config/build.sh
```diff
@@ -74,19 +74,25 @@ if [ ! -f "${BUILD_ENV}" ]; then
 fi
 source ${BUILD_ENV}
 
-# Clear out old builds
-if [ -d "${GEOPM_INSTALL}" ]; then
+# Check that GEOPM_INSTALL is defined unless we are skipping install
```
bgeltz, line 77: A small comment saying this is an additional check for GEOPM_INSTALL above and beyond what's happening in the build_env.sh would be nice.

bgeltz: And some testing of all the different combinations of these env variables.

service/docs/source/devel.rst
bgeltz: Open issue on investigating this.

bgeltz: Default off until the next release?

service/docs/source/devel.rst
bgeltz, line 149: The option to ```--disable-systemd``` should be high in this list since it is most likely to be used on older systems.

bgeltz: The option to ```--disable-systemd``` should be high in this list since it is most likely to be used on older systems.

RUNTIME.md
bgeltz: I like being explicit about the "GEOPM HPC package" or "...runtime package"

RUNTIME.md
```diff
@@ -0,0 +1,337 @@
...
+tool is used to launch an MPI application while enabling the GEOPM
+runtime to create a GEOPM Controller thread on each compute node.  The
+Controller loads plugins and executes the Agent algorithm to control
+the compute application.  The geopmlaunch(1) command is part of the
+geopmpy python package that is included in the GEOPM installation.
+See the GEOPM overview man page for further documentation and links:
+[geopm(7)](https://geopm.github.io/man/geopm.7.html).
+
+The GEOPM runtime provides some build in algorithms, each is called an
+"Agent" that implements the geopm::Agent class interface.  A developer
```
bgeltz: Man page reference and links?

RUNTIME.md
```diff
@@ -0,0 +1,337 @@
...
+PlatformIO interface is provided by the GEOPM Service package which is
+contained in the service directory of the GEOPM repository.  The
+PlatformIO abstraction enables Agent implementations to be ported to
+different hardware platforms without modification.
+
+The libgeopm library can be called directly or indirectly within MPI
+applications to enable application feedback for informing the control
+decisions.  The indirect calls are facilitated by GEOPM's integration
+with MPI and OpenMP through their profiling decorators, and the direct
+calls are made through the geopm_prof_c(3) or geopm_fortran(3)
```
bgeltz: Man pages links?

RUNTIME.md
```diff
@@ -0,0 +1,337 @@
...
+interfaces.  Marking up a compute application with profiling
+information through these interfaces can enable better integration of
+the GEOPM runtime with the compute application and more precise
+control.
```
bgeltz: Tutorial plug?

RUNTIME.md
```diff
@@ -0,0 +1,337 @@
...
+    WARNING: The lock bit for the PKG_POWER_LIMIT MSR is set.  The power_balancer
+             and power_governor agents will not function properly until this is cleared.
+
+Please enable RAPL in your BIOS and if such an option doesn't exist please
+contact your BIOS vendor to obtain a RAPL supported BIOS.
+
+For additional information, please contact the GEOPM team.
+
+
+### User Environment
```
bgeltz: Does this info still belong here since it duplicates part of what's discussed in the devel.rst?

RUNTIME.md
```diff
@@ -0,0 +1,337 @@
...
+made prior to running any GEOPM tools:
+
+    export LD_LIBRARY_PATH=$GEOPM_PREFIX/lib:$LD_LIBRARY_PATH
+    export PATH=$GEOPM_PREFIX/bin:$PATH
+    export PYTHONPATH=$(ls -d $GEOPM_PREFIX/lib/python*/site-packages | tail -n1):$PYTHONPATH
+
+Use a PYTHONPATH that points to the site-packages created by the geopm build.
+The version created is for whichever version of python 3 was used in the
+configure step.  If a different version of python is desired, override the
+default with the --with-python option in the configure script.
```
bgeltz: Configure option to devel.rst?

RUNTIME.md
```diff
@@ -0,0 +1,337 @@
...
+administrator.  This feature is required by the GEOPM runtime if the
+GEOPM Service is not active on the system.  Alternately, the access
+management for MSRs may be configured by the system administrator
+using the GEOPM Service if it is active.
+
+The msr-safe kernel driver is distributed with OpenHPC and can be
+installed using the RPMs distributed there.  The source code for the
+driver can be found here at the link below.
+
+[msr-safe repo](https://github.com/scalability-llnl/msr-safe)
```
bgeltz: Repo URL needs to be updated.

configure.ac
```diff
@@ -218,6 +218,28 @@ fi
 AC_SUBST([enable_fortran])
 AM_CONDITIONAL([ENABLE_FORTRAN], [test "x$enable_fortran" = "x1"])
 
+AC_ARG_ENABLE([geopmd-local],
```
bgeltz, line 224: Comment on ordering of this option vs. with-geopmd below being vital to the logic functioning correctly here.

bgeltz (changes requested):

service/docs/source/devel.rst
cmcantalupo: I added a lot of documentation that should help if the user needs to disable access to the service in when building the service.

RUNTIME.md
cmcantalupo: These are all linked from the geopm(7) web page which has an explicit hyperlink in the rst file.  Perhaps we can open an issue to deal with the multitude of references to man pages that do not have hyperlinks.

service/docs/source/devel.rst
```diff
@@ -21,7 +21,7 @@ module.
 Developer Build Process
 -----------------------
 
-The simplest procedure for building all of the software in the GEOPM
+The basic procedure for building all of the software in the GEOPM
```
bgeltz, line 24: :)

service/docs/source/requires.rst
```diff
@@ -40,77 +59,75 @@ the requirement must be removed from the file
...
-  .. code-block:: bash
 
-      yum install openmpi-devel elfutils-libelf-devel ruby-devel rubygems
+Run Requirements
+----------------
 
+There are some time-of-use requirements for both the GEOPM Service and
+the GEOPM HPC Runtime.  The use of MSRs enables many important
+features of the GEOPM Service, although access to MSRs is strictly
+required only for the GEOPM HPC Runtime.  There may be some
```
bgeltz:
```suggestion
features of the GEOPM Service, although access to MSRs is strictly
required only for the GEOPM HPC Runtime, gathering telemetry through
``geopmread``, setting hardware controls with ``geopmwrite``, or using
GEOPM's Python APIs to achieve similar functionality as ``geopmread`` and ``geopmwrite``.  There may be some
```

bgeltz (changes requested):

bgeltz (approved):

Issue #1903, opened 2021-09-30
Introduce new C++ class: DCGMIOGroup


lhlawson:
> estimate 2
Parent feature: #1902 

Introduce a new C++ class DCGMIOGroup with the following features:
- Ability to query DCGM device pool for metric values
- Ability to set DCGM device pool control values

Done criteria:
- [ ] Pure virtual interface class has been added with factory method
- [ ] Unit test constructor provides dependency injection
- [ ] Unit tests added for all class methods except: non-test
      constructors, private methods or classes that provide a
      last-level-abstraction of an external dependency (e.g. SDBus or
      NVMLDevicePool)
- [ ] Non-exempt methods with logic branches have multiple tests
- [ ] All dependency classes have already been merged
- [ ] All public methods have doxygen comments describing behavior,
      inputs, outputs and exceptions raised
- [ ] Implementation file has `#include "config.h"`

Issue #1902, opened 2021-09-30
As a user of NVIDIA Hardware I would like to be able to access DCGM statistics


lhlawson:
**Is your feature request related to a problem? Please describe.**
I would like to be able to access the additional profiling metrics on NVIDIA Hardware exposed by the [Data Center GPU Manager (DCGM)](https://docs.nvidia.com/datacenter/dcgm/latest/). 

**Describe the solution you'd like**
The addition of an IO Group that provides the capability to query DCGM statistics from a standalone invocation of the DCGM Manager on the local node.

**Describe alternatives you've considered**
Alternatively DCGM support could be added directly to future agent(s) that use profiling metrics.  These metrics do not appear to be exposed by NVIDIA through any other means.  Using a DCGM policy outside of GEOPM is a possibility but has several drawbacks.   

**Additional context**
This feature is currently under development [here](https://github.com/lhlawson/geopm/blob/public-lhlawson-bt-agent/src/DCGMIOGroup.cpp) for research purposes.

PR #1914, opened 2021-10-01
Added DCGMIOGroup for NVIDIA Profiling Metrics

lhlawson:

Addition of a DCGM IOGroup that connects to a local standalone nv-hostengine in order to provide NVIDIA Profiling metrics.
Requires PR #1917 

Details of change
- Fulfills Feature #1902 from github issues.
- Fixes Issue #1903 from github issues.

Remaining opens:
- [ ] Save/Restore of sample rate.  This is complicated by not having a clear way to query these values - we can reset to default values (1s, etc) or the agent preferred values (100ms, etc)

### DIFF
--- a/service/src/DCGMIOGroup.cpp
+++ b/service/src/DCGMIOGroup.cpp
@@ -62,7 +62,7 @@ namespace geopm
         , m_dcgm_device_pool(device_pool)
         , m_is_batch_read(false)
         , m_signal_available({{"DCGM::SM_ACTIVE", {
-                                  "SM activity expressed as a ratio of cycles",
+                                  "Streaming Multiprocessor activity expressed as a ratio of cycles",
                                   {},
                                   [this](unsigned int domain_idx) -> double
                                   {
@@ -73,7 +73,7 @@ namespace geopm
                                   string_format_double
                                   }},
                               {"DCGM::SM_OCCUPANCY", {
-                                  "Warp residency expressed as a ratio of maximum warps per cycles",
+                                  "Warp residency expressed as a ratio of maximum warps",
                                   {},
                                   [this](unsigned int domain_idx) -> double
                                   {
@@ -84,7 +84,7 @@ namespace geopm
                                   string_format_double
                                   }},
                               {"DCGM::DRAM_ACTIVE", {
-                                  "DRAM Send & Receive expresed as a ratio of cycles",
+                                  "DRAM send & receive expressed as a ratio of cycles",
                                   {},
                                   [this](unsigned int domain_idx) -> double
                                   {
@@ -420,17 +420,7 @@ namespace geopm
     // platform settings
     void DCGMIOGroup::restore_control(void)
     {
-        // The 1 second update rate used here is the default sample rate discussed
-        // in the DCGM user guide.
-        m_dcgm_device_pool.update_rate(1e6); //1 second
-
-        // The max samples restore value is based upon the defaults used in DCGM
-        // sdk example code for field value access provided as part of the DCGM install
-        m_dcgm_device_pool.max_samples(3600); //3600 samples
-
-        // The max storage time restore value is based upon the defaults used in DCGM
-        // sdk example code for field value access provided as part of the DCGM install
-        m_dcgm_device_pool.max_storage_time(3600); //3600 seconds
+        m_dcgm_device_pool.polling_disable();
     }
 
     void DCGMIOGroup::save_control(const std::string &save_path)
--- a/service/src/DCGMIOGroup.cpp
+++ b/service/src/DCGMIOGroup.cpp
@@ -1,0 +1,544 @@
+/*
+ * Copyright (c) 2015 - 2021, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+
+#include <cmath>
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <cstring>
+#include <sched.h>
+#include <errno.h>
+
+#include "DCGMIOGroup.hpp"
+#include "DCGMDevicePool.hpp"
+#include "geopm/PlatformTopo.hpp"
+#include "geopm/Exception.hpp"
+#include "geopm/Agg.hpp"
+#include "geopm/Helper.hpp"
+
+namespace geopm
+{
+    DCGMIOGroup::DCGMIOGroup()
+        : DCGMIOGroup(platform_topo(), dcgm_device_pool())
+    {
+    }
+
+    // Set up mapping between signal and control names and corresponding indices
+    DCGMIOGroup::DCGMIOGroup(const PlatformTopo &platform_topo, DCGMDevicePool &device_pool)
+        : m_platform_topo(platform_topo)
+        , m_dcgm_device_pool(device_pool)
+        , m_is_batch_read(false)
+        , m_signal_available({{"DCGM::SM_ACTIVE", {
+                                  "SM activity expressed as a ratio of cycles",
+                                  {},
+                                  [this](unsigned int domain_idx) -> double
+                                  {
+                                      return this->m_dcgm_device_pool.field_value(
+                                                   domain_idx, geopm::DCGMDevicePool::M_SM_ACTIVE);
+                                  },
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"DCGM::SM_OCCUPANCY", {
+                                  "Warp residency expressed as a ratio of maximum warps per cycles",
+                                  {},
+                                  [this](unsigned int domain_idx) -> double
+                                  {
+                                      return this->m_dcgm_device_pool.field_value(
+                                                   domain_idx, geopm::DCGMDevicePool::M_SM_OCCUPANCY);
+                                  },
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"DCGM::DRAM_ACTIVE", {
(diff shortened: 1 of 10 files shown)


lhlawson:
PR to be broken into two separate PRs: DCGMDevicePool PR #1917 and DCGMIOGroup PR

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,541 @@
...
+    {
+    }
+
+    // Implemented to allow an IOGroup to restore previously saved
+    // platform settings
+    void DCGMIOGroup::restore_control(void)
+    {
+        m_dcgm_device_pool.update_rate(1*1e6); //1 second
+        m_dcgm_device_pool.max_samples(0); //0 = no limit
+        m_dcgm_device_pool.max_storage_time(10); //10 seconds.
```
cmcantalupo: Are these published default values?  What is the source for these values?

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,541 @@
...
+    {
+        throw Exception("DCGMIOGroup::save_control()",
+                        GEOPM_ERROR_NOT_IMPLEMENTED, __FILE__, __LINE__);
+    }
+
+    void DCGMIOGroup::restore_control(const std::string &save_path)
+    {
+        throw Exception("DCGMIOGroup::restore_control()",
+                        GEOPM_ERROR_NOT_IMPLEMENTED, __FILE__, __LINE__);
+    }
```
cmcantalupo: Since this IOGroup does not require any external state to save (noop) or restore (values compiled into binary), save_control(string) should be a noop and restore_control(string) should passthrough restore_control().

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    int DCGMIOGroup::signal_behavior(const std::string &signal_name) const
+    {
+        return IOGroup::M_SIGNAL_BEHAVIOR_VARIABLE;
+    }
+
+    // Name used for registration with the IOGroup factory
+    std::string DCGMIOGroup::plugin_name(void)
+    {
+        return "dcgm";
```
cmcantalupo: We use all-caps for the plugin_name() for all IOGroups except the ones you wrote @lhlawson.  Sorry I'm just catching this now, but for consistency, we should return an all-caps plugin_name for DGGMIOGroup, LevelZeroIOGroup and the NVMLIOgroup.  This string appears as a prefix to the signal and control names, which are also all-caps by convention.

service/test/DCGMIOGroupTest.cpp
```diff
@@ -0,0 +1,327 @@
...
+    DCGMIOGroup dcgm_io(*m_platform_topo, *m_device_pool);
+
+    double mock_rate = 100;
+    double mock_time = 6000;
+    double mock_samples = 60000;
+
+    batch_value[(dcgm_io.push_control("DCGM::FIELD_UPDATE_RATE",
+                                    GEOPM_DOMAIN_BOARD, 0))] = mock_rate;
+    EXPECT_CALL(*m_device_pool,
+                update_rate(mock_rate*1e6)).Times(1);
```
cmcantalupo:
```suggestion
                update_rate(mock_rate * 1e6)).Times(1);
```

service/test/DCGMIOGroupTest.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+    batch_value[(dcgm_io.push_control("DCGM::MAX_STORAGE_TIME",
+                                    GEOPM_DOMAIN_BOARD, 0))] = mock_time;
+    EXPECT_CALL(*m_device_pool,
+                max_storage_time(mock_time)).Times(1);
+
+    batch_value[(dcgm_io.push_control("DCGM::MAX_SAMPLES",
+                                    GEOPM_DOMAIN_BOARD, 0))] = mock_samples;
```
cmcantalupo: fix white space

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,541 @@
...
+        : DCGMIOGroup(platform_topo(), dcgm_device_pool())
+    {
+    }
+
+    // Set up mapping between signal and control names and corresponding indices
+    DCGMIOGroup::DCGMIOGroup(const PlatformTopo &platform_topo, DCGMDevicePool &device_pool)
+        : m_platform_topo(platform_topo)
+        , m_dcgm_device_pool(device_pool)
+        , m_is_batch_read(false)
+        , m_signal_available({{"DCGM::SM_ACTIVE", {
```
cmcantalupo, line 64: it is a bit arbitrary, but it might be nice for this data to live in a public method outside of the constructor.  If this were the case, you would be able to make assertions about the data in the available signals and controls.  It would also be a bit more readable.

cmcantalupo (commented):
Couple of small comments.

service/src/DCGMIOGroup.cpp
lhlawson: The update rate is the default value specified by the DCGM user guide [here](https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-user-guide/feature-overview.html#profiling-metrics)

The maximum samples (DCGM variable maxKeepSamples) and maximum storage time (DCGM variable maxKeepAge) don't appear to have a published default value, so a 'sensible' value was chosen from our testing.  The dcgmWatchFields interface using these values is [here](https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-api/group__DCGMAPI__FI.html#group__DCGMAPI__FI_1gd04fa06405af6165fd85310dee9eae2a)

service/src/DCGMIOGroup.cpp
lhlawson: Addressed for DCGM as part of this PR.  I'll put up a separate PR to update other IOGroups to LEVELZERO and NVML.

lhlawson: #2065 filed to track NVML and LEVELZERO change

service/src/DCGMIOGroup.cpp
lhlawson: This has been further tweaked to use the 'example' values provided in the DCGM codebase as no other default value could be found.

service/src/DCGMIOGroup.hpp
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright (c) 2015 - 2021, Intel Corporation
```
bgeltz:
```suggestion
 * Copyright (c) 2015 - 2022, Intel Corporation
```

This change went in for the rest of the files in the repo recently.  All of the license headers for this PR need to be updated, please.

service/src/DCGMIOGroup.hpp
```diff
@@ -0,0 +1,122 @@
...
+            DCGMIOGroup();
+            DCGMIOGroup(const PlatformTopo &platform_topo, DCGMDevicePool &device_pool);
+            ~DCGMIOGroup();
+            std::set<std::string> signal_names(void) const override;
+            std::set<std::string> control_names(void) const override;
+            bool is_valid_signal(const std::string &signal_name) const override;
+            bool is_valid_control(const std::string &control_name) const override;
+            int signal_domain_type(const std::string &signal_name) const override;
+            int control_domain_type(const std::string &control_name) const override;
+            int push_signal(const std::string &signal_name, int domain_type, int domain_idx)  override;
```
bgeltz:
```suggestion
            int push_signal(const std::string &signal_name, int domain_type, int domain_idx) override;
```

service/src/IOGroup.cpp
```diff
@@ -51,6 +51,9 @@
 #ifdef GEOPM_CNL_IOGROUP
 #include "CNLIOGroup.hpp"
 #endif
+#ifdef GEOPM_ENABLE_DCGM
```
bgeltz, line 54: The corresponding changes to enable this in ```service/configure.ac``` are not present in this PR.

Is this configure time option going to be automatically enabled when ```--enable-nvml``` is used or will there be 2 options (e.g. ```--enable-nvml``` and ```--enable-dcgm```)?

Would there be a reason not to enable DCGM if NVML support was requested?

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,551 @@
...
+                                  "SM activity expressed as a ratio of cycles",
```
bgeltz: I think we need some documentation somewhere (or at least a pointer to whatever is public) about how to interpret these signals.  My assumption reading this is that the ratio is inclusively defined [0, 1] but it would be good to say that somewhere.

What's SM?

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,551 @@
...
+                                  [this](unsigned int domain_idx) -> double
+                                  {
+                                      return this->m_dcgm_device_pool.sample(
+                                                   domain_idx, geopm::DCGMDevicePool::M_FIELD_ID_SM_OCCUPANCY);
+                                  },
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"DCGM::DRAM_ACTIVE", {
+                                  "DRAM Send & Receive expresed as a ratio of cycles",
```
bgeltz:
```suggestion
                                  "DRAM send & receive expressed as a ratio of cycles",
```

or maybe

```suggestion
                                  "DRAM bandwidth expressed as a ratio of cycles",
```

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,551 @@
...
+                                  [this](unsigned int domain_idx) -> double
+                                  {
+                                      return this->m_dcgm_device_pool.sample(
+                                                   domain_idx, geopm::DCGMDevicePool::M_FIELD_ID_SM_ACTIVE);
+                                  },
+                                  Agg::average,
+                                  string_format_double
+                                  }},
+                              {"DCGM::SM_OCCUPANCY", {
+                                  "Warp residency expressed as a ratio of maximum warps per cycles",
```
bgeltz:
```suggestion
                                  "Warp residency expressed as a ratio of maximum warps per cycle",
```

One cycle or many cycles?

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,551 @@
...
+                                  }},
+                             })
+        , m_control_available({{"DCGM::FIELD_UPDATE_RATE", {
+                                    "Rate at which field data is polled in seconds",
+                                    {},
+                                    Agg::expect_same,
+                                    string_format_double
+                                    }},
+                               {"DCGM::MAX_STORAGE_TIME", {
+                                    "Maximum time field data is stored in seconds",
```
bgeltz, line 105: Is there an upper limit on this field?

bgeltz (changes requested):

service/src/DCGMIOGroup.cpp
lhlawson: Your reading is correct.  SM is a streaming multiprocesser, the naming is based on the underlying NVIDIA signal names and descriptions.  

If we can find a good programmatic way to pull descriptions from the nvidia libraries (similar to the Intel oneAPI libraries) we may want to just provide the official descriptions on top of a note of how we're interpreting them (i.e. are we scaling them from 0-100 to 0-1, etc)

https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-user-guide/feature-overview.html#profiling

service/src/IOGroup.cpp
lhlawson: Corresponding change was included when we added the device pool in #1917 
https://github.com/lhlawson/geopm/blob/840c2dc20afe0709e5193684984373352d020416/service/configure.ac#L166

DCGM has additional software requirements and system setup requirements which may impact performance so they are separate at this time.

service/src/DCGMIOGroup.cpp
lhlawson: ratio of maximum warps (per cycles is incorrect)

lhlawson:
Review notes:

1. Implement a lazy init on the devicepool signals (moving the dcgmwatch & related calls out of the DCGMDevicePool constructor)
2. Restore control should call unwatch (but not destroy dcgmgroups)
3. Potentially add IO Group warning if the read rate is less than the devpool polling time.  
4. Potentially update DCGMDevicePool defaults to the current 'restore control' defaults

service/src/DCGMIOGroup.cpp
lhlawson: None provided in the NVIDIA docs, so uint32_t max.

bgeltz (dismissed):

service/src/DCGMIOGroup.cpp
```diff
@@ -0,0 +1,541 @@
...
+                throw Exception("DCGMIOGroup::" + std::string(__func__) + "Handling not defined for "
+                                + control_name, GEOPM_ERROR_LOGIC, __FILE__, __LINE__);
+    #endif
+        }
+    }
+
+    // Implemented to allow an IOGroup to save platform settings before starting
+    // to adjust them
+    void DCGMIOGroup::save_control(void)
+    {
```
dannosliwcd, line 416: Maybe add a comment here to justify the empty save/restore.

If I remember right from one of your previous explanations, these controls don't need to be explicitly saved/restored because the DCGM field group takes care of that when we create/destroy it, right?

dannosliwcd (dismissed):
LGTM. Minor/optional documentation suggestion in the comments.

service/src/DCGMIOGroup.cpp
lhlawson: Good idea.  
With the most recent updates we're calling the dcgm 'unwatch' in order to restore the system to the state prior to usage of the IOGroup (i.e. disabling the dcgm geopm_fields group).  This

dannosliwcd (approved):

PR #1917, opened 2021-10-05
Addition of DCGMDevicePool for NVIDIA Profiling Metrics

### LINKED ISSUE #1904: Add new external dependency: Data Center GPU Manager (DCGM)
> estimate 1
Parent [feature/story]: #1902 

Done criteria:
- [ ] Use of interface exposed by external dependency is compatible
      with the BSD 3 clause license
- https://github.com/NVIDIA/DCGM/blob/master/LICENSE points to http://www.apache.org/licenses/LICENSE-2.0
- [ ] Shared object dependencies have been added to the build system
      in this or a prior PR
- [ ] Dependency has one or more last-level-abstraction classes if
      used in C++
- [ ] Dependency may have last-level-abstraction when used in Python
      if this enables testing
- [ ] All existing features continue to work when dependency is not
      available
- [ ] If an external python module dependency was introduced, the new
      module is added to the requirements.txt and is available via PyPI
      (pip).
- [ ] Any external package required by geopmdpy for functionality of
      the GEOPM service must be provided by RPM or debian packaging from
      standard Linux distros.

### LINKED ISSUE #1915: Introduce new C++ class: DCGMDevicePool
> estimate 1
Parent [feature/story]: #1902

Introduce new C++ class DCGMDevicePool with the following features:
- Ability to query DCGM profiling from a standalone invocation of DCGM
- Ability to set DCGM watch field sample time, sample age, and sample rate

Done criteria:

- [ ] Pure virtual interface class has been added with factory method
      NVMLDevicePool)

lhlawson:

Addition of a DCGM Device pool that connects to a local standalone nv-hostengine in order to query NVIDIA Profiling metrics.  This PR is just the device pool and is not currently used by any classes in the code base.  This is a prerequisite for PR #1914.

Details of change:
- Added DCGMDevicePool class and throw.
- The user is expected to have setup a local instance of nv-hostengine for the device pool to connect to.  

Related Features:
- Relates to #1902
- Fixes #1915
- Fixes #1904

Long term it may be worth considering adding a way for the user to specify an IP address for the nv-hostengine.

### DIFF
--- a/service/src/DCGMDevicePool.cpp
+++ b/service/src/DCGMDevicePool.cpp
@@ -58,57 +58,83 @@ namespace geopm
         m_dcgm_field_ids[M_SM_ACTIVE] = DCGM_FI_PROF_SM_ACTIVE;
         m_dcgm_field_ids[M_SM_OCCUPANCY] = DCGM_FI_PROF_SM_OCCUPANCY;
         m_dcgm_field_ids[M_DRAM_ACTIVE] = DCGM_FI_PROF_DRAM_ACTIVE;
-        m_dcgm_field_ids[M_PCIE_TX_BYTES] = DCGM_FI_PROF_PCIE_TX_BYTES;
-        m_dcgm_field_ids[M_PCIE_RX_BYTES] = DCGM_FI_PROF_PCIE_RX_BYTES;
 
         dcgmReturn_t result;
 
         //Initialize DCGM
         result = dcgmInit();
-        check_dcgm_result(result, "Error Initializing DCGM.");
+        check_dcgm_result(result, "Error Initializing DCGM.", __LINE__);
 
         // We are assuming a local version of DCGM.  This could transition to a
         // dcgmStartEmbedded at a later date.
         char *host_ip = "127.0.0.1";
         result = dcgmConnect(host_ip, &m_dcgm_handle);
-        check_dcgm_result(result, "Error connecting to standalone DCGM instance");
+        check_dcgm_result(result, "Error connecting to standalone DCGM instance", __LINE__);
 
         //Check all devices are DCGM enabled
         unsigned int dcgm_dev_id_list[DCGM_MAX_NUM_DEVICES];
         result = dcgmGetAllSupportedDevices(m_dcgm_handle, dcgm_dev_id_list, &m_dcgm_dev_count);
-        check_dcgm_result(result, "Error fetching DCGM supported devices.");
+        check_dcgm_result(result, "Error fetching DCGM supported devices.", __LINE__);
 
-        for(int i = 0; i < m_dcgm_dev_count; ++i) {
+        for (int i = 0; i < m_dcgm_dev_count; ++i) {
             std::vector<dcgmFieldValue_v1> field_values(sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]));
-            m_dcgm_field_values.push_back(field_values);;
+            m_dcgm_field_values.push_back(field_values);
         }
 
         //Setup Field Group
+
+        char geopm_group[] = "geopm_field_group";
         result = dcgmFieldGroupCreate(m_dcgm_handle, sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]), &m_dcgm_field_ids[0],
-                                      (char *)"geopm_fields", &m_field_group_id);
-        check_dcgm_result(result, "Error creating DCGM geopm_fields group.");
+                                      geopm_group, &m_field_group_id);
+
+        //Retry case
+        if (result == DCGM_ST_DUPLICATE_KEY) {
+#ifdef GEOPM_DEBUG
+            std::cerr << "DCGMDevicePool::" << std::string(__func__)  <<
+                         ": Duplicate field group found. " << std::endl;
+#endif
+            result = dcgmFieldGroupDestroy(m_dcgm_handle, m_field_group_id);
+            check_dcgm_result(result, "Error destroying DCGM geopm_fields group.", __LINE__);
+            result = dcgmFieldGroupCreate(m_dcgm_handle, sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]), &m_dcgm_field_ids[0],
+                                          geopm_group, &m_field_group_id);
+            check_dcgm_result(result, "Error re-creating DCGM geopm_fields group.", __LINE__);
+        }
+        else {
+            check_dcgm_result(result, "Error creating DCGM geopm_fields group.", __LINE__);
+        }
 
         // Note: Currently we are using dcgmWatchFields, but may transition to using
         //       dcgmProfWatchFields at a later date
         result = dcgmWatchFields(m_dcgm_handle, DCGM_GROUP_ALL_GPUS, m_field_group_id, m_update_freq,
                                  m_max_keep_age, m_max_keep_sample);
 
-        check_dcgm_result(result, "Error setting default watch field configuration.");
+        check_dcgm_result(result, "Error setting default watch field configuration.", __LINE__);
+
+        for (auto dcgm_field_id : m_dcgm_field_ids) {
+            DcgmFieldGetById(dcgm_field_id);
+            if (DcgmFieldGetById(dcgm_field_id)->fieldType != DCGM_FT_DOUBLE) {
+                throw Exception("DCGMDevicePool::" + std::string(__func__) + ": "
+                                "DCGM Field ID " + std::to_string(dcgm_field_id) +
+                                " field type is not double",
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
     }
 
     DCGMDevicePoolImp::~DCGMDevicePoolImp()
     {
         //Shutdown DCGM
         dcgmStatusDestroy(NULL);
+        dcgmFieldGroupDestroy(m_dcgm_handle, m_field_group_id);
         dcgmGroupDestroy(m_dcgm_handle, DCGM_GROUP_ALL_GPUS);
     }
 
-    void DCGMDevicePoolImp::check_dcgm_result(const dcgmReturn_t result, const std::string error)
+    void DCGMDevicePoolImp::check_dcgm_result(const dcgmReturn_t result, const std::string error, const int line)
     {
         if (result != DCGM_ST_OK) {
             throw Exception("DCGMDevicePool::" + std::string(__func__) + ": "
                             + error + ": " + errorString(result),
-                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                            GEOPM_ERROR_INVALID, __FILE__, line);
         }
     }
 
@@ -128,7 +154,7 @@ namespace geopm
         result = dcgmGetLatestValuesForFields(m_dcgm_handle, accel_idx,
                         &m_dcgm_field_ids[0], sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]),
                         &(m_dcgm_field_values.at(accel_idx))[0]);
-        check_dcgm_result(result, "Error getting latest values for fields in read_batch");
+        check_dcgm_result(result, "Error getting latest values for fields in read_batch", __LINE__);
     }
 
     void DCGMDevicePoolImp::field_update_rate(int accel_idx, int field_update_rate)
@@ -137,7 +163,7 @@ namespace geopm
         m_update_freq = field_update_rate;
         result = dcgmWatchFields(m_dcgm_handle, DCGM_GROUP_ALL_GPUS, m_field_group_id, m_update_freq,
                                  m_max_keep_age, m_max_keep_sample);
-        check_dcgm_result(result, "Error setting Field Update Rate");
+        check_dcgm_result(result, "Error setting Field Update Rate", __LINE__);
     }
 
     void DCGMDevicePoolImp::max_storage_time(int accel_idx, int max_storage_time)
@@ -146,7 +172,7 @@ namespace geopm
         m_max_keep_age = max_storage_time;
         result = dcgmWatchFields(m_dcgm_handle, DCGM_GROUP_ALL_GPUS, m_field_group_id, m_update_freq,
                                  m_max_keep_age, m_max_keep_sample);
-        check_dcgm_result(result, "Error setting Max Storage Time");
+        check_dcgm_result(result, "Error setting Max Storage Time", __LINE__);
     }
 
     void DCGMDevicePoolImp::max_samples(int accel_idx, int max_samples)
@@ -155,6 +181,6 @@ namespace geopm
         m_max_keep_sample = max_samples;
         result = dcgmWatchFields(m_dcgm_handle, DCGM_GROUP_ALL_GPUS, m_field_group_id, m_update_freq,
                                  m_max_keep_age, m_max_keep_sample);
-        check_dcgm_result(result, "Error setting Max Samples");
+        check_dcgm_result(result, "Error setting Max Samples", __LINE__);
     }
 }
--- a/service/src/DCGMDevicePool.cpp
+++ b/service/src/DCGMDevicePool.cpp
@@ -56,7 +56,7 @@ namespace geopm
     {
         m_dcgm_field_ids[M_FIELD_ID_SM_ACTIVE] = DCGM_FI_PROF_SM_ACTIVE;
         m_dcgm_field_ids[M_FIELD_ID_SM_OCCUPANCY] = DCGM_FI_PROF_SM_OCCUPANCY;
-        m_dcgm_field_ids[M_FIELD_ID_SM_DRAM_ACTIVE] = DCGM_FI_PROF_DRAM_ACTIVE;
+        m_dcgm_field_ids[M_FIELD_ID_DRAM_ACTIVE] = DCGM_FI_PROF_DRAM_ACTIVE;
 
         dcgmReturn_t result;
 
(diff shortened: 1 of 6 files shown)


service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,160 @@
...
+        char *host_ip = "127.0.0.1";
+        result = dcgmConnect(host_ip, &m_dcgm_handle);
+        check_dcgm_result(result, "Error connecting to standalone DCGM instance");
+
+        //Check all devices are DCGM enabled
+        unsigned int dcgm_dev_id_list[DCGM_MAX_NUM_DEVICES];
+        result = dcgmGetAllSupportedDevices(m_dcgm_handle, dcgm_dev_id_list, &m_dcgm_dev_count);
+        check_dcgm_result(result, "Error fetching DCGM supported devices.");
+
+        for(int i = 0; i < m_dcgm_dev_count; ++i) {
```
dannosliwcd: There's a missing space at `for (`. Also some inconsistent spacing after `//` style comments, and I think this project uses pre-c++11-style nested templates (space between closing `> >`).

You can let a script handle formatting issues. I have a `clang-format` configuration file at  https://gist.github.com/dannosliwcd/4b0c593968820438889c6461395c0eca that I've been tweaking whenever I receive code style suggestions in code review.

service/src/DCGMDevicePoolImp.hpp
```diff
@@ -0,0 +1,75 @@
...
+            double m_max_keep_age;
+            int m_max_keep_sample;
+            int m_dcgm_dev_count;
+
+            dcgmHandle_t m_dcgm_handle;
+            dcgmFieldGrp_t m_field_group_id;
+
+            unsigned short m_dcgm_field_ids[M_FIELD_IDS];
+
+            std::vector<std::vector<dcgmFieldValue_v1>> m_dcgm_field_values;
```
dannosliwcd, line 72: Maybe add a comment about how this is indexed. I think the inner one is obviously indexed by value, but the outer vector's index isn't as clear without looking where it is already used.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,160 @@
...
+            std::vector<dcgmFieldValue_v1> field_values(sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]));
+            m_dcgm_field_values.push_back(field_values);;
```
dannosliwcd: Double `;;` here. Might get handled by clang-format but I'm not sure.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        //Setup Field Group
+        result = dcgmFieldGroupCreate(m_dcgm_handle, sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]), &m_dcgm_field_ids[0],
+                                      (char *)"geopm_fields", &m_field_group_id);
```
dannosliwcd: `(char *)"geopm_fields"` seems risky, toward hitting undefined behavior. DCGM docs say it's input only, but I'm not sure why they don't have it as const. Might be safer to put that string in a local array and send that array's address to this function.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,160 @@
...
+        result = dcgmWatchFields(m_dcgm_handle, DCGM_GROUP_ALL_GPUS, m_field_group_id, m_update_freq,
+                                 m_max_keep_age, m_max_keep_sample);
+
+        check_dcgm_result(result, "Error setting default watch field configuration.");
+    }
+
+    DCGMDevicePoolImp::~DCGMDevicePoolImp()
+    {
+        //Shutdown DCGM
+        dcgmStatusDestroy(NULL);
```
dannosliwcd: Does the field group also need to be destroyed?

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,160 @@
...
+        dcgmGroupDestroy(m_dcgm_handle, DCGM_GROUP_ALL_GPUS);
+    }
+
+    void DCGMDevicePoolImp::check_dcgm_result(const dcgmReturn_t result, const std::string error)
+    {
+        if (result != DCGM_ST_OK) {
+            throw Exception("DCGMDevicePool::" + std::string(__func__) + ": "
+                            + error + ": " + errorString(result),
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
dannosliwcd: All errors will show `check_dcgm_result`, line 109. Might pass through the function name and line number as arguments to this function.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    int DCGMDevicePoolImp::dcgm_device() const {
+        return m_dcgm_dev_count;
+    }
+
+    double DCGMDevicePoolImp::field_value(int accel_idx, int geopm_field_id) const {
+        double result;
```
dannosliwcd: Maybe throw an exception in debug builds if type is not DCGM_FT_DOUBLE

dannosliwcd: Maybe also a not-just-debug check that all fields are double at initialization time, if that's possible.

service/src/DCGMDevicePool.cpp
lhlawson: Yes!

service/src/DCGMDevicePool.cpp
lhlawson: Moved to doing prechecking of all fields at init time.  Removed the PCIE_* fields as they are NOT double formatted.  We can re-add them if needed and add additional checking for the appropriate data type.

lhlawson:
A few updates as part of the review process:
- Updated initialization to handle the case where a field group needs to be destroyed and recreated
- Updated initialization to verify all FIELD IDS are of types we handle.  
- Removed PCIE metrics

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+            DCGMDevicePool() = default;
+            virtual ~DCGMDevicePool() = default;
+
+            virtual int dcgm_device() const = 0;
+            virtual double field_value(int accel_idx, int geopm_field_id) const = 0;
+            virtual void update_field_value(int accel_idx) = 0;
+            virtual void field_update_rate(int accel_idx, int field_update_rate) = 0;
+            virtual void max_storage_time(int accel_idx, int max_storage_time) = 0;
+            virtual void max_samples(int accel_idx, int max_samples) = 0;
```
cmcantalupo: Please add doxygen comments.

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace geopm
+{
+    class DCGMDevicePool
+    {
+        public:
+            enum geopm_dcgm_field_ids_e {
+                M_SM_ACTIVE,
+                M_SM_OCCUPANCY,
+                M_DRAM_ACTIVE,
+                M_FIELD_IDS
```
cmcantalupo: Add doxygen

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef DCGMDEVICEPOOL_HPP_INCLUDE
+#define DCGMDEVICEPOOL_HPP_INCLUDE
+
+namespace geopm
+{
+    class DCGMDevicePool
```
cmcantalupo, line 43: Add doxygen

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,186 @@
...
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+    }
+
+    DCGMDevicePoolImp::~DCGMDevicePoolImp()
+    {
+        //Shutdown DCGM
+        dcgmStatusDestroy(NULL);
+        dcgmFieldGroupDestroy(m_dcgm_handle, m_field_group_id);
```
dannosliwcd: Github is not letting me comment on line 127. This comment goes with line 127.

`dcgmStatusDestroy` accepts a `dcgmStatus_t` argument, rather than a pointer. Also, there is no call to `dcgmStatusCreate`.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,186 @@
...
+            check_dcgm_result(result, "Error re-creating DCGM geopm_fields group.", __LINE__);
+        }
+        else {
+            check_dcgm_result(result, "Error creating DCGM geopm_fields group.", __LINE__);
+        }
+
+        // Note: Currently we are using dcgmWatchFields, but may transition to using
+        //       dcgmProfWatchFields at a later date
+        result = dcgmWatchFields(m_dcgm_handle, DCGM_GROUP_ALL_GPUS, m_field_group_id, m_update_freq,
+                                 m_max_keep_age, m_max_keep_sample);
```
dannosliwcd, line 108: If I understand the docs, I think this is requesting that DCGM cache up to m_max_keep_sample (100) most-recent values, internally sampled every 1000 usec, keeping up to m_max_keep_age (1) seconds (i.e., up to 1000, were it not for the explicit sample limit).

Does this mean that each call to `dcgmGetLatestValuesForFields` is ignoring 99 samples from each call to DCGM? Could we alternatively get sub-geopm-sample-rate information by using `dcgmGetValuesSince_v2` that gets an array of dcgm samples that were observed between geopm requests? This could be useful for observing spikes in non-monotonically-counting signals, like the utilization signals. In that case, we could report the average or peak utilization between geopm samples instead of the point-in-time utilization when geopm gets its samples.

The docs aren't clear if that's the difference between those functions in their long-form descriptions. This is all just assumptions based on their names.

lhlawson: In testing using dcgmGetValuesSince wasn't necessary after increasing default polling rate to ever 100ms.  Some experimental agents currently use a lower polling rate, but can use the controls provided through an upcoming IOGroup #1914 to set the rate as needed.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+
+    int DCGMDevicePoolImp::dcgm_device() const {
+        return m_dcgm_dev_count;
+    }
+
+    double DCGMDevicePoolImp::field_value(int accel_idx, int geopm_field_id) const {
+        double result;
+        result = m_dcgm_field_values.at(accel_idx).at(geopm_field_id).value.dbl;
```
dannosliwcd: Can also return NAN if .status is not DCGM_ST_OK. Hopefully doesn't happen now that we have the lower update rate, but might be good in case anyone changes the rate.

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+            //         has been called
+            /// @param [in] accel_idx The index indicating a particular
+            ///        accelerator.
+            /// @return The value for the specified field
+            virtual double sample_field_value(int accel_idx, int geopm_field_id) const = 0;
+
+            /// @brief Query DCGM for the latest value for an accelerator.
+            ///        Note that this is the last value DCGM cahced.  This
+            //         updates the DCGM device pool stored value that is provided
+            //         via the field_value function
```
lhlawson:
```suggestion
            //         via the sample_field_value function
```

dannosliwcd (dismissed):
No additional comments from me. I've been using a slightly-stale version of this for about a week, and it seems to be working as expected since the change to a 100ms sample period.

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+                M_SM_OCCUPANCY,
+                /*!
+                 * @brief Field ID associated with DCGM DRAM Active metrics
+                 */
+                M_DRAM_ACTIVE,
+                /*!
+                 * @brief Number of valid field ids
+                 */
+                M_FIELD_IDS
+            };
```
cmcantalupo:
```suggestion
            enum m_field_id_e {
                /*!
                 * @brief Field ID associated with DCGM SM Active metrics
                 */
                M_FIELD_ID_SM_ACTIVE,
                /*!
                 * @brief Field ID associated with SM Occupancy metrics
                 */
                M_FIELD_ID_SM_OCCUPANCY,
                /*!
                 * @brief Field ID associated with DCGM DRAM Active metrics
                 */
                M_FIELD_ID_DRAM_ACTIVE,
                /*!
                 * @brief Number of valid field ids
                 */
                M_NUM_FIELD_ID
            };
```

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+            /// @brief Number of accelerators that support DCGM on the platform.
+            /// @return Number of accelerators.
+            virtual int dcgm_device() const = 0;
+            /// @brief Get the value for the provided geopm_field_id.
+            ///        This value should not change unless update_field_value
+            //         has been called
+            /// @param [in] accel_idx The index indicating a particular
+            ///        accelerator.
+            /// @return The value for the specified field
+            virtual double sample_field_value(int accel_idx, int geopm_field_id) const = 0;
```
cmcantalupo:
```suggestion
            virtual double sample(int accel_idx, int field_id) const = 0;
```

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+                 */
+                M_FIELD_IDS
+            };
+
+            DCGMDevicePool() = default;
+            virtual ~DCGMDevicePool() = default;
+
+            /// @brief Number of accelerators that support DCGM on the platform.
+            /// @return Number of accelerators.
+            virtual int dcgm_device() const = 0;
```
cmcantalupo:
```suggestion
            virtual int num_device() const = 0;
```

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+            /// @return The value for the specified field
+            virtual double sample_field_value(int accel_idx, int geopm_field_id) const = 0;
+
+            /// @brief Query DCGM for the latest value for an accelerator.
+            ///        Note that this is the last value DCGM cahced.  This
+            //         updates the DCGM device pool stored value that is provided
+            //         via the sample_field_value function
+            /// @param [in] accel_idx The index indicating a particular
+            ///        accelerator.
+            virtual void update_field_value(int accel_idx) = 0;
```
cmcantalupo: update_field_value -> update?

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+            /// @brief Set field update rate for DCGM devices.
+            /// @param [in] field_update_rate DCGM update rate in microseconds.
+            virtual void field_update_rate(int field_update_rate) = 0;
```
cmcantalupo: field_update_rate -> update_rate?  What is this rate determining?  I'm really not sure what this method does.

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+            /// @brief Set maximum storage time for for DCGM devices.
+            /// @param [in] max_storage_time maximum storage time in seconds
+            virtual void max_storage_time(int max_storage_time) = 0;
```
cmcantalupo, line 98: I don't understand what "storage time" means, please define

service/src/DCGMDevicePool.hpp
```diff
@@ -0,0 +1,99 @@
...
+            /// @brief Set maximum samples to store for for DCGM devices.
+            /// @param [in] max_samples maximun number of samples to store
```
cmcantalupo, line 102: I'm not sure what this method does either

cmcantalupo (commented):
I have made some suggestions to the DCGMDevicePool header.  I'm sure these will impact the implementation in this PR and also the implementationf or the DGCMIOGroup.  Please address.  I would like better documentation for the APIs before reviewing the implementation.

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,188 @@
...
+        result = dcgmConnect(host_ip, &m_dcgm_handle);
+        check_dcgm_result(result, "Error connecting to standalone DCGM instance", __LINE__);
+
+        //Check all devices are DCGM enabled
+        unsigned int dcgm_dev_id_list[DCGM_MAX_NUM_DEVICES];
+        result = dcgmGetAllSupportedDevices(m_dcgm_handle, dcgm_dev_id_list, &m_dcgm_dev_count);
+        check_dcgm_result(result, "Error fetching DCGM supported devices.", __LINE__);
+
+        for (int i = 0; i < m_dcgm_dev_count; ++i) {
+            std::vector<dcgmFieldValue_v1> field_values(sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]));
```
cmcantalupo:
```suggestion
            std::vector<dcgmFieldValue_v1> field_values(M_NUM_FIELD_ID);
```

service/src/DCGMDevicePool.cpp
```diff
@@ -0,0 +1,188 @@
...
+            m_dcgm_field_values.push_back(field_values);
+        }
+
+        //Setup Field Group
+
+        char geopm_group[] = "geopm_field_group";
+        result = dcgmFieldGroupCreate(m_dcgm_handle, sizeof(m_dcgm_field_ids)/sizeof(m_dcgm_field_ids[0]), &m_dcgm_field_ids[0],
```
cmcantalupo: please get rid of all of these sizeof(...) / sizeof(...)

service/src/DCGMDevicePool.hpp
lhlawson: Storage time is defined by the [DCGM API](https://docs.nvidia.com/datacenter/dcgm/1.6/dcgm-api/group__DCGMAPI__FI.html#group__DCGMAPI__FI_1gd04fa06405af6165fd85310dee9eae2a), it's how long the DCGM process will store those samples before dropping them.

cmcantalupo (approved):
LGTM

Issue #1923, opened 2021-10-07
theta issues: geopmlaunch aprun and tutorial build broken


bgeltz:
**Describe the bug**
I tried to use ``geopmlaunch`` to run the tutorials and I expected it to work.
Instead the tutorials are producing various errors upon execution.

**GEOPM version**
Tip of dev branch; formerly the geopm-service branch.

**Expected behavior**
The tutorials to run and pass with a return code of 0.

**Actual behavior**
The tutorials do not run.

**Error messages produced**
```
[bgeltz@thetamom3 tutorial](geopm-service)$ GEOPM_LAUNCHER="aprun" ./tutorial_0.sh
<geopmpy.launcher> <geopm> geopmpy.launcher: CPU governor mismatch: All compute nodes do not have the same governor.
(['', 'aprun -N 1 -n 1 -- cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor', '', 'performance', 'Application 24622550 resources: utime ~0s, stime ~1s, Rss ~7076, inblocks ~468, outblocks ~0'])
```

Then once the above is resolved:

```
[bgeltz@thetamom3 tutorial](geopm-service)$ GEOPM_LAUNCHER="aprun" ./tutorial_0.sh
Warning: <geopm> geopmpy.launcher: User did not provide OMP_NUM_THREADS, using 60

OMP_PROC_BIND=true GEOPM_CTL=process GEOPM_PROFILE= GEOPM_REPORT=tutorial_0_report GEOPM_TRACE=tutorial_0_trace KMP_AFFINITY=disabled MV2_ENABLE_AFFINITY=0 KMP_WARNINGS=FALSE aprun -N 5 -n 10 --cpu-binding 1:4-18,68-82,132-146,196-210:19-33,83-97,147-161,211-225:34-48,98-112,162-176,226-240:49-63,113-127,177-191,241-255 -e LD_PRELOAD='libgeopm.so' -- ./tutorial_0

[...]

PR #1926, opened 2021-10-08
Resolve theta build/run issues

### LINKED ISSUE #1924: Fix for bug #1923
> estimate 0
Fixes #1923, a bug report from github issues.

There were changes in the ``launcher.py`` needed to resolve:
1. The check to ensure that the CPU governor is the same on all the compute nodes in an allocation needs to be updated such that is suppresses the informational job completion message at the end of aprun jobs.
2. Logic that attempts to re-quote quoted arguments before concatenating them needs to be considered for the usage of ``LD_PRELOAD='libgeopm.so'`` when the argv string is constructed.

Additionally, the build syntax in ``tutorial_build_intel.sh`` needs to be able to handle when ``${CC}`` is set to something with a dash argument.

Done criteria:

- [x] Details about the bug are located in the above bug report
- [ ] ~Regression test for the issue was added~
  - Not possible until we setup regular/automatic testing.
- [x] Fix has been verified
- [x] Tutorials compile
- [x] Tutorials run and produce a return code of 0.

bgeltz:
- Fixes #1924, #1923.

### DIFF
--- a/integration/config/build.sh
+++ b/integration/config/build.sh
@@ -128,7 +128,7 @@ build(){
 
 # Run the service build
 cd service
-CC=gcc CXX=g++ build "${GEOPM_SERVICE_CONFIG_OPTIONS}"
+CFLAGS= CXXFLAGS= CC=gcc CXX=g++ build "${GEOPM_SERVICE_CONFIG_OPTIONS}"
 
 # Run the base build
 cd ${GEOPM_SOURCE}
--- a/scripts/geopmpy/launcher.py
+++ b/scripts/geopmpy/launcher.py
@@ -482,7 +482,7 @@ class Launcher(object):
         if stderr_encoding is None:
             stderr_encoding = locale.getpreferredencoding()
 
-        argv_mod = [self.launcher_command()]
+        argv_mod = self.launcher_command()
         if self.is_override_enabled:
             argv_mod.extend(self.launcher_argv(False))
             argv_mod.extend(self.argv_unparsed)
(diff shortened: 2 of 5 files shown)


integration/config/build.sh
```diff
@@ -128,7 +128,7 @@ build(){
 
 # Run the service build
 cd service
-CC=gcc CXX=g++ build "${GEOPM_SERVICE_CONFIG_OPTIONS}"
```
cmcantalupo, line 131: I wonder if we need this change in geopm-service.spec.in?

scripts/geopmpy/launcher.py
```diff
@@ -992,7 +993,7 @@ def launcher_command(self):
         """
         Returns 'srun', the name of the SLURM MPI job launch application.
         """
-        return 'srun'
+        return ['srun']
```
cmcantalupo: It is not very intuitive that a method that is called "launcher_command" should return a list of strings with one element. At the very least, all of the documentation for these methods would have to change to describe the return type as a list, but I'm really not sure why the caller can't be responsible for creating the list that will become argv.  

The way this is resolved for all of the other launcher command line options is to add a new method to the launcher base that returns empty list e.g. quiet_option() and then for the AprunLauncher this would be implemented as return ['-q'].  I'm not saying the launcher class is beautiful, but that is how it works.

cmcantalupo (changes requested):
Thanks for the debug.  I know using Popen without a shell interpreter is pretty confusing, and that is the root cause of the LD_PRELOAD issue.  

I know the launcher is far from perfect, but I'll like to keep the implementation at least consistent.  Please add a new method to the launcher base class.  Thanks!

cmcantalupo (approved):

Issue #1861, opened 2021-09-15
Enable integration testing of the service itself


cmcantalupo:
> estimate 2

Parent issue: #1829
Relates to #1896 

Note that test_su_give_access.sh requires that it is run as the root user, and is otherwise expected to fail.  

- [x] Enable test user to install the geopm-service RPMs.
- [x] #1925
- [x] Execute the geopm/service/integration/test scripts and check return code.
- [x] Run these tests in the nightly set.

PR #1927, opened 2021-10-08
Create pytest test harness for bash integration tests

### LINKED ISSUE #1925: Add pytest test harness to execute each bash script and also geopm/service/integration/test/test_controller.py
Relates to #1861

cmcantalupo:
- Relates to #1861 issue, task from github issues
- Fixes #1925 change request from github issues

Adds a test harness that uses subTest feature to programmatically add tests that execute bash scripts that match test_*.sh pattern.


service/integration/test/TestBashExamples.py
```diff
@@ -0,0 +1,63 @@
...
+            with self.subTest(bash_example=os.path.basename(script)):
+                pid = subprocess.Popen([script],
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE)
+                out, err = pid.communicate()
+                output = f'\n\nSTDOUT:\n{out.decode("utf-8")}\n\nSTDERR:\n{err.decode("utf-8")}'
+                if pid.returncode != 0:
+                    self.assertEqual(0, pid.returncode, output)
+                else:
+                    sys.stdout.write(output)
```
cmcantalupo:
```suggestion
                self.assertEqual(0, pid.returncode, output)
                sys.stdout.write(output)
```

cmcantalupo: This seems better, but I will have to test.

service/integration/test/TestBashExamples.py
```diff
@@ -0,0 +1,63 @@
...
+    def test_all(self):
+        for script in self.bash_tests:
+            script_name = os.path.basename(script)
+            with self.subTest(bash_example=script_name):
+                pid = subprocess.Popen([script],
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE)
+                out, err = pid.communicate()
+                out = out.decode("utf-8")
+                err = err.decode("utf-8")
```
dannosliwcd: We can have popen do the encoding/decoding for us if we tell it to handle io in text mode. If our minimum version is 3.7, we can set text=True. If not, we can use universal_newlines=True (same purpose, just a non-obvious name).

dannosliwcd (dismissed):
LGTM. Minor suggestion doesn't affect correctness.

service/integration/test/TestBashExamples.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_all(self):
+        for script in self.bash_tests:
+            script_name = os.path.basename(script)
+            with self.subTest(bash_example=script_name):
+                pid = subprocess.Popen([script],
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE)
+                out, err = pid.communicate()
+                out = out.decode("utf-8")
```
bgeltz: We do this in the launcher:
```
stdout_encoding = locale.getpreferredencoding()
```
Then:
```
stdout.write(stdout_bytes.decode(encoding=stdout_encoding))
```

bgeltz (changes requested):

bgeltz (approved):

Issue #1829, opened 2021-09-01
As a developer of geopm I would like to have automated CI that covers the service


cmcantalupo:
Once we merge the geopm-service branch into the dev branch we want all of our existing automation to continue to work.  It would be great to have automation that also covers the new testing in the service directory.

- Integration testing
+ Run integration tests with msr-safe but without service, and run with service but without msr-safe.

PR #1929, opened 2021-10-09
Interogate procfs for capabilities rather than restricting DBus APIs to root

### LINKED ISSUE #1896: Modifying the access control lists for geopm service requires actual root user id
Use of sudo to modify the access lists for the geopm service does not work.  The service as implemented currently has a requirement that the user is actually root.  We may be able to work around this issue by checking capabilities of the process unless the DBus configuration is also a problem.

Known issues that will result from merging solution:

- Cannot easily automate tests for features that require that they are run as root
- System administration is typically done with effective user id, not actual user id.

cmcantalupo:
- Relates to #1829 story from github issues
- Fixes #1896 change request from github issues


service/geopmdpy/service.py
```diff
@@ -830,3 +838,16 @@ def _get_pid(self, call_info):
         """
         unique_name = call_info['sender']
         return self._dbus_proxy.GetConnectionUnixProcessID(unique_name)
+
+    def _check_cap_sys_admin(self, call_info, api_name):
```
dannosliwcd, line 842: Does this also work if the old workflow is followed, running as root?

cmcantalupo: Yes, the root user shows the CAP_SYS_ADMIN bit in the procfs entry.

service/geopmdpy/access.py
```diff
@@ -97,7 +97,7 @@ def set_group_signals(self, group, signals):
         try:
             self._geopm_proxy.PlatformSetGroupAccess(group, signals, current_controls)
         except DBusError as ee:
-            raise RuntimeError('Failed to set group signal access list, request must be made by root user') from ee
+            raise RuntimeError('Failed to set group signal access list, request must be made by user with CAP_SYS_ADMIN') from ee
```
cmcantalupo: Tell the user how to get CAP_SYS_ADMIN, i.e. run with sudo or root.

service/integration/test/test_su_give_access.sh
```diff
@@ -64,19 +64,15 @@ test_error() {
...
 # SAVE INITIAL ACCESS SETTINGS
 geopmaccess > ${SAVE_SIGNALS}
 geopmaccess -c > ${SAVE_CONTROLS}
 
 # REMOVE CONTROL FROM ACCESS LIST FOR READING AND WRITING
 grep -v ${CONTROL} ${SAVE_SIGNALS} > ${TEST_SIGNALS}
 grep -v ${CONTROL} ${SAVE_CONTROLS} > ${TEST_CONTROLS}
-geopmaccess -w < ${TEST_SIGNALS}
-geopmaccess -w -c < ${TEST_CONTROLS}
+sudo geopmaccess -w < ${TEST_SIGNALS}
```
dannosliwcd: Would this result in a password prompt mid-test? Is that able to be managed by our test automation?

cmcantalupo: Yes...fixed by adding -n.  Then we are going to put geopmaccess in the list of commands that the test user can sudo.

service/geopmdpy/service.py
```diff
@@ -830,3 +838,16 @@ def _get_pid(self, call_info):
...
+        has_admin = False
+        cap = 0
+        cap_sys_admin = 0x00200000
```
bgeltz, line 845: A pointer to the manual or hyperlink to where this is defined would be great.

bgeltz (dismissed):
LGTM.  I won't know if this is all that's needed until I start working it into the nightlies.  I'm fine to open a new PR with those changes if you'd prefer to merge this now.

bgeltz (approved):

PR #1932, opened 2021-10-11
Get PMC MSR bit width from CPUID

### LINKED ISSUE #176: Prototype setting programmable counters in independent test

### LINKED ISSUE #1931: MSR::IA32_PMC0:PERFCTR is not always the right width
**Describe the bug**
I tried to compare the fixed-counter and programmable-counter variants of instructions retired. The fixed-counter variant produced consistent differences across a sleep, but the programmable counter often reported negative differences. This happens because GEOPM discards upper bits that are in use for some architectures.

**Possible Fix**
SDM vol 3b, section 18 specifies that we should read `CPUID.0AH:EAX[23:16]` to learn how many bits are available for programmable counters (48 on our skylake nodes). service/src/msr_data_arch.cpp is currently configured to read 31 bits for `MSR::IA32_PMC0:PERFCTR`. 

MSRIOGroup.cpp already has a workflow for MSRs with bit width that depends on CPUID. See the handlers for `QM_EVTSEL:RMID` and `PQR_ASSOC:RMID` for reference.

**Steps to reproduce**
```
# Read the fixed counter for retired instructions.
for i in {1..5}; do i1=$(geopmread INSTRUCTIONS_RETIRED board 0) && sleep 2 && i2=$(geopmread INSTRUCTIONS_RETIRED board 0) && echo "FIXED COUNTER Instructions: $((i2-i1))"; done
```
```
# Enable INST_RETIRED.ANY_P. See https://download.01.org/perfmon/SKX/skylakex_core_v1.24.json
geopmwrite MSR::IA32_PERFEVTSEL0:EVENT_SELECT board 0 0xc0
geopmwrite MSR::IA32_PERFEVTSEL0:UMASK board 0 0x00
geopmwrite MSR::IA32_PERFEVTSEL0:USR board 0 1
geopmwrite MSR::IA32_PERFEVTSEL0:OS board 0 1
geopmwrite MSR::IA32_PERFEVTSEL0:EN board 0 1
geopmwrite  MSR::PERF_GLOBAL_CTRL:EN_PMC0 board 0 1
# Read the PMC for retired instructions. Should match the measurements from before.
for i in {1..5}; do i1=$(geopmread MSR::IA32_PMC0:PERFCTR board 0) && sleep 2 && i2=$(geopmread MSR::IA32_PMC0:PERFCTR board 0) && echo "Programmable Counter Instructions: $((i2-i1))"; done
```

**GEOPM version**
1.1.0+dev1567gc9a848c8

dannosliwcd:
- Resolves #176 by adding an integration test for the instructions retired programmable counter.
- Fixes #1931 by using width of MSR::IA32_PMC0:PERFCT that is reported by CPUID.

### DIFF
--- a/service/src/MSRIOGroup.cpp
+++ b/service/src/MSRIOGroup.cpp
@@ -100,6 +100,7 @@ namespace geopm
         , m_time_zero(std::make_shared<geopm_time_s>(time_zero()))
         , m_time_batch(std::make_shared<double>(NAN))
         , m_rdt_info(get_rdt_info())
+        , m_pmc_bit_width(get_pmc_bit_width())
         , m_derivative_window(8)
         , m_sleep_time(0.005)  // 5000 us
     {
@@ -702,6 +703,21 @@ namespace geopm
         };
 
         return rdt;
+    }
+
+    uint32_t MSRIOGroup::get_pmc_bit_width(void)
+    {
+        uint32_t leaf, subleaf = 0;
+        uint32_t eax, ebx, ecx, edx = 0;
+
+        leaf = 0x0A;
+        subleaf = 0;
+
+        __cpuid_count(leaf, subleaf, eax, ebx, ecx, edx);
+
+        // SDM vol 3b, section 18 specifies where to find how many PMC bits are
+        // available
+        return (eax >> 16) & 0xff;
     }
 
     void MSRIOGroup::register_signal_alias(const std::string &signal_name,
@@ -1319,6 +1335,20 @@ namespace geopm
                         throw Exception(except.str(), GEOPM_ERROR_INVALID, __FILE__,
                                         __LINE__);
                     }
+                }
+
+                if (string_begins_with(msr_field_name, "IA32_PMC") &&
+                    string_ends_with(msr_field_name, ":PERFCTR")) {
+                    if (m_pmc_bit_width > 0) {
+                        end_bit = begin_bit + m_pmc_bit_width - 1;
+                    }
+#ifdef GEOPM_DEBUG
+                    else {
+                        std::cerr << "Warning: <geopm> CPUID specified 0 bits for "
+                                  << msr_field_name << "; using the default width: "
+                                  << (end_bit - begin_bit + 1) << std::endl;
+                    }
+#endif
                 }
 
                 add_msr_field_signal(msr_name, sig_ctl_name, domain_type,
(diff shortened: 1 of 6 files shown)


service/src/msr_data_arch.cpp
```diff
@@ -814,7 +814,7 @@ namespace geopm
             "fields": {
                 "PERFCTR": {
                     "begin_bit": 0,
-                    "end_bit":   31,
+                    "end_bit":   47,
```
lhlawson, line 817: Is there any way to communicate to the end-user that this value won't be used?

service/src/MSRIOGroup.cpp
```diff
@@ -1320,6 +1336,12 @@ namespace geopm
                     }
                 }
 
+                if (string_begins_with(msr_field_name, "IA32_PMC") &&
+                    string_ends_with(msr_field_name, ":PERFCTR") &&
+                    m_pmc_bit_width > 0) {
+                    end_bit = begin_bit + m_pmc_bit_width - 1;
+                }
```
lhlawson: Should we emit a specific warning in debug mode if m_pmc_bit_wdith == 0 to indicate we're using the JSON value?

lhlawson (commented):
LGTM overall, 2 specific questions on communication to the user.

service/src/msr_data_arch.cpp
cmcantalupo: You can't comment in json, and the "end-user" shouldn't really be poking around in this data.  We could put a comment about this fact near the code that overwrites the value?

cmcantalupo (commented):
LGTM.  Seems like @lhlawson has some feedback which can be addressed.

service/src/MSRIOGroup.cpp
dannosliwcd: Good call, will do. That's what was causing the GitHub Actions failure, by the way (CPUID was saying we get 0 bits, and we don't support 0-bit MSRs)

service/src/MSRIOGroup.cpp
```diff
@@ -1320,6 +1336,19 @@ namespace geopm
...
+#ifdef GEOPM_DEBUG
+                else {
+                    std::cerr << "Warning: <geopm> CPUID specified 0 bits for "
+                              << msr_field_name << "; using the default width: "
+                              << (end_bit - begin_bit + 1) << std::endl;
```
dannosliwcd: This needs to be inside the begins/endswith check.

cmcantalupo (dismissed):
LGTM

cmcantalupo (approved):
Thanks!

Issue #1933, opened 2021-10-11
As a researcher using GEOPM I would like to run experiments on a GPU system using ParRes Kernels


scoumeri:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I need to setup explicit bash scripts to run GPU dgemm and nstream kernels and need to post-process the results to use the GEOPM analysis infrastructure scripts.

**Describe the solution you'd like**
Have ParRes Kernels integrated into GEOPM.  Have a build script to download and build ParRes cublas dgemm and nstream kernels.  Have scripts setup to enable GEOPM gpu integration/experiments.    

**Describe alternatives you've considered**
Download and build ParRes Kernel from ParRes repository external to GEOPM. Use standalone bash scripts to call geopmlaunch outside of the GEOPM infrastructure.

**Additional context**

_Note_:

Prior to creating a feature request, please see the documentation at
either of the following links:

https://github.com/geopm/geopm/blob/dev/CONTRIBUTING.rst#feature-request

https://geopm.github.io/geopmdpy/contrib.html#feature-request

about the process of developing a GEOPM feature.

PR #1936, opened 2021-10-11
Parres pr dgemm nstream

### LINKED ISSUE #1921: Introduce new Python class: parres.py
> estimate 1
Parent [feature/story]: #XXXX

Done criteria:

- [ ] build.sh properly builds parres kernels of interest
- [ ] build.sh points to a specific release or commit of underlying workload
- [ ] geopm specific patches exist and apply properly
- [ ] Parres workloads run on CUDA enabled system
- [ ] [License review](https://github.com/ParRes/Kernels/blob/default/COPYING) 
- [ ] The new class is fully compliant with Python 3.  Python 2 is not
      supported.

scoumeri:
- Relates to #1933 [bug report/feature request/story] from github issues
- Fixes #1921 change request from github issues.

[Brief description of the change]

Adding ParRes Kernel (specifically dgemm and nstream) to the integration environment for use with gpus.


integration/apps/parres/Makefile.mk
```diff
@@ -0,0 +1,38 @@
...
+#  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+EXTRA_DIST += integration/apps/parres/0001-Added-nstream-cublas-mpi-init-finalize-for-GEOPM.patch \
+	      integration/apps/parres/__init__.py \
+              integration/apps/parres/build.sh \
+              integration/apps/parres/README.md \
+	      integration/apps/parres/parres.py \
```
lhlawson: wsp

scoumeri: Fixed this

integration/apps/parres/parres.py
```diff
@@ -0,0 +1,321 @@
...
+                
+        benchmark_dir = os.path.dirname(os.path.abspath(__file__))
+
+        affinity_disable = False
+        if mpi_pin_list is not None:
+            affinity_disable = True
+
+        elif mpi_pin_default:
+            affinity_disable = True
+            mpi_pin_list = "1,2,3,21,22"
```
cmcantalupo: This data seems like it is architecture specific, can it be derived from machine file?

integration/apps/parres/parres.py
```diff
@@ -0,0 +1,321 @@
...
+                
+        benchmark_dir = os.path.dirname(os.path.abspath(__file__))
+
+        affinity_disable = False
+        if mpi_pin_list is not None:
+            affinity_disable = True
+
+        elif mpi_pin_default:
+            affinity_disable = True
+            mpi_pin_list = "1,21"
```
cmcantalupo: Derive from machine file?

integration/apps/parres/parres.py
```diff
@@ -0,0 +1,321 @@
...
+    def get_bash_exec_path(self):
+        return self.exe_path
+
+    def get_bash_exec_args(self):
+        return self.app_params
+
+    def get_bash_setup_commands(self):
+        setup_commands = textwrap.dedent('''
+        ulimit -s unlimited
+        export I_MPI_FABRICS=shm
```
cmcantalupo, line 288: Is this for single node only runs? If so, is this enforced somewhere?

cmcantalupo (commented):
Thanks for the contributions.  Couple of questions about the affinity settings and the intel specific MPI fabric setting.

integration/apps/parres/parres.py
scoumeri: This is a remnant from the Hacc script (which I wrote first but will PR later).  That had a large affinity mapping which was painful to specify (but didn't actually seem to make a difference on saruman). I just left the capability in for dgemm and nstream, but don't necessarily need to use it.  

I can get rid of the mpi_pin_default.  I can also get rid of mpi_pin_list and make the user set the env variable outside of the run.  I didn't automatically set the geopm-affinity-disable, so that right now has to happen on the command line anyway.

Lowren, any preference?

integration/apps/parres/parres.py
scoumeri: Same as above comment.

integration/apps/parres/parres.py
scoumeri: Ok, discussed with Lowren.  I'll drop the options.

integration/apps/parres/parres.py
scoumeri: This has all been run on saruman which is 1 node.  According to Lowren: As is, dgemm may work for multiple nodes, but nstream probably will only work for single node.  

Will keep it to single node and I'll add a check to enforce this.

integration/apps/parres/parres.py
scoumeri: I've updated the code to drop the pinning options.

integration/apps/parres/parres.py
scoumeri: I've updated the code to drop the pinning options.

integration/apps/parres/parres.py
scoumeri: Added a check to make sure it's running on a single node.

integration/apps/parres/parres.experiment.setup
```diff
@@ -0,0 +1,37 @@
...
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+#clears application clock settings
+nvidia-smi -rac
+#clears clock settings
+nvidia-smi -rgc
+#sets power limit
+nvidia-smi -pl 300
```
lhlawson: RAC is a user level control only if root allows it.  
RGC and PL are root level controls.  May be worthwhile to call this out.

integration/apps/parres/parres.py
```diff
@@ -0,0 +1,296 @@
...
+            subprocess.call(self._parres_exp_setup, shell=True)
+
+        
+    
+    def experiment_teardown(self, output_dir):
+        if not self._parres_teardown is None:
+            os.chmod(self._parres_teardown, 0o755)
+            subprocess.call(self._parres_teardown, shell=True)
+
+
```
lhlawson, line 194:
```suggestion
def parse_fom(self, log_path):
        result = None
        key = 'AVG Rate (MF/s): '
        with open(log_path) as fid:
            for line in fid.readlines():
                if key in line:
                    result = line.split(': ')[1]
                    result = float(result.split(' ')[0])
                    break
        return result
```

lhlawson: Suggestion to add figure of merit.  Untested.

integration/apps/parres/parres.py
```diff
@@ -0,0 +1,296 @@
...
```
lhlawson, line 305:
```suggestion
```suggestion
def parse_fom(self, log_path):
        result = None
        key = 'Rate (MB/s): '
        with open(log_path) as fid:
            for line in fid.readlines():
                if key in line:
                    result = line.split(': ')[1]
                    result = float(result.split(' ')[0])
                    break
        return result
```
```

integration/apps/parres/parres.experiment.setup
scoumeri: Added comment in file about needing root level access.

integration/apps/parres/parres.py
scoumeri: Added FOM code.

integration/apps/parres/parres.py
scoumeri: Added FOM code.

integration/apps/parres/parres.py
scoumeri: Trying to resolve conversation.

scoumeri:
Made additional change to switch nstream from cublas to cuda as it seems to have better performance.

lhlawson:
> 
> 
> Made additional change to switch nstream from cublas to cuda as it seems to have better performance.

LGTM, using the patch to add a new file should work (could do a cp & modify patch instead).

cmcantalupo (approved):
Thanks @scoumeri

PR #1939, opened 2021-10-12
Fix the path for test logs printed in GitHub Actions

### LINKED ISSUE #1938: GitHub actions do not show failure logs for failed service gtests
**Describe the bug**
I committed a change that failed some gtests on GitHub actions, but it did not print the expected failure logs. The build.yml is calling `cat test/service/gtest_links/*.log || true`, which does not exist. It should call `cat service/test/gtest_links/*.log || true` instead.

dannosliwcd:
- Fixes #1938
- The GitHub build workflow attempts to cat some log files after a failure is encountered. One of the file paths has its directories reversed. This change puts the reversed directories back in order.


cmcantalupo (approved):
Thanks!

Issue #1941, opened 2021-10-12
As a NVIDIA HW User I would like to be able to query minimum and maximum frequencies


lhlawson:
**Is your feature request related to a problem? Please describe.**
Currently there's no way in GEOPM to query min & max NVIDIA GPU Frequency. 

**Describe the solution you'd like**
The addition of NVML::FREQUENCY_MIN and NVML::FREQUENCY_MAX, which return the minimum and maximum allowed frequency for the associated GPU.

**Describe alternatives you've considered**
nvidia-smi provides this functionality, but requires the user to hardcode this information in an agent (or pass it in via a policy) .

**Additional context**
 This feature already exists for LevelZero as LEVELZERO::FREQUENCY_MIN_GPU

PR #1944, opened 2021-10-12
Addition of NVML Minimum and Maximum Frequency Querying

### LINKED ISSUE #1942: Added frequency_supported_sm methods to NVMLIOGroup C++ class
> estimate 1
Parent [feature/story]: #1941 

Done criteria:
- [ ] Mock objects derived from class have methods added
- [ ] New public methods have doxygen comments describing
      behavior, inputs, outputs and exceptions raised

lhlawson:
- Relates to #1941 
- Fixes #1942 

Adds a devicepool call for supported SM frequencies at the current Memory frequency (queried automatically).  
Adds IO Group usage of the above to provide a minimum and maximum frequency value.  

This approach may not function as expected if multiple memory frequencies are supported.  
Remaining Opens: 
- [x] Testing of the new IO Group signals
- [ ] Handling of multiple memory frequencies, possibly through adding a new signal that re-polls for min and maximum at the current memory frequency.

### DIFF
--- a/service/src/NVMLDevicePool.cpp
+++ b/service/src/NVMLDevicePool.cpp
@@ -150,6 +150,38 @@ namespace geopm
         return (uint64_t)result;
     }
 
+    std::vector<double> NVMLDevicePoolImp::frequency_supported_sm(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        unsigned int count = 100;
+        unsigned int *supported_freqs;
+        supported_freqs = new unsigned int[count];
+        std::vector<double> result;
+
+        nvmlReturn_t nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
+                                                                        frequency_status_mem(accel_idx),
+                                                                        &count, supported_freqs);
+        if (nvml_result == NVML_SUCCESS) {
+            for (unsigned int i = 0; i<count; i++) {
+                result.push_back(supported_freqs[i]);
+            }
+        }
+        else if (nvml_result == NVML_ERROR_INSUFFICIENT_SIZE) {
+            supported_freqs = new unsigned int[count];
+            nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
+                                                               frequency_status_mem(accel_idx),
+                                                               &count, supported_freqs);
+            for (unsigned int i = 0; i<count; i++) {
+                result.push_back(supported_freqs[i]);
+            }
+        }
+        check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
+                          ": NVML failed to get SM Frequency for accelerator " +
+                          std::to_string(accel_idx) + ".", __LINE__);
+        return result;
+    }
+
+
     uint64_t NVMLDevicePoolImp::utilization(int accel_idx) const
     {
         check_accel_range(accel_idx);
@@ -298,7 +330,7 @@ namespace geopm
         std::vector<int> result;
         unsigned int temp = M_MAX_CONTEXTS;
 
-                nvmlProcessInfo_t *process_info_list;
+        nvmlProcessInfo_t *process_info_list;
         process_info_list = new nvmlProcessInfo_t[temp];
 
         nvmlReturn_t nvml_result = nvmlDeviceGetComputeRunningProcesses(m_nvml_device[accel_idx], &temp, &process_info_list[0]);
--- a/service/src/NVMLDevicePool.cpp
+++ b/service/src/NVMLDevicePool.cpp
@@ -58,6 +58,7 @@ namespace geopm
 
     NVMLDevicePoolImp::NVMLDevicePoolImp(const int num_cpu)
         : M_MAX_CONTEXTS(64)
+        , M_MAX_FREQUENCIES(200)
         , M_NUM_CPU(num_cpu)
     {
         nvmlReturn_t nvml_result;
@@ -153,15 +154,15 @@ namespace geopm
     std::vector<unsigned int> NVMLDevicePoolImp::frequency_supported_sm(int accel_idx) const
     {
         check_accel_range(accel_idx);
-        unsigned int count = 100;
+        unsigned int count = M_MAX_FREQUENCIES;
         std::vector<unsigned int> supported_freqs(count);
 
         nvmlReturn_t nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
                                                                         frequency_status_mem(accel_idx),
                                                                         &count, supported_freqs.data());
+        supported_freqs.resize(count);
 
         if (nvml_result == NVML_ERROR_INSUFFICIENT_SIZE) {
-            supported_freqs.resize(count);
             nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
                                                                frequency_status_mem(accel_idx),
                                                                &count, supported_freqs.data());
--- a/service/src/NVMLDevicePool.cpp
+++ b/service/src/NVMLDevicePool.cpp
@@ -150,35 +150,27 @@ namespace geopm
         return (uint64_t)result;
     }
 
-    std::vector<double> NVMLDevicePoolImp::frequency_supported_sm(int accel_idx) const
+    std::vector<unsigned int> NVMLDevicePoolImp::frequency_supported_sm(int accel_idx) const
     {
         check_accel_range(accel_idx);
         unsigned int count = 100;
-        unsigned int *supported_freqs;
-        supported_freqs = new unsigned int[count];
-        std::vector<double> result;
+        std::vector<unsigned int> supported_freqs(count);
 
         nvmlReturn_t nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
                                                                         frequency_status_mem(accel_idx),
-                                                                        &count, supported_freqs);
-        if (nvml_result == NVML_SUCCESS) {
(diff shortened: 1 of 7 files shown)


service/src/NVMLDevicePool.cpp
```diff
@@ -150,6 +150,38 @@ namespace geopm
         return (uint64_t)result;
     }
 
+    std::vector<double> NVMLDevicePoolImp::frequency_supported_sm(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        unsigned int count = 100;
+        unsigned int *supported_freqs;
+        supported_freqs = new unsigned int[count];
```
dannosliwcd: Does NVML take ownership of this pointer? If yes, add a comment noting that. If no, this and line 170 result in memory leaks.

If nvml does not take ownership, you can replace with a `std::vector<unsigned int>`. If using a vector, you can use `supported_freqs.reserve(count)` in place of the re-allocation and `supported_freqs.data()` in place of the array pointer. A vector copy constructor can convert to your `result`'s double type, as in `result = std::vector<double>(supported_freqs.begin(), supported_freqs.end())`

service/src/NVMLDevicePool.cpp
```diff
@@ -150,6 +150,38 @@ namespace geopm
...
+        std::vector<double> result;
+
+        nvmlReturn_t nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
+                                                                        frequency_status_mem(accel_idx),
+                                                                        &count, supported_freqs);
+        if (nvml_result == NVML_SUCCESS) {
```
dannosliwcd: Can avoid the duplicate result assignment code with something like:

try to load the array, update nvml_result
if result is insufficient size: retry load, update nvml_result
if result is success, copy to the result vector and return (or skip the temporary assign before returning, as in `check_nvml_result(...); return std::vector<double>(supported_freqs.begin(), supported_freqs.end())`

service/src/NVMLIOGroup.cpp
```diff
@@ -155,6 +156,20 @@ namespace geopm
...
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::max,
+                                  string_format_double
+                                  }},
+                              {"NVML::FREQUENCY_MIN", {
+                                  "Streaming multiprocessor Minimum frequency in hertz",
+                                  {},
+                                  GEOPM_DOMAIN_BOARD_ACCELERATOR,
+                                  Agg::max,
```
dannosliwcd: Seems like either this or the other should be `Agg::min`, depending on whether you want the parent domain to report the union or the intersection of available settings.

lhlawson: I've moved them to be expect_same, as we don't plan on supporting systems with a mixture of GPU types with varying min/max ranges.

service/src/NVMLDevicePool.cpp
lhlawson: https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g673c4a61192307ffc47a764071cfac00

NVML returns the supported clocks into this data structure.

service/src/NVMLDevicePool.cpp
```diff
@@ -150,6 +150,30 @@ namespace geopm
         return (uint64_t)result;
     }
 
+    std::vector<unsigned int> NVMLDevicePoolImp::frequency_supported_sm(int accel_idx) const
+    {
+        check_accel_range(accel_idx);
+        unsigned int count = 100;
```
cmcantalupo: Where is this magic number coming from? Prefer if it was a `static const` member variable.  Alternately can we query the return size by passing count as 0 and always call the function twice?

service/src/NVMLDevicePool.cpp
```diff
@@ -150,6 +150,30 @@ namespace geopm
...
+        std::vector<unsigned int> supported_freqs(count);
+
+        nvmlReturn_t nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
+                                                                        frequency_status_mem(accel_idx),
+                                                                        &count, supported_freqs.data());
+
+        if (nvml_result == NVML_ERROR_INSUFFICIENT_SIZE) {
+            supported_freqs.resize(count);
```
cmcantalupo, line 164: is the intent to increase count?  It is still 100.

service/src/NVMLDevicePool.cpp
```diff
@@ -150,6 +150,30 @@ namespace geopm
...
+            nvml_result = nvmlDeviceGetSupportedGraphicsClocks(m_nvml_device.at(accel_idx),
+                                                               frequency_status_mem(accel_idx),
+                                                               &count, supported_freqs.data());
+        }
+        check_nvml_result(nvml_result, GEOPM_ERROR_RUNTIME, "NVMLDevicePool::" + std::string(__func__) +
+                          ": NVML failed to get SM Frequency for accelerator " +
+                          std::to_string(accel_idx) + ".", __LINE__);
+
+        return supported_freqs;
```
cmcantalupo, line 174: don't we want to
```
supported_freqs.resize(count);
```
before returning a length 100 vector?

service/src/NVMLIOGroup.cpp
```diff
@@ -204,6 +219,17 @@ namespace geopm
...
+
+        for (int domain_idx = 0; domain_idx < m_platform_topo.num_domain(GEOPM_DOMAIN_BOARD_ACCELERATOR); ++domain_idx) {
+            std::vector<unsigned int> supported_frequency = m_nvml_device_pool.frequency_supported_sm(domain_idx);
+            if (supported_frequency.size() == 0) {
+                throw Exception("NVMLIOGroup::" + std::string(__func__) +
+                                ": No supported frequencies found for accelerator " + std::to_string(domain_idx),
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+            m_supported_freq.push_back(supported_frequency);
+            std::sort(m_supported_freq.at(domain_idx).begin(), m_supported_freq.at(domain_idx).end());
```
cmcantalupo: why not sort prior to the push_back?

service/src/NVMLIOGroup.cpp
```diff
@@ -480,6 +506,16 @@ namespace geopm
         if (signal_name == "NVML::FREQUENCY" || signal_name == "FREQUENCY_ACCELERATOR") {
             result = (double) m_nvml_device_pool.frequency_status_sm(domain_idx)*1e6;
         }
+        else if (signal_name == "NVML::FREQUENCY_MIN") {
+            if (m_supported_freq.at(domain_idx).size() != 0) {
+                result = (double) m_supported_freq.at(domain_idx).front()*1e6;
```
cmcantalupo:
```suggestion
                result = 1e6 * m_supported_freq.at(domain_idx).front();
```

service/src/NVMLIOGroup.cpp
```diff
@@ -480,6 +506,16 @@ namespace geopm
...
+            }
+        }
+        else if (signal_name == "NVML::FREQUENCY_MAX") {
+            if (m_supported_freq.at(domain_idx).size() != 0) {
+                result = (double) m_supported_freq.at(domain_idx).back()*1e6;
```
cmcantalupo:
```suggestion
                result = 1e6 * m_supported_freq.at(domain_idx).back();
```

service/src/NVMLIOGroup.hpp
```diff
@@ -88,6 +88,7 @@ namespace geopm
             const NVMLDevicePool &m_nvml_device_pool;
             bool m_is_batch_read;
             std::vector<uint64_t> m_initial_power_limit;
+            std::vector<std::vector<unsigned int>> m_supported_freq;
```
cmcantalupo:
```suggestion
            std::vector<std::vector<unsigned int> > m_supported_freq;
```

service/test/NVMLIOGroupTest.cpp
```diff
@@ -265,21 +272,27 @@ TEST_F(NVMLIOGroupTest, read_signal)
...
     NVMLIOGroup nvml_io(*m_platform_topo, *m_device_pool);
+    std::sort(mock_supported_freq.begin(), mock_supported_freq.end());
 
     for (int accel_idx = 0; accel_idx < num_accelerator; ++accel_idx) {
         double frequency = nvml_io.read_signal("NVML::FREQUENCY", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx);
         double frequency_alias = nvml_io.read_signal("FREQUENCY_ACCELERATOR", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx);
         EXPECT_DOUBLE_EQ(frequency, frequency_alias);
         EXPECT_DOUBLE_EQ(frequency, mock_freq.at(accel_idx)*1e6);
 
+        double frequency_min = nvml_io.read_signal("NVML::FREQUENCY_MIN", GEOPM_DOMAIN_BOARD_ACCELERATOR, accel_idx);
```
cmcantalupo, line 290: please try to add spaces around binary operators in this file.

cmcantalupo (commented):
Thanks for the addition.  Some comment inline

service/src/NVMLDevicePool.cpp
lhlawson: It's just that, a magic number to start with that is a reasonable size, similar to M_MAX_CONTEXTS.  

We can query the return size by passing count as 0, but this will result in always calling the NVML library twice, which can be time consuming (though we have not check the time of this particular function call).

service/src/NVMLDevicePool.cpp
lhlawson: Re-arrange the code a bit so only a single resize should be needed.

cmcantalupo (approved):

cmcantalupo (approved):
LGTM

PR #1953, opened 2021-10-13
Enable exception handling when acclerator support fails

### LINKED ISSUE #1952: If a GPU driver is not present after building with --enable-nvml/levelzero GEOPM will not function
**Describe the bug**
Building on a head node with the appropriate GPU libraries but without GPUs results in an error such as:
```
~/geopm_public_lhlawson> ./geopmread
  what():  <geopm> Runtime error: LevelZero::LevelZeroImp: LevelZero Driver failed to initialize. Level Zero Error: ZE_RESULT_ERROR_UNINITIALIZED: at geopm/src/LevelZero.cpp:62                                                                                                 Aborted (core dumped)
```
**GEOPM version**
1.1.0+dev1567gc9a848c8

**Expected behavior**
The accelerator topo should handle the exception and default to the behavior of the Null accelerator topo (i.e. no GPUs)

**Actual behavior**
Throw from LevelZero.cpp (or NVMLDevicePool.cpp) leading to a core dump.

**Error messages produced**
```
~/geopm_public_lhlawson> ./geopmread
  what():  <geopm> Runtime error: LevelZero::LevelZeroImp: LevelZero Driver failed to initialize. Level Zero Error: ZE_RESULT_ERROR_UNINITIALIZED: at geopm/src/LevelZero.cpp:62                                                                                                 Aborted (core dumped)
```

cmcantalupo:
- If LevelZero or NVML are enabled at configure time, but not
  at run time, PlatformIO will fail to load.
- Fixes #1952

### DIFF
--- a/service/src/AcceleratorTopo.cpp
+++ b/service/src/AcceleratorTopo.cpp
@@ -32,10 +32,7 @@
 
 #include "config.h"
 
-#include <fstream>
-#include <iostream>
-#include <sstream>
-#include <string>
+#include <geopm/Helper.hpp>
 
 #include "geopm/Exception.hpp"
 #include "AcceleratorTopoNull.hpp"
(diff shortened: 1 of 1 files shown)


service/src/AcceleratorTopo.cpp
```diff
@@ -47,13 +44,19 @@ namespace geopm
 {
     const AcceleratorTopo &accelerator_topo(void)
     {
+        std::unique_ptr<AcceleratorTopo> instance;
```
cmcantalupo: Um... this is not a singlton.

service/src/AcceleratorTopo.cpp
```diff
@@ -45,15 +42,28 @@
...
 namespace geopm
 {
-    const AcceleratorTopo &accelerator_topo(void)
+    static std::unique_ptr<AcceleratorTopo> make_unique_accelerator_topo(void)
     {
+        std::unique_ptr<AcceleratorTopo> result;
+        try {
 #ifdef GEOPM_ENABLE_NVML
-        static NVMLAcceleratorTopo instance;
+            result = geopm::make_unique<NVMLAcceratorTopo>();
```
lhlawson:
```suggestion
            result = geopm::make_unique<NVMLAcceleratorTopo>();
```

lhlawson:
Tested on a V100 enabled system:
- [x] with NVML - no error, normal function
- [x] without NVML - no error, normal function

Tested on a non-GPU system:
- [x] with levelzero - error message is caught
`Warning: <geopm> Failed to load levelzero IOGroup.  GEOPM may not work properly unless an alternate IOGroup plugin is loaded to provide signals/controls required by the Controller and Agent.
The error was: <geopm> Runtime error: LevelZero::LevelZeroImp: LevelZero Driver failed to initialize. Level Zero Error: ZE_RESULT_ERROR_UNINITIALIZED: at geopm/src/LevelZero.cpp:62`

- [x] without levelzero - default GEOPM behavior

lhlawson (approved):

Issue #1856, opened 2021-09-15
Define last level abstraction for sigwait(2) and sigwaitinfo(2) sigtimedwait(2) sigaction(2) kill(2)


cmcantalupo:
> estimate 3

Parent issue: #1857 

We will need a class that abstracts the system calls around waiting for signals to occur.  It is expected that issue #1687 will require the following four calls: sigwait(2), sigwaitinfo(2), and sigtimedwait(2) and sigaction(2) kill(2).  Since this is a last level abstraction, we don't need unit tests, but the implementation should be very thin.

Done criteria:
- [x] Define POSIXSignal class header that enables the required features
- [x] Implement the class with thin wrappers around system calls
- [x] Push PR into geopm-service branch (or dev if already merged)

cmcantalupo:
Work is started here:

https://github.com/cmcantalupo/geopm/tree/issue-1856

Krebrov001:
Note to self:

## Unit Tests

 - Testing the `POSIXSignalImp`

- we should create unit tests for the two helper methods in *this particular pull request*
  - `reduce_info()`
  - `make_sigset()`

PR #1957, opened 2021-10-19
Issue-1856

Krebrov001:
- Relates to #1856 from github issues

Implemented POSIXSIgnalImp methods


Krebrov001:
@cmcantalupo Should we add a test for `sigtimedwait()` for not just an error, but a valid use case? We could `kill(getpid(), SIGUSR1)` to set the `sigtimed wait()` and `sigwaitinfo()` also.

Issue #1860, opened 2021-09-15
Fine tune BatchProtocol class that is sketched out on the batch-server branch of cmcantalupo


cmcantalupo:
> estimate 2

Discussion in architecture meeting lead to some proposed changes to the current sketch:
https://github.com/cmcantalupo/geopm/tree/batch-server

- Base class for Protocol that has two derived classes, one for server and one for client
- Clean up notes and comments put in place live during the architecture meeting.

Parent story: #1857

cmcantalupo:
Moved working branch here:

https://github.com/cmcantalupo/geopm/tree/issue-1860

PR #1961, opened 2021-10-21
Add helper methods for batch server startup

### LINKED ISSUE #1960: Add chown to SharedMemory C++ class, uid/gid getters to Helper.cpp
> estimate 2
Parent story: #1857
Related task: #1860

Done criteria:
1. SharedMemory can be chowned from one uid/gid to another.
2. The UID and GID can be queried from the PID.

This is all necessary to get the shmem segments for batch server communication up and running with the right permissions and owners.

- [x] Unit tests added for new class methods except: non-test
      last-level-abstraction of an external dependency (e.g.
      SDBus or NVMLDevicePool)
- [x] Non-exempt methods with logic branches have multiple
      tests
- [ ] ~Mock objects derived from class have methods added~ No mocks.
- [x] New public methods have doxygen comments describing
      behavior, inputs, outputs and exceptions raised

bgeltz:
- Relates to #1860.
- Fixes #1960.

- Adds SharedMemory:chown, geopm::pid_to_uid, and geopm::pid_to_gid.
- Adds new SharedMemory API to create segments without world permissions: SharedMemory::make_unique_owner_secure.

### DIFF
--- a/service/src/SharedMemory.cpp
+++ b/service/src/SharedMemory.cpp
@@ -113,7 +113,7 @@ namespace geopm
 
     }
 
-    void SharedMemoryImp::create_memory_region(const std::string &shm_key, size_t size, bool world_perms)
+    void SharedMemoryImp::create_memory_region(const std::string &shm_key, size_t size, bool is_secure)
     {
         if (!size) {
             throw Exception("SharedMemoryImp: Cannot create shared memory region of zero size", GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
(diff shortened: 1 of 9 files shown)


service/src/SharedMemory.cpp
```diff
@@ -115,7 +122,13 @@ namespace geopm
         m_size = size + M_LOCK_SIZE;
 
         mode_t old_mask = umask(0);
-        int shm_id = shm_open(m_shm_key.c_str(), O_RDWR | O_CREAT | O_EXCL, S_IRUSR | S_IWUSR | S_IRGRP| S_IWGRP | S_IROTH| S_IWOTH);
+        int shm_id = 0;
+        if (world_perms) {
```
cmcantalupo: This looks backward to me according to the definition of world_perms in the doxygen.  Maybe we can have the "world_perms" be "is_secure", and then this will make sense.  I am in favor of dropping `S_IRGRP` and `S_IWGRP` in the secure case.

service/src/SharedMemoryImp.hpp
```diff
@@ -58,7 +58,8 @@ namespace geopm
             ///        an inter-process shared memory region.
             /// @param [in] shm_key Shared memory key to create the region.
             /// @param [in] size Size of the region to create.
-            void create_memory_region(const std::string &shm_key, size_t size);
+            /// @param [in] world_perms Allow world r/w.
```
cmcantalupo:
```suggestion
            /// @param [in] is_secure If true, restrict access to the owner only.
```

service/src/SharedMemoryImp.hpp
```diff
@@ -58,7 +58,8 @@ namespace geopm
...
+            void create_memory_region(const std::string &shm_key, size_t size, bool world_perms);
```
cmcantalupo:
```suggestion
            void create_memory_region(const std::string &shm_key, size_t size, bool is_secure);
```

cmcantalupo (changes requested):
Looks good other than the variable name and documentation.  Also suggestion to drop group permissions in the secure case as well.

cmcantalupo (approved):
LGTM

PR #1968, opened 2021-10-26
Fixed the command line arguments for Arithmetic Intensity benchmark.

fkeceli:
The command line options for this integration test were not being passed to the app but instead being consumed by the launcher.

### DIFF
--- a/integration/apps/arithmetic_intensity/arithmetic_intensity.py
+++ b/integration/apps/arithmetic_intensity/arithmetic_intensity.py
@@ -72,6 +72,9 @@ def setup_run_args(parser):
     parser.add_argument('--slowdown', type=float, default=1,
         help='When imbalance is present, this specifies the amount of work for slow ranks'
              ' to perform, as a factor of the amount of work the fast ranks perform.')
+    parser.add_argument('--base-internal-iterations', type=int,
+        help='How many iterations to perform in the inner loop, for the fast set of ranks '
+             '(all ranks if there is no imbalance).')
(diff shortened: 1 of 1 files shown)


integration/apps/arithmetic_intensity/arithmetic_intensity.py
```diff
@@ -54,67 +54,78 @@ def exec_path(run_type):
...
+                        help='Number of physical cores to reserve for the app. '
+                             'If not defined, all nodes but one will be reserved (leaving one '
                              'node for GEOPM).')
     parser.add_argument('--distribute-slow-ranks', action='store_true',
                         help='Distribute slow ranks across nodes and packages. Otherwise, slow '
                              'ranks are assigned to fill nodes and packages (default: False).')
+    parser.add_argument('--slowdown', type=float, default=1,
+        help='When imbalance is present, this specifies the amount of work for slow ranks'
+             ' to perform, as a factor of the amount of work the fast ranks perform.')
+    parser.add_argument('--slow-ranks', type=int, default=0,
```
dannosliwcd, line 78: Please add --base-internal-iterations (https://github.com/dannosliwcd/arithmetic-intensity/blob/main/main.cpp#L61)

That option is useful for recently-discussed experiments where we want to decrease float count and increase iteration count without increasing the count of synchronization points.

dannosliwcd (approved):
LGTM. I think this covers all of the benchmark's options, plus moves the args to the generated .sh scripts instead of the job launcher's command options.

bgeltz (approved):

Issue #1973, opened 2021-10-27
As a developer of the BatchServer/BatchClient I would like a separate class that manages interprocess request notifications


cmcantalupo:
**What problem does the story solve? Please describe.**
It is a little unclear about how to resolve the security requirements for sending inter-process notifications about read/write/ready status.  If we could isolate that logic to a separate class (e.g. BatchStatus?), these requirements will not impact the implementation of the BatchServer or BatchClient classes and improve testability.

**Describe the tasks that will be required to complete the story**
- [x] #1974
- [x] #1975
- [x] #1976
- [x] #1986

**Describe alternatives you've considered**
Alternatively we can leave the dependency on POSIXSignal in the BatchServer/BatchClient classes.

**Additional context**
This refactor will enable support for FIFO or pipe implementations of the status update mechanism rather than POSIX signals.

PR #1978, opened 2021-10-28
Design an abstraction around notifications for batch server/client

### LINKED ISSUE #1974: Design an interface for BatchStatus
> estimate 1

- [x] Write BatchStatus.hpp header interfaces (something like): 
    ```
    BatchStatus::send_message(int pid, int msg);
    BatchStatus::receive_message(msg);
    ```
- [x] Add doxygen for all interfaces and the class
- [x] Refactor BatchServer and BatchClient to use new interface to make sure interface is sufficient (not part of PR).

cmcantalupo:
- Remove most direct use of POSIXSignal from the BatchServer
  and BatchClient implementations.
- Define new class interface for status updates.
- Update implementation of BatchServer and BatchClient to
  use BatchStatus interface.
- BatchStatus could be implemented with POSIXSignal class
- BatchStatus may also be implemented with mkfifo(2) using
  the same key used for shared memory
- Relates to #1973
- Fixes #1974

### DIFF
--- a/service/Makefile.am
+++ b/service/Makefile.am
@@ -144,6 +144,7 @@ libgeopmd_la_SOURCES = $(include_HEADERS) \
                        src/AcceleratorTopoNull.cpp \
                        src/AcceleratorTopoNull.hpp \
                        src/Agg.cpp \
+                       src/BatchStatus.hpp \
                        src/CNLIOGroup.cpp \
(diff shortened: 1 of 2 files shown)


cmcantalupo:
duplicate, merged with implementation

Issue #1962, opened 2021-10-21
As a user of the ServiceIOGroup I would like to be able to use the batch interface


cmcantalupo:
**What problem does the story solve?**
I'm always frustrated when I would like to write a C++ Agent algorithm that uses controls or signals that require the GEOPM Service (e.g. LevelZero related telemetry and hardware settings).  

**Describe the tasks that will be required to complete the story**
- [x] #1959
- [x] #1960
- [x] #1989
- [x] #1986
- [x] #1963
- [x] #1964
- [x] #1988
- [x] #1965
- [x] #1966
- [x] #1967

**Describe alternatives you've considered**
Alternatively we could add a DBus interface that does not use inter-process shared memory or signalling.  This would be less performant, but may have fewer bugs related multi-threading issues.

**Additional context**
Current working branch for this is is #1958, but this should be factored out into pull requests that resolve each of the issues above.

PR #1992, opened 2021-11-04
Implement test and document BatchServerClass

### LINKED ISSUE #1963: Implement the BatchServer class including unit tests and documentation.
> estimate 3

Parent story: #1962 

- [x] Pure virtual interface class has been added with factory method
- [x] Unit test constructor provides dependency injection
- [x] Unit tests added for all class methods except: non-test
      NVMLDevicePool)
- [x] Non-exempt methods with logic branches have multiple tests
- [x] All dependency classes have already been merged
- [x] All public methods have doxygen comments describing behavior,
- [x] Implementation file has `#include "config.h"`

cmcantalupo:
- Relates to #1962 
- Fixes #1963


service/src/BatchServer.cpp
```diff
@@ -0,0 +1,258 @@
...
+        , m_pio(pio)
+        , m_signal_shmem(signal_shmem)
+        , m_control_shmem(control_shmem)
+        , m_batch_status(batch_status)
+        , m_server_key(std::to_string(m_client_pid))
+        , m_server_pid(0)
+        , m_is_active(false)
+    {
+        bool is_test = (m_batch_status != nullptr);
+        if (!is_test) {
```
Krebrov001: Perhaps invert the if statement?

service/src/BatchServer.hpp
```diff
@@ -0,0 +1,167 @@
...
+            std::vector<geopm_request_s> m_signal_config;
+            std::vector<geopm_request_s> m_control_config;
+            PlatformIO &m_pio;
+            std::shared_ptr<SharedMemory> m_signal_shmem;
+            std::shared_ptr<SharedMemory> m_control_shmem;
+            std::shared_ptr<BatchStatus> m_batch_status;
+            std::string m_server_key;
+            int m_server_pid;
+            bool m_is_active;
+            std::vector<int> m_signal_idx;
```
Krebrov001:
```cpp
// Each element in the vector is a handle for the request that is pushed onto the stack.
```

service/src/BatchServer.hpp
```diff
@@ -0,0 +1,167 @@
...
+            std::vector<int> m_control_idx;
```
Krebrov001: Would be good to include doxygen comments to describe what exactly is stored in here. And rename them as a vector of **handles**.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+    void BatchServerImp::update_and_write(void)
+    {
+        if (m_control_config.size() == 0) {
+            return;
+        }
+
+        double *buffer = (double *)m_control_shmem->pointer();
+        int buffer_idx = 0;
+        for (const auto &idx : m_control_idx) {
```
Krebrov001: I would like to rename `idx` to `handle`. And rename `m_control_idx` to `m_control_handle`.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,258 @@
...
+    void BatchServerImp::read_and_update(void)
+    {
+        if (m_signal_config.size() == 0) {
+            return;
+        }
+
+        m_pio.read_batch();
+        double *buffer = (double *)m_signal_shmem->pointer();
+        int buffer_idx = 0;
+        for (const auto &idx : m_signal_idx) {
```
Krebrov001: Similar for the signals.

service/src/BatchServer.hpp
```diff
@@ -0,0 +1,167 @@
...
+        private:
+            void push_requests(void);
+            void read_and_update(void);
+            void update_and_write(void);
+            void create_shmem(void);
+            void check_invalid_signal(void);
+            void check_return(int ret, const std::string &func_name) const;
+
+            int m_client_pid;
+            std::vector<geopm_request_s> m_signal_config;
```
Krebrov001:
```cpp
const std::vector<geopm_request_s> m_signal_config;
const std::vector<geopm_request_s> m_control_config;
```

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    void BatchServerImp::read_and_update(void)
+    {
+        if (m_signal_config.size() == 0) {
+            return;
+        }
+
+        m_pio.read_batch();
+        double *buffer = (double *)m_signal_shmem->pointer();
```
Krebrov001: Can we rename this to `shared_memory_buffer`?

service/test/MockSharedMemory.hpp
```diff
@@ -0,0 +1,71 @@
...
+        MOCK_METHOD(std::string, key, (), (const, override));
+        MOCK_METHOD(size_t, size, (), (const, override));
+        MOCK_METHOD(std::unique_ptr<geopm::SharedMemoryScopedLock>,
+                    get_scoped_lock, (), (override));
+        MOCK_METHOD(void, unlink, (), (override));
+        MOCK_METHOD(void, chown, (const unsigned int gid, const unsigned int uid),
+                    (override));
+
+    protected:
+        std::vector<char> m_buffer;
```
Krebrov001, line 68: Perhaps we can use `std::array` here instead, since the size of the vector is not changing?

service/test/BatchServerTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+    m_signal_config = {geopm_request_s {1, 0, "CPU_FREQUENCY"},
+                       geopm_request_s {2, 1, "TEMPERATURE"}};
+    m_control_config = {geopm_request_s {1, 0, "MAX_CPU_FREQUENCY"}};
+    m_batch_server = std::make_shared<BatchServerImp>(m_client_pid,
+                                                      m_signal_config,
+                                                      m_control_config,
+                                                      *m_pio_ptr,
+                                                      m_batch_status,
+                                                      m_signal_shmem,
+                                                      m_control_shmem);
```
Krebrov001: Just my style.

service/test/MockSharedMemory.hpp
cmcantalupo: Not new code, we can open an issue to change this mock implementation, but I don't see what would motivate that change.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,250 @@
...
+    BatchServerImp::BatchServerImp(int client_pid,
+                                   const std::vector<geopm_request_s> &signal_config,
+                                   const std::vector<geopm_request_s> &control_config)
+        : BatchServerImp(client_pid, signal_config, control_config,
+                         platform_io(), nullptr, nullptr, nullptr)
+    {
+        // Fork the server when calling real constructor.
+        auto setup = [this]() {this->create_shmem();};
+        auto run = [this]() {this->run_batch();};
+        m_server_pid = fork_with_setup(setup, run);
```
dannosliwcd: `BatchServerImp::fork_with_setup()` is able to directly call `create_shmem()` and `run_batch`. The `std::function` arguments don't seem necessary.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,250 @@
...
+                                                                        m_server_key))
+        , m_server_pid(0)
+        , m_is_active(true)
+    {
+
+    }
+
+    BatchServerImp::~BatchServerImp()
+    {
+        stop_batch();
```
dannosliwcd: This calls `BatchStatusImp::send_message` and `BatchStatusImp::receive_message`, which may throw exceptions. Consider wrapping this `stop_batch()` call in a `try...except{log the error}` block, to avoid throwing from a destructor (a possible source of undefined-behavior bugs in cases were a destructor is called due to some other exception, then raises a nested exception from itself).

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    int BatchServerImp::fork_with_setup(std::function<void(void)> setup,
+                                        std::function<void(void)> run)
+    {
+        int pipe_fd[2];
+        check_return(pipe(pipe_fd), "pipe(2)");
+        int forked_pid = fork();
+        check_return(forked_pid, "fork(2)");
+        if (forked_pid == 0) {
+            check_return(close(pipe_fd[0]), "close(2)");
```
dannosliwcd, line 331: Looks like the reason for this pipe is to ensure that the parent process does not proceed until the child process has created the shared memory. Why not just create the shared memory before forking?

service/src/BatchServer.cpp
dannosliwcd: Also, is it possible to hang on destruction if it's waiting for a message that never comes?

dannosliwcd (commented):
Partially done. Left off at BatchServerTest.cpp

service/src/BatchServer.cpp
cmcantalupo: This is to enable testing.  We can substitute other void(void) functions and test that forking them works.

cmcantalupo: A good question.  The answer is that we do not want the lifetime of the shared memory regions to be coupled to the lifetime of the geopmd systemd service process (this should be since server boot).  If we create them after we fork, they should be cleaned up when the forked server ends.

service/src/BatchServer.cpp
cmcantalupo: A good question.  The answer is that we do not want the lifetime of the shared memory regions to be coupled to the lifetime of the geopmd systemd service process (this should be since server boot).  If we create them after we fork, they should be cleaned up when the forked server ends.

service/src/BatchServer.cpp
cmcantalupo: Good idea!

cmcantalupo (commented):
Thanks for the feedback Daniel

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,250 @@
...
+    {
+        return m_server_key;
+    }
+
+    void BatchServerImp::stop_batch(void)
+    {
+        if (m_is_active) {
+            m_batch_status->send_message(M_MESSAGE_QUIT);
+            m_batch_status->receive_message(M_MESSAGE_CONTINUE);
+            m_is_active = false;
```
cmcantalupo: This message is going the wrong way.  The caller of this method is the parent process that forked the batch server (i.e. the primary geopmd process).  It doesn't make sense for geopmd to send a message to the client process to "quit".  This should be probably be resolved with a signal sent by geopmd and a signal handler on the batch server process.

cmcantalupo (commented):
The BatchServerImp::stop_batch() is not working in integration tests.  The reason is because of the comment above.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    void BatchServerImp::run_batch(void)
+    {
+        push_requests();
+        m_batch_status->send_message(M_MESSAGE_CONTINUE);
```
cmcantalupo: This messsage is extra, no continue should be sent at this point

dannosliwcd (commented):
Finished the pass where I left off. No additional comments from the remaining files.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,338 @@
...
+    {
+        if (m_server_pid != 0 && m_is_active) {
+            try {
+                stop_batch();
+            }
+            catch (const Exception &ex) {
+                std::cerr << "Warning: <geopm> BatchServerImp::~BatchServerImp(): Exception thrown in destructor: " <<  ex.what();
+            }
+            catch (...) {
+                std::cerr << "Warning: <geopm> BatchServerImp::~BatchServerImp(): Non-GEOPM exception thrown in destructor";
```
cmcantalupo: add `\n` to end of messages

service/test/BatchServerTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+        MAP_SHARED | MAP_ANONYMOUS,  // int flags
+        -1,                          // int fd
+        0                            // off_t offset
+    );
+    size_t &counter = *counter_mem;
+    counter = 0;
+
+    std::function<void(void)> setup = [&counter](void)
+    {
+        if(counter == 0u) {
```
cmcantalupo:
```suggestion
        if (counter == 0u) {
```

service/test/BatchServerTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+TEST_F(BatchServerTest, fork_with_setup)
+{
+    size_t *counter_mem = (size_t*) mmap(
+        NULL,                        // void *addr
+        sizeof(size_t),              // size_t length
+        PROT_READ  | PROT_WRITE,     // int prot
+        MAP_SHARED | MAP_ANONYMOUS,  // int flags
+        -1,                          // int fd
+        0                            // off_t offset
+    );
```
cmcantalupo:
```suggestion
        0);                          // off_t offset
```

service/test/BatchServerTest.cpp
```diff
@@ -74,20 +73,19 @@ class BatchServerTest : public ::testing::Test
         std::vector<geopm_request_s> m_signal_config;
         std::vector<geopm_request_s> m_control_config;
         std::shared_ptr<BatchServerImp> m_batch_server;
+        std::shared_ptr<BatchServerImp> m_batch_server_signals;
+        std::shared_ptr<BatchServerImp> m_batch_server_controls;
+        std::shared_ptr<BatchServerImp> m_batch_server_pid;
```
Krebrov001, line 78: What is the pro of having three different `BatchServerImp` objects, instead of having only one, whose definition changes based on the test fixture? It seems that only one of the three objects is used at any given time.

service/test/BatchServerTest.cpp
```diff
@@ -285,34 +264,28 @@ TEST_F(BatchServerTest, run_batch_read_empty)
...
 
     EXPECT_CALL(*m_batch_status,
-                send_message(BatchStatus::M_MESSAGE_CONTINUE))
+                send_message(BatchStatus::M_MESSAGE_QUIT))
         .Times(1)
         .RetiresOnSaturation();
 
-    m_batch_server->run_batch();
-
-    EXPECT_EQ(result.size(), m_signal_shmem->size());  // both should be == 0
```
Krebrov001: Is that the reason why we cannot check the size, because `nullptr` is passed in as the argument for the `m_signal_shmem` in the `BatchServerImp::BatchServerImp()`?

service/test/BatchServerTest.cpp
```diff
@@ -242,23 +237,11 @@ TEST_F(BatchServerTest, run_batch_read)
...
     std::vector<double> result = {};
 
-    // {}
-    for (const auto &request : m_signal_config) {
-        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
-                                            request.domain_idx))
-            .WillOnce(Return(idx))
-            .RetiresOnSaturation();
-        ++idx;
-    }
```
Krebrov001: I notice that we avoid pushing the signals by not having that for loop at all, since our `m_signal_config` is still initialized with the proper values. And the empty vector is passed into the constructor instead.
```cpp
m_signal_config = {geopm_request_s {1, 0, "CPU_FREQUENCY"},
                       geopm_request_s {2, 1, "TEMPERATURE"}};
```

service/test/BatchServerTest.cpp
```diff
@@ -395,77 +352,78 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+        testing::Mock::AllowLeak(m_pio_ptr.get());
+        testing::Mock::AllowLeak(m_batch_status.get());
+        testing::Mock::AllowLeak(m_posix_signal.get());
     };
 
-    int forked_pid = m_batch_server->fork_with_setup(setup, run);
+    int forked_pid = m_batch_server_pid->fork_with_setup(setup, run);
     waitpid(forked_pid, NULL, 0);
 
     ASSERT_EQ(2u, counter);
```
Krebrov001:
```suggestion
    EXPECT_EQ(2u, counter);
```

service/test/BatchServerTest.cpp
```diff
@@ -395,77 +352,78 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
-        ++counter;
+        if (counter == 1u) {
+            ++counter;
+        }
+        this->TearDown();
+        testing::Mock::AllowLeak(m_signal_shmem.get());
+        testing::Mock::AllowLeak(m_control_shmem.get());
+        testing::Mock::AllowLeak(m_pio_ptr.get());
+        testing::Mock::AllowLeak(m_batch_status.get());
+        testing::Mock::AllowLeak(m_posix_signal.get());
```
Krebrov001: Can we do something like
```cpp
this->m_signal_shmem.reset();
```

service/src/BatchServer.cpp
```diff
@@ -78,7 +79,20 @@ namespace geopm
...
         };
         auto run = [this]() {
-            this->run_batch();
+            try {
+                this->run_batch();
+            }
+            catch(...) {
+                if (m_is_client_waiting) {
+                    std::cerr << "Warning: <geopm>: " << __FILE__ << ":" << __LINE__
+                              << " Batch server was terminated while client was waiting: sending client quit message\n";
```
Krebrov001: Minor note, it seems that the `BatchServer` may be invoked by the `geopmd` as just a server process, without any connection to the command line ... perhaps we could use the `syslog()` instead for writing error messages? I have used it before.

service/test/BatchServerTest.cpp
```diff
@@ -285,34 +264,28 @@ TEST_F(BatchServerTest, run_batch_read_empty)
...
     EXPECT_CALL(*m_batch_status,
-                send_message(BatchStatus::M_MESSAGE_CONTINUE))
+                send_message(BatchStatus::M_MESSAGE_QUIT))
         .Times(1)
         .RetiresOnSaturation();
 
-    m_batch_server->run_batch();
-
-    EXPECT_EQ(result.size(), m_signal_shmem->size());  // both should be == 0
+    m_batch_server_controls->run_batch();
```
Krebrov001, line 271: Hmm ... do we want this one to be `run_batch()` or `event_loop()`? It seems that we aren't handling any theoretical thrown exceptions in this particular unit test. I assume that these exceptions can occur upon receipt of any signal other than `SIGTERM`? Maybe it can be used to test the failure condition?

service/src/BatchServer.cpp
```diff
@@ -144,42 +162,85 @@ namespace geopm
...
+    void BatchServerImp::write_message(char out_message)
+    {
+        if (m_is_client_waiting) {
+            try {
+                m_batch_status->send_message(out_message);
+                m_is_client_waiting = false;
+            }
+            catch (const Exception &ex) {
+                // If we were not interupted SIGTERM rethrow
+                if (ex.err_value() != EINTR ||
```
Krebrov001: Why are we not setting the `in_message = BatchStatus::M_MESSAGE_TERMINATE;` in this situation?

Krebrov001 (commented):
Thank you Chris for providing me the opportunity to review your pull request.

Krebrov001:
In general, I would like to add the doxygen `@throws` to all methods which may throw an exception, either directly or indirectly.
https://www.doxygen.nl/manual/commands.html

service/src/BatchServer.cpp
```diff
@@ -78,7 +79,20 @@ namespace geopm
             this->create_shmem();
         };
         auto run = [this]() {
-            this->run_batch();
+            try {
+                this->run_batch();
+            }
+            catch(...) {
+                if (m_is_client_waiting) {
```
dannosliwcd: This looks redundant with the check inside `run_batch`, so one of these two loggers can probably be removed. This one _should_ be dead code in its current state since the other one clears `m_is_client_waiting` before rethrowing. 

I think removing either one would be fine. I'd probably favor removing the deeper-in-the-callstack log-and-rethrows since it should result in fewer instances of manually bubbling errors up a callstack.

service/src/BatchServer.cpp
```diff
@@ -0,0 +1,373 @@
...
+        int forked_pid = fork();
+        check_return(forked_pid, "fork(2)");
+        if (forked_pid == 0) {
+            check_return(close(pipe_fd[0]), "close(2)");
+            setup();
+            char msg = BatchStatus::M_MESSAGE_CONTINUE;
+            check_return(write(pipe_fd[1], &msg, 1), "write(2)");
+            check_return(close(pipe_fd[1]), "close(2)");
+            run();
+            exit(0);
```
cmcantalupo: use _exit() and reset the shared memory pointers.

service/src/BatchServer.cpp
cmcantalupo: This detail is managed by dasbus.

service/src/BatchServer.cpp
cmcantalupo: I have refactored this code a bit in an attempt to clarify what is going on for the reader.  Let me know if it makes more sense now.

service/src/BatchServer.cpp
cmcantalupo: I have reset the shared memory pointers and am now calling _Exit().  This should clean up any stale keys that could get left around in certain error conditions, but does not call the destructor for all of the IOGroups registered in PlatformIO :)

service/test/BatchServerTest.cpp
cmcantalupo: This enables us to easily write more test fixtures that use any of these objects, and avoids issues around calling the destructor if we overwrite a the m_batch_server to set it to a new value.  It also allows the TearDown function to be responsible for more of the clean up after each test, thus reducing lines of code in each test fixture.

service/test/BatchServerTest.cpp
cmcantalupo: Would be great to get better testing of cases where mock objects in the unit tests are made to throw exceptions when they are called.

service/test/BatchServerTest.cpp
cmcantalupo: If there are any copies of the shared pointer elsewhere, the reset() will not be sufficient.  In this case we know that the _Exit() call will not clean up these mocks, and so we can use the gmock feature.

service/src/BatchServer.cpp
```diff
@@ -177,17 +174,15 @@ namespace geopm
...
-            }
-            catch (const Exception &ex) {
-                // If we were not interupted SIGTERM rethrow
-                if (ex.err_value() != EINTR ||
-                    g_sigterm_count == 0) {
-                    throw;
-                }
+        try {
+            m_batch_status->send_message(out_message);
+	    m_is_client_waiting = false;
```
cmcantalupo: whitespace!

service/test/BatchServerTest.cpp
cmcantalupo: m_signal_shmem and m_control_shmem should *not* be leaked since they are managed by the forked process explicitly.  These two lines should be removed.

dannosliwcd (approved):

PR #1996, opened 2021-11-06
Implement the BatchClient class with tests and documentation

### LINKED ISSUE #1964: Implement the BatchClient class including unit tests and documentation.
> estimate 2

Parent story: #1962 

- [x] Pure virtual interface class has been added with factory method
- [x] Unit test constructor provides dependency injection
- [x] Unit tests added for all class methods except: non-test
      NVMLDevicePool)
- [x] Non-exempt methods with logic branches have multiple tests
- [x] All dependency classes have already been merged
- [x] All public methods have doxygen comments describing behavior,
- [x] Implementation file has `#include "config.h"`

cmcantalupo:
- Relates to #1962 
- Fixes #1964 
- Requires #1992

### DIFF
--- a/service/src/BatchClient.cpp
+++ b/service/src/BatchClient.cpp
@@ -99,12 +99,12 @@ namespace geopm
 
     void BatchClientImp::write_batch(std::vector<double> settings)
     {
-        if (m_num_control == 0) {
-            return;
-        }
         if (settings.size() != (size_t)m_num_control) {
             throw Exception("BatchClientImp::write_batch(): settings vector length does not match the number of configured controls",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        if (m_num_control == 0) {
+            return;
         }
         double *buffer = (double *)m_control_shmem->pointer();
         std::copy(settings.begin(), settings.end(), buffer);
--- a/service/src/BatchClient.cpp
+++ b/service/src/BatchClient.cpp
@@ -114,6 +114,9 @@ namespace geopm
 
     void BatchClientImp::stop_batch(void)
     {
+        // Note that all requests sent to the batch server block on
+        // the client side until the server has completed the
+        // request. This is even true for the request to quit.
         m_batch_status->send_message(BatchStatus::M_MESSAGE_QUIT);
         m_batch_status->receive_message(BatchStatus::M_MESSAGE_QUIT);
     }
--- a/service/src/BatchClient.cpp
+++ b/service/src/BatchClient.cpp
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2015 - 2021, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of Intel Corporation nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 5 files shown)


service/test/BatchClientTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+        .Times(1);
+    m_batch_client->write_batch(settings_expect);
+    double *shmem_buffer =(double *)m_control_shmem->pointer();
+    EXPECT_EQ(settings_expect[0], shmem_buffer[0]);
+}
+
+TEST_F(BatchClientTest, read_batch_empty)
+{
+    std::vector<double> result_expect;
+    EXPECT_CALL(*m_batch_status, send_message(_))
```
Krebrov001, line 131: What is `send_message(_)`?

service/src/BatchClient.hpp
```diff
@@ -0,0 +1,128 @@
...
+            BatchClientImp(const std::string &server_key, double timeout,
+                           int num_signal, int num_control);
+            BatchClientImp(int num_signal, int num_control,
+                           std::shared_ptr<BatchStatus> batch_status,
+                           std::shared_ptr<SharedMemory> signal_shmem,
+                           std::shared_ptr<SharedMemory> control_shmem);
+            virtual ~BatchClientImp() = default;
+            std::vector<double> read_batch(void) override;
+            void write_batch(std::vector<double> settings) override;
+            void stop_batch(void) override;
```
Krebrov001, line 118: I would like to add a unit test for `stop_batch()`.

service/src/BatchClient.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    void BatchClientImp::write_batch(std::vector<double> settings)
+    {
+        if (m_num_control == 0) {
+            return;
+        }
+        if (settings.size() != (size_t)m_num_control) {
+            throw Exception("BatchClientImp::write_batch(): settings vector length does not match the number of configured controls",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
Krebrov001, line 104: I want to add a unit test for `write_batch()` that would trigger the exception to be thrown when `std::vector<double> settings` with wrong number of elements is passed in.

Krebrov001: Maybe it would be useful to move the `if (m_num_control == 0)` below this check to have the error condition get triggered even if `std::vector<double> settings = {1.2, 2.3}` is passed in, when there are zero controls?

Krebrov001:
In general, I would like to add the doxygen `@throws` to all methods which may throw an exception, either directly or indirectly.
https://www.doxygen.nl/manual/commands.html

service/src/BatchClient.cpp
```diff
@@ -0,0 +1,120 @@
...
+        double *buffer = (double *)m_control_shmem->pointer();
+        std::copy(settings.begin(), settings.end(), buffer);
+        m_batch_status->send_message(BatchStatus::M_MESSAGE_WRITE);
+        m_batch_status->receive_message(BatchStatus::M_MESSAGE_CONTINUE);
+    }
+
+    void BatchClientImp::stop_batch(void)
+    {
+        m_batch_status->send_message(BatchStatus::M_MESSAGE_QUIT);
+        m_batch_status->receive_message(BatchStatus::M_MESSAGE_QUIT);
```
dannosliwcd, line 121: Will this result in left-over resources on the server side if a client terminates without calling stop_batch()?

I think you already answered this when reviewing the server side, but I can't remember. Maybe a comment here (or when creating this connection) could help.


Also why does this need to receive a quit message? Is there some subsequent work that fails if the server hasn't yet ack'd the stop request?

dannosliwcd (dismissed):
Just raised two questions. LGTM if those are non-issues.

service/src/BatchClient.cpp
cmcantalupo: The geopmd process checks that all PID's that have open sessions are active once per second.  If a PID that has an open session terminates, geopmd will stop any batch server running for that session and restore all control values.  This is done with the callback registration below:

https://github.com/geopm/geopm/blob/dev/service/geopmdpy/service.py#L680-L687

There is a problem though... the close_session() method does not shut down the open batch server as it should...

https://github.com/geopm/geopm/blob/dev/service/geopmdpy/service.py#L419-L428

cmcantalupo: Addressed this issue with the rather large patch added to the next PR, #1998

https://github.com/geopm/geopm/pull/1998/commits/b797fba5e412953875aa8442e837251b7dd84c95

Since this does not impact the BatchClient implementation, this can be merged with the #1998

cmcantalupo: Oh and the QUIT message. We send that message whenever the client has made a request and the batch server is ending.  In the highlighted case, we send back the QUIT message when the client requested it.  If the server is interrupted by geopmd while servicing a  READ or WRITE request from the client, the client will wait for the CONTINUE message, but actually receive the QUIT message.  This will be an error on the client side *unless* the request that got interrupted was the QUIT request...

dannosliwcd (approved):

PR #2007, opened 2021-11-19
Improve test coverage for `ServiceIOGroupTest` and `BatchServerTest`:

Krebrov001:
Improve test coverage for `ServiceIOGroupTest`:
 - `ServiceIOGroupTest.read_signal_exception`
 - `ServiceIOGroupTest.write_control_exception`
 - `ServiceIOGroupTest.save_control`
 - `ServiceIOGroupTest.restore_control`

   - Relates to #2016  [bug report/feature request/story] from github issues

### DIFF
--- a/service/src/BatchServer.cpp
+++ b/service/src/BatchServer.cpp
@@ -93,6 +93,7 @@ namespace geopm
         auto setup = [this]() {
             this->child_register_handler();
             this->create_shmem();
+            return BatchStatus::M_MESSAGE_CONTINUE;
         };
         auto run = [this]() {
             this->run_batch();
@@ -369,7 +370,7 @@ namespace geopm
         m_posix_signal->sig_action(signo, &action, nullptr);
     }
 
-    int BatchServerImp::fork_with_setup(std::function<void(void)> setup,
+    int BatchServerImp::fork_with_setup(std::function<char(void)> setup,
                                         std::function<void(void)> run)
     {
         int pipe_fd[2];
@@ -378,8 +379,7 @@ namespace geopm
         check_return(forked_pid, "fork(2)");
         if (forked_pid == 0) {
             check_return(close(pipe_fd[0]), "close(2)");
-            setup();
-            char msg = BatchStatus::M_MESSAGE_CONTINUE;
+            char msg = setup();  // BatchStatus::M_MESSAGE_CONTINUE
             check_return(write(pipe_fd[1], &msg, 1), "write(2)");
             check_return(close(pipe_fd[1]), "close(2)");
             run();
--- a/service/src/BatchServer.cpp
+++ b/service/src/BatchServer.cpp
@@ -69,6 +69,9 @@ static void action_sigchld(int signo, siginfo_t *siginfo, void *context)
             g_sigchld_status = child_status;
         }
     }
+    if (child_pid == -1) {
+        g_sigchld_status = errno ? errno : GEOPM_ERROR_RUNTIME;
+    }
     ++g_sigchld_count;
 }
 
--- a/service/src/BatchServer.cpp
+++ b/service/src/BatchServer.cpp
@@ -250,6 +250,7 @@ namespace geopm
                     out_message = BatchStatus::M_MESSAGE_QUIT;
                     break;
                 case BatchStatus::M_MESSAGE_TERMINATE:
+                    out_message = BatchStatus::M_MESSAGE_TERMINATE;
                     break;
                 default:
                     throw Exception("BatchServerImp::run_batch(): Received unknown response from client: " +
--- a/service/src/geopm/Exception.hpp
+++ b/service/src/geopm/Exception.hpp
@@ -119,6 +119,13 @@ namespace geopm
             /// The error code associated with the exception.
             int m_err;
     };
+
+    /// @brief Function that converts an error code into an error message.
+    ///
+    /// @param error_value The error code associated with the exception.
+    ///
+    /// @return string The error message associated with the error code.
+    std::string error_message(int error_value);
 }
 
 #endif
--- a/service/test/BatchServerTest.cpp
+++ b/service/test/BatchServerTest.cpp
@@ -461,6 +461,78 @@ TEST_F(BatchServerTest, run_batch_write_empty)
 }
 
 /**
+ * @test Provides coverage for the case when BatchStatus::M_MESSAGE_TERMINATE is read message
+ *       by the server in BatchServerImp::event_loop()
+ */
+TEST_F(BatchServerTest, receive_message_terminate)
+{
+    InSequence sequence;
+
+    int idx = 0;
+    std::vector<double> result = {240.042, 250.052};
+
+    for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    for (const auto &request : m_control_config) {
+        EXPECT_CALL(*m_pio_ptr, push_control(request.name, request.domain,
+                                             request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(Return(BatchStatus::M_MESSAGE_TERMINATE))
+        .RetiresOnSaturation();
+
+    m_batch_server->run_batch();
+}
+
+/**
+ * @test Provides coverage for the case when not a valid message is read
+ *       by the server in BatchServerImp::event_loop()
+ */
+TEST_F(BatchServerTest, receive_message_default)
+{
+    InSequence sequence;
+
+    int idx = 0;
+    std::vector<double> result = {240.042, 250.052};
+
+    for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    for (const auto &request : m_control_config) {
+        EXPECT_CALL(*m_pio_ptr, push_control(request.name, request.domain,
+                                             request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(Return(127))  // a "random" number
+        .RetiresOnSaturation();
+
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_batch_server->run_batch(),
+        GEOPM_ERROR_RUNTIME,
+        "BatchServerImp::run_batch(): Received unknown response from client: "
+    );
+}
+
+/**
  * @throws geopm Exception to simulate failure of read(2) system call in BatchStatusImp::receive_message()
  */
 ACTION(batch_status_read_EINTR)
--- a/service/test/BatchServerTest.cpp
+++ b/service/test/BatchServerTest.cpp
@@ -79,6 +79,7 @@ class BatchServerTest : public ::testing::Test
         std::shared_ptr<BatchServerImp> m_batch_server_signals;
         std::shared_ptr<BatchServerImp> m_batch_server_controls;
         std::shared_ptr<BatchServerImp> m_batch_server_pid;
+        std::shared_ptr<BatchServerImp> m_batch_server_child_process;
 
 };
 
@@ -133,6 +134,16 @@ void BatchServerTest::SetUp()
                                                           m_signal_shmem,
                                                           m_control_shmem,
                                                           m_server_pid);
+
+    m_batch_server_child_process = std::make_shared<BatchServerImp>(getpid(),
+                                                                    m_signal_config,
+                                                                    m_control_config,
+                                                                    *m_pio_ptr,
+                                                                    m_batch_status,
+                                                                    m_posix_signal,
+                                                                    m_signal_shmem,
+                                                                    m_control_shmem,
+                                                                    0);
 }
 
 
@@ -153,6 +164,7 @@ void BatchServerTest::TearDown()
     m_batch_server_signals.reset();
     m_batch_server_controls.reset();
     m_batch_server_pid.reset();
+    m_batch_server_child_process.reset();
 }
 
 /**
@@ -187,6 +199,49 @@ TEST_F(BatchServerTest, stop_batch)
                           BatchStatus::M_MESSAGE_TERMINATE));
     m_batch_server->stop_batch();
     EXPECT_FALSE(m_batch_server->is_active());
+}
+
+/**
+ * @throws geopm Exception to simulate failure of sigqueue(2) system call in POSIXSignalImp::sig_queue()
+ */
+ACTION(sig_queue_SIGTERM)
+{
+    throw geopm::Exception(
+        "POSIXSignal(): POSIX signal function call "
+        "sigqueue()"
+        " returned an error",
+        EINVAL,
+        __FILE__,
+        __LINE__
+    );
+}
+
+/**
+ * @test First check that BatchServerImp::stop_batch() throws an exception if it is called from child process.
+ *       Then check that activate the catch block in BatchServerImp::stop_batch() by failure of POSIXSignalImp::sig_queue()
+ */
+TEST_F(BatchServerTest, stop_batch_exception)
+{
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_batch_server_child_process->stop_batch(),
+        GEOPM_ERROR_RUNTIME,
+        "BatchServerImp::stop_batch(): must be called from parent process, not child process"
+    );
+
+    EXPECT_CALL(
+        *m_posix_signal.get(),
+        sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE)
+    )
+    .Times(1)
+    .WillRepeatedly(sig_queue_SIGTERM());
+
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_batch_server->stop_batch(),
+        EINVAL,
+        "POSIXSignal(): POSIX signal function call "
+        "sigqueue()"
+        " returned an error"
+    );
 }
 
 /**
--- a/service/test/BatchServerTest.cpp
+++ b/service/test/BatchServerTest.cpp
@@ -775,12 +775,11 @@ TEST_F(BatchServerTest, fork_with_setup)
         return BatchStatus::M_MESSAGE_CONTINUE;
     };
 
-    std::function<void(void)> run = [&counter, this](void)
+    std::function<void(void)> run = [&counter](void)
     {
         if (counter == 1u) {
             ++counter;
         }
-        this->TearDown();
     };
 
     int forked_pid = m_batch_server_pid->fork_with_setup(setup, run);
@@ -1021,7 +1020,9 @@ int BatchServerTest::fork_other(std::function<void(int, int)> child_process_func
             child_process_func(read_pipe_fd, main_pid);  // pass read end of pipe
             close(read_pipe_fd);  // close read end of pipe
         }
-        exit(EXIT_SUCCESS);
+        // Use _Exit() instead of exit() to prevent Google Test from complaining about
+        // leaked mock objects.
+        _Exit(EXIT_SUCCESS);
     // parent process //
     } else {
         if (child_is_server) {
@@ -1117,12 +1118,6 @@ TEST_F(BatchServerTest, fork_and_terminate_child)
 
         // The server process is stopped at BatchStatus::receive_message() inside BatchServerImp::read_message()
         m_batch_server_test->run_batch();
-
-        // Allow Leak the mock objects in the child process.
-        testing::Mock::AllowLeak(m_pio_ptr.get());
-        testing::Mock::AllowLeak(m_batch_status.get());
-        testing::Mock::AllowLeak(m_signal_shmem.get());
-        testing::Mock::AllowLeak(m_control_shmem.get());
     };
 
     /// This function contains the parent process, which is the client.
@@ -1170,10 +1165,6 @@ TEST_F(BatchServerTest, fork_and_terminate_parent)
         sigval value;
         value.sival_int = BatchStatus::M_MESSAGE_TERMINATE;
         sigqueue(server_pid, SIGTERM, value);
-
-        // Allow Leak the mock shared memory objects in the child process.
-        testing::Mock::AllowLeak(m_signal_shmem.get());
-        testing::Mock::AllowLeak(m_control_shmem.get());
     };
 
     /// This function contains the parent process, which is the server.
--- a/service/test/BatchServerTest.cpp
+++ b/service/test/BatchServerTest.cpp
@@ -767,11 +767,12 @@ TEST_F(BatchServerTest, fork_with_setup)
     size_t &counter = *counter_mem;
     counter = 0;
 
-    std::function<void(void)> setup = [&counter](void)
+    std::function<char(void)> setup = [&counter](void)
     {
         if (counter == 0u) {
             ++counter;
         }
+        return BatchStatus::M_MESSAGE_CONTINUE;
     };
 
     std::function<void(void)> run = [&counter, this](void)
@@ -789,6 +790,36 @@ TEST_F(BatchServerTest, fork_with_setup)
 
     munmap(counter_mem, sizeof(size_t));
     counter_mem = nullptr;
+}
+
+/**
+ * @test This is similar to BatchServerTest.fork_with_setup, except that it checks an error case.
+ *       In the BatchServerImp::fork_with_setup() if the parent process received something other than
+ *       BatchStatus::M_MESSAGE_CONTINUE over the pipe, it would cause an Exception to be thrown.
+ *       This test guarantees coverage of the error case.
+ */
+TEST_F(BatchServerTest, fork_with_setup_exception)
+{
+    std::function<char(void)> setup = [](void)
+    {
+        // return something other than BatchStatus::M_MESSAGE_CONTINUE
+        // so that it would write that bad message to the pipe in the child process
+        // and the parent process would throw the Exception.
+        return BatchStatus::M_MESSAGE_QUIT;
+    };
+
+    std::function<void(void)> run = [](void)
+    {
+    };
+
+    // wait for any child process unless it gets overwritten with the actual PID of the child process.
+    int forked_pid = 0;
+    GEOPM_EXPECT_THROW_MESSAGE(
+        (forked_pid = m_batch_server_pid->fork_with_setup(setup, run)),
+        GEOPM_ERROR_RUNTIME,
+        "BatchServerImp: Receivied unexpected message from batch server at startup: \""
+    );
+    waitpid(forked_pid, NULL, 0);
 }
 
 /**
--- a/service/test/BatchServerTest.cpp
+++ b/service/test/BatchServerTest.cpp
@@ -48,7 +48,10 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
+#include <cstring>    // for strlen(), strncmp()
 #include <string>     // for std::string
+#include <sstream>    // for std::stringstream
+#include <iostream>   // for std::streambuf, std::cerr, std::flush
 
 using testing::InSequence;
 using testing::Return;
@@ -710,4 +713,165 @@ TEST_F(BatchServerTest, fork_with_setup)
 
     munmap(counter_mem, sizeof(size_t));
     counter_mem = nullptr;
+}
+
+/**
+ * @class cerr_redirect
+ *
+ * @details This class facilitates redirect the stderr to an internal buffer for recording purposes.
+ */
+class cerr_redirect {
+    public:
+        /**
+         * @brief The redirect is NOT active when the object is created.
+         */
+        cerr_redirect()
+        {
+            cerr_backup = nullptr;
+            is_active = false;
+        }
+
+        /**
+         * @brief Restore the backup streambuf if it was not restored already.
+         */
+        ~cerr_redirect()
+        {
+            if (cerr_backup) {
+                std::cerr.rdbuf(cerr_backup);
+            }
+            is_active = false;
+        }
+
+        /**
+         * @brief Call this function first. Activates the redirect.
+         *
+         * @details After calling this function, all std::cerr will be redirected to the internal stringstream.
+         *          This function saves the backup streambuf of std::cerr, and resets it to the internal stringstream buffer.
+         */
+        void open_redirect(void)
+        {
+            if (!is_active) {
+                // Flush the remaining stderr into the terminal.
+                std::cerr << std::flush;
+                // Clear the stringstream
+                buffer.clear();
+                buffer.str(std::string());
+                // Set the streambuf of the std::cerr equivalent to the streambuf of the stringstream.
+                // Return the former streambuf of the std::cerr prior to the change.
+                cerr_backup = std::cerr.rdbuf( buffer.rdbuf() );
+                is_active = true;
+            }
+        }
+
+        /**
+         * @brief This function deactivates the redirect. It should be called after open_redirect().
+         *
+         * @details This function restores the backup streambuf associated with the stderr, and closes the stringstream redirect.
+         *
+         * @return The internal string of the stringstream, which recorded the redirected output.
+         */
+        std::string close_redirect(void)
+        {
+            if (is_active)
+            {
+                // Flush the remaining stderr into the streambuf.
+                std::cerr << std::flush;
+                // Restore the streambuf of the std::cerr to it's original state.
+                std::cerr.rdbuf(cerr_backup);
+                is_active = false;
+
+                return buffer.str();
+            } else {
+                return std::string();
+            }
+        }
+    private:
+        /// records the redirected stderr into it's own streambuf
+        std::stringstream buffer;
+        /// the backup streambuf associated with the stderr
+        std::streambuf* cerr_backup;
+        /// is true if the redirect is active
+        bool is_active;
+};
+
+/**
+ * @throws (int)2 to simulate failure of sigqueue(2) system call in POSIXSignalImp::sig_queue()
+ */
+ACTION(sig_queue_throw_2)
+{
+    throw 2;
+}
+
+/**
+ * @test Check coverage of BatchServerImp::~BatchServerImp()
+ *       Create new BatchServer objects, and have BatchServerImp::stop_batch() throw exceptions in the destructor.
+ */
+TEST_F(BatchServerTest, destructor_exceptions)
+{
+    ///
+    /// Testing catch (const Exception &ex) {}
+    ///
+
+    // Create new BatchServer object
+    std::shared_ptr<BatchServerImp> m_batch_server_exception = std::make_shared<BatchServerImp>(
+        m_client_pid,
+        m_signal_config,
+        m_control_config,
+        *m_pio_ptr,
+        m_batch_status,
+        m_posix_signal,
+        m_signal_shmem,
+        m_control_shmem,
+        m_server_pid
+    );
+
+    EXPECT_CALL(
+        *m_posix_signal.get(),
+        sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE)
+    )
+    .Times(1)
+    .WillRepeatedly(sig_queue_SIGTERM());
+
+    const char* expected_stderr_1 = "Warning: <geopm> BatchServerImp::~BatchServerImp(): Exception thrown in destructor: ";
+    size_t stderr_1_length = strlen(expected_stderr_1);
+
+    cerr_redirect c_redirect;
+    c_redirect.open_redirect();
+    m_batch_server_exception.reset();  // calling BatchServerImp::~BatchServerImp()
+    std::string actual_stderr_1 = c_redirect.close_redirect();
+    // Compare if the recorded stderr matches the expected stderr.
+    EXPECT_EQ(0, strncmp(expected_stderr_1, actual_stderr_1.c_str(), stderr_1_length));
+
+    ///
+    /// Testing catch (...) {}
+    ///
+
+    // Create new BatchServer object
+    m_batch_server_exception.reset(new BatchServerImp(
+        m_client_pid,
+        m_signal_config,
+        m_control_config,
+        *m_pio_ptr,
+        m_batch_status,
+        m_posix_signal,
+        m_signal_shmem,
+        m_control_shmem,
+        m_server_pid
+    ));
+
+    EXPECT_CALL(
+        *m_posix_signal.get(),
+        sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE)
+    )
+    .Times(1)
+    .WillRepeatedly(sig_queue_throw_2());
+
+    expected_stderr_1 = "Warning: <geopm> BatchServerImp::~BatchServerImp(): Non-GEOPM exception thrown in destructor\n";
+    stderr_1_length = strlen(expected_stderr_1);
+
+    c_redirect.open_redirect();
+    m_batch_server_exception.reset();  // calling BatchServerImp::~BatchServerImp()
+    actual_stderr_1 = c_redirect.close_redirect();
+    // Compare if the recorded stderr matches the expected stderr.
+    EXPECT_EQ(0, strncmp(expected_stderr_1, actual_stderr_1.c_str(), stderr_1_length));
 }
--- a/service/test/BatchServerTest.cpp
+++ b/service/test/BatchServerTest.cpp
@@ -69,6 +69,7 @@ class BatchServerTest : public ::testing::Test
     protected:
         void SetUp();
         void TearDown();
+        int fork_other(std::function<void(int, int)> child_process_func, std::function<void(int, int)> parent_process_func, bool child_is_server);
         std::shared_ptr<MockPlatformIO> m_pio_ptr;
         std::shared_ptr<MockBatchStatus> m_batch_status;
         std::shared_ptr<MockPOSIXSignal> m_posix_signal;
@@ -701,6 +702,9 @@ TEST_F(BatchServerTest, read_batch_exception)
         .Times(1)
         .WillRepeatedly(read_batch_mbox_failed_EINVAL());
 
+    EXPECT_CALL(*m_batch_status, send_message(BatchStatus::M_MESSAGE_QUIT))
+        .Times(1);
+
     GEOPM_EXPECT_THROW_MESSAGE(
         m_batch_server->run_batch(),
         EINVAL,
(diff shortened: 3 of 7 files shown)


service/test/BatchServerTest.cpp
```diff
@@ -426,3 +788,466 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        // and receive_message() throws an exception in the check_return().
+        // This exception is caught in the BatchServerImp::read_message()
+        // and char in_message = BatchStatus::M_MESSAGE_TERMINATE is returned to BatchServerImp::event_loop()
+        EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(
+            [&write_pipe_fd](){
+                /* Extra code for synchronizing the server. */
+               char unique_char = '!';
+               int ret = write(write_pipe_fd, &unique_char, sizeof(unique_char));
+               ret = ret;
```
bgeltz:
```suggestion
```

This line is breaking the clang build.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +788,466 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        // Allow Leak the mock objects in the child process.
+        testing::Mock::AllowLeak(m_pio_ptr.get());
+        testing::Mock::AllowLeak(m_batch_status.get());
+        testing::Mock::AllowLeak(m_signal_shmem.get());
+        testing::Mock::AllowLeak(m_control_shmem.get());
+    };
+
+    /// This function contains the parent process, which is the client.
+    /// It takes the read_pipe_fd and the PID of the child process, which is the server.
+    std::function<void(int, int)> parent_process_func = [this](int read_pipe_fd, int server_pid)
```
bgeltz:
```suggestion
    std::function<void(int, int)> parent_process_func = [](int read_pipe_fd, int server_pid)
```

Unless you're going to use `this`, it must be omitted since warnings are treated as errors.  It's an unused variable otherwise.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +788,466 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        // and receive_message() throws an exception in the check_return().
+        // This exception is caught in the BatchServerImp::read_message()
+        // and char in_message = BatchStatus::M_MESSAGE_TERMINATE is returned to BatchServerImp::event_loop()
+        EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(
+            [&write_pipe_fd](){
+                /* Extra code for synchronizing the server. */
+               char unique_char = '!';
+               int ret = write(write_pipe_fd, &unique_char, sizeof(unique_char));
+               ret = ret;
```
bgeltz:
```suggestion
```

This is also breaking the clang build.

bgeltz (changes requested):
Please address these compilation issues.

service/test/ServiceIOGroupTest.cpp
```diff
@@ -181,6 +181,37 @@ TEST_F(ServiceIOGroupTest, read_signal_behavior)
                                "BAD SIGNAL not valid for ServiceIOGroup");
 }
 
+TEST_F(ServiceIOGroupTest, read_signal_exception)
+{
+    // !is_valid_signal(signal_name)
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_serviceio_group->read_signal("NUM_VACCUM_TUBES", 4, 0),
```
bgeltz:
```suggestion
        m_serviceio_group->read_signal("NUM_VACUUM_TUBES", 4, 0),
```

I know it's a nit, but it's spelled "vacuum".  This does not affect the functionality.

service/test/ServiceIOGroupTest.cpp
```diff
@@ -192,6 +223,39 @@ TEST_F(ServiceIOGroupTest, write_control)
     }
 }
 
+TEST_F(ServiceIOGroupTest, write_control_exception)
+{
+    // !is_valid_control(control_name)
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_serviceio_group->write_control("NUM_VACCUM_TUBES", 4, 0, 7.0),
```
bgeltz:
```suggestion
        m_serviceio_group->write_control("NUM_VACUUM_TUBES", 4, 0, 7.0),
```

I know it's a nit, but it's spelled "vacuum".  This does not affect the functionality.

service/test/ServiceIOGroupTest.cpp
```diff
@@ -264,13 +328,15 @@ TEST_F(ServiceIOGroupTest, write_batch)
 TEST_F(ServiceIOGroupTest, save_control)
 {
     /// These are noops, make sure mock objects are not called into
+    /// We want it to fail if it tries to open the non-existent file path.
```
bgeltz, line 313: Where are we testing that this failure actually happens?  I see we open a bad path here, but I also expected something like ```GEOPM_EXPECT_THROW_MESSAGE(...)```.

Same comment for the `restore_control` test below.

service/test/BatchServerTest.cpp
```diff
@@ -47,6 +48,11 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
+#include <cstring>    // for strlen(), strncmp()
+#include <string>     // for std::string
+#include <sstream>    // for std::stringstream
+#include <iostream>   // for std::streambuf, std::cerr, std::flush
```
bgeltz:
```suggestion
#include <cstring>
#include <string>
#include <sstream>
#include <iostream>
```

These comments are unnecessary.  Please remove them.

service/test/BatchServerTest.cpp
```diff
@@ -63,6 +69,7 @@ class BatchServerTest : public ::testing::Test
     protected:
         void SetUp();
         void TearDown();
+        int fork_other(std::function<void(int, int)> child_process_func, std::function<void(int, int)> parent_process_func, bool child_is_server);
```
bgeltz:
```suggestion
        int fork_other(std::function<void(int, int)> child_process_func,
                       std::function<void(int, int)> parent_process_func,
                       bool child_is_server);
```

service/test/BatchServerTest.cpp
```diff
@@ -150,8 +168,12 @@ void BatchServerTest::TearDown()
     m_batch_server_signals.reset();
     m_batch_server_controls.reset();
     m_batch_server_pid.reset();
+    m_batch_server_child_process.reset();
 }
 
+/**
+ * @test BatchServerImp::server_pid() in isolation
```
bgeltz:
```suggestion
 * @test Check BatchServerImp::server_pid() in isolation
```

service/test/BatchServerTest.cpp
```diff
@@ -160,12 +182,18 @@ TEST_F(BatchServerTest, get_server_pid)
     EXPECT_EQ(m_server_pid, server_pid);
 }
 
+/**
+ * @test BatchServerImp::get_server_key() in isolation
```
bgeltz:
```suggestion
 * @test Check BatchServerImp::get_server_key() in isolation
```

service/test/BatchServerTest.cpp
```diff
@@ -160,12 +182,18 @@ TEST_F(BatchServerTest, get_server_pid)
...
+ */
 TEST_F(BatchServerTest, get_server_key)
 {
     std::string key = m_batch_server->server_key();
     EXPECT_EQ(std::to_string(m_client_pid), key);
 }
 
+/**
+ * @test Check BatchServerImp::stop_batch() to send SIGTERM signal to the server.
```
bgeltz:
```suggestion
 * @test Check that BatchServerImp::stop_batch() sends a SIGTERM signal to the server.
```

service/test/BatchServerTest.cpp
```diff
@@ -177,6 +205,57 @@ TEST_F(BatchServerTest, stop_batch)
     EXPECT_FALSE(m_batch_server->is_active());
 }
 
+/**
+ * @throws geopm Exception to simulate failure of sigqueue(2) system call in POSIXSignalImp::sig_queue()
```
bgeltz:
```suggestion
 * @throws geopm::Exception() to simulate a failure of the sigqueue(2) system call in POSIXSignalImp::sig_queue()
```

service/test/BatchServerTest.cpp
```diff
@@ -177,6 +205,57 @@ TEST_F(BatchServerTest, stop_batch)
...
+        " returned an error",
+        EINVAL,
+        __FILE__,
+        __LINE__
+    );
+}
+
+/**
+ * @test First check that BatchServerImp::stop_batch() throws an exception if it is called from child process.
+ *       Then check that activate the catch block in BatchServerImp::stop_batch() by failure of POSIXSignalImp::sig_queue()
```
bgeltz:
```suggestion
 * @test Check that BatchServerImp::stop_batch() throws an exception if it is called from the child process.
 *       Then check the catch block in BatchServerImp::stop_batch() by forcing a failure of POSIXSignalImp::sig_queue().
```

service/test/BatchServerTest.cpp
```diff
@@ -235,6 +313,14 @@ TEST_F(BatchServerTest, run_batch_read)
     EXPECT_EQ(result[1], data_ptr[1]);
 }
 
+/**
+ * @test Check BatchServerImp::run_batch() when there are no signals and you try to read.
+ *       First the requests only controls are filled up.
```
bgeltz:
```suggestion
 *       First the control requests are populated.
```

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+/**
+ * @test Provides coverage for the case when not a valid message is read
+ *       by the server in BatchServerImp::event_loop()
+ */
+TEST_F(BatchServerTest, receive_message_default)
+{
+    InSequence sequence;
+
+    int idx = 0;
+    std::vector<double> result = {240.042, 250.052};
```
bgeltz:
```suggestion
```
Not used.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+    EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(Return(BatchStatus::M_MESSAGE_TERMINATE))
+        .RetiresOnSaturation();
+
+    m_batch_server->run_batch();
+}
+
+/**
+ * @test Provides coverage for the case when not a valid message is read
+ *       by the server in BatchServerImp::event_loop()
```
bgeltz:
```suggestion
 * @test Provides coverage for the case when an invalid message is read
 *       by the server in the BatchServerImp::event_loop().
```

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+
+    EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(Return(127))  // a "random" number
+        .RetiresOnSaturation();
+
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_batch_server->run_batch(),
+        GEOPM_ERROR_RUNTIME,
+        "BatchServerImp::run_batch(): Received unknown response from client: "
+    );
```
bgeltz: If the unknown response message (127) is in the exception text, then we should verify it is there.

It would be good if the "random" status message (127) was stored in a variable and that variable was used with `GEOPM_EXPECT_THROW_MESSAGE` to ensure that the proper unknown response was received (127 in this case).

The way the test is now it asserts that *some* exception is generated because of an unknown response, but we are specifically sending 127, so we should verify that.

If the 127 does not already show up in the exception text, then ignore this comment.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+}
+
+/**
+ * @throws geopm Exception to simulate failure of read(2) system call in BatchStatusImp::receive_message()
```
bgeltz:
```suggestion
 * @throws geopm::Exception() to simulate a failure of the read(2) system call in BatchStatusImp::receive_message().
```

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+/**
+ * @test Testing BatchStatus::receive_message() to throw an exception inside of BatchServerImp::read_message()
+ *       This is needed to activate the catch blocks in BatchServerImp::read_message() and BatchServerImp::run_batch()
+ */
+TEST_F(BatchServerTest, receive_message_exception)
+{
+    InSequence sequence;
+
+    int idx = 0;
+    std::vector<double> result = {240.042, 250.052};
```
bgeltz:
```suggestion
```
Not used.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+
+    for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    for (const auto &request : m_control_config) {
```
bgeltz:
```suggestion
    idx = 0;
    for (const auto &request : m_control_config) {
```

When pushing signals and controls, make sure the index is reset in between.  It doesn't matter for this test because you are not verifying the indices, but it will lead to confusing behavior if this code is used in other tests that do utilizes the indices.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
```
bgeltz, line 526:
```suggestion
    idx = 0;
    for (const auto &request : m_control_config) {
```

When pushing signals and controls, make sure the index is reset in between.  It doesn't matter for this test because you are not verifying the indices, but it will lead to confusing behavior if this code is used in other tests that do utilizes the indices.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+/**
+ * @test Testing BatchStatusImp::send_message() to throw an exception inside of BatchServerImp::write_message()
+ *       This is needed to activate the catch blocks in BatchServerImp::write_message() and BatchServerImp::run_batch()
+ */
+TEST_F(BatchServerTest, write_message_exception)
+{
+    InSequence sequence;
+
+    int idx = 0;
+    std::vector<double> result = {240.042, 250.052};
```
bgeltz:
```suggestion
```
Not used.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+
+    for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    for (const auto &request : m_control_config) {
```
bgeltz, line 567:
```suggestion
    idx = 0;
    for (const auto &request : m_control_config) {
```

When pushing signals and controls, make sure the index is reset in between.  It doesn't matter for this test because you are not verifying the indices, but it will lead to confusing behavior if this code is used in other tests that do utilizes the indices.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+/**
+ * @test Testing SSTIOImp::read_batch() to throw an exception inside of BatchServerImp::read_and_update()
+ *       This is needed to activate if (m_is_client_waiting) branch inside the catch block in BatchServerImp::run_batch()
+ */
+TEST_F(BatchServerTest, read_batch_exception)
+{
+    InSequence sequence;
+
+    int idx = 0;
+    std::vector<double> result = {240.042, 250.052};
```
bgeltz:
```suggestion
```
Not used.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+
+    for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+    }
+
+    for (const auto &request : m_control_config) {
```
bgeltz, line 615:
```suggestion
    idx = 0;
    for (const auto &request : m_control_config) {
```

When pushing signals and controls, make sure the index is reset in between.  It doesn't matter for this test because you are not verifying the indices, but it will lead to confusing behavior if this code is used in other tests that do utilizes the indices.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +461,263 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_batch_server->run_batch(),
+        EINVAL,
+        "SSTIOImp::read_batch() mbox read failed"
+    );
+}
+
+/**
+ * @test Check if we can create the shared memory.
+ *       Check if the shared memory files for both the signals and controls were created in the /dev/shm/
```
bgeltz:
```suggestion
 *       Check if the shared memory files for both the signals and controls were created in the /dev/shm/.
```

service/test/BatchServerTest.cpp
```diff
@@ -177,6 +205,57 @@ TEST_F(BatchServerTest, stop_batch)
...
+ACTION(sig_queue_SIGTERM)
+{
+    throw geopm::Exception(
+        "POSIXSignal(): POSIX signal function call "
+        "sigqueue()"
+        " returned an error",
+        EINVAL,
+        __FILE__,
+        __LINE__
+    );
```
bgeltz:
```suggestion
    throw geopm::Exception("POSIXSignal(): POSIX signal function call sigqueue() returned an error",
                           EINVAL, __FILE__, __LINE__);
```

service/test/BatchServerTest.cpp
```diff
@@ -177,6 +205,57 @@ TEST_F(BatchServerTest, stop_batch)
...
+ * @test First check that BatchServerImp::stop_batch() throws an exception if it is called from child process.
+ *       Then check that activate the catch block in BatchServerImp::stop_batch() by failure of POSIXSignalImp::sig_queue()
+ */
+TEST_F(BatchServerTest, stop_batch_exception)
+{
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_batch_server_child_process->stop_batch(),
+        GEOPM_ERROR_RUNTIME,
+        "BatchServerImp::stop_batch(): must be called from parent process, not child process"
+    );
```
bgeltz:
```suggestion
    GEOPM_EXPECT_THROW_MESSAGE(m_batch_server_child_process->stop_batch(),
                               GEOPM_ERROR_RUNTIME, 
                               "BatchServerImp::stop_batch(): must be called from parent process, not child process");
```

bgeltz: Please use this style for spacing and line termination for function calls.  Please let me know if you have questions.

There are many calls to `GEOPM_EXPECT_THROW_MESSAGE` in this file.  Please fix all of them.

bgeltz (changes requested):

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+ *       BatchStatus::M_MESSAGE_CONTINUE over the pipe, it would cause an Exception to be thrown.
+ *       This test guarantees coverage of the error case.
+ */
+TEST_F(BatchServerTest, fork_with_setup_exception)
+{
+    std::function<char(void)> setup = [](void)
+    {
+        // return something other than BatchStatus::M_MESSAGE_CONTINUE
+        // so that it would write that bad message to the pipe in the child process
+        // and the parent process would throw the Exception.
```
bgeltz:
```suggestion
```

That comment is redundant. The information is already discussed in the docstring.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+    std::function<void(void)> run = [](void)
+    {
+    };
+
+    // wait for any child process unless it gets overwritten with the actual PID of the child process.
+    int forked_pid = 0;
+    GEOPM_EXPECT_THROW_MESSAGE(
+        (forked_pid = m_batch_server_pid->fork_with_setup(setup, run)),
+        GEOPM_ERROR_RUNTIME,
+        "BatchServerImp: Receivied unexpected message from batch server at startup: \""
```
bgeltz: Can you add to the expected text here about the specific unexpected message (M_MESSAGE_QUIT) that was received?  Is that text already provided in the exception similar to my other comment?

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+    );
+    waitpid(forked_pid, NULL, 0);
+}
+
+/**
+ * @class cerr_redirect
+ *
+ * @details This class facilitates redirect the stderr to an internal buffer for recording purposes.
+ */
+class cerr_redirect {
```
bgeltz:
```suggestion
class CerrRedirect {
```
Use camel case for C++ classes.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+    public:
+        /**
+         * @brief The redirect is NOT active when the object is created.
+         */
+        cerr_redirect()
```
bgeltz:
```suggestion
        CerrRedirect()
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        {
+            cerr_backup = nullptr;
+            is_active = false;
+        }
+
+        /**
+         * @brief Restore the backup streambuf if it was not restored already.
+         */
+        ~cerr_redirect()
```
bgeltz:
```suggestion
        ~CerrRedirect()
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        {
+            if (!is_active) {
+                // Flush the remaining stderr into the terminal.
+                std::cerr << std::flush;
+                // Clear the stringstream
+                buffer.clear();
+                buffer.str(std::string());
+                // Set the streambuf of the std::cerr equivalent to the streambuf of the stringstream.
+                // Return the former streambuf of the std::cerr prior to the change.
+                cerr_backup = std::cerr.rdbuf( buffer.rdbuf() );
```
bgeltz:
```suggestion
                cerr_backup = std::cerr.rdbuf(buffer.rdbuf());
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+            if (is_active)
+            {
+                // Flush the remaining stderr into the streambuf.
+                std::cerr << std::flush;
+                // Restore the streambuf of the std::cerr to it's original state.
+                std::cerr.rdbuf(cerr_backup);
+                is_active = false;
+
+                return buffer.str();
+            } else {
```
bgeltz:
```suggestion
            } 
            else {
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+         * @brief This function deactivates the redirect. It should be called after open_redirect().
+         *
+         * @details This function restores the backup streambuf associated with the stderr, and closes the stringstream redirect.
+         *
+         * @return The internal string of the stringstream, which recorded the redirected output.
+         */
+        std::string close_redirect(void)
+        {
+            if (is_active)
+            {
```
bgeltz:
```suggestion
            if (is_active) {
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+                std::cerr.rdbuf(cerr_backup);
+                is_active = false;
+
+                return buffer.str();
+            } else {
+                return std::string();
+            }
+        }
+    private:
+        /// records the redirected stderr into it's own streambuf
```
bgeltz: You used the triple slash here to indicate that it's a docstring, but then there's no @keyword here.  Please insert the missing @keyword or remove these lines.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        bool is_active;
+};
+
+/**
+ * @throws (int)2 to simulate failure of sigqueue(2) system call in POSIXSignalImp::sig_queue()
+ */
+ACTION(sig_queue_throw_2)
+{
+    throw 2;
+}
```
bgeltz: It seems odd to me to mix these the usage of these ACTION's with the usage of lambda's.  Both are used in this file, and I think it would be good to use one or the other.  My preference is with lambda's.  I think this is one we should discuss as a group with @cmcantalupo.

The usage of these ACTIONs doesn't add anything new AFAICT.  It's just more complexity.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+ *       Create new BatchServer objects, and have BatchServerImp::stop_batch() throw exceptions in the destructor.
+ */
+TEST_F(BatchServerTest, destructor_exceptions)
+{
+    ///
+    /// Testing catch (const Exception &ex) {}
+    ///
+
+    // Create new BatchServer object
+    std::shared_ptr<BatchServerImp> m_batch_server_exception = std::make_shared<BatchServerImp>(
```
bgeltz:
```suggestion
    std::shared_ptr<BatchServerImp> batch_server_exception = std::make_shared<BatchServerImp>(
```
This variable is not a class member.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        m_client_pid,
+        m_signal_config,
+        m_control_config,
+        *m_pio_ptr,
+        m_batch_status,
+        m_posix_signal,
+        m_signal_shmem,
+        m_control_shmem,
+        m_server_pid
+    );
```
bgeltz:
```suggestion
        m_server_pid);
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+    EXPECT_CALL(
+        *m_posix_signal.get(),
+        sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE)
+    )
+    .Times(1)
+    .WillRepeatedly(sig_queue_SIGTERM());
```
bgeltz:
```suggestion
    EXPECT_CALL(*m_posix_signal.get(),
                sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE))
        .Times(1)
        .WillRepeatedly(sig_queue_SIGTERM());
```
Please follow the style of other invocations of EXPECT_CALL.  They should all look similar to this.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+    const char* expected_stderr_1 = "Warning: <geopm> BatchServerImp::~BatchServerImp(): Exception thrown in destructor: ";
+    size_t stderr_1_length = strlen(expected_stderr_1);
+
+    cerr_redirect c_redirect;
+    c_redirect.open_redirect();
+    m_batch_server_exception.reset();  // calling BatchServerImp::~BatchServerImp()
+    std::string actual_stderr_1 = c_redirect.close_redirect();
+    // Compare if the recorded stderr matches the expected stderr.
+    EXPECT_EQ(0, strncmp(expected_stderr_1, actual_stderr_1.c_str(), stderr_1_length));
```
bgeltz: Since you're using `std::string`'s in the CerrRedirect class, it seems logical to stick with them and not use a mix of `std::string`'s and C strings.

I would change all of this code to use `std::string`'s and then remove all of the C comparison functions and use `EXPECT_EQ` with `std::string` objects.  This is the pattern we use in other tests.
https://google.github.io/googletest/reference/assertions.html#c-strings

This will also cleanup your `#include`'s.  Our general rule is to avoid using C-style code unless it's absolutely necessary for interacting with some API.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        m_client_pid,
+        m_signal_config,
+        m_control_config,
+        *m_pio_ptr,
+        m_batch_status,
+        m_posix_signal,
+        m_signal_shmem,
+        m_control_shmem,
+        m_server_pid
+    ));
```
bgeltz:
```suggestion
        m_server_pid));
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+    EXPECT_CALL(
+        *m_posix_signal.get(),
+        sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE)
+    )
+    .Times(1)
+    .WillRepeatedly(sig_queue_throw_2());
```
bgeltz: Wrong style.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+            child_process_func(write_pipe_fd, main_pid);  // pass write end of pipe
+            close(write_pipe_fd);  // close write end of pipe
+        } else {  // parent is server
+            close(write_pipe_fd);  // close write end of pipe
+            child_process_func(read_pipe_fd, main_pid);  // pass read end of pipe
+            close(read_pipe_fd);  // close read end of pipe
+        }
+        // Use _Exit() instead of exit() to prevent Google Test from complaining about
+        // leaked mock objects.
+        _Exit(EXIT_SUCCESS);
```
bgeltz, line 936: I want to understand why this is necessary.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+    // parent process //
+    } else {
```
bgeltz:
```suggestion
    }
    else {
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        if (child_is_server) {
+            close(write_pipe_fd);  // close write end of pipe
+            parent_process_func(read_pipe_fd, result);  // pass read end of pipe
+            close(read_pipe_fd);  // close read end of pipe
+        } else {  // parent is server
```
bgeltz:
```suggestion
        }
        else {  // parent is server
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+            client_pid,
+            m_signal_config,
+            m_control_config,
+            *m_pio_ptr,
+            m_batch_status,
+            nullptr,           // Create a real POSIXSignal because we want to use sig_action()
+            m_signal_shmem,
+            m_control_shmem,
+            m_server_pid
+        );
```
bgeltz:
```suggestion
            m_server_pid);
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+        // There is no EXPECT_CALL for m_posix_signal->make_sigset() and m_posix_signal->sig_action()
+        // because we are using the real POSIXSignal instead of the mock object.
+
+        // register the action_sigterm()
+        // This uses the real POSIXSignal instead of the mock object.
+        m_batch_server_test->child_register_handler();
+
+        int idx = 0;
+
```
bgeltz:
```suggestion
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +790,456 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
+        }
+
+        for (const auto &request : m_control_config) {
```
bgeltz, line 995:
```suggestion
        idx = 0;
        for (const auto &request : m_control_config) {
```

bgeltz (changes requested):

service/test/BatchServerTest.cpp
cmcantalupo: _Exit() should be called by a forked process to allow the statically allocated variables to be destroyed by the parent process, not by the forked process.

service/test/ServiceIOGroupTest.cpp
cmcantalupo: Note we are checking that the ServiceIOGroup::save_control() method does not do anything.  We provide the bad file path to make sure that the method does not try to use the path.

service/test/ServiceIOGroupTest.cpp
```diff
@@ -181,6 +181,37 @@ TEST_F(ServiceIOGroupTest, read_signal_behavior)
...
 }
 
+TEST_F(ServiceIOGroupTest, read_signal_exception)
+{
+    // !is_valid_signal(signal_name)
+    GEOPM_EXPECT_THROW_MESSAGE(
+        m_serviceio_group->read_signal("NUM_VACUUM_TUBES", 4, 0),
+        GEOPM_ERROR_INVALID,
+        "ServiceIOGroup::read_signal(): signal name \""
+    );
```
bgeltz:
```suggestion
    GEOPM_EXPECT_THROW_MESSAGE(m_serviceio_group->read_signal("NUM_VACUUM_TUBES", 4, 0),
                               GEOPM_ERROR_INVALID, "ServiceIOGroup::read_signal(): signal name");
```

Please use this style for spacing and line endings.  Please change all invocations of ```GEOPM_EXPECT_THROW_MESSAGE``` to match this style.  Look at how other tests use ```GEOPM_EXPECT_THROW_MESSAGE``` and let me know if you have questions about the style.

Additionally, I changed the string in the last argument slightly.  I removed '\"'.  It seems like you started to set this up to expect a specific message in the exception text (I'm assuming "NUM_VACUUM_TUBES"), but then stopped.  What was the intention here?

service/test/BatchServerTest.cpp
```diff
@@ -160,39 +179,77 @@ TEST_F(BatchServerTest, get_server_pid)
...
+    EXPECT_CALL(*m_posix_signal,
+                sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE))
+        .Times(1)
+        .WillRepeatedly(
+            /// @throws geopm::Exception to simulate a failure of the sigqueue(2) system call in POSIXSignalImp::sig_queue()
+            [](){
+                throw geopm::Exception("POSIXSignal(): POSIX signal function call sigqueue() returned an error",
+                           EINVAL, __FILE__, __LINE__);
+            }
+        );
```
bgeltz:
```suggestion
        .WillRepeatedly(Throw(geopm::Exception("POSIXSignal(): POSIX signal function call sigqueue() returned an error",
                              EINVAL, __FILE__, __LINE__)));
```

Please remove all these lambdas and use the ```Throw()``` action.
https://google.github.io/googletest/reference/actions.html

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +437,227 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+        ++idx;
+    }
+
+    EXPECT_CALL(*m_batch_status, receive_message())
+        .WillOnce(Return(127))  // a "random" number
+        .RetiresOnSaturation();
+
+    GEOPM_EXPECT_THROW_MESSAGE(m_batch_server->run_batch(),
+                               GEOPM_ERROR_RUNTIME,
+                               "BatchServerImp::run_batch(): Received unknown response from client: " "127");
```
bgeltz: Please move the random number (127) into a variable, and use that variable with ```EXPECT_CALL``` and ```GEOPM_EXPECT_THROW_MESSAGE```.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +437,227 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+    // This is where it throws the message.
+    EXPECT_CALL(*m_batch_status, receive_message())
+        .Times(1)
+        .WillRepeatedly(
+            /// @throws geopm::Exception() to simulate a failure of the read(2) system call in BatchStatusImp::receive_message().
+            [](){
+                throw geopm::Exception("BatchStatusImp: System call failed: " "read(2)",
+                           EINTR, __FILE__, __LINE__);
+                return '\0';
+            }
```
bgeltz: Remove this lambda and use the Throw action similar to the above suggestion.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +437,227 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+    EXPECT_CALL(*m_batch_status,
+                send_message(BatchStatus::M_MESSAGE_QUIT))
+        .Times(1)
+        .WillRepeatedly(
+            /// @throws geopm::Exception to simulate failure of write(2) system call in BatchStatusImp::send_message()
+            [](){
+                throw geopm::Exception("BatchStatusImp: System call failed: " "write(2)",
+                           EINTR, __FILE__, __LINE__);
+            }
+        );
```
bgeltz: Remove this lambda and use the Throw action similar to the above suggestion.

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +437,227 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+
+    EXPECT_CALL(*m_batch_status,
+                send_message(BatchStatus::M_MESSAGE_QUIT))
+        .Times(1)
+        .RetiresOnSaturation();
+
+    GEOPM_EXPECT_THROW_MESSAGE(m_batch_server->run_batch(),
+                               EINTR,
+                               "BatchStatusImp: System call failed: " "write(2)");
```
bgeltz:
```suggestion
                               "BatchStatusImp: System call failed: write(2)");
```

service/test/BatchServerTest.cpp
```diff
@@ -359,6 +437,227 @@ TEST_F(BatchServerTest, run_batch_write_empty)
...
+
+    // This is where it throws the message.
+    EXPECT_CALL(*m_pio_ptr, read_batch())
+        .Times(1)
+        .WillRepeatedly(
+            /// @throws geopm::Exception to simulate failure of ioctl(2) system call in SSTIoctlImp::mbox() in SSTIOImp::read_batch()
+            [](){
+                throw geopm::Exception("SSTIOImp::read_batch() mbox read failed",
+                           EINVAL, __FILE__, __LINE__);
+            }
```
bgeltz: Remove this lambda and use the Throw action similar to the above suggestion.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+    std::function<void(void)> run = [](void)
+    {
+    };
+
+    // wait for any child process unless it gets overwritten with the actual PID of the child process.
+    int forked_pid = 0;
+    GEOPM_EXPECT_THROW_MESSAGE((forked_pid = m_batch_server_pid->fork_with_setup(setup, run)),
+                               GEOPM_ERROR_RUNTIME,
+                               "BatchServerImp: Receivied unexpected message from batch server at startup: \"" "113" "\"");
```
bgeltz:
```suggestion
                               "BatchServerImp: Receivied unexpected message from batch server at startup: \" 113\"");
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+                return std::string();
+            }
+        }
+    private:
+        // records the redirected stderr into it's own streambuf
+        std::stringstream buffer;
+        // the backup streambuf associated with the stderr
+        std::streambuf* cerr_backup;
+        // is true if the redirect is active
+        bool is_active;
```
bgeltz: All class members should start with "m_":
```suggestion
        bool m_is_active;
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+            }
+            else {
+                return std::string();
+            }
+        }
+    private:
+        // records the redirected stderr into it's own streambuf
+        std::stringstream buffer;
+        // the backup streambuf associated with the stderr
+        std::streambuf* cerr_backup;
```
bgeltz:
```suggestion
        std::streambuf* m_cerr_backup;
```
All class members should start with "m_":

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+                return buffer.str();
+            }
+            else {
+                return std::string();
+            }
+        }
+    private:
+        // records the redirected stderr into it's own streambuf
+        std::stringstream buffer;
```
bgeltz:
```suggestion
        std::stringstream m_buffer;
```
All class members should start with "m_":

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+    std::shared_ptr<BatchServerImp> batch_server_exception = std::make_shared<BatchServerImp>(
+        m_client_pid,
+        m_signal_config,
+        m_control_config,
+        *m_pio_ptr,
+        m_batch_status,
+        m_posix_signal,
+        m_signal_shmem,
+        m_control_shmem,
+        m_server_pid);
```
bgeltz: Please follow this style when there are a large number of arguments:
https://github.com/geopm/geopm/blob/07334f10777d04914179bb8d8649f3b6d6f92488/service/test/BatchServerTest.cpp#L94-L102

Please change all occurrences if there are multiple.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        );
+
+    std::string expected_stderr_1 = "Warning: <geopm> BatchServerImp::~BatchServerImp(): Exception thrown in destructor: ";
+
+    CerrRedirect c_redirect;
+    c_redirect.open_redirect();
+    batch_server_exception.reset();  // calling BatchServerImp::~BatchServerImp()
+    std::string actual_stderr_1 = c_redirect.close_redirect();
+    // Compare if the recorded stderr matches the expected stderr.
+    EXPECT_EQ(true, geopm::string_begins_with(actual_stderr_1, expected_stderr_1));
```
bgeltz: Use ```EXPECT_TRUE``` for assertions like this.
https://google.github.io/googletest/reference/assertions.html

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+            }
+        );
+
+    expected_stderr_1 = "Warning: <geopm> BatchServerImp::~BatchServerImp(): Non-GEOPM exception thrown in destructor\n";
+
+    c_redirect.open_redirect();
+    batch_server_exception.reset();  // calling BatchServerImp::~BatchServerImp()
+    actual_stderr_1 = c_redirect.close_redirect();
+    // Compare if the recorded stderr matches the expected stderr.
+    EXPECT_STREQ(expected_stderr_1.c_str(), actual_stderr_1.c_str());
```
bgeltz: There's no need to use C strings nor ```EXPECT_STREQ```.  

Instead, use ```EXPECT_EQ``` and the string object.
https://google.github.io/googletest/reference/assertions.html#c-strings

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+
+    EXPECT_CALL(*m_posix_signal,
+                sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE))
+        .Times(1)
+        .WillRepeatedly(
+            /// @throws geopm::Exception to simulate a failure of the sigqueue(2) system call in POSIXSignalImp::sig_queue()
+            [](){
+                throw geopm::Exception("POSIXSignal(): POSIX signal function call sigqueue() returned an error",
+                           EINVAL, __FILE__, __LINE__);
+            }
```
bgeltz: Remove this lambda and use the Throw action similar to the above suggestion.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        m_server_pid));
+
+    EXPECT_CALL(*m_posix_signal,
+                sig_queue(m_server_pid, SIGTERM, BatchStatus::M_MESSAGE_TERMINATE))
+        .Times(1)
+        .WillRepeatedly(
+            /// @throws (int)2 to simulate failure of sigqueue(2) system call in POSIXSignalImp::sig_queue()
+            [](){
+                throw 2;
+            }
```
bgeltz: Remove this lambda and use the Throw action similar to the above suggestion.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        // This uses the real POSIXSignal instead of the mock object.
+        batch_server_test->child_register_handler();
+
+        int idx = 0;
+        for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
```
bgeltz:
```suggestion
            ++idx;
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        }
+
+        idx = 0;
+        for (const auto &request : m_control_config) {
+        EXPECT_CALL(*m_pio_ptr, push_control(request.name, request.domain,
+                                             request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
```
bgeltz:
```suggestion
            ++idx;
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+                sleep(1024);
+                throw geopm::Exception("BatchStatusImp: System call failed: " "read(2)",
+                                       EINTR, __FILE__, __LINE__);
+
+                // This is irrelevant, it doesn't get returned because of the throw statement.
+                // This is just to match the return type of BatchStatusImp::receive_message()
+                return '$';
+            }
+        )
+        .RetiresOnSaturation();
```
bgeltz:
```suggestion
        EXPECT_CALL(*m_batch_status, receive_message())
            .WillOnce([&write_pipe_fd]() {
                /* Extra code for synchronizing the server. */
                char unique_char = '!';
                write(write_pipe_fd, &unique_char, sizeof(unique_char));
                sleep(1024);
                throw geopm::Exception("BatchStatusImp: System call failed: " "read(2)",
                                       EINTR, __FILE__, __LINE__);
                // This is irrelevant, it doesn't get returned because of the throw statement.
                // This is just to match the return type of BatchStatusImp::receive_message()
                return '$';
            }).RetiresOnSaturation();
```

For any lambdas that cannot be converted to ```Throw``` actions, please follow this style.  Here's an example from another test:
https://github.com/geopm/geopm/blob/07334f10777d04914179bb8d8649f3b6d6f92488/service/test/SSTIOTest.cpp#L229-L233

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        // This uses the real POSIXSignal instead of the mock object.
+        batch_server_test->child_register_handler();
+
+        int idx = 0;
+        for (const auto &request : m_signal_config) {
+        EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
+                                            request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
```
bgeltz:
```suggestion
            ++idx;
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        }
+
+        idx = 0;
+        for (const auto &request : m_control_config) {
+        EXPECT_CALL(*m_pio_ptr, push_control(request.name, request.domain,
+                                             request.domain_idx))
+            .WillOnce(Return(idx))
+            .RetiresOnSaturation();
+        ++idx;
```
bgeltz:
```suggestion
            ++idx;
```

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+                sleep(1024);
+                throw geopm::Exception("BatchStatusImp: System call failed: " "read(2)",
+                                       EINTR, __FILE__, __LINE__);
+
+                // This is irrelevant, it doesn't get returned because of the throw statement.
+                // This is just to match the return type of BatchStatusImp::receive_message()
+                return '$';
+            }
+        )
+        .RetiresOnSaturation();
```
bgeltz: Please change this to follow the style to the one above.

service/test/BatchServerTest.cpp
```diff
@@ -426,3 +730,441 @@ TEST_F(BatchServerTest, fork_with_setup)
...
+        std::shared_ptr<BatchServerImp> batch_server_test = std::make_shared<BatchServerImp>(
+            client_pid,
+            m_signal_config,
+            m_control_config,
+            *m_pio_ptr,
+            m_batch_status,
+            nullptr,           // Create a real POSIXSignal because we want to use sig_action()
+            m_signal_shmem,
+            m_control_shmem,
+            m_server_pid);
```
bgeltz: Please update this to follow the style of the above suggestion.

bgeltz (changes requested):

service/test/BatchServerTest.cpp
```diff
@@ -271,21 +331,29 @@ TEST_F(BatchServerTest, run_batch_read_empty)
...
+    for (const auto &request : m_signal_config) {
         EXPECT_CALL(*m_pio_ptr, push_signal(request.name, request.domain,
                                             request.domain_idx))
             .WillOnce(Return(idx))
             .RetiresOnSaturation();
         ++idx;
     }
 
-     for (const auto &request : m_control_config) {
+    // idx should be the same when we use write_batch()
```
bgeltz, line 357:
```suggestion
```

This comment doesn't make sense to me.  What is requiring this to be true?

bgeltz (changes requested):
This is starting to look pretty good now.

I'm good to merge after either dropping or justifying that comment.

bgeltz (dismissed):

service/src/BatchServer.cpp
```diff
@@ -69,6 +69,9 @@ static void action_sigchld(int signo, siginfo_t *siginfo, void *context)
             g_sigchld_status = child_status;
         }
     }
+    if (child_pid == -1) {
+        g_sigchld_status = GEOPM_ERROR_RUNTIME;
+    }
```
cmcantalupo:
```suggestion
    if (child_pid == -1) {
        g_sigchld_status = errno ? errno : GEOPM_ERROR_RUNTIME;
    }
```

service/src/geopm/Exception.hpp
```diff
@@ -119,6 +119,11 @@ namespace geopm
             /// The error code associated with the exception.
             int m_err;
     };
+
+    /// @param error_value The error code associated with the exception.
```
cmcantalupo, line 125:
```suggestion
    /// @brief Function that converts an error code into an error message
    /// @param error_value The error code associated with the exception.
```

cmcantalupo: There is an equivalent C interface here:

https://github.com/geopm/geopm/blob/dev/service/src/geopm_error.h#L60

perhaps it would be best to match the name:

geopm::get_error_string() -> geopm::error_message()

cmcantalupo (commented):
Couple of minor suggestions.

service/src/geopm/Exception.hpp
bgeltz: The interface Chris is referencing is implemented here:
https://github.com/geopm/geopm/blob/dev/service/src/Exception.cpp#L71

@cmcantalupo Why have two different implementations though?  My thought was to change the test to use the C interface and remove this one.

cmcantalupo: It is useful to be able to directly use the std::string type and not have a fixed length C string.  Additionally the C interface provides the message from the last handled exception (useful for our external C interfaces), but this is not helpful in a context where there may not be a handled exception.

cmcantalupo (dismissed):
Thanks! Looks good.  Please adjust the history as you see fit, and then we can merge it in.

cmcantalupo (approved):

Krebrov001:
Pull request successfully merged into `dev` branch.

PR #2013, opened 2021-11-24
Unlink shared memory keys after BatchClient attaches

### LINKED ISSUE #2012: The shared memory keys opened by the BatchClient are not unlinked
**Describe the bug**
I tried to use the BatchServer and I expected that all shared memory keys would be cleaned up instead none of the keys are unlinked

**GEOPM version**
705bb36fe

**Expected behavior**
Keys should always be cleaned up, even if the client does not connect.

**Actual behavior**
No calls to the unlink method are made for the SharedMemory objects that support the batch server.

cmcantalupo:
- Also unlink in BatchServer destructor in case BatchClient never connects.
- Fixes #2012

### DIFF
--- a/service/src/BatchClient.cpp
+++ b/service/src/BatchClient.cpp
@@ -82,7 +82,12 @@ namespace geopm
         , m_signal_shmem(signal_shmem)
         , m_control_shmem(control_shmem)
     {
-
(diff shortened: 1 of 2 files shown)


Krebrov001 (approved):
Thank you for unlinking the shared memory keys.

Issue #2018, opened 2021-11-29
As a developer of GEOPM python agents I would like to be able to generate a standard report


cmcantalupo:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to write integration tests for the python agents and I cannot leverage the integration infrastructure for the C++ agents.

**Describe the solution you'd like**
I would like to be able to call into the geopm::Reporter class from python and generate a standard GEOPM report using the C++ implementation for the Reporter.

**Describe alternatives you've considered**
Alternatively we could write an implementation of the Reporter class and its dependencies in python, but this seems difficult to maintain and develop.

**Additional context**
POC draft is posted here:
    https://github.com/cmcantalupo/geopm/tree/python-reporter

PR #2019, opened 2021-11-29
Add python hooks that enable use of the C++ reporter

cmcantalupo:
- Relates to #2018


scripts/geopmpy/reporter.py
```diff
@@ -0,0 +1,72 @@
...
+int geopm_reporter_init(void);
+int geopm_reporter_update(void);
+int geopm_reporter_generate(const char *profile_name,
+                            const char *agent_name,
+                            size_t result_max,
+                            char *result);
+""")
+_dl = get_dl('libgeopmpolicy.so')
+
+def init():
```
cmcantalupo, line 49: Missing doc strings!

service/geopmdpy/gffi.py
```diff
@@ -0,0 +1,52 @@
...
+#  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+import cffi
+gffi = cffi.FFI()
+_dl = dict()
+
+def get_dl(so_name):
```
cmcantalupo: add doc strings

src/geopm_reporter.h
```diff
@@ -0,0 +1,53 @@
...
+#define GEOPM_REPORTER_H_INCLUDE
+
+#include <cstdlib>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+int geopm_reporter_init(void);
```
cmcantalupo, line 43: Neither are there dox strings nor a man page, there should probably be one or the other. That being said, these APIs may change wrt #2039.

src/ApplicationSampler.cpp
```diff
@@ -182,8 +182,7 @@ namespace geopm
     void ApplicationSamplerImp::update(const geopm_time_s &curr_time)
     {
         if (!m_status || !m_sampler) {
-            throw Exception("ApplicationSamplerImp::" + std::string(__func__) + "(): cannot read process info before connect().",
-                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            return;
```
bgeltz, line 185: It's too bad there isn't a way to tell the ApplicationSampler to get out of the way without removing these exceptions.  Turning them into no-ops seems like it might come back to bite us later if we're chasing app connect / data flow issues.

Maybe this is fine since app connect is pretty well ironed out at this point, and we don't expect any issues.

src/Reporter.cpp
```diff
@@ -203,7 +256,9 @@ namespace geopm
         };
 
         std::vector<region_info> region_ordered;
-        auto region_name_set = application_io.region_name_set();
+        GEOPM_DEBUG_ASSERT(region_name_set.size() == 0 ||
+                           m_proc_region_agg != nullptr,
+                           "ReporterImp::create_report(): region set is not empty, but region aggregtator pointer is null");
```
bgeltz:
```suggestion
                           "ReporterImp::create_report(): region set is not empty, but region aggregator pointer is null");
```

src/Reporter.cpp
```diff
@@ -272,28 +300,31 @@ namespace geopm
...
-        auto unmarked_data = get_region_data(GEOPM_REGION_HASH_UNMARKED);
-        yaml_write(report, M_INDENT_UNMARKED_FIELD, unmarked_data);
-        // agent extensions for unmarked
-        const auto &it = agent_region_report.find(GEOPM_REGION_HASH_UNMARKED);
-        if (it != agent_region_report.end()) {
-            yaml_write(report, M_INDENT_UNMARKED_FIELD, agent_region_report.at(GEOPM_REGION_HASH_UNMARKED));
-        }
+        double epoch_count = m_platform_io.sample(m_epoch_count_idx);
+        // Do not add epoch or unmarked section if no application attached
+        if (!std::isnan(epoch_count)) {
```
bgeltz, line 305: What if an app attaches, but does *not* have epoch markup?

test/ReporterTest.cpp
```diff
@@ -328,7 +328,7 @@ TEST_F(ReporterTest, generate)
 
     // Other calls
     EXPECT_CALL(m_tree_comm, overhead_send()).WillOnce(Return(678 * 56));
-    EXPECT_CALL(*m_comm, rank()).WillOnce(Return(0));
+    EXPECT_CALL(*m_comm, rank()).WillRepeatedly(Return(0));
```
bgeltz, line 331: What warranted this change?

bgeltz (changes requested):

src/Reporter.cpp
cmcantalupo: The epoch_count is zero, not NAN.

test/ReporterTest.cpp
cmcantalupo: This is a side effect of the refactor of the Reporter into functions that pass the Comm rather than having a local variable rank that is used repeatedly.

src/Reporter.cpp
cmcantalupo: thanks!

src/ApplicationSampler.cpp
cmcantalupo: Yeah, I don't see a better solution.

bgeltz (approved):

PR #2032, opened 2021-12-06
Integration test fixes for running w/o msr-safe privs

bgeltz:
A collection of fixes for the integration tests and CI.


src/Reporter.cpp
```diff
@@ -366,7 +366,7 @@ namespace geopm
 
         comm->gatherv((void *) (host_report.c_str()), sizeof(char) * buffer_size,
                       (void *) report_buffer.data(), buffer_size_array, buffer_displacement, 0);
-        report_buffer.back() = '\0';
+        report_buffer.push_back('\0');
```
cmcantalupo: This *could* require the creation of a copy of the report.  Probably should just set the last character to null conditioned on 
```
if (report_buffer.size() != 0) {
    report_buffer.back() = '\0';
}
```

src/ProfileIOGroup.cpp
```diff
@@ -176,12 +177,12 @@ namespace geopm
 
         if (m_do_read[M_SIGNAL_REGION_HASH]) {
             for (int idx = 0; idx < m_num_cpu; ++idx) {
-                m_per_cpu_sample[M_SIGNAL_REGION_HASH][idx] = hash_to_signal(m_application_sampler.cpu_region_hash(idx));
```
cmcantalupo: This change will impact how hash invalid and hint inactive are reported during normal use.  Why do we want this change?

cmcantalupo (commented):
Thanks for finding all these issues!  Just a couple of questions.

cmcantalupo (approved):

PR #2043, opened 2021-12-09
Tie write lock access to the session leader PID

### LINKED ISSUE #2037: As a user of the service I would like to enable multiple PIDs within the same TTY session to be able to write controls
I'm always frustrated when I have a use case where more than one PID would like to write to the controls, or when a control is reverted immediately upon process termination.  This also avoids the current "normal" use case for geopmsession when writing controls which requires a kill signal to be sent to the geopmsession process.

Instead of gating access to the write lock based on PID, use the process group determine ability to write controls and also use the process group lifetime to determine when control values are restored.

Considered allowing all children of the PID that first asks for the write lock to be enabled.

**Additional context**
This change in geopm service requirements will resolve a number of issues that arise when running the GEOPM HPC Runtime integration tests using the GEOPM service to get access to MSR controls.

cmcantalupo:
- Fixes #2037

Pretty simple change that will provide the feature described in issue #2037

### DIFF
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -756,11 +756,9 @@ class PlatformService(object):
             if self._write_pid is not None:
                 raise RuntimeError(f'The PID {client_pid} requested write access, but the geopm service already has write mode client with SID {self._write_pid}')
             if client_sid != client_pid:
-                self._active_sessions.add_client(
-                    client_sid,
-                    self._active_sessions.get_signals(client_pid),
-                    self._active_sessions.get_controls(client_pid),
-                    self._watch_client(client_sid))
+                uid = os.stat(f'/proc/{client_sid}/status').st_uid
+                user = pwd.getpwuid(uid).pw_name
+                self.open_session(user, client_sid)
             self._active_sessions.set_write_client(client_sid)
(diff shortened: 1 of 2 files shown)


service/geopmdpy/service.py
```diff
@@ -751,11 +751,18 @@ def _get_user_groups(self, user):
...
+        if self._write_pid != client_sid:
             if self._write_pid is not None:
-                raise RuntimeError(f'The PID {client_pid} requested write access, but the geopm service already has write mode client with PID {self._write_pid}')
-            self._active_sessions.set_write_client(client_pid)
-            self._write_pid = client_pid
+                raise RuntimeError(f'The PID {client_pid} requested write access, but the geopm service already has write mode client with SID {self._write_pid}')
+            if client_sid != client_pid:
+                self._active_sessions.add_client(
+                    client_sid,
+                    self._active_sessions.get_signals(client_pid),
```
cmcantalupo: Probably should figure out the permissions for the session leader PID independently.  This implementation is passing the lists for the requesting PID.  Note these lists will only be used in the case where the session leader PID makes a request.

bgeltz (approved):

PR #2024, opened 2021-12-02
Gpu frequency sweep

scoumeri:
- Relates to #2022  [bug report/feature request/story] from github issues
- Fixes #2023  change request from github issues.
- Follow on to PR #2010  

[Brief description of the change]
Added gpu frequency sweep capability to integration/experiments 

_Note_:

Prior to creating a pull request, please see the documentation at
either of the following links:

https://github.com/geopm/geopm/blob/dev/CONTRIBUTING.rst#change-request

https://geopm.github.io/contrib.html#change-request

about the process to change the GEOPM repository.

PR #2010, opened 2021-11-22
Public lhlawson cpugpuuncore fixed freq

scoumeri:
- Fixes #2008 
- Fixes #2009  change request from github issues.

[Brief description of the change]
Added fixed frequency agent to set accelerator, cpu, and uncore frequencies.  This is being set as a draft PR as the unit tests are not completed yet.

PR #1945, opened 2021-10-12
Reporter update for accelerator alias signals

lhlawson:
- Relates to #1940.  This PR will not close this feature on its own.  
- Related to #1943 
- Addresses #1946 
- Addresses #1653
- Follow on to PR #1944 

This PR adds per signal NVML behavior required for proper reporting, aliases that may be used in either NVML or LevelZero (but have not yet been propagated to the LevelZeroIOGroup), and finally adds support for accelerator power, frequency, and energy in the report (Reporter.cpp). 

Tasks keeping this in draft:
- [x] Unit testing of aliases
- [x] Unit or integration testing of report capability

PR #2045, opened 2021-12-14
Addition of the GPU Activity Agent to steer GPU frequency based upon GPU Compute Activity

### LINKED ISSUE #1906: As a user of GPU hardware I would like an agent that dynamically modulates GPU frequency in active regions based on GPU Compute Activity
GPU frequency steering during periods of idleness as described in #1905 does not reduce frequency in active regions (i.e. when GPU Utilization is non-zero).  In order to reduce frequency in the active regions additional metrics from are required, such as those discussed in #1902.  An agent that uses metrics of HW activity such compute (or SM) activity could provide better overall energy savings than an agent that uses utilization alone to decrease power only when the GPU is idle..

An agent that uses HW telemetry gathered from the GPU to reduce energy consumption by automatically selecting a lower frequency when doing so would have minimal impact on the performance of the GPU.  One such metric is SM_ACTIVE provided by DCGM, but this may be generalized to any Execution Unit activity metric.

#1905 with hand tuning of frequency in active regions is an option, but requires extensive workload characterization.

**Additional context**
This feature is a predicated on #1904 for NVIDIA GPUs, and may lead to updates of the work completed in #1787 for Intel GPUs.

Tasks:
- [x] #2646
- [x] #2647

### LINKED ISSUE #1908: Introduce new C++ class: GPUActivityAgent
> estimate 1
Parent [feature/story]: #1906 

Introduce new C++ class GPUActivityAgent with the following features:
- Monitoring of GPU Utilization and GPU Compute Activity
- Setting of GPU frequency based upon GPU Utilization and GPU Compute Activity

The primary differentiator between this agent and #1907 is that this agent targets dynamically changing frequency during regions of active compute, where #1907 uses fixed frequencies when the GPU is active (though it may be a series of fixed frequencies for varying levels of utilization).

Done criteria:
- [ ] Pure virtual interface class has been added with factory method
      NVMLDevicePool)
- [ ] Implementation modulates frequency to a fixed value when the GPU utilization is low
- [ ] Implementation modulates frequency to a fixed value when the GPU utilization is high
- [ ] Implementation further modulates frequency based upon compute activity (hw metrics) when GPU utilization is high

lhlawson:
- Fixes #1906  
- Fixes #1908

- Relies upon changes made in PR #1914 #1917 #1945 and should not be integrated prior to review those PRs
- Agent operates best when using Fefficient information obtained from GPU frequency sweeps.  Experiments providing GPU frequency sweeps through the integration infrastructure are introduced in PR #2024, #2031.  These may rely upon PR #2010 

This PR introduces an agent that modulates the GPU frequency within a set of user specified bounds based upon the GPU_COMPUTE_ACTIVITY signal.  For NVIDIA systems this is the SM_ACTIVE signal.  The user may also provide an optional input 'phi' which can bias the agents frequency selection towards energy efficiency (phi of 1.0) or performance (phi of 0.0).  The default behavior if phi = 0.5

Primary changes:
- addition of GEOPM_CXX_MAN_GPUActivityAgent.3.rst
- addition of geopm_agent_gpu_activity.7.rst
- src/GPUActivityAgent.cpp
- src/GPUActivityAgent.hpp
- test/GPUActivityAgentTest.cpp
- integration/experiment/energy_efficiency/Makefile.mk
- integration/experiment/energy_efficiency/gpu_activity.py
- integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm.py
- integration/experiment/energy_efficiency/run_gpu_activity_parres_nstream.py
- integration/frequency_sweep/gen_gpu_activity_policy_recommendation.py

Additional tasks to consider:
- [x] Integration tests based upon the run_gpu_activity_parress_dgemm/stream.py files
- [x] Experiment scripts to find the Fe (freq_efficient) policy input based upon the GPU frequency sweep of parres dgemm
- [x] Experiment script updates (energy_efficiency/gpu_activity.py) to automatically provide the 'optimal' config instead of a simple phi sweep with default HPE sx40 values.


integration/experiment/energy_efficiency/gpu_activity.py
```diff
@@ -0,0 +1,142 @@
...
+def trace_signals():
+    return []
+
+def launch_configs(output_dir, app_conf):
+    mach = machine.init_output_dir(output_dir)
+    sys_min = mach.frequency_min()
+    sys_max = mach.frequency_max()
+    sys_sticker = mach.frequency_sticker()
+    config_list = [
+                     {"ACCELERATOR_FREQ_MAX": 1530000000,
```
lhlawson: Update util.py to query frequency min, max, etc for GPUs

ronn/geopm_agent_gpu_activity.7.ronn
```diff
@@ -0,0 +1,161 @@
...
+the compute activity of each GPU as provided by the ACCELERATOR_COMPUTE_ACTIVITY
+signal and modified by the UTILIZATION_ACCELERATOR signal.
+
+The GPUActivityAgent scales frequency in the range of Fe to Fmax, where Fmax
+is provided via the policy as ACCELERATROR_FREQ_MAX and Fe is provided via
+the policy as ACCELERATOR_FREQ_EFFICIENT.  Low activity regions (compute activity
+of 0.0) run at the Fe frequency, high activity regions (compute activity of 1.0)
+run at the Fmax frequency, and regions in between the extremes run at a frequency
+selected via
+F = = Fe + (Fmax - Fe) * ACCELERATOR_COMPUTE_ACTIVITY/UTILIZATION_ACCELERATOR.
```
dannosliwcd:
```suggestion
F = Fe + (Fmax - Fe) * ACCELERATOR_COMPUTE_ACTIVITY/UTILIZATION_ACCELERATOR.
```

ronn/geopm_agent_gpu_activity.7.ronn
```diff
@@ -0,0 +1,161 @@
...
+UTILIZATION_ACCELERATOR is used to scale the ACCELERATOR_COMPUTE_ACTIVITY in order
+to scale frequency selection with the percentage of time a kernel is runnning on
```
dannosliwcd:
```suggestion
to scale frequency selection with the percentage of time a kernel is running on
```

ronn/geopm_agent_gpu_activity.7.ronn
```diff
@@ -0,0 +1,161 @@
...
+## AGENT BEHAVIOR HIGHLIGHTS
+Policies and samples are vectors of double precision values where each
+value has a meaning defined by the **geopm::Agent(3)** implementation.
+The Agent interface also provides methods for the Agent to extend
+reports and traces with additional Agent-specific information.
+
+* `Agent Name`:
+  Set the `--geopm-agent` launch option or `GEOPM_AGENT` environment
+  variable to "energy_efficient" and the Controller will select the
+  EnergyEfficientAgent for its control handler.  See
```
dannosliwcd: Can probably emit the --geopm-agent stuff, or make it specific to this agent

ronn/geopm_agent_gpu_activity.7.ronn
```diff
@@ -0,0 +1,161 @@
...
+  `ACCELERATOR_FREQ_EFFICIENT`: The minimum frequency in hertz that the algorithm is
+                                allowed to choose.  If NAN is passed, it will use
+                                (maximum frequency + minimum frequency) / 2 by default.
+
+
+  `ACCELERATOR_PHI`: The performance bias knob.  The value must be between
+                     0.0 and 1.0. If NAN is passed, it will use 0.5 by default.
+
+* `Agent Sample Definitions`:
+  N/A
```
dannosliwcd: Looks like the agent adds all of its traced signals as samples, outputting differenced values for energy and raw values for the rest.

service/docs/source/GEOPM_CXX_MAN_GPUActivityAgent.3.rst
```diff
@@ -0,0 +1,98 @@
...
+manual.
+
+Note that the ``GPUActivityAgent`` class is derived from `geopm::Agent(3) <GEOPM_CXX_MAN_Agent.3>`_ class.
+
+SYNOPSIS
+--------
+
+#include `<geopm/GPUActivityAgent.hpp> <https://github.com/geopm/geopm/blob/dev/src/GPUActivityAgent.hpp>`_\
+
+Link with ``-lgeopm`` **(MPI)** or ``-lgeopmpolicy`` **(non-MPI)**
```
bgeltz, line 11: Please add that this requires either `--enable-nvml` or `--enable-levelzero` at configure time.

service/docs/source/reference.rst
```diff
@@ -79,6 +79,7 @@ GEOPM Manual Pages
    GEOPM_CXX_MAN_EnergyEfficientAgent.3
    GEOPM_CXX_MAN_EnergyEfficientRegion.3
    GEOPM_CXX_MAN_Exception.3
+   GEOPM_CXX_MAN_GPUActivityAgent.3
```
bgeltz: Perhaps this should also be added to the `GEOPM HPC Runtime Agents` section below.

service/docs/source/reference.rst
```diff
@@ -79,6 +79,7 @@ GEOPM Manual Pages
    GEOPM_CXX_MAN_EnergyEfficientAgent.3
    GEOPM_CXX_MAN_EnergyEfficientRegion.3
    GEOPM_CXX_MAN_Exception.3
```
bgeltz: I'm also just noticing that the FrequencyMapAgent is not listed here.

service/docs/source/GEOPM_CXX_MAN_GPUActivityAgent.3.rst
```diff
@@ -0,0 +1,98 @@
...
+
+
+.. code-block:: c++
```
bgeltz: I went looking at the other docu to see if we listed out all the agent methods, and I don't see that elseware.  I question the value of this block.

To me it's just confusing.  I can always go read the header file if I'm curious.  My vote would be to remove this code block section.

bgeltz: I see it's in there for the EE agent now.  I think it should be removed from there as well.

service/docs/source/GEOPM_CXX_MAN_GPUActivityAgent.3.rst
```diff
@@ -0,0 +1,98 @@
...
+`geopm_agent_gpu_activity(7) <geopm_agent_gpu_activity.7.html>`_ man page.
+
+For more details on the implementation, see the doxygen
+page at <https://geopm.github.io/dox/classgeopm_1_1_gpu_activity.html>.
+
+CLASS METHODS
+-------------
+
+**TODO**
+
```
bgeltz:
```suggestion
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+
+
+
+
+DESCRIPTION
+-----------
+
+The goal of this Agent is to save GPU energy by scaling GPU frequency based upon
+the compute activity of each GPU as provided by the GPU_COMPUTE_ACTIVITY
+signal and modified by the GPU_UTILIZATIOR signal.
```
bgeltz:
```suggestion
signal and modified by the GPU_UTILIZATION signal.
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+
+The **GPUActivityAgent** scales frequency in the range of Fe to Fmax, where Fmax
+is provided via the policy as ``GPU_FREQ_MAX`` and Fe is provided via
+the policy as ``GPU_FREQ_EFFICIENT``.  Low activity regions (compute activity
+of 0.0) run at the Fe frequency, high activity regions (compute activity of 1.0)
+run at the Fmax frequency, and regions in between the extremes run at a frequency
```
bgeltz:
```suggestion
run at the Fmax frequency, and regions in between the extremes run at a frequency (F)
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+selected using the equation
```
bgeltz:
```suggestion
selected using the equation:
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+
+
+DESCRIPTION
+-----------
+
+The goal of this Agent is to save GPU energy by scaling GPU frequency based upon
+the compute activity of each GPU as provided by the GPU_COMPUTE_ACTIVITY
+signal and modified by the GPU_UTILIZATIOR signal.
+
+The **GPUActivityAgent** scales frequency in the range of Fe to Fmax, where Fmax
```
bgeltz:
```suggestion
The **GPUActivityAgent** scales frequency in the range of ``Fe`` to ``Fmax``, where ``Fmax``
```

I think it makes it more readable when the variables are offset from the rest of the text.  I'm in favor of doing this everywhere in this doc.

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+The GPUActivityAgent provides an optional input of phi that allows for biasing the
+frequency range used by the agent.  The default phi value of 0.5 provides frequency
+selection in the full range from Fe to Fmax.  A phi value less than 0.5 biases the
+agent towards higher frequencies by increasing the Fe value provided by the policy.
+In the extreme case (phi of 0) Fe will be raised to Fmax.  A phi value greater than
+0.5 biases the agent towards lower frequencies by reducing the Fmax value provided
+by the policy.  In the extreme case (phi of 1.0) Fmax will be lowered to Fe.
+
+For NVIDIA based systems the GPUActivityAgent attempts to set the
+DCGM::FIELD_UPDATE_RATE to 100ms, DCGM::MAX_STORAGE_TIME to 1s, and DCGM::MAX_SAMPLES
```
bgeltz:
```suggestion
``DCGM::FIELD_UPDATE_RATE`` to 100ms, ``DCGM::MAX_STORAGE_TIME`` to 1s, and ``DCGM::MAX_SAMPLES``
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+to 100.  While the DCGM documentation indicates that users should 'generally' query
```
bgeltz:
```suggestion
to 100.  While the DCGM documentation indicates that users should generally query
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+no faster than 100ms, the interface allows for setting the polling rate in the
+microsecond range. If the agent is intended to be used with workloads that exhibit
+extremely short phase behavioor a 1ms polling rate can be used.
```
bgeltz:
```suggestion
extremely short phase behavior a 1ms polling rate can be used.
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+     Per GPU time during the Region
+     of Interest (ROI) where ROI is determined as the
+     first sample of GPU activity to the last sample of GPU
+     activity.
+  ``GPU # Active Region Start Time``\ :
+     Per GPU start time for the Region
+     of Interest (ROI) where ROI is determined as the
+     first sample of GPU activity to the last sample of GPU
+     activity.
+  ``GPU # Active Region Start Time``\ :
```
bgeltz:
```suggestion
  ``GPU # Active Region Stop Time``\ :
```

service/docs/source/conf.py
```diff
@@ -140,6 +140,7 @@ def get_description(rst_file_name):
     "GEOPM_CXX_MAN_EnergyEfficientAgent.3",
     "GEOPM_CXX_MAN_EnergyEfficientRegion.3",
     "GEOPM_CXX_MAN_Exception.3",
```
bgeltz, line 152: Surprised Fmap is missing.  Not this PR's problem though.

service/docs/source/geopm.7.rst
```diff
@@ -268,87 +270,87 @@ GEOPM ENVIRONMENT VARIABLES
 ---------------------------
 
 
-* 
+*
```
bgeltz: Looks like this is all just dangling whitespace.

Was this here for a reason @cmcantalupo?

src/GPUActivityAgent.hpp
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (c) 2015, 2016, 2017, 2018, 2019, Intel Corporation
```
bgeltz:
```suggestion
 * Copyright (c) 2015 - 2022, Intel Corporation
```

Please make this change everywhere that it's needed in this PR.

src/GPUActivityAgent.hpp
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+#ifndef GPUACTIVITYAGENT_HPP_INCLUDE
+#define GPUACTIVITYAGENT_HPP_INCLUDE
+
+#include <vector>
+
+#include "Agent.hpp"
+#include "geopm_time.h"
+#include "config.h"
```
bgeltz: Please move `config.h` to the cpp.  Make it the first include in the list.

In general take a look at:
https://github.com/geopm/geopm/blob/dev/service/src/PlatformIO.cpp#L33

I'm trying to do this in all the files I edit when I see it.  I'm trying to follow the ordering pattern in deea1df1025c56390cfaaf16a138f4897f4b0bbd:

    config.h, PIOImp.hpp,
    C includes, STL C++ includes,
    GEOPM public includes,
    GEOPM private C includes, GEOPM private C++ includes.

src/GPUActivityAgent.hpp
```diff
@@ -0,0 +1,134 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef GPUACTIVITYAGENT_HPP_INCLUDE
+#define GPUACTIVITYAGENT_HPP_INCLUDE
+
+#include <vector>
```
bgeltz, line 10:
```suggestion
#include <vector>
#include <functional>
```

Needed for `std::function`

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,460 @@
...
+        if (all_names.find("DCGM::FIELD_UPDATE_RATE") != all_names.end()) {
+            // DCGM documentation indicates that users should query no faster than 100ms
+            // even though the interface allows for setting the polling rate in the us range.
+            // In practice reducing below the 100ms value has proven functional, but should only
+            // be attempted if there is a proven need to catch short phase behavior that cannot
+            // be accomplished with the default settings.
+            m_platform_io.write_control("DCGM::FIELD_UPDATE_RATE", GEOPM_DOMAIN_BOARD, 0, 0.1); //100ms
+            m_platform_io.write_control("DCGM::MAX_STORAGE_TIME", GEOPM_DOMAIN_BOARD, 0, 1);
+            m_platform_io.write_control("DCGM::MAX_SAMPLES", GEOPM_DOMAIN_BOARD, 0, 100);
+        }
```
bgeltz: Thinking about this code block when we're using LevelZero and not NVML...

I'm wondering if this code that effectively initializes the specific IOGroup (DCGM in this case) actually belongs inside the IOGroup itself.  It struck me as odd that the agent here needs to understand the specifics of the IOGroup that is currently compiled in.

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,460 @@
...
+        if (phi > 0.5) {
+            //Energy Biased.  Scale F_max down to F_efficient based upon phi value
+            //Active region phi usage
+            f_max = std::max(f_efficient, f_max - f_range * (phi-0.5) / 0.5);
+        }
+        else if (phi < 0.5) {
+            //Perf Biased.  Scale F_efficient up to F_max based upon phi value
+            //Active region phi usage
+            f_efficient = std::min(f_max, f_efficient + f_range * (0.5-phi) / 0.5);
+        }
```
bgeltz, line 169: What about when `phi = 0.5`?

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+        //Update Policy
+        in_policy[M_POLICY_GPU_FREQ_MAX] = f_max;
+        in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = f_efficient;
+
+    }
+
+    // Distribute incoming policy to children
+    void GPUActivityAgent::split_policy(const std::vector<double>& in_policy,
+                                    std::vector<std::vector<double> >& out_policy)
```
```suggestion
                                        std::vector<std::vector<double> >& out_policy)
```

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,460 @@
...
+            double gpu_compute_activity = m_gpu_compute_activity.at(domain_idx).value;
+            // gpu Utilization - Used to scale activity for short GPU phases
+            double gpu_utilization = m_gpu_utilization.at(domain_idx).value;
+
+            // Default to F_max
+            double f_request = m_f_max;
+
+            if (!std::isnan(gpu_compute_activity)) {
+                // Boundary Checking
+                gpu_compute_activity = std::min(gpu_compute_activity, 1.0);
```
bgeltz: Should this be an error condition?  When `gpu_compute_activity` is > 1.0?

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,460 @@
...
+                // frequency sensitive.  If a workload has a compute activity of 0.5, and
+                // is resident on the GPU for 50% of cycles (0.5) it is treated as having
+                // a 1.0 compute activity value
+                //
+                // For inactive regions the frequency selection is simply the efficient
+                // frequency from system characterization.
+                //
+                // This approach assumes the efficient frequency is suitable as both a
+                // baseline for active regions and and inactive regions. This is generally
+                // true of the efficient frequency is low power enough at idle due to clock
```
```suggestion
                // true when the efficient frequency consumes low power at idle due to clock
```

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,460 @@
...
+                else {
+                    if (m_gpu_active_region_stop.at(domain_idx) == 0) {
+                        m_gpu_active_region_stop.at(domain_idx) = m_time.value;
+                        m_gpu_active_energy_stop.at(domain_idx) = m_gpu_energy.at(domain_idx).value;
+                    }
+                }
+            }
+
+            // Frequency clamping
+            f_request = std::min(f_request, m_f_max);
```
bgeltz, line 287: Might want to add a `ifdef GEOPM_DBUG` with some prints that your frequency was clamped one way or the other unless it's easy to determine from the reports and traces.

bgeltz (changes requested):

src/GPUActivityAgent.cpp
lhlawson: When phi=0.5 we allow the agent to steer frequency in the range of Fe to Fmax without modifying either Fe or Fmax.

src/GPUActivityAgent.cpp
lhlawson: gpu_compute_activity is provided by DCGM through the nv-hostengine process.  Depending on how nv-hostengine is configured and how noisy the data hw & nv-hostengine return there's the possibility the values could exceed 1.0.  When seen in testing this behavior was transient so the approach of simply clamping to the maximum expected value of 1.0 was used vs GEOPM throwing an error due to bad data inputs.

The end result of gpu_compute_activity > 1.0 is the algorithm selecting the maximum frequency (being performance biased) provided by the user/admin.

src/GPUActivityAgent.cpp
lhlawson: For this agent the current default values used by the DCGMDevicePool in dev should suffice, but the user may want to change them if their workload has shorter phases.

We could remove this code if we both assume that a user will do a `geompwrite DCGM::FIELD_UPDATE_RATE board 0 <better value>` prior to their if there's a known better value _and the DCGM setting is persistent for the duration of their time on the system_.  Whenever the dcgmWatchFields function is called in the DCGMDevicePool it needs to provide the three values being written, and prior to the GEOPM service I don't believe there a way to modify those outside the agent that would persist until the agent ran.

service/docs/source/GEOPM_CXX_MAN_GPUActivityAgent.3.rst
lhlawson: Added that it requires --enable-nvml and --enable-dcgm on NVIDIA systems.  
Agent and/or IO Group updates are likely required for LevelZero systems, I'll file an issue/feature request for that update (including documentation)

service/docs/source/GEOPM_CXX_MAN_GPUActivityAgent.3.rst
lhlawson: Agreed.

lhlawson:
Latest updates include a SAMPLE_PERIOD policy value that allows the end user to specify sample period/wait time.  This should help address any control loop timing issues that stem from vastly different GPU counts or GPU signal & control access time varying from system to system.

src/GPUActivityAgent.cpp
lhlawson: Added a "GPU Clipped Frequency Request" value to the report when GEOPM_DEBUG is defined.  I also moved the per GPU Region of interest start & stop time to the GEOPM_DEBUG case, as per GPU region time is enough for most non-debug cases.

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,464 @@
...
+                            std::to_string(in_policy[M_POLICY_GPU_FREQ_MAX]) +
+                            ".", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        // Not all gpus provide an 'efficient' frequency signal, and the
+        // value provided by the policy may not be valid.  In this case approximating
+        // f_efficient as midway between F_min and F_max is reasonable.
+        if (std::isnan(in_policy[M_POLICY_GPU_FREQ_EFFICIENT])) {
+            in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = (in_policy[M_POLICY_GPU_FREQ_MAX]
+                                                             + gpu_min_freq) / 2;
```
```suggestion
            in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = (in_policy[M_POLICY_GPU_FREQ_MAX] + gpu_min_freq) / 2;
```

or align it one space in from the open paren on the line above:
```suggestion
            in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = (in_policy[M_POLICY_GPU_FREQ_MAX]
                                                      + gpu_min_freq) / 2;
```

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,464 @@
...
+        return {{"Wait time (sec)", std::to_string(M_WAIT_SEC)}};
+    }
+
+    // Adds number of frquency requests to the per-node section of the report
+    std::vector<std::pair<std::string, std::string> > GPUActivityAgent::report_host(void) const
+    {
+        std::vector<std::pair<std::string, std::string> > result;
+
+        result.push_back({"GPU Frequency Requests", std::to_string(m_gpu_frequency_requests)});
+#ifdef GEOPM_DEBUG
```
bgeltz: Using a debug build is going to slow down the overall runtime.  I'm not sure how big the impact would be on this Agent without doing some testing. It seems that if you're trying to study something around clipping, you ought to use a release build to ensure things run at full speed.

Why not just include this all the time?

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,464 @@
...
+        for (int domain_idx = 0; domain_idx < M_NUM_GPU; ++domain_idx) {
+            double energy_stop = m_gpu_active_energy_stop.at(domain_idx);
+            double energy_start = m_gpu_active_energy_start.at(domain_idx);
+            double region_stop = m_gpu_active_region_stop.at(domain_idx);
+            double region_start =  m_gpu_active_region_start.at(domain_idx);
+            result.push_back({"GPU " + std::to_string(domain_idx) +
+                              " Active Region Energy", std::to_string(energy_stop - energy_start)});
+            result.push_back({"GPU " + std::to_string(domain_idx) +
+                              " Active Region Time", std::to_string(region_stop - region_start)});
+#ifdef GEOPM_DEBUG
```
bgeltz: Again, these region times might look pretty different between a debug and release build.  If that matters, I'd just do this all the time.

src/GPUActivityAgent.cpp
lhlawson: As long as it doesn't add too much to the report including it all the time seems reasonable.

src/GPUActivityAgent.cpp
lhlawson: These I'd omit, as ACTIVE REGION TIME provides most of the relevant information.

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+      Set the ``--geopm-agent`` launch option or ``GEOPM_AGENT`` environment
+      variable to ``"gpu_activity"`` and the Controller will select the
+      EnergyEfficientAgent for its control handler.  See
+      `geopm_launch(1) <geopm_launch.1.html>`_ and `geopm(7) <geopm.7.html>`_ for more information about
+      launch options and environment variables.
+
+*
+  **Agent Policy Definitions**:
+
+      The ``Fe``, ``Fmax``, and  and maximum frequency are policy values.
```
lhlawson:
```suggestion
      ``Fe``, ``Fmax``, and ``phi`` are policy values.
```

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,158 @@
...
+
+  ``GPU_FREQ_EFFICIENT``\ :
+      The minimum frequency in hertz that the algorithm is
+      allowed to choose.  If NAN is passed, it will use
+      (maximum frequency + minimum frequency) / 2 by default.
+
+  ``GPU_PHI``\ :
+      The performance bias knob.  The value must be between
+      0.0 and 1.0. If NAN is passed, it will use 0.5 by default.
+
```
lhlawson, line 89: Add Sample Rate

integration/experiment/energy_efficiency/gpu_activity.py
```diff
@@ -0,0 +1,128 @@
...
+def trace_signals():
+    return []
+
+def launch_configs(output_dir, app_conf):
+    mach = machine.init_output_dir(output_dir)
+    sys_min = mach.frequency_min()
+    sys_max = mach.frequency_max()
+    sys_sticker = mach.frequency_sticker()
+    config_list = [
+                     {"GPU_FREQ_MAX": float("NAN"),
```
lhlawson: Add command line argument to provide GPU_FREQ_MAX and GPU_FREQ_EFFICIENT.   Generate this list from phi 0-1 in 0.1 increments instead of hard-coding

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,146 @@
...
+--------
+
+:doc:`geopm(7) <geopm.7.html>`\ ,
+:doc:`geopm_agent_monitor(7) <geopm_agent_monitor.7.html>`\ ,
+:doc:`geopm_agent_energy_efficient(7) <geopm_agent_energy_efficient.7.html>`\ ,
+:doc:`geopm::Agent(3) <GEOPM_CXX_MAN_Agent.3.html>`\ ,
+:doc:`geopm_agent_c(3) <geopm_agent_c.3.html>`\ ,
+:doc:`geopm_prof_c(3) <geopm_prof_c.3.html>`\ ,
+:doc:`geopmagent(1) <geopmagent.1.html>`\ ,
+:doc:`geopmlaunch(1) <geopmlaunch.1.html>`
```
lhlawson:
```suggestion
:doc:`geopm(7) <geopm.7>`\ ,
:doc:`geopm_agent_monitor(7) <geopm_agent_monitor.7>`\ ,
:doc:`geopm_agent_energy_efficient(7) <geopm_agent_energy_efficient.7>`\ ,
:doc:`geopm::Agent(3) <GEOPM_CXX_MAN_Agent.3>`\ ,
:doc:`geopm_agent_c(3) <geopm_agent_c.3>`\ ,
:doc:`geopm_prof_c(3) <geopm_prof_c.3>`\ ,
:doc:`geopmagent(1) <geopmagent.1>`\ ,
:doc:`geopmlaunch(1) <geopmlaunch.1>`
```

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,453 @@
...
+    // Used by the factory to create objects of this type
+    std::unique_ptr<Agent> GPUActivityAgent::make_plugin(void)
+    {
+        return geopm::make_unique<GPUActivityAgent>();
+    }
+
+    // Describes expected policies to be provided by the resource manager or user
+    std::vector<std::string> GPUActivityAgent::policy_names(void)
+    {
+        return {"GPU_FREQ_MAX", "GPU_FREQ_EFFICIENT", "GPU_PHI", "SAMPLE_PERIOD"};
```
lhlawson:
```suggestion
        return {"GPU_FREQ_MAX", "GPU_FREQ_EFFICIENT", "GPU_PHI"};
```
Post 2.0 Agents are expected to use a more general geopmlaunch option to set sample period/control rate.  SAMPLE_PERIOD should be removed from the agent and supporting collateral

integration/test/test_gpu_activity_agent.py
```diff
@@ -0,0 +1,165 @@
...
+from types import SimpleNamespace
+
+import geopmpy.agent
+import geopmpy.io
+from yaml import load
+try:
+    from yaml import CSafeLoader as SafeLoader
+except ImportError:
+    from yaml import SafeLoader
+from collections import defaultdict
```
lhlawson:
```suggestion
```

integration/test/test_gpu_activity_agent.py
```diff
@@ -0,0 +1,165 @@
...
+        experiment_args = SimpleNamespace(
+            output_dir=output_dir,
+            node_count=1,
+            parres_cores_per_node=None,
+            parres_gpus_per_node=None,
+            parres_cores_per_rank=1,
+            parres_init_setup=None,
+            parres_exp_setup=None,
+            parres_teardown=None,
+            parres_args=None,
```
lhlawson, line 76: It's beneficial to increase the parres nstream input size due to the relatively short runtime with the default settings, but not necessarily required.

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+            m_gpu_freq_max_control.push_back(control{m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL",
+                                                     GEOPM_DOMAIN_GPU,
+                                                     domain_idx), NAN});
+        }
+    }
+
+    // Validate incoming policy and configure default policy requests.
+    void GPUActivityAgent::validate_policy(std::vector<double> &in_policy) const
+    {
+        assert(in_policy.size() == M_NUM_POLICY);
```
avilcheslopez: GEOPM_DEBUG_ASSERT

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+        //Update Policy
+        in_policy[M_POLICY_GPU_FREQ_MAX] = f_max;
+        in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = f_efficient;
+    }
+
+    // Distribute incoming policy to children
+    void GPUActivityAgent::split_policy(const std::vector<double>& in_policy,
+                                        std::vector<std::vector<double> >& out_policy)
+    {
+        assert(in_policy.size() == M_NUM_POLICY);
```
avilcheslopez: GEOPM_DEBUG_ASSERT

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+    // Indicate whether to send samples up to the parent
+    bool GPUActivityAgent::do_send_sample(void) const
+    {
+        return false;
+    }
+
+    void GPUActivityAgent::adjust_platform(const std::vector<double>& in_policy)
+    {
+        assert(in_policy.size() == M_NUM_POLICY);
```
avilcheslopez: GEOPM_DEBUG_ASSERT

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+            if (!std::isnan(gpu_core_activity)) {
+                // Boundary Checking
+                gpu_core_activity = std::min(gpu_core_activity, 1.0);
+
+                // Frequency selection is based upon the gpu compute activity.
+                // For active regions this means that we scale with the amount of work
+                // being done (such as SM_ACTIVE for NVIDIA GPUs).
+                //
+                // The compute activity is scaled by the GPU Utilization, to help
+                // address the issues that come from workloads have short phases that are
```
avilcheslopez: ...workloads ^that^ have...

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+                // frequency sensitive.  If a workload has a compute activity of 0.5, and
+                // is resident on the GPU for 50% of cycles (0.5) it is treated as having
+                // a 1.0 compute activity value
+                //
+                // For inactive regions the frequency selection is simply the efficient
+                // frequency from system characterization.
+                //
+                // This approach assumes the efficient frequency is suitable as both a
+                // baseline for active regions and and inactive regions. This is generally
```
avilcheslopez: and and

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+                ++m_gpu_frequency_clipped;
+            }
+            f_request = std::min(f_request, m_f_max);
+            f_request = std::max(f_request, m_f_efficient);
+
+            // Store frequency request
+            gpu_freq_request.push_back(f_request);
+        }
+
+        if (!gpu_freq_request.empty()) {
```
avilcheslopez: Why is this check necessary? Is it possible M_NUM_GPU could be 0? Will this agent load if there are no GPUs available?

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,436 @@
...
+    // If controls have a valid updated value write them.
+    bool GPUActivityAgent::do_write_batch(void) const
+    {
+        return m_do_write_batch;
+    }
+
+    // Read signals from the platform and calculate samples to be sent up
+    void GPUActivityAgent::sample_platform(std::vector<double> &out_sample)
+    {
+        assert(out_sample.size() == M_NUM_SAMPLE);
```
avilcheslopez: GEOPM_DEBUG_ASSERT

avilcheslopez (commented):
Just leaving some minor comments...

src/GPUActivityAgent.cpp
lhlawson: It's no longer necessary.  In previous revisions it was possible to hit a condition where the gpu_freq_request had not been initialized.  It may now be removed.

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,440 @@
...
+        , M_WAIT_SEC(0.020) // 20ms wait default
+        , M_POLICY_PHI_DEFAULT(0.5)
+        , M_GPU_ACTIVITY_CUTOFF(0.05)
+        , M_NUM_GPU(m_platform_topo.num_domain(GEOPM_DOMAIN_GPU))
+        , m_do_write_batch(false)
+    {
+        geopm_time(&m_last_wait);
+    }
+
+    // Push signals and controls for future batch read/write
```
avilcheslopez: This comment seems to apply to the init_platform_io() method.

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,440 @@
...
+                           "GPUActivityAgent::split_policy(): policy vector incorrectly sized");
+        for (auto &child_pol : out_policy) {
+            child_pol = in_policy;
+        }
+    }
+
+    // Indicate whether to send the policy down to children
+    bool GPUActivityAgent::do_send_policy(void) const
+    {
+        return true;
```
avilcheslopez: Should this be false? Policy needs to be sent once in the beginning.

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,440 @@
...
+        for (int domain_idx = 0; domain_idx < M_NUM_GPU; ++domain_idx) {
+            m_gpu_active_region_start.push_back(0.0);
+            m_gpu_active_region_stop.push_back(0.0);
+            m_gpu_active_energy_start.push_back(0.0);
+            m_gpu_active_energy_stop.push_back(0.0);
+        }
+
+        if (level == 0) {
+            init_platform_io();
+        }
```
avilcheslopez, line 64: Is level == 0 the root? Does this agent need to act differently for children and parent?

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,439 @@
...
+                                         GEOPM_DOMAIN_GPU,
+                                         domain_idx), NAN});
+            m_gpu_energy.push_back({m_platform_io.push_signal("GPU_ENERGY",
+                                    GEOPM_DOMAIN_GPU,
+                                    domain_idx), NAN});
+        }
+
+        m_time = {m_platform_io.push_signal("TIME", GEOPM_DOMAIN_BOARD, 0), NAN};
+
+        for (int domain_idx = 0; domain_idx < M_NUM_GPU; ++domain_idx) {
```
avilcheslopez: Since this "for" loop iterates over the same values as the one above, it seems like they could be merged.

lhlawson:
- [x] Add info on system characterization to docs (what workload, how, etc)
- [x] Add info on generating policy
- [x] JSON policy requires double quotes for gen_pol*

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,161 @@
...
+
+
+
+
+
+Description
+-----------
+
+The goal of this Agent is to save GPU energy by scaling GPU frequency based upon
+the compute activity of each GPU as provided by the GPU_COMPUTE_ACTIVITY
```
avilcheslopez: GPU_CORE_ACTIVITY

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,161 @@
...
+signal and modified by the GPU_UTILIZATION signal.
+
+The **GPUActivityAgent** scales frequency in the range of ``Fe`` to ``Fmax``, where ``Fmax``
+is provided via the policy as ``GPU_FREQ_MAX`` and ``Fe`` is provided via
+the policy as ``GPU_FREQ_EFFICIENT``.  Low activity regions (compute activity
+of 0.0) run at the ``Fe`` frequency, high activity regions (compute activity of 1.0)
+run at the ``Fmax`` frequency, and regions in between the extremes run at a frequency (F)
+selected using the equation:
+
+``F = Fe + (Fmax - Fe) * GPU_COMPUTE_ACTIVITY/GPU_UTILIZATION``
```
avilcheslopez: GPU_CORE_ACTIVITY

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,161 @@
...
+
+``GPU_UTILIZATION`` is used to scale the ``GPU_COMPUTE_ACTIVITY`` in order
```
avilcheslopez: GPU_CORE_ACTIVITY

integration/experiment/energy_efficiency/gpu_activity.py
```diff
@@ -0,0 +1,89 @@
...
+                                enable_traces=args.enable_traces,
+                                enable_profile_traces=args.enable_profile_traces)
+
+def main(app_conf, **defaults):
+    parser = argparse.ArgumentParser()
+    setup_run_args(parser)
+    parser.set_defaults(**defaults)
+    args, extra_args = parser.parse_known_args()
+    launch(app_conf=app_conf, args=args,
+           experiment_cli_args=extra_args)
```
avilcheslopez, line 89: Note: minor opportunity to refactor, as all experiment dir main modules have the exact same ``main()``.

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+        void SetUp();
+        void TearDown();
+        static const int M_NUM_CPU = 1;
+        static const int M_NUM_BOARD = 1;
+        static const int M_NUM_GPU = 1;
+        std::unique_ptr<GPUActivityAgent> m_agent;
+        std::vector<double> m_default_policy;
+        size_t m_num_policy;
+        double m_freq_min;
+        double m_freq_max;
```
avilcheslopez: Could & should be const

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+    EXPECT_EQ(m_freq_max, policy[FREQ_MAX]);
+    EXPECT_EQ((policy[FREQ_MAX] + m_freq_min) / 2, policy[FREQ_EFFICIENT]);
+    EXPECT_EQ(0.5, policy[PHI]);
+
+    // non-default policy is accepted
+    // setup & load policy
+    policy[FREQ_MAX] = m_freq_max;
+    policy[FREQ_EFFICIENT] = m_freq_max / 2;
+    policy[PHI] = 0.1;
+    m_agent->validate_policy(policy);
```
avilcheslopez: Expect that it doesn't throw. Same applies to all calls of validate_policy(), really.

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+TEST_F(GPUActivityAgentTest, adjust_platform_high)
+{
+    //TODO: Setup f_max, min etc...via read
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_min));
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
+    std::vector<double> policy;
+    policy = m_default_policy;
```
avilcheslopez: Merge these two lines: std::vector<double> policy = m_default_policy

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
+
+    const std::vector<double> empty(m_num_policy, NAN);
+    std::vector<double> policy;
+
+    // default policy is accepted
+    // load default policy
+    policy = m_default_policy;
```
avilcheslopez: Merge these two lines: std::vector policy = m_default_policy

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+TEST_F(GPUActivityAgentTest, adjust_platform_medium)
+{
+    //TODO: Setup f_max, min etc...via read
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_min));
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
+    std::vector<double> policy;
+    policy = m_default_policy;
```
avilcheslopez: Merge these two lines: std::vector policy = m_default_policy

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+TEST_F(GPUActivityAgentTest, adjust_platform_low)
+{
+    //TODO: Setup f_max, min etc...via read
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_min));
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
+    std::vector<double> policy;
+    policy = m_default_policy;
```
avilcheslopez: Merge these two lines: std::vector policy = m_default_policy

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+TEST_F(GPUActivityAgentTest, adjust_platform_zero)
+{
+    //TODO: Setup f_max, min etc...via read
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_min));
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
+    std::vector<double> policy;
+    policy = m_default_policy;
```
avilcheslopez: Merge these two lines: std::vector policy = m_default_policy

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+TEST_F(GPUActivityAgentTest, adjust_platform_signal_out_of_bounds)
+{
+    //TODO: Setup f_max, min etc...via read
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_min));
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
+    std::vector<double> policy;
+    policy = m_default_policy;
```
avilcheslopez: Merge these two lines: std::vector policy = m_default_policy

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+            PHI = 2,
+        };
+
+        void SetUp();
+        void TearDown();
+        static const int M_NUM_CPU = 1;
+        static const int M_NUM_BOARD = 1;
+        static const int M_NUM_GPU = 1;
+        std::unique_ptr<GPUActivityAgent> m_agent;
+        std::vector<double> m_default_policy;
```
avilcheslopez: Could and should be const

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+    policy[FREQ_MAX] = NAN;
+    policy[FREQ_EFFICIENT] = NAN;
+    policy[PHI] = 1.1;
+    GEOPM_EXPECT_THROW_MESSAGE(m_agent->validate_policy(policy), GEOPM_ERROR_INVALID,
+                               "POLICY_GPU_PHI value out of range");
+}
+
+TEST_F(GPUActivityAgentTest, adjust_platform_high)
+{
+    //TODO: Setup f_max, min etc...via read
```
avilcheslopez: Is this really necessary?

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+                .WillRepeatedly(Return(m_freq_max - 1)); //Any non-m_freq_max value will work here
+    EXPECT_CALL(*m_platform_io, sample(GPU_ENERGY_IDX))
+                .WillRepeatedly(Return(123456789));
+    m_agent->sample_platform(tmp);
+
+    //Adjust
+    //Check frequency
+    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQUENCY_CONTROL_MIN_IDX, m_freq_max)).Times(M_NUM_GPU);
+    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQUENCY_CONTROL_MAX_IDX, m_freq_max)).Times(M_NUM_GPU);
+    m_agent->adjust_platform(policy);
```
avilcheslopez: Opportunity to refactor: similar code across all "adjust_platform_*" tests

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+TEST_F(GPUActivityAgentTest, adjust_platform_high)
+{
+    //TODO: Setup f_max, min etc...via read
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_min));
+    EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", _, _)).WillRepeatedly(
+                Return(m_freq_max));
+    EXPECT_CALL(*m_platform_topo, num_domain(GEOPM_DOMAIN_BOARD)).WillRepeatedly(Return(M_NUM_BOARD));
```
avilcheslopez: This block of code is repeated at the beginning of every test, so it could be moved into the ``SetUp()`` method.

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+using ::testing::_;
+using ::testing::Invoke;
+using ::testing::Sequence;
+using ::testing::Return;
+using ::testing::AtLeast;
+using geopm::GPUActivityAgent;
+using geopm::PlatformTopo;
+
+class GPUActivityAgentTest : public :: testing :: Test
```
avilcheslopez, line 36: Additional tests could be added for:
- do_send_policy()
- do_send_sample()
- policy_names()
- report_host()?
- do_write_batch()?

integration/experiment/gpu_frequency_sweep/gen_gpu_activity_policy_recommendation.py
```diff
@@ -0,0 +1,93 @@
...
+    df = extract_columns(full_df)
+    gpu_freq_efficient = policy_efficient_energy(df)
+    # This script is not intended to provide an assessment of maximum frequency,
+    # or phi.  As such NAN is provided for these policy values.
+    # The associated agent interprets these as the related system or agent default
+    # config, ex: GPU_FREQ_MAX: NAN --> HW Maximum Frequency, GPU_PHI --> 0.5
+    # (balanced mode).
+    policy = {"GPU_FREQ_MAX" : float('nan'),
+                "GPU_FREQ_EFFICIENT" : gpu_freq_efficient,
+                "GPU_PHI" : float('nan')}
```
lhlawson:
```suggestion
              "GPU_FREQ_EFFICIENT" : gpu_freq_efficient,
              "GPU_PHI" : float('nan')}
```

integration/test/test_gpu_activity_agent.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # TODO:
+        # Once the GPU Frequency sweep infrastructure is added a full
+        # characterization integration test should be added that does
+        # a GPU frequency sweep of parres dgemm, parses the frequency
+        # sweep using the gen_gpu_activity_policy_recommendation.py
+        # script, and uses the output provided.
+        #
+        # This is a less time consuming version of that approach,
+        # where efficient freq is estimated based on min and max
```
lhlawson, line 49: This should be tackled after merging #2024

integration/test/test_gpu_activity_agent.py
```diff
@@ -0,0 +1,138 @@
...
+from integration.test import util
+from integration.test import geopm_test_launcher
+from experiment.energy_efficiency import gpu_activity
+from apps.parres import parres
+
+@util.skip_unless_config_enable('beta')
+@util.skip_unless_do_launch()
+@util.skip_unless_gpu()
+@util.skip_unless_workload_exists("apps/parres/Kernels/Cxx11/dgemm-mpi-cublas")
+@util.skip_unless_workload_exists("apps/parres/Kernels/Cxx11/nstream-mpi-cuda")
```
lhlawson, line 31: If we remove these the test will fail on run instead of skipping.  This might be a better default behavior if we only plan to run this script when the workloads have been built.

avilcheslopez: Note from Brad: leave skips. File issue for feature request: "as a developer of GEOPM I would like to be able to run weekly tests vs. actual workloads in the integration apps directory."

Makefile.am
```diff
@@ -115,6 +115,7 @@ roff_man_inst = man/GEOPM_CXX_MAN_Agent.3 \
                 # end
 
 roff_man_noinst = man/GEOPM_CXX_MAN_Comm.3 \
+                  man/GEOPM_CXX_MAN_GPUActivityAgent.3 \
```
bgeltz: Any man pages we need to add should also be behind the beta flag.  This should removed, and the chapter 7 included in the ```roff_man_beta``` section.

bgeltz (changes requested):

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -0,0 +1,162 @@
...
+
+Example Policy
+--------------
+
+An example policy generated using the parres dgemm workloads is
+provided below. ::
+
+    {"GPU_FREQ_MAX": NaN,
+     "GPU_FREQ_EFFICIENT": 982000000.0,
+      "GPU_PHI": NaN}
```
bgeltz:
```suggestion
     "GPU_PHI": NaN}
```
Off by 1

integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm.py
```diff
@@ -0,0 +1,24 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_activity.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_dgemm_appconf(mach, args)
+    gpu_activity.launch(app_conf=app_conf, args=args,
+                   experiment_cli_args=extra_args)
```
bgeltz: Indentation issue

integration/experiment/energy_efficiency/run_gpu_activity_parres_nstream.py
```diff
@@ -0,0 +1,24 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_activity.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_nstream_appconf(mach, args)
+    gpu_activity.launch(app_conf=app_conf, args=args,
+                   experiment_cli_args=extra_args)
```
bgeltz: Indentation issue

src/GPUActivityAgent.hpp
```diff
@@ -0,0 +1,110 @@
...
+            const PlatformTopo &m_platform_topo;
+            geopm_time_s m_last_wait;
+            double M_WAIT_SEC;
+            const double M_POLICY_PHI_DEFAULT;
+            const double M_GPU_ACTIVITY_CUTOFF;
+            const int M_NUM_GPU;
+            bool m_do_write_batch;
+            bool m_do_send_policy;
+
+            struct signal
```
bgeltz:
```suggestion
            struct m_signal
```

src/GPUActivityAgent.hpp
```diff
@@ -0,0 +1,110 @@
...
+            {
+                int batch_idx;
+                double value;
+            };
+
+            struct control
```
bgeltz:
```suggestion
            struct m_control
```
We should've done this for the CPU-CA too.

test/GPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,319 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "config.h"
+
+#include <stdlib.h>
+#include <iostream>
+#include <fstream>
```
bgeltz:
```suggestion
```
Can these all be removed?  If not, remove as much as possible.

bgeltz (changes requested):

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    // Indicate whether to send the policy down to children
+    bool GPUActivityAgent::do_send_policy(void) const
+    {
+        return m_do_send_policy;
+    }
+
+    void GPUActivityAgent::aggregate_sample(const std::vector<std::vector<double> > &in_sample,
+                                        std::vector<double>& out_sample)
```
bgeltz: Indentation

bgeltz (changes requested):

integration/experiment/gpu_frequency_sweep/gen_gpu_activity_policy_recommendation.py
```diff
@@ -0,0 +1,93 @@
...
+
+    try:
+        df = geopmpy.io.RawReportCollection('*report', dir_name=args.path).get_app_df()
+    except RuntimeError:
+        sys.stderr.write('<geopm> Error: No report data found in ' + path + \
+                         '; run a frequency sweep before using this analysis.\n')
+        sys.exit(1)
+
+    output = main(df)
+    sys.stdout.write("AGENT POLICY:\n")
```
dannosliwcd, line 92:
```suggestion
```
If the tool only prints the generated policy to stdout, then users can `./gen_gpu_activity_policy_recommendation.py > /path/to/policy/file`

dannosliwcd: Could alternatively add a `--output=/path/to/output/file` option. But for a tool that generates the contents of only  one file, `generate > file` feels more typical.

integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm.py
```diff
@@ -0,0 +1,24 @@
...
+from apps.parres import parres
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_activity.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_dgemm_appconf(mach, args)
```
lhlawson:
```suggestion
    app_conf = parres.create_dgemm_cuda_appconf(mach, args)
```

integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm.py
```diff
@@ -0,0 +1,24 @@
...
+'''
+Run ParRes dgemm with the gpu_activity agent.
+'''
+import argparse
+
+from experiment import machine
+from experiment.energy_efficiency import gpu_activity
+from apps.parres import parres
+
+
```
lhlawson: rename file to run_gpu_activity_parres_dgemm_cuda.py

integration/experiment/energy_efficiency/run_gpu_activity_parres_nstream.py
```diff
@@ -0,0 +1,24 @@
...
+'''
+Run ParRes nstream with the gpu_activity agent.
+'''
+import argparse
+
+from experiment import machine
+from experiment.energy_efficiency import gpu_activity
+from apps.parres import parres
+
+
```
lhlawson: rename file to run_gpu_activity_parres_nstream_cuda.py

integration/experiment/energy_efficiency/run_gpu_activity_parres_nstream.py
```diff
@@ -0,0 +1,24 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_activity.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_nstream_appconf(mach, args)
```
lhlawson:
```suggestion
    app_conf = parres.create_nstream_cuda_appconf(mach, args)
```

integration/experiment/energy_efficiency/Makefile.mk
```diff
@@ -17,6 +18,8 @@ EXTRA_DIST += integration/experiment/energy_efficiency/barrier_frequency_sweep.p
               integration/experiment/energy_efficiency/run_cpu_activity_nasft.py \
               integration/experiment/energy_efficiency/run_cpu_activity_nekbone.py \
               integration/experiment/energy_efficiency/run_cpu_activity_pennant.py \
+              integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_cuda.py \
+              integration/experiment/energy_efficiency/run_gpu_activity_parres_nstream_cuda.py \
```
bgeltz, line 22:
```suggestion

              integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_cuda.py \
              integration/experiment/energy_efficiency/run_gpu_activity_parres_nstream_cuda.py \
                            integration/experiment/energy_efficiency/run_cpu_activity_pennant.py \
```

integration/experiment/Makefile.mk
```diff
@@ -21,4 +21,5 @@ include integration/experiment/frequency_sweep/Makefile.mk
 include integration/experiment/monitor/Makefile.mk
 include integration/experiment/power_sweep/Makefile.mk
 include integration/experiment/trace_analysis/Makefile.mk
+include integration/experiment/gpu_frequency_sweep/Makefile.mk
```
bgeltz:
```suggestion
include integration/experiment/monitor/Makefile.mk
include integration/experiment/gpu_frequency_sweep/Makefile.mk
include integration/experiment/power_sweep/Makefile.mk
include integration/experiment/trace_analysis/Makefile.mk
```

copying_headers/MANIFEST.EXEMPT
```diff
@@ -62,6 +62,7 @@ integration/experiment/power_sweep/README.md
 integration/experiment/README.md
 integration/experiment/trace_analysis/README.md
 integration/experiment/uncore_frequency_sweep/README.md
+integration/experiment/gpu_frequency_sweep/README.md
```
bgeltz: Move this up under the monitor.  Please

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,434 @@
...
+                // frequency from system characterization.
+                //
+                // This approach assumes the efficient frequency is suitable as both a
+                // baseline for active regions and inactive regions. This is generally
+                // true when the efficient frequency consumes low power at idle due to clock
+                // gating or other hardware PM techniques.
+                //
+                // If f_efficient does not meet these criteria this behavior can still be
+                // achieved through tracking the GPU Utilization signal and setting frequency
+                // to a separate idle value (f_idle) during regions where GPU Utilizaiton is
```
```suggestion
                // to a separate idle value (f_idle) during regions where GPU Utilization is
```

bgeltz (changes requested):

bgeltz (approved):

Issue #1675, opened 2021-06-16
As a developer of the GEOPM service I would like nightly integration tests use a fully installed service


cmcantalupo:
Nightly testing should be implemented which installs the geopm service and then runs a battery of targeted integration tests against the service.     

The test-session user has been added to the system an this user is exercising the GEOPM HPC Runtime integration tests.  These tests are not all passing or skipped, and the reporting does not include slack or email broadcast.

We should also have some integration tests that do not rely on the GEOPM HPC Runtime but rather, use only interfaces provided by the geopm-service packages.

PR #2055, opened 2021-12-22
Remove all references to geopm_context

### LINKED ISSUE #2033: Test failures related to import geopm_context
**Describe the bug**
I tried to run `test_environment.py` and I expected the test to pass instead it fails as it's unable to find the `integration` module.

**GEOPM version**
`1.1.0+dev1674g69e2becc`
https://github.com/bgeltz/geopm/tree/issue-2015

**Expected behavior**
When this test is run through the nightlies, the following is observed in the output log (test_loop_output.txt):
```
test_geopm_environment (test_environment.TestIntegrationEnvironment)
Test behavior of geopm environment files. ... skipped 'Cannot write to path: /etc/geopm/environment-default.json'
```

**Actual behavior**
Running the test in isolation provides the following output:
```
$ python3 ./test_environment.py
  File "./test_environment.py", line 39, in <module>
    from integration.test import geopm_context
ModuleNotFoundError: No module named 'integration'
```

Removing all but this Python test file from the ```__main__.py``` provides similar output:

```
[bgeltz@mcfly6 test](issue-2015)$ python3 .
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "./__main__.py", line 54, in <module>
    from test_environment import *
  File "./test_environment.py", line 39, in <module>
    from integration.test import geopm_context
ModuleNotFoundError: No module named 'integration'
```

**Additional context**
This issue stems from hacks to the ```PYTHONPATH``` that happen in some tests but not others, and the order in which the tests fire.

E.g. Examining `test_monitor.py` you see:
https://github.com/geopm/geopm/blob/73a7ec4ae2bb9e68cb91a2d1c8151346e6148f1f/integration/test/test_monitor.py#L43-L44

If *any* test executes the ```sys.path.append(...)``` logic (as in the running of the entire suite with the nightlies), then this problem goes away.  The ordering in which these test fire may hide this problem since some tests have the ```sys.path.append(...)``` and some do not.

To further illustrate, I can work around the problem by performing the following:
```
[bgeltz@mcfly5 test](issue-2015)$ echo $PYTHONPATH
/home/bgeltz/build/geopm/lib/python3.6/site-packages:/home/bgeltz/git/geopm/integration
[bgeltz@mcfly5 test](issue-2015)$ export PYTHONPATH=${HOME}/git/geopm:${PYTHONPATH}
[bgeltz@mcfly5 test](issue-2015)$ python3 ./test_environment.py -v
test_geopm_environment (__main__.TestIntegrationEnvironment)
Test behavior of geopm environment files. ... skipped 'Cannot write to path: /etc/geopm/environment-default.json'

Ran 1 test in 0.000s

OK (skipped=1)
```

The following tests have related issues:
```
[bgeltz@mcfly test](issue-2015)$ grep -nir "from integration.test"
test_environment.py:39:from integration.test import geopm_context
test_progress.py:41:from integration.test import geopm_context
test_progress.py:42:from integration.test import util
test_template.py.in:44:from integration.test import geopm_context
test_template.py.in:48:from integration.test import util
test_template.py.in:52:    from integration.test import geopm_test_launcher
test_hint_time.py:44:from integration.test import geopm_context
test_hint_time.py:49:from integration.test import util
test_hint_time.py:54:    from integration.test import geopm_test_launcher
test_monitor.py:44:from integration.test import geopm_context
test_ompt.py:38:from integration.test import geopm_context
test_profile_overflow.py:46:from integration.test import geopm_context
test_fmap_short_region_slop.py:66:    from integration.test import geopm_context
test_fmap_short_region_slop.py:72:from integration.test import util
test_fmap_short_region_slop.py:76:    from integration.test import geopm_test_launcher
```

**Resolution**
One of the following resolutions should be implemented pending feasibility:
1. All occurrences of `from integration.test import geopm_context` should be changed to just `import geopm_context`.
2. If the `from...` syntax is required, ensure that the `sys.path.append(...)` logic is also included.
3. Is `geopm_context` still needed moving forward?  Can this all be removed?

cmcantalupo:
- This file was used to add source code locations to
  the PYTHONPATH while the test was executing.
- Require that user have geopmdpy and geopmpy in
  their PYTHONPATH in order to run tests
- Move skipped integrations tests into separate
  directory and get rid of imports from __main__.py
- Use unittest discover mechanism to find integration
  tests for the HPC runtime.
- Require that the geopm source code repo is in
  the python path to give access to integration
  test modules that are not installed
- Modify imports in tests accordingly
- Relates to #1675
- Fixes #2033


integration/test/README.md
```diff
@@ -2,33 +2,10 @@ GEOPM Integration Tests
...
-times and checks that the GEOPM report and trace files demonstrate
-features of the GEOPM runtime.
 
 ENVIRONMENT VARIABLES
 ---------------------
 
-### GEOPM_LAUNCHER
-Used to force the integration tests to target a specific Launcher
-implementation as supported by the Launcher factory.  See
-**geopmlaunch(1)** for more details.
```
cmcantalupo: Leave in...This should still work and be relevant.

integration/test/__main__.py
```diff
@@ -1,77 +0,0 @@
-#!/usr/bin/env python3
```
cmcantalupo, line 1: This will break the other nightlies.  They need to be reconfigured to use the unittest discover feature.

cmcantalupo (commented):
The tutorial integration test is still failing:

```
OMP_PROC_BIND=true GEOPM_CTL=process GEOPM_PROFILE= GEOPM_AGENT=power_governor GEOPM_POLICY=tutorial_power_policy.json GEOPM_REPORT=tutorial_3_governed_report GEOPM_TRACE=tutorial_3_governed_trace KMP_AFFINITY=disabled MV2_ENABLE_AFFINITY=0 KMP_WARNINGS=FALSE srun -N 2 -n 10 --cpu-bind v,mask_cpu:0x2,0xffc00000000ffc,0x3ff000000003ff000,0x3ff000000003ff000000,0xffc00000000ffc00000000 --export=LD_PRELOAD=libgeopm.so,ALL -- ./tutorial_3

srun: Loaded msrsave restore plugin.
cpu-bind=MASK - mcfly1, task  0  0 [62251]: mask 0x2 set
cpu-bind=MASK - mcfly2, task  8  3 [4453]: mask 0x3ff000000003ff000000 set
cpu-bind=MASK - mcfly1, task  3  3 [62254]: mask 0x3ff000000003ff000000 set
cpu-bind=MASK - mcfly2, task  9  4 [4454]: mask 0xffc00000000ffc00000000 set
cpu-bind=MASK - mcfly1, task  1  1 [62252]: mask 0xffc00000000ffc set
cpu-bind=MASK - mcfly2, task  5  0 [4450]: mask 0x2 set
cpu-bind=MASK - mcfly1, task  2  2 [62253]: mask 0x3ff000000003ff000 set
cpu-bind=MASK - mcfly2, task  7  2 [4452]: mask 0x3ff000000003ff000 set
cpu-bind=MASK - mcfly1, task  4  4 [62255]: mask 0xffc00000000ffc00000000 set
cpu-bind=MASK - mcfly2, task  6  1 [4451]: mask 0xffc00000000ffc set
Error: <geopm> Runtime error: SDBus: Failed to call sd-bus function sd_bus_call_method(), error:-5 name: not.known.Error.RuntimeError: The PID 62251 requested write access, but the geopm service already has write mode client with PID or SID of 48295: at src/SDBus.cpp:59
MPI_COMM_WORLD size: 8
Beginning loop of 500 iterations.
```

cmcantalupo:
```# SUCCESS: test-service.sh Thu Dec 23 08:25:10 PST 2021```
The integration tests are passing.  Just need to fix the unit test.

File: integration/test/README.md [hunk 3]
```diff
@@ -44,19 +26,37 @@ application to wrap target job execution.
...
+
+# Set PATH values
+export LD_LIBRARY_PATH=${GEOPM_INSTALL}/lib:${LD_LIBRARY_PATH}
+export PYTHONPATH=${GEOPM_PYTHONPATH}:${PYTHONPATH}
+export PATH=${GEOPM_INSTALL}/bin:${PATH}
+
+python3 -m unittest discover \
+        --top-level-directory ${GEOPM_SOURCE} \
+        --start-directory ${GEOPM_SOURCE}/integration/test \
+        --pattern 'test_*.py' \
```
bgeltz: According to the [docs](https://docs.python.org/3/library/unittest.html#unittest-test-discovery), the default pattern is `test*.py` when it is not specified.

Is there a reason that is not sufficient?

integration/test/README.md
```diff
@@ -44,19 +26,37 @@ application to wrap target job execution.
               $PATH_TO_DDT_CLIENT_SESSION_FILE" \
               ./geopm_test_integration.py -v TestIntegration.test_count`
```
bgeltz: Isn't this use case broken with this PR?

PYTHONPATH isn't set to anything special in these cases.

integration/test/README.md
```diff
@@ -44,19 +26,37 @@ application to wrap target job execution.
...
-
-    $ git clean -fd . ; GEOPM_RUN_LONG_TESTS=true ./geopm_test_integration.py -v TestIntegration.test_power_consumption
-
+EXAMPLE
+-------
+
+```
+#!/bin/bash
+
+# Set to geopm source code location
```
bgeltz: I expected to see this here:
```
if [ -f ${HOME}/.geopmrc ]; then
    source ${HOME}/.geopmrc
    source ${GEOPM_SOURCE}/config/run_env.sh
else
    exit -1
fi
```
Then you don't need to hardcode the other env vars below.

integration/test/README.md
```diff
@@ -44,19 +26,37 @@ application to wrap target job execution.
...
+export GEOPM_SOURCE=/home/test/Git/geopm
+
+# Set to the install prefix provided to configure
+export GEOPM_INSTALL=/home/test/build/geopm
+
+# Find location of geopmdpy site-packages
+export GEOPM_PYTHONPATH=$(dirname $(find ${GEOPM_INSTALL} -name geopmdpy -type d))
```
bgeltz: This is different than the code that's already in `run_env.sh` that accomplishes the same thing.  We should only have it in one spot.

service/geopmdpy/service.py
```diff
@@ -759,6 +759,7 @@ def _write_mode(self, client_pid):
             write_pid = client_sid
         else:
             write_pid = client_pid
+        self.check_client(self._write_pid)
```
bgeltz, line 762: This should really be in it's own PR.  It has nothing to do with `geopm_context`.

service/geopmdpy_test/TestPlatformService.py
```diff
@@ -497,7 +498,7 @@ def test_start_batch_write_blocked(self):
         self.open_mock_session('other', other_pid)
         with mock.patch('geopmdpy.pio.write_control', return_value=[]) as mock_write_control, \
              mock.patch('geopmdpy.pio.save_control', return_value=[]), \
-             mock.patch('os.getsid', return_value=other_pid) as mock_getsid:
+             mock.patch('os.getsid', return_value=client_sid) as mock_getsid:
```
bgeltz, line 501: Are any of the changes here expressly related to the `geopm_context` removal?

bgeltz (changes requested):
I'm curious why we have a mixture of skip decorators with py files in test, and some other py files moved into test_skipped.

Seems we should do one or the other?

service/geopmdpy_test/TestPlatformService.py
cmcantalupo: This is to support the change to check that the PID holding the write lock is still active.

integration/test/README.md
cmcantalupo: This is a README with instructions on how to use our integration tests...I'm not sure it is helpful to refer to bash script...

service/geopmdpy/service.py
cmcantalupo: I can split it out, but it will be a two line commit (this one and the one below that you commented on).

integration/test/README.md
cmcantalupo: I'm just trying to keep the instructions self contained and simple.  I'm not sure referring to two other bash scripts, one of which is not part of our repository, is helpful.

integration/test/README.md
cmcantalupo: No, we could omit line 52.

cmcantalupo:
The skipped files primarily do not use the unittest infrastructure, but rather were dropped in the integration/test directory and not added to the __main__.py imports.  These were never executed as part of our integration test scripting.

integration/test/README.md
cmcantalupo: yes, these comments should be updated

service/geopmdpy/service.py
bgeltz: I'm fine to leave this in as long as we have an issue for a future test.

integration/test/README.md
bgeltz: geopm_test_integration.py doesn't exist
The PYTHONPATH isn't set.

2 different examples showing how to do this via unittest discovery and by manually executing a single py file would be helpful.

bgeltz (changes requested):

service/geopmdpy/service.py
cmcantalupo: https://github.com/geopm/geopm/issues/2062

integration/README.md
```diff
@@ -38,6 +38,13 @@ The [apps](apps) directory is where information about
 applications is located.  Each application supported has its own
 sub-directory within the apps directory.
 
+## `config`
+
+The bash scripts in the [config](config) directory set environment
```
bgeltz:
```suggestion
The `bash` scripts in the [config](config) directory set environment
```

integration/config/README.md
```diff
@@ -0,0 +1,70 @@
+GEOPM Environment Configuration
+===============================
+
+This directory contains a collection of `bash` scripts that may be
+sourced to set up a users environment.  These scripts may be sourced
```
bgeltz:
```suggestion
sourced to set up a user's environment.  These scripts may be sourced
```

integration/config/README.md
```diff
@@ -0,0 +1,70 @@
...
+within a user's shell in order to build GEOPM, or to run against a
+locally installed version of GEOPM.
+
+Please refer to the [README](../README.md) one directory up for
+information about how to define envirnment variables relevent to these
```
bgeltz:
```suggestion
information about how to define environment variables relevant to these
```

bgeltz (changes requested):

bgeltz (approved):

PR #2063, opened 2022-01-10
Implement SaveControl class for saving control settings to disk

cmcantalupo:
- Relates to #2057 story from github issues
- Fixes #2058

PR #2070, opened 2022-01-14
Fix for the tutorial build

bgeltz:
- Broken since 32dbf1cb9 in #2063.
- These tutorial IOGroups were simply missed with the `name()` API updates.
- Additionally address compile time warnings generated with GNU around:
  - `%d` vs. `%ld` format specifiers to printf.
  - Defined but not used functions if OMP is available.
  - Missing return code.
- Address GNU/distro specific issues with linking.  On ubuntu with GNU/GCC, the libraries have to be last.  This does not impact other flavors tested (i.e. SLES).


cmcantalupo (approved):
👍

Issue #2093, opened 2022-02-02
As a system administrator I would like to be sure that the access list files used by the service are securely manipulated


cmcantalupo:
**What problem does the story solve? Please describe.**
The current implementation does not check permissions, ownership, or group of the access files.  The configuration directory is also created without restricted permissions.  

**Describe the tasks that will be required to complete the story**
- [x] #2094
- [x] #2095
- [x] #2096
- [x] #2097
- [x] #2098

```
def secure_makedirs(path):
    """Securely make a directory"""
     pass

def secure_makefile(path, contents):
        """Securely output contents to file atomically"""
        pass

def secure_readfile(path):
      """Securely read a file"""
      pass

```

**Describe alternatives you've considered**
Duplicating the code for accessing the files and adding a separate module for the AccessFiles class.

**Additional context**
#2084

PR #2100, opened 2022-02-03
Add secure functions to varrun.py

### LINKED ISSUE #2094: Split directory creation code and file creation code from ActiveSessions class into separate free functions described below (secure_*()) located in the same module
> estimate 1

- [x] move existing code and refactor use to match input parameters
- [x] Provide doc strings that match the requirements for validation
- [x] Write positive and negative unit tests for each of these functions that mock out system calls (reuse some code already existing for ActiveSessions unit tests)
- [x] Mock use of secure_*() calls in the unit tests for ActiveSessions (maybe?)

```
def secure_make_dirs(path):
    """Securely make a directory"""
     pass

def secure_make_file(path, contents):
    """Securely output contents to file atomically"""
    pass

def secure_read_file(path):
    """Securely read a file"""
    pass

```

cmcantalupo:
- Relates to #2093
- Fixes #2094


service/geopmdpy_test/TestActiveSessions.py
```diff
@@ -162,155 +162,6 @@ def create_json_file(self, directory, filename, contents, permissions=0o600):
             else:
                 json.dump(contents, file)
 
-    def test_default_creation(self):
```
cmcantalupo, line 165: Please leave these tests.  They should continue to pass without modification, and we will update them later to mock the secure_*() methods

service/geopmdpy_test/TestActiveSessions.py
```diff
@@ -339,16 +192,21 @@ def test_creation_bad_session_perms(self):
         # first time in __init__() it is called with a string to determine the directory
         # second time in _load_session_file() it is called with a file descriptor to determine the file
         # based on the data type we use a different mock for the os.stat()
-        side_effect = lambda path: type(path) is int and session_2_mock or dir_mock
```
cmcantalupo, line 344: What was the reason that this unit test for the ActiveSessions class need to change with the refactor?  I don't expect TestActiveSessions.py to change at all for this PR.

Krebrov001: The implementation of `secure_read_file()` has changed, so the unit tests should be updated accordingly. We see new mocks added for `stat.S_ISREG` and `stat.S_ISLNK`. And we see an additional mock for `os.path.isdir()` which is calleld twice, once for the directory, once for the file, so we are having `os_path_isdir_effect` to distinguish between these two use cases.

Krebrov001 (dismissed):
I don't see anything wrong here.

service/geopmdpy/varrun.py
```diff
@@ -52,6 +52,189 @@
...
+            sys.stderr.write(f'Warning: <geopm-service> {path} is a symbolic link, the link will be renamed to {renamed_path}')
+            sys.stderr.write(f'Warning: <geopm-service> the symbolic link points to {os.readlink(path)}')
+        # If it's a so-called "regular file"
+        elif not os.path.isdir(path):
+            is_valid = False
+            sys.stderr.write(f'Warning: <geopm-service> {path} is not a directory, it will be renamed to {renamed_path}')
+        # If it's a directory
+        else:
+            st = os.stat(path)
+            # If the permissions is not what we wanted
```
bgeltz:
```suggestion
            # If the permissions are not what we wanted
```

service/geopmdpy/varrun.py
```diff
@@ -52,6 +52,189 @@
...
+                sys.stderr.write(f'Warning: <geopm-service> {path} has wrong user owner, it will be renamed to {renamed_path}')
+                sys.stderr.write(f'Warning: <geopm-service> the wrong user owner was {user_owner}')
+                is_valid = False
+            # If the group owner is not what we wanted
+            group_owner = st.st_gid
+            if group_owner != os.getgid():
+                sys.stderr.write(f'Warning: <geopm-service> {path} has wrong group owner, it will be renamed to {renamed_path}')
+                sys.stderr.write(f'Warning: <geopm-service> the wrong group owner was {group_owner}')
+                is_valid = False
+        # If one of the three above branches revealed invalid file
```
bgeltz:
```suggestion
        # If one of the three above branches revealed an invalid file
```

service/geopmdpy/varrun.py
```diff
@@ -52,6 +52,189 @@
...
+        - The permissions for the file are 0o600
+
+    If the existing path is determined to be insecure, a warning is
+    printed to syslog and the existing file will be renamed to
+    `<path>-<UUID>-INVALID` so that it may be audited later, but not
+    used.  The warning message will display the reason why the file
+    was not secure and None is returned.
+
+    If the path points to an existing file that is determined to be
+    secure then the contents of the file is returned.
```
bgeltz:
```suggestion
    secure then the contents of the file are returned.
```

service/geopmdpy_test/TestSecureFiles.py
```diff
@@ -0,0 +1,591 @@
...
+
+        Test calls secure_make_dirs() when the geopm-service
+        provided path is not a directory, but a link. It asserts that
+        a warning message is printed to standard error and that a
+        new directory is created while the link is renamed.
+
+        """
+        sess_path = f'{self._TEMP_DIR.name}/geopm-service-link'
+        os.symlink('/root', sess_path)
+        with mock.patch('os.path.islink', return_value=True) as mock_os_path_islink, \
```
bgeltz: It might be better here to use wraps instead of outright mocking of os.path.islink.

service/geopmdpy_test/TestSecureFiles.py
```diff
@@ -0,0 +1,591 @@
...
+        provided path is not a directory, but a file. It asserts that
+        a warning message is printed to standard error and that a
+        new directory is created while the file is renamed.
+
+        """
+        sess_path = f'{self._TEMP_DIR.name}/geopm-service.txt'
+        filename = Path(sess_path)
+        filename.touch(exist_ok=True)
+
+        with mock.patch('os.path.islink', return_value=False) as mock_os_path_islink, \
```
bgeltz: Samee here.

bgeltz (changes requested):

service/geopmdpy_test/TestSecureFiles.py
Krebrov001: What do you mean by wraps?

service/geopmdpy/varrun.py
```diff
@@ -190,6 +191,9 @@ def secure_read_file(path):
         if os.path.islink(path):
             sys.stderr.write(f'Warning: <geopm-service> {path} is a symbolic link, it will be renamed to {renamed_path}')
             sys.stderr.write(f'Warning: <geopm-service> the symbolic link points to {os.readlink(path)}')
+        # If it is a fifo
+        elif stat.S_ISFIFO(os.stat(path).st_mode):
```
cmcantalupo: Should check for stat.S_ISREG() before opening it.  This will cover the case of a link or a FIFO.  More verbose error messaging we can check each of these individually first, but I'm also okay with a single check that it is a regular file.

bgeltz: Also \n's

bgeltz (approved):

Issue #2062, opened 2022-01-06
As a user the GEOPM service I would like to be confident that the implementation of the write lock is free of race conditions


cmcantalupo:
**What problem does the story solve? Please describe.**
When a user wants to write to controls through the GEOPM service, the write lock must be free.  Without an integration test that proves that when this lock is under contention the behavior is well defined, we cannot be confident that there are no existing race conditions.

**Describe the tasks that will be required to complete the story**
Implement integration tests that use the write lock under contention and show that the behavior is well understood.

**Describe alternatives you've considered**
These tests help provide QoS guarantees for the service that enable security requirements that are not flexible.

**Additional context**
A change goes in with PR #2055 will remove the race condition that could occur if a process holding the write lock is killed and another process requests the lock within the one second interval for the periodic check done by GLib.  It would be good to have a check that show that this fix is working.

PR #2103, opened 2022-02-08
Change how the GEOPM Service control lock / save / restore logic works

### LINKED ISSUE #2092: Running the test_su_restart.sh test causes the service to be unable to restart
**Describe the bug**
I tried to run the test_su_restart.sh test in the service integration tests and I expected it to pass and for the system to be up instead the test claims to pass after outputting a lot of error messages, and the service will not restart.  The symptom seems to be a session file lists itself as "rw" but the the SAVE_FILES directory is empty.

**GEOPM version**
#2084 

**Expected behavior**
The test successfully restarts the service and restores the control values.

**Actual behavior**

**Error messages produced**

**Additional context**
This test script is only present in the pull request #2084

cmcantalupo:
- Relates to #2062
- Fixes #2092

### DIFF
--- a/service/geopmdpy/varrun.py
+++ b/service/geopmdpy/varrun.py
@@ -57,6 +57,7 @@ import psutil
 import uuid
 import grp
 import pwd
+import fcntl
 
 from . import pio
 
@@ -135,6 +136,7 @@ def secure_make_dirs(path):
     else:
         os.mkdir(path, mode=0o700)
 
+
 def secure_make_file(path, contents):
     """Securely and atomically create a file containing a string
 
@@ -155,10 +157,13 @@ def secure_make_file(path, contents):
     """
     temp_path = f'{path}-{uuid.uuid4()}-tmp'
     old_mask = os.umask(0o077)
-    with open(os.open(temp_path, os.O_CREAT | os.O_WRONLY, 0o600), 'w') as file:
-        file.write(contents)
-    os.rename(temp_path, path)
-    os.umask(old_mask)
+    try:
+        with open(os.open(temp_path, os.O_CREAT | os.O_WRONLY, 0o600), 'w') as file:
+            file.write(contents)
+        os.rename(temp_path, path)
+    finally:
+        os.umask(old_mask)
+
 
 def secure_read_file(path):
     """Securely read a file into a string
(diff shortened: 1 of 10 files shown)


service/geopmdpy/varrun.py
```diff
@@ -571,11 +533,110 @@ def _load_session_file(self, sess_path):
...
-                sess.pop('batch_server')
+           os.rename(sess_path, renamed_path)
+
+
+class WriteLock(object):
+    def __init__(self, var_path=GEOPM_SERVICE_VAR_PATH):
+        self._VAR_PATH = var_path
+        self._LOCK_PATH = os.path.join(self._VAR_PATH, "CONTROL_LOCK")
+        # Create an empty lock file if it does not exist
+        lock_pid = self.try_lock()
```
cmcantalupo: assumes that /var/run/geopm-service exists, probably should call the secure_make_dirs() method here.

service/integration/test/test_su_restart.sh
```diff
@@ -0,0 +1,68 @@
...
+    exit -1
+}
+
+# RUN WRITE SESSION TEST AND MAKE SURE IT PASSES
+${TEST_SCRIPT} &
+sleep 1
+sudo systemctl stop geopm
+sleep 1
+sudo systemctl start geopm
+wait
```
cmcantalupo: Need to check on the return value from the background process

service/geopmdpy/varrun.py
```diff
@@ -179,64 +184,58 @@ def secure_read_file(path):
...
     contents = None
+    if is_secure_path(path):
+        with open(path) as fid:
+            if is_secure_file(path, fid):
+                contents = fid.read()
+    return contents
+
 
-    # If the path exists
+def is_secure_path(path):
```
cmcantalupo, line 206: missing doc strings

service/geopmdpy/varrun.py
```diff
@@ -179,64 +184,58 @@ def secure_read_file(path):
...
-
-    # If the path does not exist
     else:
         sys.stderr.write(f'Warning: <geopm-service> {path} does not exist\n')
+    return result
 
-    return contents
+
+def is_secure_file(path, fid):
+    result = False
```
cmcantalupo, line 258: missing doc strings

service/geopmdpy/varrun.py
```diff
@@ -700,11 +661,98 @@ def _load_session_file(self, sess_path):
...
-            if batch_pid is not None:
-                sess.pop('batch_server')
+           os.rename(sess_path, renamed_path)
 
-        self._sessions[client_pid] = dict(sess)
-        self._update_session_file(client_pid)
+
+class WriteLock(object):
+    def __init__(self, var_path=GEOPM_SERVICE_VAR_PATH):
+        self._VAR_PATH = var_path
```
cmcantalupo, line 941: Add doc strings

bgeltz (changes requested):
Why is all the man page work in this PR?  It's unrelated to the write lock.

service/geopmdpy/varrun.py
```diff
@@ -700,11 +661,98 @@ def _load_session_file(self, sess_path):
...
+        self._LOCK_PATH = os.path.join(self._VAR_PATH, "CONTROL_LOCK")
+        self._fid = None
+
+    def __enter__(self):
+        old_mask = os.umask(0o077)
+        try:
+            trial_count = 0
+            while self._fid is None and trial_count < 2:
```
bgeltz, line 959: Why try twice?  Why not more/less?
Is a sleep necessary between trials?

service/geopmdpy/varrun.py
```diff
@@ -700,11 +661,98 @@ def _load_session_file(self, sess_path):
...
+        Args:
+            pid (int): The PID to assign the lock
+
+        Returns:
+            int: The PID of the session that holds the write lock upon return
+                 or None if the write lock is not held
+
+        """
+        file_pid = None
+        self._fid.seek(0, os.SEEK_END)
```
bgeltz: Why seek() instead of readline()?  It might simplify this logic.

service/geopmdpy/varrun.py
cmcantalupo: We will try twice, the first time may be foiled by a bad file, if the second attempt is also foiled, then we want to give up.  We want to try more than once to avoid a bad file causing the lock operation to fail.

We don't need to sleep because the is_secure_file() call will rename it to an invalid name.

service/geopmdpy/varrun.py
cmcantalupo: Updated implementation, should be a bit easier to read.  Thanks

service/geopmdpy/varrun.py
```diff
@@ -700,11 +692,185 @@ def _load_session_file(self, sess_path):
...
+                    # File is insecure and was renamed, try again
+                    self._fid.close()
+                    self._fid = None
+                trial_count += 1
+        finally:
+            os.umask(old_mask)
+        if self._fid == None:
+            self._rpc_unlock()
+            raise RuntimeError('Unable to open write lock securely after two tries')
+        fcntl.lockf(self._fid, fcntl.LOCK_EX)
```
bgeltz, line 977: If this is only remaining to future proof the service should it need to be multi-process, a comment explaining as much would be very beneficial to all our memories. :)

service/geopmdpy_test/TestWriteLock.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        with WriteLock(self._sess_path) as write_lock:
+            self.assertIsNone(write_lock.try_lock())
+            out_pid = write_lock.try_lock(self._orig_pid)
+            self.assertEqual(self._orig_pid, out_pid)
+            try:
+                with WriteLock(self._sess_path) as other_lock:
+                    self.fail('Able to create nested WriteLock context')
+            except RuntimeError as ex:
+                self.assertEqual('Attempt to modify control lock file concurrently', str(ex))
```
Krebrov001:
```python
self.assertEqual('Attempt to modify control lock file in the same process', str(ex))
```

bgeltz (approved):

PR #2129, opened 2022-02-22
Add geopm namespace to make_unique() call in Tracer.cpp

cmcantalupo:
- Disambiguate it from std::make_unique() available after c++14


dannosliwcd (approved):
LGTM

PR #2138, opened 2022-03-08
Add a configure option that will disable sphinx documentation

### LINKED ISSUE #2137: As a devloper of GEOPM I would like be able to compile the GEOPM Service without sphinx dependencies
On some systems using the Python sphinx documentation may be problematic due to unavailable dependencies.  I would like support to build the other GEOPM Service software components without creating the documentation.

Add a configure option to the service build that will copy the ReST version of the man pages into the installed man page path and do not build web documentation.  This will result in man pages installed which have subpar formatting, but will still be mostly readable.  

A distribution tarball includes the man pages, and could be generated on another system, but this is not great for a development cycle. 

**Additional context**
Similar feature used to exist to disable ronn requirement.

cmcantalupo:
- Fixes #2137

### DIFF
--- a/service/configure.ac
+++ b/service/configure.ac
@@ -74,6 +74,20 @@ fi
 [enable_coverage="0"]
 )
 
+AC_ARG_ENABLE([docs],
+  [AS_HELP_STRING([--disable-docs], [Disable the use of the Python sphinx tool to generate documentation])],
+[if test "x$enable_docs" = "xno" ; then
+  enable_docs="0"
+else
+  enable_docs="1"
+fi
+],
+[enable_docs="1"]
+)
+
+AC_SUBST([enable_docs])
+AM_CONDITIONAL([ENABLE_DOCS], [test "x$enable_docs" = "x1"])
+
 AC_ARG_ENABLE([systemd],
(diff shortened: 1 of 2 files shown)


bgeltz (dismissed):

service/configure.ac
```diff
@@ -74,6 +74,20 @@ fi
 [enable_coverage="0"]
 )
 
+AC_ARG_ENABLE([docs],
+  [AS_HELP_STRING([--disable-docs], [Disable the use of the Python sphinx tool to generate documentation])],
```
dannosliwcd, line 78: In the latest release (https://github.com/geopm/geopm/blob/v1.1.0/configure.ac#L274), this was `--disable-doc`.

I was going to recommend keeping the old name or adding an alias to avoid breaking any existing build scripts that someone might have. But on second thought, I think that doesn't matter since the build process is different for the next release.

dannosliwcd (approved):
LGTM. The build option's name changes from release to release, but I think this is a non-issue since the build steps are changing anyways.

service/configure.ac
cmcantalupo: Thanks for pointing that out.  The build target is now "make docs" so I think this should be clearer to users who are not familiar with our old options...however, I can see the argument to keep it the same as it used to be.

Issue #2130, opened 2022-02-22
As a user of PlatformIO I would like requests to fall back on the other IOGroups that support the request when they fail


cmcantalupo:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I try to use the NVMLIOGroup to set frequency limits, and I'm not able to even though the ServiceIOGroup provides access.  This problem exists for any IOGroup that can be loaded without privilege, but provides some signals or controls that require it.  

**Describe the solution you'd like**
Requests to PlatformIO for high level aliases for signals or controls that fail due to permissions problems, should handle the exception and attempt the same operation through the ServiceIOGroup.  This will require that the PlatformIO::push_*() interfaces attempt to read or write to verify permissions.

**Describe alternatives you've considered**

1. Restrict output of `IOGroup::signal_names()` and `IOGroup::control_names()` to those accessible by the calling process.  

2. Add explicit interface: 

```
    std::set<string> IOGroup::service_signals(void); 
    std::set<string> IOGroup::service_controls(void); 
```

3. Express Linux Capability requirements for each signal and control: 

```
    uint64_t IOGroup::signal_capabilities(const std::string signal_name); 
    uint64_t IOGroup::control_capabilities(const std::string control_name); 
```

4. Express a load priority rank so that signal and control load order is explicit and decoupled from IOGroup load order 

```
    int IOGroup::signal_priority(const std::string signal_name); 
    int IOGroup::control_priority(const std::string control_name); 
```

**Additional context**

Issue #1690, opened 2021-06-17
As a user of PlatformIO from libgeopm or libgeopmpolicy, I would like to use the ServiceIOGroup to provide signals/controls from IOGroups that can be loaded in user space


cmcantalupo:
IOGroups should all be modified so that only signals and controls that are actually available are published by the object when the IOGroup is loaded.

- This will enable service to provide exactly the signals and controls that the end user does not have access to without the service, but allow faster non-service path when it can be used.
- It also avoids the case where a user can load an IOGroup, but fails to read or write in user space for a particular signal or control provided by the IOGroup.  If the signals and controls are published, the ServiceIOGroup will not override them, so the user cannot use the service to succeed with the same IOGroup created by geopmd as user root.

cmcantalupo:
The solution is slightly different than proposed, but #2144 effectively resolves this issue.

PR #2144, opened 2022-03-10
Implement IOGroup fallback in PlatformIO

### LINKED ISSUE #1813: Accelerator IOGroups (NVML & LevelZero) do not hide controls that the user does not have permission to use
Parent issue: #1690 

Both NVML and LevelZero provide a mechanism to determine if users have access to various controls.  These should be used to selectively remove controls from the m_control_available list if the user does not have the appropriate permissions, vs the current approach of try/catch handling of 'INSUFFICIENT PERMISSION' errors for things such as save & restore.

bgeltz:
- Relates to #2130, #1690.
- Fixes #1813.

- This affects the following APIs: read_signal(), write_control(), push_signal(), push_control().
- When a signal or control is requested through one of the above APIs, all IOGroups will be evaluated to see if they support the request.
  + The domain of the last registered plugin in the IOGroupFactory that supports the control or signal is used to determine the "native domain".  Any subsequent IOGroups that do not match this domain exactly are disregarded.
  + When read_signal() or write_control() is invoked, if there is an issue with the attempted call with the first IOGroup, the next in the list will be evaluated until all IOGroups that have been previously identified are exhausted.  An exception is raised in this instance.
  + Otherwise the IOGroup that is able to service the request will do so.

### DIFF
--- a/service/src/PlatformIO.cpp
+++ b/service/src/PlatformIO.cpp
@@ -225,8 +225,13 @@ namespace geopm
                         read_signal_ok = true;
                     }
                     catch (const geopm::Exception &ex) {
-                        err_msg += std::string(ex.what()) + "\n";
-                        // TODO Modify EpochIOGroup to throw a known code so that push will succeed below
+                        if (ex.err_value() == GEOPM_ERROR_NOT_IMPLEMENTED) {
+                            // IOGroups may not support read_signal()
+                            read_signal_ok = true;
+                        }
+                        else {
+                            err_msg += std::string(ex.what()) + "\n";
+                        }
                     }
                     if (read_signal_ok == true) {
                         int group_signal_idx = ii->push_signal(signal_name, domain_type, domain_idx);
@@ -327,14 +332,19 @@ namespace geopm
                     try {
                         // Attempt to read then write the control to ensure batch writes will succeed
                         val = ii->read_signal(control_name, domain_type, domain_idx);
+                        ii->write_control(control_name, domain_type, domain_idx, val);
                         read_signal_ok = true;
                     }
                     catch (const geopm::Exception &ex) {
-                        err_msg += std::string(ex.what()) + "\n";
-                        // TODO Modify EpochIOGroup to throw a known code so that push will succeed below
+                        if (ex.err_value() == GEOPM_ERROR_NOT_IMPLEMENTED) {
+                            // IOGroups may not support read_signal()/write_control()
+                            read_signal_ok = true;
+                        }
+                        else {
+                            err_msg += std::string(ex.what()) + "\n";
+                        }
                     }
                     if (read_signal_ok == true) {
-                        ii->write_control(control_name, domain_type, domain_idx, val);
                         int group_control_idx = ii->push_control(control_name, domain_type, domain_idx);
                         result = m_active_control.size();
                         m_existing_control[ctl_tup] = result;
--- a/service/src/PlatformIO.cpp
+++ b/service/src/PlatformIO.cpp
@@ -214,27 +214,32 @@ namespace geopm
             result = sig_tup_it->second;
         }
         std::string err_msg;
+        bool read_signal_ok = false;
         if (result == -1) {
             auto iogroups = find_signal_iogroup(signal_name);
             for (auto ii : iogroups) {
-                try {
-                    if (domain_type == ii->signal_domain_type(signal_name)) {
+                if (domain_type == ii->signal_domain_type(signal_name)) {
+                    try {
                         // Attempt to read before pushing to ensure batch reads will succeed
                         (void)ii->read_signal(signal_name, domain_type, domain_idx);
+                        read_signal_ok = true;
+                    }
+                    catch (const geopm::Exception &ex) {
+                        err_msg += std::string(ex.what()) + "\n";
+                        // TODO Modify EpochIOGroup to throw a known code so that push will succeed below
+                    }
+                    if (read_signal_ok == true) {
                         int group_signal_idx = ii->push_signal(signal_name, domain_type, domain_idx);
                         result = m_active_signal.size();
                         m_existing_signal[sig_tup] = result;
                         m_active_signal.emplace_back(ii, group_signal_idx);
                         break;
                     }
-                    else {
-                        result = push_signal_convert_domain(signal_name, domain_type, domain_idx);
-                        m_existing_signal[sig_tup] = result;
-                        break;
-                    }
                 }
-                catch (const geopm::Exception &ex) {
-                    err_msg += std::string(ex.what()) + "\n";
+                else {
+                    result = push_signal_convert_domain(signal_name, domain_type, domain_idx);
+                    m_existing_signal[sig_tup] = result;
+                    break;
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


service/src/PlatformIO.cpp
```diff
@@ -560,55 +632,55 @@ namespace geopm
...
     }
 
     std::function<std::string(double)> PlatformIOImp::format_function(const std::string &signal_name) const
     {
         std::function<std::string(double)> result;
         // PlatformIOImp forwards formatting request to underlying IOGroup
-        auto iogroup = find_signal_iogroup(signal_name);
-        if (iogroup == nullptr) {
+        auto iogroups = find_signal_iogroup(signal_name);
+        if (iogroups.size() == 0) {
```
cmcantalupo: .empty()?

service/src/PlatformIO.cpp
```diff
@@ -105,31 +102,61 @@ namespace geopm
...
+                if (result.empty()) {
+                    result.push_back(*it);
+                    native_domain = (*it)->signal_domain_type(signal_name);
+                }
+                else if ((*it)->signal_domain_type(signal_name) == native_domain) {
+                    result.push_back(*it);
+                }
+#ifdef GEOPM_DEBUG
+                else {
+                    std::cerr << "Warning: <geopm> PlatformIO::find_signal_iogroup(): "
```
cmcantalupo, line 122: Suggested warning:

Native domain differs for the same signal provided by another IOGroup.
The current implementation does not support fallback using an IOGroup
that provides a signal with a different native domain.  If this is an issue
for your use case, please request this feature.  The signal << signal_name <<
will not use << (*it)->name() for fallback if a read fails.

service/src/PlatformIO.cpp
```diff
@@ -105,31 +102,61 @@ namespace geopm
...
+                if (result.empty()) {
+                    result.push_back(*it);
+                    native_domain = (*it)->control_domain_type(control_name);
+                }
+                else if ((*it)->control_domain_type(control_name) == native_domain) {
+                    result.push_back(*it);
+                }
+#ifdef GEOPM_DEBUG
+                else {
+                    std::cerr << "Warning: <geopm> PlatformIO::find_control_iogroup(): "
```
cmcantalupo, line 155: Similar warning as above might be clearer

service/src/PlatformIO.cpp
```diff
@@ -210,26 +227,49 @@ namespace geopm
...
-                    result = m_active_signal.size();
-                    m_existing_signal[sig_tup] = result;
-                    m_active_signal.emplace_back(iogroup, group_signal_idx);
+            auto iogroups = find_signal_iogroup(signal_name);
+            for (auto ii : iogroups) {
+                if (domain_type == ii->signal_domain_type(signal_name)) {
+                    try {
+                        // Attempt to read before pushing to ensure batch reads will succeed
+                        (void)ii->read_signal(signal_name, domain_type, domain_idx);
+                        read_signal_ok = true;
```
cmcantalupo: suggested rename:

read_signal_ok -> do_push_signal

service/src/PlatformIO.cpp
```diff
@@ -210,26 +227,49 @@ namespace geopm
         if (sig_tup_it != m_existing_signal.end()) {
             result = sig_tup_it->second;
         }
+        std::string err_msg;
+        bool read_signal_ok = false;
```
cmcantalupo: Move this declaration closer to use: inside of for and if contexts, i.e. after line 235

service/src/PlatformIO.cpp
```diff
@@ -296,26 +336,52 @@ namespace geopm
         if (ctl_tup_it != m_existing_control.end()) {
             result = ctl_tup_it->second;
         }
+        std::string err_msg;
+        bool read_signal_ok = false;
```
cmcantalupo: read_signal_ok -> is_write_valid

also move declaration after line 344

service/src/PlatformIO.cpp
```diff
@@ -210,26 +227,49 @@ namespace geopm
...
         if (result == -1) {
-            auto iogroup = find_signal_iogroup(signal_name);
-            if (iogroup != nullptr) {
-                if (domain_type == iogroup->signal_domain_type(signal_name)) {
-                    int group_signal_idx = iogroup->push_signal(signal_name, domain_type, domain_idx);
-                    result = m_active_signal.size();
-                    m_existing_signal[sig_tup] = result;
-                    m_active_signal.emplace_back(iogroup, group_signal_idx);
+            auto iogroups = find_signal_iogroup(signal_name);
+            for (auto ii : iogroups) {
```
```suggestion
            for (auto ii : find_signal_iogroup(signal_name)) {
```

service/src/PlatformIO.cpp
```diff
@@ -296,26 +336,52 @@ namespace geopm
...
         if (result == -1) {
-            auto iogroup = find_control_iogroup(control_name);
-            if (iogroup != nullptr) {
-                if (iogroup->control_domain_type(control_name) == domain_type) {
-                    int group_control_idx = iogroup->push_control(control_name, domain_type, domain_idx);
-                    result = m_active_control.size();
-                    m_existing_control[ctl_tup] = result;
-                    m_active_control.emplace_back(iogroup, group_control_idx);
+            auto iogroups = find_control_iogroup(control_name);
+            for (auto ii : iogroups) {
```
```suggestion
            for (auto ii : find_control_iogroup(control_name)) {
```

service/src/PlatformIO.cpp
```diff
@@ -296,26 +336,52 @@ namespace geopm
...
+                    int val;
+                    try {
+                        // Attempt to read then write the control to ensure batch writes will succeed
+                        val = ii->read_signal(control_name, domain_type, domain_idx);
+                        ii->write_control(control_name, domain_type, domain_idx, val);
+                        read_signal_ok = true;
+                    }
+                    catch (const geopm::Exception &ex) {
+                        if (ex.err_value() == GEOPM_ERROR_NOT_IMPLEMENTED) {
+                            // IOGroups may not support read_signal()/write_control()
```
cmcantalupo: I wonder if we would actually support an IOGroup with a control that could only be written through the batch interface.  We do not currently have any IOGroups which are implemented this way.

service/src/PlatformIO.cpp
```diff
@@ -296,26 +336,52 @@ namespace geopm
...
+                        else {
+                            err_msg += std::string(ex.what()) + "\n";
+                        }
+                    }
+                    if (read_signal_ok == true) {
+                        int group_control_idx = ii->push_control(control_name, domain_type, domain_idx);
+                        result = m_active_control.size();
+                        m_existing_control[ctl_tup] = result;
+                        m_active_control.emplace_back(ii, group_control_idx);
+                        break;
```
cmcantalupo: can we get rid of the early breaks and just put a

```
if (result != -1) {
    break;
}
```
At the end of the two loops over IOGroups in push_signal() and push_control()

service/src/PlatformIO.cpp
```diff
@@ -438,17 +504,38 @@ namespace geopm
...
+                    break;
+                }
+                catch (const geopm::Exception &ex) {
+                    err_msg += std::string(ex.what()) + "\n";
+                }
+            }
         }
-        else {
-            result = iogroup->read_signal(signal_name, domain_type, domain_idx);
+        if (std::isnan(result)) {
```
cmcantalupo: We need a different variable to track if the read was successful.  NAN may be returned when no exception was raised.

service/src/PlatformIO.cpp
```diff
@@ -489,16 +576,39 @@ namespace geopm
...
+        auto iogroups = find_control_iogroup(control_name);
+        if (iogroups.empty()) {
             throw Exception("PlatformIOImp::write_control(): control name \"" + control_name + "\" not found",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        else if (iogroup->control_domain_type(control_name) != domain_type) {
-            write_control_convert_domain(control_name, domain_type, domain_idx, setting);
+
+        std::string err_msg;
+        bool write_complete = false;
```
cmcantalupo: write_complete -> is_write_complete

cmcantalupo (commented):
Looks pretty good.  One real issue:

https://github.com/geopm/geopm/pull/2144/files#r826474099

The rest are style related.

service/src/PlatformIO.cpp
bgeltz: Here I was just referencing the fact that you cannot ```write_control()``` to either the EpochIOGroup or the ProfileIOGroup.

cmcantalupo (approved):

Issue #1670, opened 2021-06-16
As a system administrator of the GEOPM service I would like to know what the implications are for adding a signal or control to an access list


cmcantalupo:
The signal and control descriptions as they are currently written do not provide the system administrator with enough information to know about the security implications of providing access.

PR #2147, opened 2022-03-11
Expose turbo ratio limit controls when they are writable

### LINKED ISSUE #2143: As a user of ISST features I would like to modify turbo ratio limits on systems that support writable turbo ratio limits
ISST features enable configuration of turbo in ways that enable some cores to exceed regular turbo ratio limits, at the expense of more strict limits on other cores. In order to allow this behavior, the package-level turbo ratio limits need to be increased. If the package-level turbo ratio limits cannot be increased, then the benefits of ISST will remain constrained by those limits.

GEOPM currently configures the `MSR::TURBO_RATIO_LIMIT::*` MSRs as read-only, but they are writable on some systems. The `MSR::PLATFORM_INFO:PROGRAMMABLE_RATIO_LIMITS_TURBO_MODE` signal indicates whether the current system supports writable limits.

This feature request is to make `TURBO_RATIO_LIMIT::*` MSRs available as controls on systems that have the `MSR::PLATFORM_INFO:PROGRAMMABLE_RATIO_LIMITS_TURBO_MODE` bit set.

dannosliwcd:
- Adds a check to the PLATFORM_INFO MSR when setting up MSRIOGroup. If
  the platform reports that turbo ratio limits are configurable, the
  turbo ratio limit MSRs are exposed as controls. Resolves https://github.com/geopm/geopm/issues/2143
- Adds documentation for turbo ratio limit MSRs. Related to https://github.com/geopm/geopm/issues/1670

### DIFF
--- a/service/src/msr_data_skx.cpp
+++ b/service/src/msr_data_skx.cpp
@@ -61,7 +61,8 @@ namespace geopm
                     "units":     "none",
                     "scalar":    1,
                     "behavior":  "constant",
-                    "writeable": false
+                    "writeable": false,
+                    "description": "Indicates whether the MSR::TURBO_RATIO_LIMIT:* signals are also available as controls."
                 },
                 "PROGRAMMABLE_TDP_LIMITS_TURBO_MODE": {
                     "begin_bit": 29,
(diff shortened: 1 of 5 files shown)


bgeltz (dismissed):
LGTM.  Nice.

service/src/msr_data_skx.cpp
```diff
@@ -61,7 +61,8 @@ namespace geopm
                     "units":     "none",
                     "scalar":    1,
                     "behavior":  "constant",
```
dannosliwcd, line 63: Realized this is no longer "constant" when writable. I need to make it variable instead.

lhlawson:
LGTM

cmcantalupo (approved):

Issue #2148, opened 2022-03-11
As a user of Intel Hardware I would like to be able to access LevelZero Metrics


lhlawson:
**Is your feature request related to a problem? Please describe.**
I would like to be able to access the additional profiling metrics on Intel Hardware exposed by the [LevelZero Tools API](https://spec.oneapi.io/level-zero/latest/tools/PROG.html#metrics).

**Describe the solution you'd like**
The addition of an IO Group that provides the capability to query Level Zero metrics per GPU.

**Describe alternatives you've considered**
Support for LevelZero Metrics could be added directly to the agents requiring these metrics.
It could be added to the existing LevelZeroIOGroup and device pool as well, as it falls under the general LevelZero umbrella.

**Additional context**
This is the Intel hardware equivalent of feature request #1902

- [ ] #2893
- [ ] #2894
- [ ] #2895
- [ ] #2896

PR #2149, opened 2022-03-11
Addition of LevelZero metric support to the LevelZero device pool and IO Group

lhlawson:
- Relates to #2148 feature quest from github issues
- Provides Intel hardware support analogous to the NVIDIA support added in #1917 and #1914

The addition of metric polling capability to the LevelZero IO Group, Device Pool, and API shim.   General Metric Streamer flowchart implemented follows the LevelZero approach shown [here](https://spec.oneapi.io/level-zero/latest/_images/tools_metric_streamer.png)

Details of change:
- Updated LevelZero.cpp to set ZET environment variables and follow the general time based metric query approach outlined in the [LevelZero Tools API](https://spec.oneapi.io/level-zero/latest/tools/PROG.html#metrics). 
- Added querying for the 'ComputeBasics' group of Intel Hardware.  An additional metric group may be queried but is not supported at this time.  
- Added signals and aliases based on these metrics for GPU_COMPUTE_ACTIVITY and GPU_COMPUTE_STALLS.  Additional signals may be created from data collected in the ComputeBasics group in future PRs.

Additional Tasks:
- [x] Define Default LevelZero Tools report/event rate based on workload performance
- [ ] Add Test for signal and control of LevelZero Tools report/event rate.
- [ ] Consider adding LEVELZERO::METRICS:METRIC:<AGGREGATION>.  Currently all signals are averaged over the number of reports received, but this could be modified to be AVG, LAST, MIN, MAX, etc
- [x] Test dynamic tracing with geopm controller (read_signal and batch read have been tested, but not with the controller) 
- [ ] Confirmation of geopm behavior with all combinations of ZES_ENABLE_SYSMAN and ZET_ENABLE_METRICS


service/src/LevelZero.cpp
```diff
@@ -172,13 +189,293 @@ namespace geopm
...
+                case ZET_VALUE_TYPE_FLOAT64:
+                    data_double = data.value.fp64;
+                    break;
+                case ZET_VALUE_TYPE_BOOL8:
+                    data_double = data.value.ui32;
+                    break;
+                default:
+                    break;
+                };
+                m_devices.at(l0_device_idx).m_metric_data.at(metric_name).push_back(data_double);
```
lhlawson: metric_data.at(metric_name) is never cleared.

lhlawson:
Replaced by newer implementation https://github.com/geopm/geopm/pull/2897

Issue #1920, opened 2021-10-06
As an HPC admin, I would like to be able to train a Neural Net to be used with the solution proposed in #1919


fkeceli:
**Is your feature request related to a problem? Please describe.**

The GEOPM Agent proposed in #1919 (NN Agent) might require a NN specifically trained for a HPC system node architecture. This may need to be done in the field by using a list of previously decided representative benchmarks

**Describe the solution you'd like**

Our proposed solution to training a NN for the GEOPM NN Agent is as follows:

1. Perform frequency sweeps on a list of representative benchmarks. The benchmarks may be CPU only, GPU only or CPU/GPU mixed (depending on the final agent architecture).
2. Find the best performing frequencies based on the objective for each benchmark
3. Annotate the trace lines of each benchmark with the best frequencies
4. Train the NN with the trace lines to output the best frequencies

In this solution we would like to have a chain of scripts that do the following:
1. Perform sweeps and dump out HW telemetry of interest in traces as well as final energy and runtime in reports
2. Find best frequencies from reports and annotate the trace lines
3. Train the NN

**Describe alternatives you've considered**
FIXME

**Additional context**
Add any other context for the feature request here.

Issue #1919, opened 2021-10-06
As an HPC admin/user, I would like a GEOPM agent that modulates the CPU and/or GPU frequencies to minimize the energy of an application, potentially subject to a limit on acceptable performance degradation, by applying learnings from known good solutions for other applications


fkeceli:
**Is your feature request related to a problem? Please describe.**

The top level problem description is given in #1918. The feature request refers to a specific solution strategy described next.

**Describe the solution you'd like**

We would like to train a Neural Net (NN) via supervised learning to make real time frequency decisions based on HW telemetry. The NN shall be trained on a representative set of benchmarks, for which the optimal solutions are found offline (via a method such as exhaustive search or analytically). Moreover, if a solution is suspected suboptimal, it can be added to the training set (requiring a solution to be found offline) to refine the NN.

**Describe alternatives you've considered**

We are exploring another solution to this problem described in #1906.

**Additional context**

--

PR #2151, opened 2022-03-11
Pass profile information from the runtime to python agents when generating reports

### LINKED ISSUE #2150: As a developer of GEOPM python agents, I would like agents to be able to generate launch-specific profile names in reports
A python agent can know a hard-coded profile name or an overridden on provided on the command line. But it is impossible for the agent to generate a default profile name based on whatever is being executed. Such default behavior would improve usability because default report profile names would help distinguish between application runs in cases where the user did not decide to manually generate profile names in advance. For example, a useful default name could be the command and arguments that are being launched under the agent.

A python agent receives generated report file contents from a call to `runtime.Controller.run(argv, policy)`, which internally calls the owning agent's `get_report()` method, which is expected to call `geopm.reporter.generate(profile_name, agent_name)`. The contents of `argv` are known by the controller, but not by the agent.

One way to enable more flexibility over profile names is to add a new optional `profile` argument to the `runtime.Controller.run` method. If no override is provided, the `runtime.Controller` can generate a profile name from the already-known `argv`, the provided or default profile can be passed as an argument to the agent's `get_report()`.

dannosliwcd:
- This change allows for default behavior where the controller uses the launched application as the profile name.
- Helps improve usability issues discovered while working on #1919 and #1920
- Resolves #2150

### DIFF
--- a/integration/test/geopm_test_launcher.py
+++ b/integration/test/geopm_test_launcher.py
@@ -187,7 +187,7 @@ class TestLauncher(object):
                     self.msr_restore()
                 raise
 
-    def get_report(self):
+    def get_report(self, profile):
         return Report(self._report_path)
 
     def get_trace(self):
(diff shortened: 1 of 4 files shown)


bgeltz (approved):
LGTM

cmcantalupo (approved):
LGTM

PR #2158, opened 2022-03-15
Use rpm --eval to discover the local rpmbuild paths in build_dasbus.sh 

bensallen:
Use rpm --eval to discover the local rpmbuild's appropriate paths in build_dabus.sh. This useful for building when a user has their own custom paths set.


cmcantalupo (approved):
Thanks for the contribution @bensallen! Looks good to me.

Issue #2172, opened 2022-03-16
As a researcher using Intel GPUs I would like the ability to set the LevelZero Performance Factor


lhlawson:
**Is your feature request related to a problem? Please describe.**
Some workloads can benefit from external performance hints to Intel GPUs via the performance factor interface.  Currently GEOPM does not support this signal or control.
Feature [documentation](https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#performance-factor)

**Describe the solution you'd like**
Expose a series of LEVELZERO::GPU_PERFORMANCE_FACTOR* and GPUCHIP_PERFORMANCE_FACTOR* signals at various domains (package/all, compute, memory) similar to the existing LevelZero utilization signals. 
Similarly a series of controls for these domains should be exposed.

**Describe alternatives you've considered**
Relying upon hardware balancing - may result in perf loss vs hand tune values
Using a single setting per device - may result in perf loss on some workloads

**Additional context**

PR #2081, opened 2022-01-19
Addition of levelzero chip energy and power support.

lhlawson:
- Relates to #1993 bug report from github issues
- Fixes #2074 change request from github issues.

Update to LevelZero Device Pool, IOGroup, and LevelZero.cpp to support chip level energy readings and derivative power signals.  

TODO: 
- [x] refactor energy readings for accelerator and accelerator_chip (currently 2 functions are used when 1 would work).
- [x] Update ordering of GEOPM domain, levelzero domain, idx, etc for all LevelZero.cpp and LevelZeroDevicePool.cpp functions to match (new functions are scattershot in their ordering).
- [x] Add support for any power_limit* functions at the chip level where appropriate.  Pushed to later PR.
- [x] Update power domain tracking for cases where ZES_AFFINITY_MASK is used in a way that forces sub-devices to act as device (i.e. =0.0)

PR #2173, opened 2022-03-16
LevelZero Performance Factor Support

lhlawson:
- Relates to #2172 feature request from github issues
- Built on top of #2081 since it provides signals with similar Device/Chip split.   

Adds support for LevelZero sysman performance factor signals and controls.  
Feature documentation [here](https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#performance-factor)

The high level is that this is a unitless value in Sysman from 0-100 that provides a tradeoff between the compute hardware and all other hardware.  100 favors compute, 0 favors support hardware.  

In the Level IO Group we support reading the associated signals and writing controls for:
- [x] The Chip Level Compute Domain

Items keeping this in draft:
- [x] Updates to MockLevelZero.hpp
- [x] Updates to LevelZeroDevicePoolTest.cpp 
- [x] Updates to LevelZeroIOGroupTest.cpp
- [x] Update signals and controls for save/restore - setconfig() may not be granted, so getconfig() may not contain the last value written as outlined [here](https://spec.oneapi.io/level-zero/latest/sysman/api.html#zesperformancefactorsetconfig)

### DIFF
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -341,7 +341,8 @@ namespace geopm
                     else if (property.engines == ZES_ENGINE_TYPE_FLAG_MEDIA) { //TODO: should this be _DMA?
                         m_devices.at(device_idx).perf_domain.at(geopm::LevelZero::M_DOMAIN_MEMORY) = handle;
                     }
-                    else if (property.engines == ZES_ENGINE_TYPE_FLAG_OTHER) { //TODO: Update to not rely on OTHER
+                    //TODO: Update to not rely on the OTHER enum if possible
+                    else if (property.engines == ZES_ENGINE_TYPE_FLAG_OTHER) {
                         m_devices.at(device_idx).perf_domain.at(geopm::LevelZero::M_DOMAIN_ALL) = handle;
                     }
                     else {
@@ -515,10 +516,10 @@ namespace geopm
     int LevelZeroImp::performance_domain_count(int geopm_domain, unsigned int l0_device_idx, int l0_domain) const
     {
         int result = 0;
-        if (geopm_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR) {
+        if (geopm_domain == GEOPM_DOMAIN_GPU) {
             result = m_devices.at(l0_device_idx).perf_domain.size();
         }
-        else if (geopm_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
+        else if (geopm_domain == GEOPM_DOMAIN_GPU_CHIP) {
             result = m_devices.at(l0_device_idx).subdevice.perf_domain.at(l0_domain).size();
         }
         return result;
@@ -531,7 +532,7 @@ namespace geopm
         zes_perf_handle_t handle;
         ze_result_t ze_result;
 
-        if (geopm_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR) {
+        if (geopm_domain == GEOPM_DOMAIN_GPU) {
             handle = m_devices.at(l0_device_idx).perf_domain.at(l0_domain);
 
             ze_result = zesPerformanceFactorGetConfig(handle, &result);
@@ -539,7 +540,7 @@ namespace geopm
                             + std::string(__func__) +
                             ": Sysman failed to get performance factor values", __LINE__);
         }
-        else if (geopm_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
+        else if (geopm_domain == GEOPM_DOMAIN_GPU_CHIP) {
             handle = m_devices.at(l0_device_idx).subdevice.perf_domain.at(l0_domain).at(l0_domain_idx);
 
             ze_result = zesPerformanceFactorGetConfig(handle, &result);
@@ -824,10 +825,10 @@ namespace geopm
         zes_perf_handle_t handle;
         ze_result_t ze_result;
 
-        if (geopm_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR) {
+        if (geopm_domain == GEOPM_DOMAIN_GPU) {
             handle = m_devices.at(l0_device_idx).perf_domain.at(l0_domain);
         }
-        else if (geopm_domain == GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP) {
+        else if (geopm_domain == GEOPM_DOMAIN_GPU_CHIP) {
             handle = m_devices.at(l0_device_idx).subdevice.perf_domain.at(l0_domain).at(l0_domain_idx);
         }
         else {
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -164,11 +164,14 @@ namespace geopm
         // TODO: When additional device types such as FPGA, MCA, and Integrated GPU are supported by GEOPM
         // This should be changed to a more general loop iterating over type and caching appropriately
         for (unsigned int gpu_idx = 0; gpu_idx < m_num_gpu; gpu_idx++) {
-            domain_cache(gpu_idx);
+            frequency_domain_cache(gpu_idx);
+            power_domain_cache(gpu_idx);
+            perf_domain_cache(gpu_idx);
+            engine_domain_cache(gpu_idx);
        }
     }
 
-    void LevelZeroImp::domain_cache(unsigned int device_idx) {
+    void LevelZeroImp::frequency_domain_cache(unsigned int device_idx) {
         ze_result_t ze_result;
         uint32_t num_domain = 0;
 
@@ -223,6 +226,11 @@ namespace geopm
                 }
             }
         }
+    }
+
+    void LevelZeroImp::power_domain_cache(unsigned int device_idx) {
+        ze_result_t ze_result;
+        uint32_t num_domain = 0;
 
         //Cache power domains
         num_domain = 0;
@@ -297,6 +305,12 @@ namespace geopm
                       cached_energy_timestamp.resize(m_devices.at(device_idx).subdevice.power_domain.size());
         }
 
+    }
+
+    void LevelZeroImp::perf_domain_cache(unsigned int device_idx) {
+        ze_result_t ze_result;
+        uint32_t num_domain = 0;
+
         //Cache performance domains
         num_domain = 0;
         ze_result = zesDeviceEnumPerformanceFactorDomains(m_devices.at(
@@ -361,7 +375,11 @@ namespace geopm
                 }
             }
         }
+    }
 
+    void LevelZeroImp::engine_domain_cache(unsigned int device_idx) {
+        ze_result_t ze_result;
+        uint32_t num_domain = 0;
 
         //Cache engine domains
         num_domain = 0;
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -511,8 +511,6 @@ namespace geopm
     int LevelZeroImp::performance_domain_count(int geopm_domain, unsigned int l0_device_idx, int l0_domain) const
     {
         int result = 0;
-        if (geopm_domain == GEOPM_DOMAIN_GPU) {
-        }
         else if (geopm_domain == GEOPM_DOMAIN_GPU_CHIP) {
             result = m_devices.at(l0_device_idx).subdevice.perf_domain.at(l0_domain).size();
         }
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -297,6 +297,98 @@ namespace geopm
                       cached_energy_timestamp.resize(m_devices.at(device_idx).subdevice.power_domain.size());
         }
 
+        //Cache performance domains
+        num_domain = 0;
+        ze_result = zesDeviceEnumPerformanceFactorDomains(m_devices.at(
+                    device_idx).device_handle, &num_domain, nullptr);
+        if (ze_result == ZE_RESULT_ERROR_UNSUPPORTED_FEATURE) {
+#ifdef GEOPM_DEBUG
+            std::cerr << "Warning: <geopm> LevelZero: Performance domain detection is "
+                         "not supported.\n";
+#endif
+        }
+        else {
+            check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                            "LevelZero::" + std::string(__func__) +
+                            ": Sysman failed to get number of domains", __LINE__);
+
+            std::vector<zes_perf_handle_t> perf_domain(num_domain);
+
+            ze_result = zesDeviceEnumPerformanceFactorDomains(m_devices.at(
+                            device_idx).device_handle, &num_domain, perf_domain.data());
+            check_ze_result(ze_result, GEOPM_ERROR_RUNTIME, "LevelZero::"
+                            + std::string(__func__) +
+                            ": Sysman failed to get domain handle(s).", __LINE__);
+
+            m_devices.at(device_idx).subdevice.
+                      perf_domain.resize(geopm::LevelZero::M_DOMAIN_SIZE);
+            m_devices.at(device_idx).
+                      perf_domain.resize(geopm::LevelZero::M_DOMAIN_SIZE);
+
+            //int num_device_perf_domain = 0;
+            //int num_subdevice_perf_domain = 0;
+            for (auto handle : perf_domain) {
+                zes_perf_properties_t property;
+                ze_result = zesPerformanceFactorGetProperties(handle, &property);
+                check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                "LevelZero::" + std::string(__func__) +
+                                ": Sysman failed to get domain performance factor properties",
+                                __LINE__);
+
+                //Finding non-subdevice domain.
+                if (property.onSubdevice == 0) {
+                    //std::cout << "\tDebug: levelZero device " <<
+                    //             std::to_string(num_device_perf_domain) <<
+                    //             ": engine_type - " << std::to_string(property.engines) << std::endl;
+                    //++num_device_perf_domain;
+
+                    if (property.engines == ZES_ENGINE_TYPE_FLAG_COMPUTE) {
+                        m_devices.at(device_idx).perf_domain.at(geopm::LevelZero::M_DOMAIN_COMPUTE) = handle;
+                    }
+                    else if (property.engines == ZES_ENGINE_TYPE_FLAG_MEDIA) { //TODO: should this be _DMA?
+                        m_devices.at(device_idx).perf_domain.at(geopm::LevelZero::M_DOMAIN_MEMORY) = handle;
+                    }
+                    else if (property.engines == ZES_ENGINE_TYPE_FLAG_OTHER) {
+                        m_devices.at(device_idx).perf_domain.at(geopm::LevelZero::M_DOMAIN_ALL) = handle;
+                    }
+                    else {
+                        throw Exception("LevelZero::" + std::string(__func__) +
+                                        ": Unsupported device level performance factor domain "
+                                        " factor domain ()" + std::to_string(property.engines) +
+                                        " detected.", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                    }
+                }
+                else {
+                    if (property.engines == ZES_ENGINE_TYPE_FLAG_COMPUTE) {
+                        m_devices.at(device_idx).subdevice.perf_domain.at(
+                                 geopm::LevelZero::M_DOMAIN_COMPUTE).push_back(handle);
+                    }
+                    else if (property.engines == ZES_ENGINE_TYPE_FLAG_MEDIA) { //TODO: should this be _DMA?
+                        m_devices.at(device_idx).subdevice.perf_domain.at(
+                                 geopm::LevelZero::M_DOMAIN_MEMORY).push_back(handle);
+                    }
+                    else {
+                        throw Exception("LevelZero::" + std::string(__func__) +
+                                        ": Unsupported subdevice level performance factor domain "
+                                        " factor domain ()" + std::to_string(property.engines) +
+                                        " detected.", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                    }
+                    //std::cout << "\tDebug: levelZero sub device " <<
+                    //             std::to_string(property.subdeviceId) <<
+                    //             ": engine_type - " << std::to_string(property.engines) << std::endl;
+                    //++num_subdevice_perf_domain;
+                }
+            }
+
+            //std::cout << "Debug: levelZero device performance factor domains: " <<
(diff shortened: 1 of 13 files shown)


service/src/LevelZeroDevicePool.cpp
```diff
@@ -355,4 +429,24 @@ namespace geopm
...
     }
+
+    void LevelZeroDevicePoolImp::performance_factor_control(int domain, unsigned int domain_idx,
+                                                            int l0_domain, double setting) const
+    {
+        if (domain == GEOPM_DOMAIN_BOARD_ACCELERATOR) {
+            check_idx_range(domain, domain_idx);
+            m_levelzero.performance_factor_control(domain, domain_idx, l0_domain, 0, setting);
+        }
+        else if (domain == GEOPM_DOMAIN_BOARD_ACCELERATOR) {
```
lhlawson: Should be _CHIP.

service/src/LevelZero.hpp
```diff
@@ -122,6 +122,23 @@ namespace geopm
             virtual uint64_t active_time_timestamp(unsigned int l0_device_idx,
                                                    int l0_domain, int l0_domain_idx) const = 0;
 
+            /// @brief Get the number of LevelZero perf domains of a certain type
+            /// @param [in] geopm_domain The GEOPM domain being targeted
+            /// @param [in] l0_device_idx The LevelZero device being targeted
+            /// @param [in] l0_domain The LevelZero domain type being targeted
+            /// @return GPU perf domain count.
+            virtual int performance_domain_count(int geopm_domain, unsigned int l0_device_idx,
+                                          int l0_domain) const = 0;
```
avilcheslopez: Nit: alignment

service/src/LevelZero.cpp
```diff
@@ -267,6 +267,72 @@ namespace geopm
             }
         }
 
+        //Cache performance domains
```
avilcheslopez, line 314: Suggestion: consider refactoring this function due to its size (for readability and maintainability). It seems like it could be broken down into smaller helper functions, one for each domain being cached.

service/src/LevelZero.cpp
```diff
@@ -392,6 +458,45 @@ namespace geopm
         return m_devices.at(l0_device_idx).subdevice.engine_domain.at(l0_domain).size();
     }
 
+    int LevelZeroImp::performance_domain_count(int geopm_domain, unsigned int l0_device_idx, int l0_domain) const
```
avilcheslopez: I'm seeing two functions hidden within this one, where their respective interfaces are not 100% aligned, resulting in an interface that could cause confusion. I would recommend breaking this out:
- ``int perf_domain_count_gpu(unsigned int l0_device_idx)``
- ``int perf_domain_count_gpu_chip(unsigned int l0_device_idx, l0_domain)``

This would avoid unnecessarily passing the ``l0_domain`` parameter when calling the sub device variant. It would also get rid of the ``geopm_domain`` parameter, which would avoid having to deal with the case of an invalid domain being passed, currently being handled by returning 0 which could lead to a "silent bug".

service/src/LevelZero.cpp
```diff
@@ -392,6 +458,45 @@ namespace geopm
...
+        if (geopm_domain == GEOPM_DOMAIN_GPU) {
+            result = m_devices.at(l0_device_idx).perf_domain.size();
+        }
+        else if (geopm_domain == GEOPM_DOMAIN_GPU_CHIP) {
+            result = m_devices.at(l0_device_idx).subdevice.perf_domain.at(l0_domain).size();
+        }
+        return result;
+    }
+
+    double LevelZeroImp::performance_factor(int geopm_domain, unsigned int l0_device_idx,
```
avilcheslopez: Prior suggestion applies here as well.

service/src/LevelZero.cpp
```diff
@@ -603,6 +707,34 @@ namespace geopm
                         ": Sysman failed to set frequency.", __LINE__);
     }
 
+    void LevelZeroImp::performance_factor_control(int geopm_domain,
```
avilcheslopez: Similar interface change suggestion as for other ``performance_factor`` functions applies here as well.

service/src/LevelZeroDevicePool.cpp
```diff
@@ -327,6 +328,30 @@ namespace geopm
         return m_levelzero.energy(domain_idx);
     }
 
+    double LevelZeroDevicePoolImp::performance_factor(int domain,
```
avilcheslopez, line 414: Same interface suggestion as in LevelZero.

service/src/LevelZeroIOGroup.cpp
```diff
@@ -972,6 +1006,21 @@ namespace geopm
...
+            try {
+                // Currently only the levelzero compute domain control is supported.
+                // As new controls are added they should be included
+                m_perf_factor.at(domain_idx) =
+                    m_levelzero_device_pool.performance_factor(GEOPM_DOMAIN_GPU_CHIP,
+                                                               domain_idx,
+                                                               geopm::LevelZero::M_DOMAIN_COMPUTE);
+            }
+            catch (const geopm::Exception &ex) {
+                throw Exception("LevelZeroIOGroup::" + std::string(__func__) + ": "
```
avilcheslopez, line 1078: ``geopm::Exception()``

Note: the exception handling here is truncating the error message generated by the underlying API (LevelZeroDevicePool, LevelZero). If we don't have a convenient mechanism for nesting and reporting exceptions, maybe append the error message from the underlying exception to this one. That way, besides reporting that the operation failed, you provide the reason why.

service/src/LevelZeroIOGroup.cpp
```diff
@@ -997,6 +1046,23 @@ namespace geopm
...
+            }
+            try {
+                // Currently only the levelzero compute domain control is supported.
+                // As new controls are added they should be included
+                m_levelzero_device_pool.performance_factor_control(GEOPM_DOMAIN_GPU_CHIP,
+                                        domain_idx,
+                                        geopm::LevelZero::M_DOMAIN_COMPUTE,
+                                        m_perf_factor.at(domain_idx));
+            }
+            catch (const geopm::Exception &ex) {
```
avilcheslopez, line 1119: Same suggestion as before

service/src/LevelZero.cpp
```diff
@@ -285,6 +293,82 @@ namespace geopm
...
+        //Cache performance domains
+        num_domain = 0;
+        ze_result = zesDeviceEnumPerformanceFactorDomains(m_devices.at(
+                    device_idx).device_handle, &num_domain, nullptr);
+        if (ze_result == ZE_RESULT_ERROR_UNSUPPORTED_FEATURE) {
+#ifdef GEOPM_DEBUG
+            std::cerr << "Warning: <geopm> LevelZero: Performance domain detection is "
+                         "not supported.\n";
+#endif
+        }
```
avilcheslopez: Nit: it kind of bugs me that this will be an empty ``if`` when ``GEOPM_DEBUG`` is not defined. Is there a better way to structure this to avoid that. I see the other instances below avoid this situation, for example.

More importantly, should this just be a warning? How will the layer of code depending on this realize the feature is not supported?

service/src/LevelZero.cpp
```diff
@@ -285,6 +293,82 @@ namespace geopm
...
+                        m_devices.at(device_idx).subdevice.perf_domain.at(
+                                 geopm::LevelZero::M_DOMAIN_COMPUTE).push_back(handle);
+                    }
+#ifdef GEOPM_DEBUG
+                    else {
+                         std::cerr << "Warning: <geopm> LevelZero:" <<
+                                      " Unsupported sub-device level performance factor domain (" <<
+                                      std::to_string(property.engines) << ") detected." << std::endl;
+                    }
+#endif
```
avilcheslopez, line 354: I'm seeing that in cases like this one, considering that the ``else`` branch just provides a warning, it's possible for handles to be left with the default value with which the vector initialize them. I see later that these handles could be passed to the "ze" API functions. Can we rely on those correctly dealing with spurious handles and returning a proper error? Besides those calls, are there any other uses in the code that might assume the handles are filled in such that it would run into an unexpected error?

service/src/LevelZero.cpp
avilcheslopez: To be addressed as part of #2689

service/src/LevelZero.cpp
avilcheslopez: To be addressed as part of #2689

service/src/LevelZero.cpp
avilcheslopez: To be addressed as part of #2689

service/src/LevelZeroDevicePool.cpp
avilcheslopez: To be addressed as part of #2689

service/src/LevelZero.cpp
```diff
@@ -426,6 +510,45 @@ namespace geopm
...
+    int LevelZeroImp::performance_domain_count(int geopm_domain, unsigned int l0_device_idx, int l0_domain) const
+    {
+        int result = 0;
+        if (geopm_domain == GEOPM_DOMAIN_GPU) {
+            result = m_devices.at(l0_device_idx).perf_domain.size();
+        }
+        else if (geopm_domain == GEOPM_DOMAIN_GPU_CHIP) {
+            result = m_devices.at(l0_device_idx).subdevice.perf_domain.at(l0_domain).size();
+        }
+        return result;
```
avilcheslopez: This will silently proceed when an invalid domain is provided and return a default result of 0. Looking at it from an interface consistency point of view, there are other methods that throw in this situation, instead: the preexisting ``num_gpu()`` method and the newly added ``performance_factor_control()``. What criteria do you use to determine which behavior applies? If it's not very clear and there's potential for confusion, consider addressing this consistency issue.

lhlawson:
Splitting perf_factor & other levelzero API into GPU & GPU_CHIP functions tracked as part of #2689

service/src/LevelZero.cpp
lhlawson: If the performance domain is not detected then the m_devices perf domain structure size will be 0.  
The performance_domain_count function is used to check that the tracking structure is non-zero before any related calls.  

From a user perspective the IOGroup will not expose the performance factor signals (they'll be removed during signal trimming)

service/src/LevelZero.cpp
lhlawson: The _domain_count functions are intended to be used to check if it is sensible to call into any of the APIs around the associated feature (power, perf_factor, frequency, etc), hence they return 0.  The LevelZeroDevicePool::check_domain_exists will throw an error in this case.  

The control functions have additional checking around them as a safety precaution for conditions that are specifically not supported to avoid ending up in a case where something is changed on the system due to bad write.

service/src/LevelZero.cpp
lhlawson: These calls are checked using the check_domain_count function in LevelZeroDevicePool.  These checks were moved here to enable unit testing of those functions (as LevelZero.cpp is not unit-testable).  
We should not rely on the API dealing with spurious handles cleanly.

service/src/LevelZero.hpp
```diff
@@ -181,11 +196,23 @@ namespace geopm
             /// @param [in] l0_device_idx The index indicating a particular
             ///        Level Zero GPU.
             /// @param [in] domain The domain type being targeted
+            /// @param [in] domain_idx The domain being targeted
```
avilcheslopez: l0_domain
l0_domain_idx

service/src/LevelZero.hpp
```diff
@@ -181,11 +196,23 @@ namespace geopm
...
             /// @param [in] range_max Max target frequency in MHz.
             virtual void frequency_control(unsigned int l0_device_idx, int l0_domain,
                                            int l0_domain_idx, double range_min,
                                            double range_max) const = 0;
+
+            /// @brief Set the performance factor for the LevelZero device.
+            /// @param [in] l0_device_idx The index indicating a particular
+            ///        Level Zero accelerator.
+            /// @param [in] domain The level zero domain type being targeted
+            /// @param [in] domain_idx The level zero domain being targeted
```
avilcheslopez: l0_domain
l0_domain_idx

service/src/LevelZeroIOGroup.cpp
```diff
@@ -1010,7 +1044,24 @@ namespace geopm
             catch (const geopm::Exception &ex) {
                 throw Exception("LevelZeroIOGroup::" + std::string(__func__) + ": "
                                 + " Failed to fetch frequency control range for "
-                                " GPU_CHIP domain " + std::to_string(domain_idx),
+                                " GPU_CHIP domain " + std::to_string(domain_idx) +
+                                "due to exception.  Exception: " + ex.what(),
```
avilcheslopez: Suggestion: "due to exception: " + ex.what()

Same suggestion applies in changes below.

Disregard if this is already an established pattern followed for many exceptions in the file or LevelZero code in general.

avilcheslopez (dismissed):

service/src/LevelZero.cpp
```diff
@@ -285,6 +293,68 @@ namespace geopm
...
+
+            for (auto handle : perf_domain) {
+                zes_perf_properties_t property = {};
+                ze_result = zesPerformanceFactorGetProperties(handle, &property);
+                check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                "LevelZero::" + std::string(__func__) +
+                                ": Sysman failed to get domain performance factor properties",
+                                __LINE__);
+
+                //Finding non-subdevice domain.
```
```suggestion
                //Finding subdevice domain.
```

service/src/LevelZero.hpp
```diff
@@ -130,6 +130,21 @@ namespace geopm
...
+            /// @param [in] l0_domain The LevelZero domain type being targeted
+            /// @return GPU perf domain count.
+            virtual int performance_domain_count(unsigned int l0_device_idx,
+                                                 int l0_domain) const = 0;
+            /// @brief Get the performance factor value of various LevelZero domains
+            /// @param [in] l0_device_idx The LevelZero device being targeted
+            /// @param [in] l0_domain The LevelZero domain type being targeted
+            /// @param [in] l0_domain_idx The index indicating a particular
+            ///        Level Zero domain.
+            /// @return Accelerator or subdevice performance factor value
```
bgeltz:
```suggestion
            /// @return GPU or subdevice performance factor value
```

service/src/LevelZeroDevicePool.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "config.h"
 
+#include <iostream>
```
bgeltz:
```suggestion
```

service/src/LevelZeroDevicePool.cpp
```diff
@@ -392,6 +393,30 @@ namespace geopm
...
+                            " is not supported for the performance factor domain.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        std::pair<unsigned int, unsigned int> dev_subdev_idx_pair;
+        dev_subdev_idx_pair = subdevice_device_conversion(domain_idx);
+
+        check_domain_exists(m_levelzero.performance_domain_count(
+                                        dev_subdev_idx_pair.first, l0_domain),
+                                        __func__, __LINE__);
```
bgeltz:
```suggestion
        check_domain_exists(m_levelzero.performance_domain_count(dev_subdev_idx_pair.first, l0_domain),
                                                                 __func__, __LINE__);
```

service/src/LevelZeroDevicePool.cpp
```diff
@@ -412,4 +437,24 @@ namespace geopm
...
+                            ": domain " + std::to_string(domain) +
+                            " is not supported for the performance factor domain.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        std::pair<unsigned int, unsigned int> dev_subdev_idx_pair;
+        dev_subdev_idx_pair = subdevice_device_conversion(domain_idx);
+
+        check_domain_exists(m_levelzero.performance_domain_count(
+                                        dev_subdev_idx_pair.first, l0_domain),
+                                        __func__, __LINE__);
```
bgeltz: Same as above, please.

service/src/LevelZeroDevicePool.hpp
```diff
@@ -138,6 +138,16 @@ namespace geopm
             virtual uint64_t energy_timestamp(int domain, unsigned int domain_idx,
                                               int l0_domain) const = 0;
 
+            /// @brief Get the LevelZero device performance factor
+            /// @param [in] domain The GEOPM domain type being targeted
+            /// @param [in] domain_idx The GEOPM domain index
+            ///             (i.e. accelerator or chip being targeted)
```
bgeltz:
```suggestion
            ///             (i.e. GPU or chip being targeted)
```

service/src/LevelZeroDevicePool.hpp
```diff
@@ -138,6 +138,16 @@ namespace geopm
...
+            /// @param [in] l0_domain The LevelZero domain type being targeted
+            /// @return Accelerator or chip performance value, 0-1.0
```
bgeltz:
```suggestion
            /// @return GPU or chip performance value, 0-1.0
```

service/src/LevelZeroDevicePool.hpp
```diff
@@ -150,6 +160,15 @@ namespace geopm
                                            int l0_domain, double range_min,
                                            double range_max) const = 0;
 
+            /// @brief Set performance factor for LevelZero device.
+            /// @param [in] domain The GEOPM domain type being targeted
+            /// @param [in] domain_idx The GEOPM domain index
+            ///             (i.e. accelerator or chip being targeted)
```
bgeltz, line 174:
```suggestion
            ///             (i.e. GPU or chip being targeted)
```

service/src/LevelZeroDevicePool.hpp
```diff
@@ -150,6 +160,15 @@ namespace geopm
...
+            /// @param [in] l0_domain The LevelZero domain type being targeted
+            /// @param [in] setting Performance factor target, 0-1.0
```
bgeltz:
```suggestion
            /// @param [in] setting Performance factor target, 0 - 1.0
```

service/src/LevelZeroIOGroup.cpp
```diff
@@ -428,6 +444,13 @@ namespace geopm
                                     GEOPM_DOMAIN_GPU_CHIP,
                                     Agg::expect_same,
                                     string_format_double
+                                    }},
+                               {M_NAME_PREFIX + "GPU_CORE_PERFORMANCE_FACTOR_CONTROL", {
+                                    "Performance Factor",
```
bgeltz:
```suggestion
                                    "Expresses a trade-off between energy provided to the GPU compute hardware and the supporting units",
```

Please put this and the default value in a man page.  Also some general guidance about what setting numbers closer to 0 means vs. numbers closer to 1.

service/src/LevelZeroIOGroup.cpp
```diff
@@ -475,6 +498,9 @@ namespace geopm
                      IOGroup::M_SIGNAL_BEHAVIOR_VARIABLE}},
         })
         , m_frequency_range(m_platform_topo.num_domain(GEOPM_DOMAIN_GPU_CHIP), std::make_pair(0, 0))
+        // https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#tuning-workload-performance
+        // 'The default value is 50'
```
bgeltz:
```suggestion
```
Move this to the man page.

service/src/LevelZeroIOGroup.cpp
```diff
@@ -526,6 +552,9 @@ namespace geopm
         register_signal_alias("GPU_CORE_FREQUENCY_MAX_CONTROL",
                               M_NAME_PREFIX + "GPU_CORE_FREQUENCY_MAX_CONTROL");
 
+        register_signal_alias(M_NAME_PREFIX + "GPU_CORE_PERFORMANCE_FACTOR_CONTROL",
+                              M_NAME_PREFIX + "GPU_CORE_PERFORMANCE_FACTOR");
```
bgeltz, line 573: Add docu in the man page about the alias.

Update as necessary with a comment explaining what's going on here too.

service/src/LevelZeroIOGroup.cpp
```diff
@@ -983,6 +1012,11 @@ namespace geopm
                                                       geopm::LevelZero::M_DOMAIN_COMPUTE,
                                                       curr_min / 1e6, setting / 1e6);
         }
+        else if(control_name == M_NAME_PREFIX + "GPU_CORE_PERFORMANCE_FACTOR_CONTROL") {
+            m_levelzero_device_pool.performance_factor_control(domain_type, domain_idx,
+                                                               geopm::LevelZero::M_DOMAIN_COMPUTE,
+                                                               setting*100);
```
bgeltz:
```suggestion
                                                               setting * 100);
```

service/test/LevelZeroDevicePoolTest.cpp
```diff
@@ -90,6 +100,15 @@ TEST_F(LevelZeroDevicePoolTest, subdevice_conversion_and_function)
         EXPECT_EQ((uint64_t)(value+sub_idx+num_gpu_subdevice*40), m_device_pool.active_time_timestamp(GEOPM_DOMAIN_GPU_CHIP, sub_idx, MockLevelZero::M_DOMAIN_COMPUTE));
 
         EXPECT_NO_THROW(m_device_pool.frequency_control(GEOPM_DOMAIN_GPU_CHIP, sub_idx, MockLevelZero::M_DOMAIN_COMPUTE, value, value));
+
+        EXPECT_NO_THROW(m_device_pool.frequency_control(GEOPM_DOMAIN_GPU_CHIP, sub_idx, MockLevelZero::M_DOMAIN_COMPUTE, value, value));
```
bgeltz:
```suggestion
```

service/test/LevelZeroDevicePoolTest.cpp
```diff
@@ -90,6 +100,15 @@ TEST_F(LevelZeroDevicePoolTest, subdevice_conversion_and_function)
...
+
+        EXPECT_EQ(perf_value_chip_compute[sub_idx], m_device_pool.performance_factor(GEOPM_DOMAIN_GPU_CHIP, sub_idx, MockLevelZero::M_DOMAIN_COMPUTE));
+        EXPECT_EQ(perf_value_chip_mem[sub_idx], m_device_pool.performance_factor(GEOPM_DOMAIN_GPU_CHIP, sub_idx, MockLevelZero::M_DOMAIN_MEMORY));
+
+        //TODO: Add SUBDEVICE perf factor calls
```
bgeltz:
```suggestion
```

bgeltz (changes requested):

bgeltz (dismissed):

cmcantalupo (dismissed):
FWIW LGTM however there are conflicts in the rebase.

bgeltz (approved):

Issue #1779, opened 2021-08-12
Correct affinity determining logic in launcher.py


bgeltz:
The current implementation assumes that for _i_ sockets and _j_ physical cores per socket, the logical core numbering maps to:
1. Logical core 0 to logical core _j_ => physical cores 0 to _j_.
2. Logical core _j_ to logical core 2*_j_ => HTs associated with physical cores 0 to _j_ on the first socket.
3. Logical core 2*_j_ to logical core 3*_j_ => physical cores 0 to _j_ on the second socket.
4. Logical core 3*_j_ to logical core 4*_j_ => HTs associated with physical cores 0 to _j_ on the second socket.
5. This pattern continues up to _i_ sockets. 
 
This is not a safe assumption.  This logic should be replaced by inspecting the data in ```/sys/devices/system/cpu/cpuX/topology/*``` to determine which HTs are associated with which physical cores, and which logical cores are associated with which sockets.

https://github.com/geopm/geopm/blob/79e8f963fd9b386a7576233711764b161c68c86e/scripts/geopmpy/launcher.py#L667

More info:
https://software.intel.com/content/www/us/en/develop/articles/intel-64-architecture-processor-topology-enumeration.html
Sysfs interface:
https://www.kernel.org/doc/html/latest/admin-guide/cputopology.html

PR #2177, opened 2022-03-17
If linux is not NUMA enabled, create a single numa group for all CPUs

### LINKED ISSUE #2176: GEOPM Requires NUMA nodes to be provided by lscpu
**Describe the bug**
I tried to run geopmread -d on a single socket system and I expected system information to be provided, instead I received an error stating NUMA node information was not available.

**GEOPM version**
1.1.0+devdev1879g3972475b

**Expected behavior**
Something similar to:
```
./geopmread -d
board                   1
package                 1
core                    40
cpu                     40
board_memory            1
package_memory          0
board_nic               0
package_nic             0
board_accelerator       2
package_accelerator     0
board_accelerator_chip  2
```
**Actual behavior**
```
./geopmread
  what():  <geopm> Runtime error: PlatformTopoImp: parsing lscpu output, key not found: "NUMA node(s)": at src/PlatformTopo.cpp:498
```

**Error messages produced**
  what():  <geopm> Runtime error: PlatformTopoImp: parsing lscpu output, key not found: "NUMA node(s)": at src/PlatformTopo.cpp:498

geopm was built with --enable-debug

**Additional context**
This occurred on a single socket system with attached GPUs.  
The error is accurate, lscpu does not contain NUMA Node information.  The underlying hardware may not support this or this may be an issue with the system configuration.  In either case it would be preferable for a warning to be provided and execution to continue normally, as NUMA node information is not critical for all customers.

cmcantalupo:
- Fixes #2176

### DIFF
--- a/service/test/PlatformTopoTest.cpp
+++ b/service/test/PlatformTopoTest.cpp
@@ -66,6 +66,7 @@ class PlatformTopoTest : public :: testing :: Test
         std::string m_bdx_lscpu_str;
         std::string m_ppc_lscpu_str;
         std::string m_no0x_lscpu_str;
+        std::string m_no_numa_lscpu_str;
         bool m_do_unlink;
         std::shared_ptr<MockAcceleratorTopo> m_accelerator_topo;
 };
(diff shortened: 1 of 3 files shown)


service/test/PlatformTopoTest.cpp
```diff
@@ -239,6 +240,31 @@ void PlatformTopoTest::SetUp()
...
+        "CPU MHz:               2101.000\n"
+        "CPU max MHz:           2101.0000\n"
+        "CPU min MHz:           1200.0000\n"
+        "BogoMIPS:              4190.38\n"
+        "Virtualization:        VT-x\n"
+        "L1d cache:             32K\n"
+        "L1i cache:             32K\n"
+        "L2 cache:              256K\n"
+        "L3 cache:              46080K\n"
+        "Flags:                 fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf eagerfpu pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 fma cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch epb cat_l3 cdp_l3 invpcid_single intel_pt spec_ctrl ibpb_support tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm cqm rdt_a rdseed adx smap xsaveopt cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local dtherm ida arat pln pts\n";
```
cmcantalupo, line 267: This is the output used for the new unit test, would be better to have something more representative of the failing case.

lhlawson: The failing system had two notable differences from the unit test above:
- `On-line CPU(s) list: 0-71` Was provided instead of 'On-line CPU(s) mask:'. 
- lscpu listed `CPU(s): 72` and `Core(s) per socket: 16`.  This means that for the above example `Cores*Sockets*Threads < CPU(s)`.  This triggered another exception in PlatformTopo, however this behavior seems reasonable given that lscpu is providing inconsistent data..  WIP to determine why this happens and if there's a reasonable workaround.

bgeltz (approved):
This looks fine.  I think we're going to have to overhaul this a bit when we revisit #1779, but I'm not sure.  If `lscpu` can lie to GEOPM about the cores available, then we're going to have to check sysfs.

PR #2187, opened 2022-03-22
Improve geopmaccess error message for invalid names

### LINKED ISSUE #2186: Inadequate error messaging when adding invalid signals or controls to the access list
**Describe the bug**
I tried to add signals and controls to the access list that were actually invalid and I expected an error message that describes the problem instead the CLI suggests that I run with root even though I am running as user root.  

**GEOPM version**
9c4ad58

**Expected behavior**
The error message printed to syslog includes the details that describe the invalid names that were provided.  A similar error message should be provided for the user of the geopmaccess command line tool. 

This should be effected by the geopmaccess command line tool checking the user inputs *before* sending to the DBus interface, and not to rely on the daemon to discover the bad user input.  

**Actual behavior**
No reason is provided for the failure, other than the suggestion to run as root.

**Error messages produced**
CLI interface error message without enough information:

```
Error: Failed to set group control access list, try with sudo or as "root" user (requires CAP_SYS_ADMIN)
```

**Additional context**

cmcantalupo:
- Add user side check when modifying access lists
- Fixes #2186

### DIFF
--- a/service/geopmdpy/access.py
+++ b/service/geopmdpy/access.py
@@ -92,6 +92,11 @@ class Access:
             _, current_controls = self._geopm_proxy.PlatformGetGroupAccess(group)
         except DBusError as ee:
             raise RuntimeError('Failed to read group signal access list for specified group: {}'.format(group)) from ee
+        signals_supported, _ = self._geopm_proxy.PlatformGetAllAccess()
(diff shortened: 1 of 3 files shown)


dannosliwcd (approved):
LGTM

PR #2188, opened 2022-03-22
Add OBS integration to the CI workflow

bgeltz:
On any pushes to the `dev` branch (i.e. when a PR is merged) or to the `release-v2.0-candidate` branch (manual only), a new job will start that will make the distribution tarball from the pushed commit and transmit it and the generated spec file over to OBS.

This new job will only trigger for pushes to one of the target branches.  It will *not* fire for opening a PR, pushing to a PR, or closing to a PR.

The `dev` branch builds will be here:
https://build.opensuse.org/package/show/home:geopm/geopm-service

While the v2.0 RC builds will be here:
https://build.opensuse.org/package/show/home:geopm:release-v2.0-candidate/geopm-service

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,9 +36,11 @@ on:
   push:
     branches:
       - dev
+      - release-v2.0-candidate
   pull_request:
     branches:
       - dev
+      - release-v2.0-candidate
 
 concurrency:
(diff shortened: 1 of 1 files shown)


cmcantalupo (approved):
LGTM

dannosliwcd (approved):
Nice! This should make it easier to keep up with packaging issues between releases.

PR #2193, opened 2022-03-23
Do not install the devel package in install_service.sh

### LINKED ISSUE #2197: Update install_service.sh to not install the devel package
PR #2179 introduced changes in the spec file for building the service RPMs such that the symlink for libgeopm.so to the versioned .so file was only created if the devel RPM was installed.

PR #2192 changed the cffi bindings to use the versioned .so file.

The `install_service.sh` scripts needs to be updated as it represents what we expect users to install in production.  The script should *not* install the devel package.

bgeltz:
Resolves #2197.


cmcantalupo (approved):
Lgtm

PR #2200, opened 2022-03-23
Use more scalable mechanism for validating Unix group

### LINKED ISSUE #2199: The way that Unix groups are queried in system_files.py is not supported on some distributed systems
**Describe the bug**
I tried to use the group access lists on a large HPC cluster running LDAP and I expected to be able to modify access lists for user groups instead an error was reported saying that the Unix group was not found.

**GEOPM version**
1708279

**Expected behavior**
Do not use `grp.getgrall()` to query all Unix groups which is not supported on some at scale systems.  The only reason to have this list is to verify that a group is valid., so just use `grp.getgrnam()` to verify that the user's requested group exists.

**Actual behavior**
The call to `grp.getgrall()` returns an empty list, and attempts to update access lists for individual groups will fail.

**Error messages produced**
```
dasbus.error.DBusError: Linux group is not defined: group = "valid_group"
```

**Additional context**
Thanks @bensallen for the suggested solution:
```
diff --git a/service/geopmdpy/system_files.py b/service/geopmdpy/system_files.py
index 51e950ac..5405aeed 100644
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -714,7 +714,6 @@ class AccessLists(object):
     def __init__(self, config_path=GEOPM_SERVICE_CONFIG_PATH):
         self._CONFIG_PATH = config_path
         self._DEFAULT_ACCESS = '0.DEFAULT_ACCESS'
-        self._ALL_GROUPS = [gg.gr_name for gg in grp.getgrall()]
         self._pio = pio
 
     def _validate_group(self, group):
@@ -724,7 +723,9 @@ class AccessLists(object):
             group = str(group)
             if group[0].isdigit():
                 raise RuntimeError('Linux group name cannot begin with a digit: group = "{}"'.format(group))
-            if group not in self._ALL_GROUPS:
+            try:
+                grp.getgrnam(group)
+            except KeyError:
                 raise RuntimeError('Linux group is not defined: group = "{}"'.format(group))
         return group
```

cmcantalupo:
- Fix use of system call that may not be supported
  on some distributed systems.
- This implementation is more scalable and efficient
- Rather than querying for all Unix groups, just verify
  the request from the user.
- Fixes #2199

Co-authored-by: Ben Allen <bsallen@alcf.anl.gov>

### DIFF
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -714,7 +714,6 @@ class AccessLists(object):
     def __init__(self, config_path=GEOPM_SERVICE_CONFIG_PATH):
         self._CONFIG_PATH = config_path
         self._DEFAULT_ACCESS = '0.DEFAULT_ACCESS'
-        self._ALL_GROUPS = [gg.gr_name for gg in grp.getgrall()]
         self._pio = pio
 
     def _validate_group(self, group):
@@ -724,7 +723,9 @@ class AccessLists(object):
             group = str(group)
             if group[0].isdigit():
                 raise RuntimeError('Linux group name cannot begin with a digit: group = "{}"'.format(group))
-            if group not in self._ALL_GROUPS:
+            try:
+                grp.getgrnam(group)
+            except KeyError:
                 raise RuntimeError('Linux group is not defined: group = "{}"'.format(group))
         return group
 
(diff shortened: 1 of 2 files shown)


cmcantalupo:
Thanks @bensallen

bgeltz (approved):

PR #2203, opened 2022-03-24
Run astyle across every C and C++ source file

cmcantalupo:
This commit can be easily regenerated with the astyle tool.  It should go in when there are not many outstanding PRs that would be affected.


examples/print_affinity.cpp
```diff
@@ -49,16 +49,16 @@ int main(int argc, char** argv)
 
     std::ostringstream thread_list;
 #ifdef _OPENMP
-#pragma omp parallel for ordered
```
cmcantalupo: I'm not sure exactly what astyle is doing here, maybe it is indenting due to the appearance of the `for` keyword?  Probably better to just edit out open mp pragma changes

Krebrov001 (approved):

PR #2207, opened 2022-03-25
Addressed issue 2206 - mapping aliased levelzero derivative signals

### LINKED ISSUE #2206: LEVELZERO::GPU_POWER and GPU_POWER read_batch behavior differs
**Describe the bug**
I tried to do a read_batch of GPU_POWER and I expected timing equivalent to LEVELZERO::GPU_POW#ER instead the Derivative signal read() path was used leading to a 35-40ms delay in tracing

**GEOPM version**
1.1.0+dev1888ge44df591

**Expected behavior**
GPU_POWER and LEVELZERO::GPU_POWER timing should be near identical when used in a trace.  

**Actual behavior**
GPU_POWER takes 35-40ms to return, LEVELZERO::GPU_POWER takes < 1ms.  

For debug geopmread_main.cpp was modified to call push_signal, then repeatedly call read_batch(), sample(idx), and report the timing of the calls. 
Expectation is that a read_batch & sample timing for the GPU_POWER alias will match that of LEVELZERO::GPU_POWER.  Timed values <1ms is truncated to 0ms:

```
ZES_ENABLE_SYSMAN=1 geopmread LEVELZERO::GPU_POWER board 0;
time: 0 ms, result: nan
time: 0 ms, result: <number>
time: 0 ms, result: <number>
ZES_ENABLE_SYSMAN=1 geopmread GPU_POWER board 0;
time: 35 ms, result: nan
time: 35 ms, result: nan
time: 35 ms, result: nan
```

**Additional context**
The level zero read_batch does a check of signals to make sure they're not in the derivative signal pushed list:
https://github.com/geopm/geopm/blob/7f9ed5b3793345d7e8024652f09bb272ddbed6ce/service/src/LevelZeroIOGroup.cpp#L720

This is populated when a signal is pushed if that signal is in the derivative_signal_map:
https://github.com/geopm/geopm/blob/7f9ed5b3793345d7e8024652f09bb272ddbed6ce/service/src/LevelZeroIOGroup.cpp#L659-L660

When GPU_POWER is added as an alias, it's not added to the derivative signal map by default.  Proposed solution is to check any aliases in register_signal_aliases and if the signal they're aliasing is a derivative signal add the alias to the derivative_signal_map tracking.

lhlawson:

- Relates to #2206 
- Fixes #2206 

Added code so any alias of a derivative signal is added to the derivative signals list.  This prevents the derivative signal from directly calling read() in the read_batch case.  

Tested using a geopmread_main.cpp modified to do repeated read_batch() calls (as described in #2206):
```
:~> ZES_ENABLE_SYSMAN=1 geopmread LEVELZERO::GPU_POWER board 0
time: 0 ms, result: nan
time: 0 ms, result: <number>
time: 0 ms, result: <number>
time: 0 ms, result: <number>

:~> ZES_ENABLE_SYSMAN=1 geopmread GPU_POWER board 0
time: 0 ms, result: nan
time: 0 ms, result: <number>
time: 0 ms, result: <number>
time: 0 ms, result: <number>
```

### DIFF
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -1018,8 +1018,7 @@ namespace geopm
             "\n    alias_for: " + signal_name;
 
         auto der_it = m_derivative_signal_map.find(signal_name);
-        if (der_it != m_derivative_signal_map.end())
-        {
+        if (der_it != m_derivative_signal_map.end()) {
             m_derivative_signal_map[alias_name] = der_it->second;
         }
     }
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -1010,11 +1010,18 @@ namespace geopm
             // skip adding an alias if underlying signal is not found
             return;
         }
+
         // copy signal info but append to description
         m_signal_available[alias_name] = it->second;
         m_signal_available[alias_name].m_description =
             m_signal_available[signal_name].m_description +
             "\n    alias_for: " + signal_name;
+
+        auto der_it = m_derivative_signal_map.find(signal_name);
+        if (der_it != m_derivative_signal_map.end())
+        {
+            m_derivative_signal_map[alias_name] = der_it->second;
+        }
     }
 
     void LevelZeroIOGroup::register_control_alias(const std::string &alias_name,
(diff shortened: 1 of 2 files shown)


cmcantalupo (dismissed):

service/src/LevelZeroIOGroup.cpp
```diff
@@ -1010,11 +1010,18 @@ namespace geopm
...
+
         // copy signal info but append to description
         m_signal_available[alias_name] = it->second;
         m_signal_available[alias_name].m_description =
             m_signal_available[signal_name].m_description +
             "\n    alias_for: " + signal_name;
+
+        auto der_it = m_derivative_signal_map.find(signal_name);
+        if (der_it != m_derivative_signal_map.end())
+        {
```
cmcantalupo:
```suggestion
        if (der_it != m_derivative_signal_map.end()) {
```

cmcantalupo (approved):

bgeltz (approved):

PR #2251, opened 2022-04-08
Enable reference counting for multiple nested open sessions

### LINKED ISSUE #2011: Error messages produced when a GEOPM Service session is terminated by the administrator are confusing
**Describe the bug**
I tried to run any of the client facing DBus APIs after a session was closed by an administrator and I expected to get an error message that explains that the system admin has closed the session instead a message about how a session was never opened is printed.  

Additionally, if a user has a program which calls:

```
io.github.geopm.PlatformOpenSession()
io.github.geopm.PlatformReadSignal("signal0", "board", 0)
io.github.geopm.PlatformOpenSession()
io.github.geopm.PlatformReadSignal("signal1", "board", 0)
io.github.geopm.PlatformCloseSession()
io.github.geopm.PlatformReadSignal("signal2", "board", 0)
io.github.geopm.PlatformCloseSession()
```
the call to `io.github.geopm.PlatformReadSignal("signal2", "board", 0)` will fail saying that a session must be opened, although no error will be reported when the second call to `PlatformOpenSession()` is made.

**GEOPM version**
705bb36fe6b13079fdc8a0eae3098c1a3f7b1d59

**Expected behavior**
Better error messaging over all.  It would also be preferable that there were some sort of reference counter provided for the client side open/close calls for the session.  This reference counter would have to be robust to termination calls through the `PlatformCloseSessionAdmin()` interface.

**Actual behavior**
Confusing error messages, and failure to support nested calls to PlatformOpenSession()

**Error messages produced**
There are not integration tests that produce this error.  This issue is being created based on developer analysis of the source code.  Providing integration tests that cover cases where the session is terminated by the administrator while the client is active should be included in fixing this issue.

Krebrov001:
Enable the PlatformService to support clients having multiple
independent components, which can concurrently use the single
connection. Establish a shared pointer-esque abstraction over
connections, a one to one mapping with the client process.

Add the reference count to the JSON active session schema.

- Relates to #2011 
- Fixes #2011

### DIFF
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -358,12 +358,9 @@ class PlatformService(object):
                               the session.
 
         """
-        # if the connection already exists
         if self._active_sessions.is_client_active(client_pid):
             self._active_sessions.increment_reference_count(client_pid)
         else:
-            # Establish the connection and add an active session with
-            # the initial reference count of 1.
             signals, controls = self.get_user_access(user)
             watch_id = self._watch_client(client_pid)
             self._active_sessions.add_client(client_pid, signals, controls, watch_id)
@@ -399,21 +396,14 @@ class PlatformService(object):
         """
         self._active_sessions.check_client_active(client_pid, 'PlatformCloseSession')
         reference_count = self._active_sessions.get_reference_count(client_pid)
-        # A negative reference_count is impossible because this condition
-        # is enforced by the _session_schema.
         if reference_count == 0:
-            # this is the case that the geopm daemon was killed while
-            # that in the process of _close_Session_completely so we
-            # need to call _close_Session_completely() again. The
-            # daemon died, was restarted, and found a session with
+            # The daemon died, restarted, and found a session with
             # zero reference count.
             self._close_session_completely(client_pid)
         elif reference_count == 1:
-            # last reference, close the session completely
             self._active_sessions.decrement_reference_count(client_pid)
             self._close_session_completely(client_pid)
         else:  # reference_count > 1:
-            # more than one reference, decrement the reference count
             self._active_sessions.decrement_reference_count(client_pid)
 
     def close_session_admin(self, client_pid):
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -310,7 +310,7 @@ class PlatformService(object):
         raise NotImplementedError('PlatformService: Implementation incomplete')
 
     def open_session(self, user, client_pid):
-        """Open a new session session for the client thread.
+        """Open a new session for the client thread.
 
         The creation of a session is the first step that a client
         thread makes to interact with the GEOPM service.  Each Linux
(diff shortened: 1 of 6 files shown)


service/geopmdpy/system_files.py
```diff
@@ -570,6 +572,91 @@ def set_watch_id(self, client_pid, watch_id):
...
+        We want to provide that same connection as a singleton, and keep track of
+        how many independent components have access to that same connection as a resource,
+        in order to know when to deallocate that resource when all these components
+        have disconnected from the session.
+
+        Args:
+            client_pid (int): Linux PID that opened the session
+
+        Returns:
+            int: GLib watch ID to track the session lifetime
```
cmcantalupo:
```suggestion
            int: The reference count
```

service/geopmdpy/system_files.py
```diff
@@ -570,6 +572,91 @@ def set_watch_id(self, client_pid, watch_id):
...
+
+        Raises:
+            RuntimeError: Client does not have an open session
+
+        """
+        self.check_client_active(client_pid, 'get_reference_count')
+        return self._sessions[client_pid]['reference_count']
+
+    def set_reference_count(self, client_pid, reference_count):
```
cmcantalupo, line 603: This interface only seems to be used in the test.  We can leave it, but I'm not sure how it would be used in practice.

service/geopmdpy/service.py
```diff
@@ -349,11 +358,15 @@ def open_session(self, user, client_pid):
                               the session.
 
         """
+        # if the connection already exists
```
bgeltz:
```suggestion
```

service/geopmdpy/service.py
```diff
@@ -349,11 +358,15 @@ def open_session(self, user, client_pid):
...
         if self._active_sessions.is_client_active(client_pid):
-            return
-        signals, controls = self.get_user_access(user)
-        watch_id = self._watch_client(client_pid)
-        self._active_sessions.add_client(client_pid, signals, controls, watch_id)
+            self._active_sessions.increment_reference_count(client_pid)
+        else:
+            # Establish the connection and add an active session with
+            # the initial reference count of 1.
```
bgeltz:
```suggestion
```

service/geopmdpy/service.py
```diff
@@ -370,10 +383,64 @@ def close_session(self, client_pid):
...
+        open session.
+
+        Args:
+            client_pid (int): Linux PID of the client thread
+
+        """
+        self._active_sessions.check_client_active(client_pid, 'PlatformCloseSession')
+        reference_count = self._active_sessions.get_reference_count(client_pid)
+        # A negative reference_count is impossible because this condition
+        # is enforced by the _session_schema.
```
bgeltz:
```suggestion
```

service/geopmdpy/service.py
```diff
@@ -370,10 +383,64 @@ def close_session(self, client_pid):
...
+        if reference_count == 0:
+            # this is the case that the geopm daemon was killed while
+            # that in the process of _close_Session_completely so we
+            # need to call _close_Session_completely() again. The
+            # daemon died, was restarted, and found a session with
+            # zero reference count.
```
bgeltz:
```suggestion
            # The daemon died, restarted, and found a session with
            # zero reference count.
```

service/geopmdpy/service.py
```diff
@@ -370,10 +383,64 @@ def close_session(self, client_pid):
...
+            self._close_session_completely(client_pid)
+        elif reference_count == 1:
+            # last reference, close the session completely
```
bgeltz:
```suggestion
```

service/geopmdpy/service.py
```diff
@@ -370,10 +383,64 @@ def close_session(self, client_pid):
...
+            self._active_sessions.decrement_reference_count(client_pid)
+            self._close_session_completely(client_pid)
+        else:  # reference_count > 1:
+            # more than one reference, decrement the reference count
```
bgeltz:
```suggestion
```

bgeltz (changes requested):
Do we need to add a destructor to the `PlatformService` class that calls `close_session_completely()` for all the client PIDs that are active?  I expected to see this chunk of code somewhere.

bgeltz (approved):

PR #2254, opened 2022-04-13
(abandoned) Add a FrequencyTimeBalancer class

### LINKED ISSUE #2253: Add a helper that estimates relationships between per-core frequency controls and epoch performance
> estimate 1
Parent feature: #2575

Introduce new C++ class with the following features:
- identifies CPU core frequencies that should be set to rebalance an application based on time spent in a region of interest

Done criteria:

- [ ] balance when current frequencies are all unlimited
- [ ] balance when all frequencies should go unlimited
- [ ] does not change frequencies when already balanced
- [ ] independently balances subdomains (e.g., independently balances cores within one package vs cores within another package)
- [ ] can ignore control domain indices (e.g., to avoid balancing cores that do not execute the application)
- [ ] does not use frequency-capped cores as balance reference
- [ ] resets to baseline if invariants are violated (e.g., if there is not at least one previously-unlimited core)

Some types of HPC workloads exhibit imbalance, where some workers spend time waiting for others to reach synchronization points. Imbalance can be modeled and corrected in different ways depending on which types of signals and controls are available. 

I would like to have an agent that balances time spent in a monitored region of code by modulating frequency controls. Such an agent would benefit from routines that can recommend frequency control settings based on previous frequency settings and measurements of time in the region of interest.

A solution might assume a worst-case relationship between performance and frequency, or it might attempt to model that relationship internally. Either way, it would be nice if the interface does not require knowledge of the internal modeling mechanism.

dannosliwcd:
- Selects frequencies that should be set to balance execution times
- Resolves #2253

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -434,6 +434,8 @@ libgeopmpolicy_la_SOURCES = src/Accumulator.cpp \
                             src/FrequencyGovernorImp.hpp \
                             src/FrequencyMapAgent.cpp \
                             src/FrequencyMapAgent.hpp \
+                            src/FrequencyTimeBalancer.cpp \
+                            src/FrequencyTimeBalancer.hpp \
                             src/Imbalancer.cpp \
                             src/ModelParse.cpp \
                             src/ModelParse.hpp \
--- a/src/FrequencyTimeBalancer.cpp
+++ b/src/FrequencyTimeBalancer.cpp
@@ -1,0 +1,255 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ *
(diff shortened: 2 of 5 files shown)


dannosliwcd:
Converted back to draft mode. There were are some changes to consider tradeoffs between different types of balancing decisions (e.g., when the choice of frequency on one core impacts the achievable range of frequencies on another core). I'm working through how to cleanly propagate that information to this layer, and it will probably result in at least some interface changes here.

dannosliwcd:
Closing to restart the PR series for the frequency balancer agent (tasks tracked in story #2575).

PR #2255, opened 2022-04-13
Changes to FrequencyMapAgent to solve the JSON bug

### LINKED ISSUE #1483: FrequencyMapAgent report extensions not proper YAML
The FrequencyMapAgent adds some information to the host report about the frequency map that cannot be parsed as YAML.

Related to #615

Krebrov001:
The FrequencyMapAgent adds some information to the host report about the frequency map that cannot be parsed as YAML.

This fixes the frequency map so that it becomes valid JSON, which can then be parsed by the python interpreter.

- Relates to #1483 
- Fixes #1483


src/FrequencyMapAgent.hpp
```diff
@@ -79,6 +79,11 @@ namespace geopm
             static std::unique_ptr<Agent> make_plugin(void);
             static std::vector<std::string> policy_names(void);
             static std::vector<std::string> sample_names(void);
+            void set_maps(const std::map<uint64_t, double>& hash_freq_map,
+                          const std::set<uint64_t>& default_freq_hash) {
+                m_hash_freq_map = hash_freq_map;
+                m_default_freq_hash = default_freq_hash;
+            }
```
bgeltz, line 86: I think it would be better to test this through our integration tests than to introduce a new class method that is only used in testing.

Typically when we're testing the behavior of agents, this is done through an integration test.

This would also illustrate the indentation issues I commented on.  You won't see that in a unit test.

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +364,33 @@ namespace geopm
     std::vector<std::pair<std::string, std::string> > FrequencyMapAgent::report_host(void) const
     {
         std::vector<std::pair<std::string, std::string> > result;
-        std::ostringstream oss;
-        oss << std::setprecision(M_PRECISION) << std::scientific;
```
bgeltz, line 341: This line that uses the ```set_precision``` call is important and it must be preserved.

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +364,33 @@ namespace geopm
...
+            full_map.insert({ key, m_default_freq });
+        });
+
+        std::map<std::string, Json> temp_map;
+        for (const auto &region : full_map)
+        {
+            // Json object(int(region.first));
+            // std::string key_string = object.dump();
+            // key_string.erase(std::remove(key_string.begin(), key_string.end(), '"'), key_string.end());
+            // temp_map[key_string] = region.second;
```
bgeltz: Remove unused code prior to review.  It distracts from what you are really doing with the changed code.

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +364,33 @@ namespace geopm
...
-            oss << "\n    0x" << std::hex << std::setfill('0') << std::setw(16) << std::fixed;
-            oss << region;
-            oss << std::setfill('\0') << std::setw(0) << std::scientific;
-            oss << ": " << m_default_freq;
+
+        std::map<uint64_t, double> full_map(m_hash_freq_map);
+        std::for_each(m_default_freq_hash.begin(), m_default_freq_hash.end(), [&](uint64_t key)
+        {
+            full_map.insert({ key, m_default_freq });
+        });
```
bgeltz: I'd avoid `std::for_each`.  This is much simpler:

```suggestion      
        for (const auto &region : m_default_freq_hash) {
            full_map.insert({region, m_default_freq});
        }
```

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +364,33 @@ namespace geopm
...
+
+        std::map<std::string, Json> temp_map;
+        for (const auto &region : full_map)
+        {
+            // Json object(int(region.first));
+            // std::string key_string = object.dump();
+            // key_string.erase(std::remove(key_string.begin(), key_string.end(), '"'), key_string.end());
+            // temp_map[key_string] = region.second;
+            std::ostringstream key_ss;
+            key_ss << "0x" << std::hex << std::setfill('0') << std::setw(16) << std::fixed;
```
bgeltz, line 358:
```suggestion
            key_ss << "\n    0x" << std::hex << std::setfill('0') << std::setw(16) << std::fixed;
```

This needs to remain so that the indentation in the report is still correct.

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +364,33 @@ namespace geopm
...
+            std::ostringstream value_ss;
+            value_ss << std::setfill('\0') << std::setw(0) << std::scientific;
+            value_ss << region.second;
+            std::string key_string = key_ss.str();
+            key_string.erase(std::remove(key_string.begin(), key_string.end(), '"'), key_string.end());
+            temp_map[key_string] = value_ss.str();
         }
-        oss << "\n";
-        result.push_back(std::make_pair("Frequency map", oss.str()));
+        Json my_json = Json(temp_map);
```
bgeltz, line 363: I'd prefer to see the original method of using a single `std::ostringstream` used here instead of splitting up the region and the frequency strings into separate streams.

You should be able to avoid that `erase` call, and it's overall less code.  The way you have it here will also result in improper indentation in the report which will cause a parsing problem in the `RawReport` Python code.

The resulting output string from the `std::ostringstream` could still be parsed by json11 if that is desired.

bgeltz (changes requested):

test/FrequencyMapAgentTest.cpp
```diff
@@ -485,3 +486,112 @@ TEST_F(FrequencyMapAgentTest, validate_policy)
...
+                                  "0x000000007b561f45: 2.100000e+09, "
+                                  "0x00000000a74bbf35: 1.200000e+09, "
+                                  "0x00000000d691da00: 1.900000e+09, "
+                                  "0x8000000000000000: 2.100000e+09}";
+
+    auto result = m_agent->report_host();
+    for (auto& key_value : result) {
+        key_value.second.erase(std::remove(key_value.second.begin(), key_value.second.end(), '"'), key_value.second.end());
+        // std::cerr << key_value.first << " : " << key_value.second << std::endl;
+        // std::cerr << key_value.first << " : " << reference_map << std::endl;
```
bgeltz:
```suggestion
```

src/FrequencyMapAgent.hpp
bgeltz: Don't do this, do it with a test constructor instead.

service/src/geopm/Helper.hpp
```diff
@@ -89,6 +89,16 @@ namespace geopm
...
 
+    /// @brief Finds all occurences of the old substring in a string,
+    ///        and replaces them with the new substring.
+    /// @param [in, out] source_dest The string within which to replace the substring.
+    /// @param [in] old_substring The substring to be replaced.
+    /// @param [in] new_substring The new substring to replace the old one with.
+    /// @return bool True if one or more replacements have been made.
+    bool replace_all_substrings(std::string &source_dest,
+                                const std::string &old_substring,
+                                const std::string &new_substring);
```
bgeltz: The function and its implementation should be removed.  It is not used anywhere in the implementation for the bug fix.

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +372,26 @@ namespace geopm
...
+        }
+
+        std::map<std::string, Json> temp_map;
+        for (const auto &region : full_map)
+        {
+            std::ostringstream key_ss;
+            key_ss << "0x" << std::hex << std::setfill('0') << std::setw(16) << std::fixed;
+            key_ss << region.first;
+            std::string key_string = key_ss.str();
+            key_string.erase(std::remove(key_string.begin(), key_string.end(), '"'), key_string.end());
```
bgeltz: Is this line still needed?  We want the keys (which are region hashes in this case) to be treated as strings.  Please add a comment here on why this line is still needed if it is.

test/FrequencyMapAgentTest.cpp
```diff
@@ -485,3 +486,108 @@ TEST_F(FrequencyMapAgentTest, validate_policy)
...
+    FrequencyMapAgent frequency_agent(
+        {},
+        {},
+        *m_platform_io, *m_platform_topo
+    );
+    std::string reference_map =  "{}";
+
+    auto result = frequency_agent.report_host();
+    for (auto& key_value : result) {
+        key_value.second.erase(std::remove(key_value.second.begin(), key_value.second.end(), '"'), key_value.second.end());
```
bgeltz: Same question here as before: what purpose does this line serve?  The map is empty.

bgeltz (changes requested):

service/src/geopm/Helper.hpp
Krebrov001: It would be useful later though. I spend several hours on this implementation.

src/FrequencyMapAgent.cpp
Krebrov001: You're right, this line is extraneous, I'll remove it.

test/FrequencyMapAgentTest.cpp
Krebrov001: This line removes the extra `"` characters from the string. Previously the keys used to be strings surrounded with quotes: `"0x01234567"` Now they are just `0x01234567`

src/FrequencyMapAgent.cpp
```diff
@@ -364,22 +372,25 @@ namespace geopm
...
+            key_ss << region.first;
+            std::string key_string = key_ss.str();
+            temp_map[key_string] = region.second;
         }
-        oss << "\n";
-        result.push_back(std::make_pair("Frequency map", oss.str()));
+        Json my_json = Json(temp_map);
+
+        std::string frequency_map_data = my_json.dump();
+        result.push_back(std::make_pair("Frequency map", frequency_map_data));
```
bgeltz:
```suggestion
        result.push_back(std::make_pair("Frequency map", Json(temp_map).dump()));
```

test/FrequencyMapAgentTest.cpp
```diff
@@ -485,3 +486,108 @@ TEST_F(FrequencyMapAgentTest, validate_policy)
...
+                                  "0x000000007b561f45: 2100000000, "
+                                  "0x000000007b561f47: null, "
+                                  "0x00000000a74bbf35: 1200000000, "
+                                  "0x00000000d691da00: 1900000000, "
+                                  "0x00000000d691da02: null, "
+                                  "0x8000000000000000: 2100000000}";
+
+    auto result = frequency_agent.report_host();
+    for (auto& key_value : result) {
+        key_value.second.erase(std::remove(key_value.second.begin(), key_value.second.end(), '"'), key_value.second.end());
```
bgeltz:
```suggestion
```

test/FrequencyMapAgentTest.cpp
```diff
@@ -332,12 +339,6 @@ TEST_F(FrequencyMapAgentTest, split_policy)
 #endif
 }
 
-TEST_F(FrequencyMapAgentTest, name)
-{
-    EXPECT_EQ("frequency_map", m_agent->plugin_name());
-    EXPECT_NE("bad_string", m_agent->plugin_name());
-}
-
```
bgeltz: Why did this move up?  Please put it back.  It is not related to the other changes you are making.

test/Makefile.mk
```diff
@@ -212,13 +212,17 @@ GTEST_TESTS = test/gtest_links/AccumulatorTest.empty \
...
               test/gtest_links/FrequencyGovernorTest.frequency_bounds_invalid \
               test/gtest_links/FrequencyGovernorTest.validate_policy \
+              test/gtest_links/FrequencyMapAgentTest.name \
               test/gtest_links/FrequencyMapAgentTest.adjust_platform_map \
               test/gtest_links/FrequencyMapAgentTest.adjust_platform_uncore \
-              test/gtest_links/FrequencyMapAgentTest.name \
+              test/gtest_links/FrequencyMapAgentTest.split_policy \
               test/gtest_links/FrequencyMapAgentTest.enforce_policy \
               test/gtest_links/FrequencyMapAgentTest.policy_to_json \
-              test/gtest_links/FrequencyMapAgentTest.split_policy \
```
bgeltz: Please put these (the `name` and `split_policy` tests) back in the order they were previously in.  We try to keep things arranged alphabetically by class.

bgeltz (changes requested):

test/Makefile.mk
```diff
@@ -192,6 +192,10 @@ GTEST_TESTS = test/gtest_links/AccumulatorTest.empty \
               test/gtest_links/FrequencyMapAgentTest.policy_to_json \
               test/gtest_links/FrequencyMapAgentTest.split_policy \
               test/gtest_links/FrequencyMapAgentTest.validate_policy \
+              test/gtest_links/FrequencyMapAgentTest.test_hash_freq_map \
+              test/gtest_links/FrequencyMapAgentTest.test_default_freq_hash \
+              test/gtest_links/FrequencyMapAgentTest.test_both_map_and_set \
+              test/gtest_links/FrequencyMapAgentTest.test_neither_map_nor_set \
```
bgeltz:
```suggestion
              test/gtest_links/FrequencyMapAgentTest.report_hash_freq_map \
              test/gtest_links/FrequencyMapAgentTest.report_default_freq_hash \
              test/gtest_links/FrequencyMapAgentTest.report_both_map_and_set \
              test/gtest_links/FrequencyMapAgentTest.report_neither_map_nor_set \
```

Then alphabetize the list above here.

bgeltz (approved):

bgeltz (changes requested):
Please take a look at my comment.

Also, were you going to fix the PR description?  It still says WIP.

test/Makefile.mk
```diff
@@ -190,6 +190,10 @@ GTEST_TESTS = test/gtest_links/AccumulatorTest.empty \
               test/gtest_links/FrequencyMapAgentTest.name \
               test/gtest_links/FrequencyMapAgentTest.enforce_policy \
               test/gtest_links/FrequencyMapAgentTest.policy_to_json \
+              test/gtest_links/FrequencyMapAgentTest.report_hash_freq_map \
+              test/gtest_links/FrequencyMapAgentTest.report_default_freq_hash \
+              test/gtest_links/FrequencyMapAgentTest.report_both_map_and_set \
+              test/gtest_links/FrequencyMapAgentTest.report_neither_map_nor_set \
```
bgeltz:
```suggestion
              test/gtest_links/FrequencyMapAgentTest.report_both_map_and_set \
              test/gtest_links/FrequencyMapAgentTest.report_default_freq_hash \
              test/gtest_links/FrequencyMapAgentTest.report_hash_freq_map \
              test/gtest_links/FrequencyMapAgentTest.report_neither_map_nor_set \
```

bgeltz: That's the proper alphabetical order.

bgeltz (approved):

Issue #2267, opened 2022-04-15
As an AI researcher I would like to be able to tune the hyper parameters of Pytorch models used in GEOPM


lhlawson:
**Is your feature request related to a problem? Please describe.**
Currently there is no tuning of hyper parameters proposed for ML based agents.  This would allow for tuning the models used to provide better frequency decisions

**Describe the solution you'd like**
An implementation of hyperparameter tuning around the train_*_model-pytorch.py file introduced in PR #2146 and #2217.  Preferably using the Ray Tune library

**Describe alternatives you've considered**
Using 'best  guess' hyper parameters or custom one off solutions to hyper parameter tuning.

PR #2217, opened 2022-03-30
Addition of C++ CPU Pytorch Agent Tutorial 

lhlawson:
- Relates to #2216 feature request from github issues
- Builds on PR #2044 (sweep process scripts used)

This PR introduces:
1.  A new pytorch based neural net training script that operates on the output of a modified version of the process_cpu_frequency_sweep.py script introduced in PR #2044.
2. A new C++ GEOPM agent capable of loading these models and using them to steer CPU frequency decisions
3. A short tutorial on how to use the new capabilities based on workloads (PARRES DGEMM and NSTREAM) previously added to the GEOPM Integration Infrastructure

Todo before merge:
- [ ] License review and updates. Pytorch license available [here](https://github.com/pytorch/pytorch/blob/master/LICENSE)
- [ ] Consider C++14 requirement for GEOPM - To move this Agent into the runtime the GEOPM runtime would need to support C++14 (currently using C++11). No compilation issues have been found with this change but it may cause issues on older systems.

Issues keeping this in draft:
- [x] Updates to README.md to indicate changes beyond #2146 
- [x] Validation of results using per-package CPU-NN approach or modification of approach
- [x] Breaking the PR out into two PRs - one for the CPU Pytorch Agent, one for the Node (CPU + GPU) Pytorch Agent.

PR #2268, opened 2022-04-15
Pytorch Model Hyperparameter Training

lhlawson:
- Relates to #2216 feature request from github issues
- Based on PR #2217 
- Relates to #2267 feature request from github issues

This PR updates the train_*_model-pytorch.py files as follows:
1. Introduces the ray tuning library for hyper parameter tuning.  
2. Separates training, testing, and the training loop into separate functions as required by 1
3. Separation of the training set int o a training and validation set.  
4. If a --leave-app-out is specified it is used as a test set after training

This approach to hyper parameter tuning follows the basic pytorch example [here](https://pytorch.org/tutorials/beginner/hyperparameter_tuning_tutorial.html) and the Ray Tune tutorial [here](https://docs.ray.io/en/latest/tune/getting-started.html#tune-tutorial)

Hyper parameters tuned include:
- Fully connected layer width
- Fully connected layer depth
- Learning rate
- Batch Size
- Loss Criterion
- Different NN approaches (FF, LSTM, RNN)

Potential improvements:
- Different search space selection for the config values (i.e. how we select learning rate, width, depth), potentially from [here](https://docs.ray.io/en/latest/tune/faq.html#how-do-i-choose-hyperparameter-ranges)
- Alternative to ASHA scheduler for early termination (AHB, PBT, RS, or BO)

- epoch count tuning

This PR relies on the Ray Tune library being installed, as indicated in the README.md.  Ray has an Apache 2.0 license.

Example output during run:
image

Note that if the python file is edited during an active run the following error will occur:
```
Python type cannot be used as a value:
...
    def __init__(self, width_input, width_fc, depth_fc):
        super(FFNet, self).__init__()
              ~~~~~ <--- HERE
...
```
In order to address this issue when/if it occurs:
1. The final model is saved pre-torchscript conversion.  to <input>-pre-torchscript
2. The best width, depth, batch size, and learning rate is provided at the end of hyperparameter tuning
3. A --convert-model option is added to rerun just the torchscript conversion portion.  This requires using --width and --depth:
```
--convert-model <input>-pre-torchscript --width <best_width> --depth <best_depth>
```
The --convert-model option currently limited to the Feed Forward Net.
Though Ray Tune may select LSTM the C++ CPUTorchAgent currently only supports basic Feed Forward Nets

In addition this PR provides a -force-linear-phi-freq option for GPU frequency sweep processing that modifies the existing perf-energy tradeoff frequency selection by forcing non repeating linearly spaced frequencies for every phi value between the initial perf-energy tradeoff maximum and minimum frequency identified.  In practice this should help phi have a greater dynamic range.

This PR also provides experiment scripts that use the GEOPM integration infrastructure.


tutorial/pytorch_agent/train_cpu_model-pytorch.py
```diff
@@ -0,0 +1,336 @@
...
+        x = self.norm1(x)
+        x = self.fc1(x)
+        x = self.sigmoid(x)
+        # This approach doesn't work for torchscript - torch.jit.script(best_model)
+        #for d in range(self.depth_fc):
+        #    x = (self.fc_list[d])(x)
+        #    x = self.sigmoid(x)
+        for d in range(self.depth_fc):
+            x = self.fc2(x)
+            x = self.sigmoid(x)
```
lhlawson: Using a list of fc layers does not seem to agree with torchscript.  We may have to chose between hyperparameter tuning & saving to a script model.  This would result in a two pass system (tune, then update the model depth & train with new settings for a model usable by the agent)

lhlawson: Solved in commit bbe5cb2

tutorial/pytorch_agent/train_cpu_model-pytorch.py
```diff
@@ -74,15 +97,20 @@ def main():
     #Print phi to phi-freq mapping
     print(df_traces.pivot_table('phi-freq', 'phi', 'app-config'))
 
+    if args.convert_model is not None:
+        model = P3Net(width_input=len(X_columns), width_fc=args.width, depth_fc=args.depth)
+        model.load_state_dict(torch.load(args.convert_model))
+        model_to_script(model, args.output)
+        exit(1)
```
dannosliwcd: This looks like a successful termination case. Should it `exit(0)` instead?

Also, I think `exit()` usually [only exists in interactive python shells](https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module). Maybe `sys.exit()` instead.

tutorial/pytorch_agent/train_cpu_model-pytorch.py
```diff
@@ -110,96 +137,218 @@ def main():
...
+    with torch.no_grad():
+        for idx, (inputs, target_control) in enumerate(test_loader):
+            prediction_total += inputs.size(0)
+            # Run inputs through model, save prediction
+            predicted_control = model(inputs)
+            # Round to nearest 100 MHz increment
+            predicted_control = np.round(predicted_control,1)
+
+            # Generate stats
+            prediction_correct += (target_control == predicted_control).sum().item()
```
dannosliwcd, line 434: Should this be scored with a metric that gives partial credit for an almost-right answer instead of no credit? E.g., MAE or MSE, as in train(). If the answer is no, maybe add a comment noting why.

tutorial/pytorch_agent/train_cpu_model-pytorch.py
```diff
@@ -110,96 +137,218 @@ def main():
...
+                    prediction_max = predicted_control.max()
+                if (prediction_min > predicted_control.min()):
+                    prediction_min = predicted_control.min()
+
+        print('\ttest output min: {:.2f}, max: {:.2f}.'.format(prediction_min, prediction_max))
+    return prediction_correct/prediction_total
+
+class P3Net(nn.Module):
+    def __init__(self, width_input, width_fc, depth_fc):
+        super(P3Net, self).__init__()
```
dannosliwcd: Maybe related to the error message you're seeing: Why not `super().__init__()`?

tutorial/pytorch_agent/train_cpu_model-pytorch.py
lhlawson: Great question!  This is the final test against data that was not used in training and validation, and is only hit if --leave-app-out was specified.  We're rounding any frequency decisions to the nearest 100MHz to match common CPU controls which provides some leeway.  All that said, I'm in favor of passing back loss as proposed.  

In the training loop we are passing back this same accuracy metric and using it for model selection and early termination via ASHA.  We are also passing back loss using MSE there, but it is only used for reporting.  We could change lines 166-167 to as called out in the TODO:
```
metric="accuracy", #TODO: we may want to switch to using metric="loss", mode="min"
mode="max",
```

To have ASHA use the loss instead, but we should investigate the tradeoff there

dannosliwcd: Is STEP_SIZE a universal limitation across all of our CPU frequency control interfaces (beyond just the ones we already have IOGroups for)?

The reason I ask is because I wonder if the 100 MHz rounding behavior belongs in a controller helper's code (e.g., an agent, the FrequencyGovernor, or similar) instead. Then this (or other frequency recommenders) could recommend the best frequency for a given decision and the governor or other class could decide what to do when the exact frequency can't be specified.

tutorial/pytorch_agent/train_cpu_model-pytorch.py
lhlawson: super().__init__() may work here as well.  In testing it would occasionally hit the same error however.

lhlawson:
An LSTM Net and RNN have been added. They appear to function, but require a normalization layer (preferably through nn.LayerNorm) and likely requirefurther tuning and refinement.  

The CPUTorchAgent only supports FF nets (it would require a change in input prepping to handle LSTM or RNN) at this time, so this is primarily an example of how we may add additional models for selection in tuning.

tutorial/pytorch_agent/train_cpu_model-pytorch.py
lhlawson: I believe some client hardware uses 50MHz steps.  On the non GPU side we know that NVIDIA hardware supported by the NVMLIOGroup has 7 & 8 MHz steps, and these changes should be moved to the train_gpu_model-pytorch.py code eventually.
I've updated the code to use a single 'evaluate()' function that provides accuracy and loss, which may be used instead of accuracy.

alawibaba:
This is great! I've started testing this as a drop-in for the hyperparameter sweep script that I was working on. I had one minor comment: I think this *should* work with RNNs without modification -- I think the torchscript model will handle maintaining the state for us. I'll test that and report back.

alawibaba (dismissed):

Issue #2259, opened 2022-04-13
Fix all issues identified by static analysis in geopm-service C++ implementation


cmcantalupo:
> estimate 1
Parent story: #2219

Done criteria:
- [x] Pull requests are merged which fix all real issues in geopm/service/src
- [x] False positives are dealt with in the infrastructure

cmcantalupo:
Related to #2269

PR #2276, opened 2022-04-19
Clean up C++ issues exposed by static analysis

cmcantalupo:
- Relates to #2259

### DIFF
--- a/service/src/Exception.cpp
+++ b/service/src/Exception.cpp
@@ -105,6 +105,7 @@ namespace geopm
 
     Exception &Exception::operator=(const Exception &other)
     {
+        std::runtime_error::operator=(other);
         m_err = other.m_err;
         return *this;
     }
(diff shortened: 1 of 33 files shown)


bgeltz (dismissed):
LGTM.  It's probably a good idea to at least run test_monitor.py before merging since there were changes to MSRIO.

service/src/Exception.cpp
```diff
@@ -103,6 +103,12 @@ namespace geopm
 
     }
 
+    Exception &Exception::operator=(const Exception &other)
+    {
+        m_err = other.m_err;
+        return *this;
+    }
```
avilcheslopez, line 111: Should this call the base class' assignment operator as well?

avilcheslopez (commented):
Just a minor observation

avilcheslopez (approved):
Looks good to me!

service/src/Exception.cpp
cmcantalupo: Thanks!

Issue #2260, opened 2022-04-13
Fix all issues in geopm HPC runtime C++ implementation discovered during static analysis


cmcantalupo:
> estimate 1
Parent story: #2219

Done criteria:
- [x] Pull requests are merged which fix all real issues in geopm/src and geopm/tutorials
- [x] False positives are dealt with in the infrastructure

PR #2279, opened 2022-04-21
Fix static issues discovered

cmcantalupo:
- Relates to #2259
- Relates to #2260

### DIFF
--- a/examples/timed_region.cpp
+++ b/examples/timed_region.cpp
@@ -36,6 +36,7 @@ int main(int argc, char**argv)
     iterations = 0;
     while (iterations < total_iterations) {
         geopm_tprof_post();
+        ++iterations;
     }
     geopm_prof_exit(region_id[1]);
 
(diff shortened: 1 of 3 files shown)


examples/timed_region.cpp
```diff
@@ -46,6 +47,7 @@ int main(int argc, char**argv)
     iterations = 0;
     while (iterations < total_iterations) {
         geopm_tprof_post();
+        ++iterations;
     }
     geopm_prof_exit(region_id[2]);
```
avilcheslopez, line 53: Not really part of this review, but worth pointing out:

I see this block of code repeated verbatim three times in this file. It might be worth factoring it out (especially if it is more "generalizable"). If it had been, the fix would've had to be made in only one place.

avilcheslopez (approved):
Just a minor note not really part of this review that might be worth considering for a future change.

examples/timed_region.cpp
cmcantalupo: Probably worth considering the removal of the "examples" directory for 2.0.  It is kind of a mess of random use cases.

PR #2281, opened 2022-04-22
Remove write feature from geopmsession command line tool

### LINKED ISSUE #2104: As a user of GEOPM I would like a single command line interface to set controls
The use case for the "write mode" of geopmsession is much less clear now that the write lock is not coupled directly to the process that writes (but rather the process leader).  The original use case for geopmsession's write mode was a work around for this problem.  

I suggest we de-feature geopmsession, and modify all examples and integration tests that use the write mode of geopmsession to use the geopmwrite command line tool instead.

We could also leave the feature, but I think the documentation will be confusing for users who are used geopmwrite and the requirement to kill the geopmsession process running in the background cannot be done without emitting warning messages.

**Additional context**
It may also be good to remove any direct use of the service proxy from geompsession, and have it use the pio interface instead.  Originally the pio interface was not part of geopmdpy, but now it is.

cmcantalupo:
- Fixes #2104

### DIFF
--- a/service/docs/source/geopmsession.1.rst
+++ b/service/docs/source/geopmsession.1.rst
@@ -1,26 +1,18 @@
-.. role:: raw-html-m2r(raw)
-   :format: html
 
-
-geopmsession(1) -- Command line interface for the geopm service read/write features
+geopmsession(1) -- Command line interface for the geopm service batch read features
 ===================================================================================
-
-
-
-
 
 
 DESCRIPTION
 -----------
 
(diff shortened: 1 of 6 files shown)


avilcheslopez (commented):
Just for my understanding and maybe to get more context:

Since geopmsession allows performing batch reads in this more interactive mode, how would the user be able to do the same with batch writes (or is that not a common use case)?

avilcheslopez (approved):

Issue #2257, opened 2022-04-13
Fix all issues in geopmpy discovered during static analysis


cmcantalupo:
> estimate 3
Parent story: #2219

Done criteria:
- [x] Pull requests are merged which fix all real issues in geopmpy
- [x] False positives are dealt with in the infrastructure

Issue #2256, opened 2022-04-13
Fix all issues in geopmdpy discovered during static analysis


cmcantalupo:
> estimate 3
Parent story: #2219

Done criteria:
- [ ] Pull requests are merged which fix all real issues in geopmdpy
- [ ] False positives are dealt with in the infrastructure

PR #2283, opened 2022-04-23
Address PlatformTopo issues around the lscpu cache

### LINKED ISSUE #1548: lscpu cache persists on reboot
On systems where the contents of /tmp persist between reboots (which is most systems by default), the lscpu cache may be incorrect because of e.g. BIOS changes.

A solution should be developed to ensure that the cache is correct.  Presently we've discussed 2 possibilities:
1. ~~Remove the cache at the end of a run, always.~~
2. Compare the cache file's last modified time against the system uptime, and if the uptime is newer remove the old file.

https://man7.org/linux/man-pages/man2/sysinfo.2.html

### LINKED ISSUE #2269: lscpu cache is created insecurely
**Describe the bug**
I tried to use `/tmp/geopm-topo-cache` and I expected it to be created in a secure manner instead it is readable/writable by the world.

**GEOPM version**
dev

**Expected behavior**
The topo cache should be created securely:
1. When the service starts up, the service should create the cache with secure permissions in the `/var/run` directory for the service.
2. When the runtime starts up it should first check `/var/run` for the topo cache.  If it is not present, then the runtime should create the cache securely in `/tmp` for the specific user who is initiating the run.  The UID should be appended to the file name to avoid name collisions.
3. If the runtime wrote the cache to `/tmp` it should be removed at the end of a run.
4. When the service is shutdown, the file in `/var/run` should be removed?

**Actual behavior**
The file is permissively created in `/tmp` and persists forever.

**Additional context**
Relates to #1548

bgeltz:
- Resolves #1548.
- Resolves #2269.
- Related to #2256.
- Related to #2257.

### DIFF
--- a/service/geopmdpy/topo.py
+++ b/service/geopmdpy/topo.py
@@ -211,7 +211,7 @@ def create_cache():
 
     """
     global _dl
-    err = _dl.geopm_topo_create_cache_service()
+    err = _dl.geopm_topo_create_cache()
     if err < 0:
-        raise RuntimeError("geopm_topo_create_cache_service() failed: {}".format(
+        raise RuntimeError("geopm_topo_create_cache() failed: {}".format(
             error.message(err)))
--- a/service/geopmdpy/topo.py
+++ b/service/geopmdpy/topo.py
@@ -201,8 +201,8 @@ def create_cache():
     functions in the topo module as well as any use of the GEOPM
     runtime.  File permissions of the cache file are set to
     "-rw-r--r--", i.e. 644. The path for the cache file is
-    /run/geopm-service/geopm-topo-cache.  If thie file is older than
-    the last boot it will be rengerated.  If the file exists but has
+    /run/geopm-service/geopm-topo-cache.  If this file is older than
+    the last boot it will be regenerated.  If the file exists but has
     improper permissions is will be regenerated.  If the file has been
     created since the last boot with the correct permissions, no
     operation will be performed.  To force the creation of a new cache
--- a/service/geopmdpy/topo.py
+++ b/service/geopmdpy/topo.py
@@ -200,14 +200,18 @@ def create_cache():
     exist.  This cache file will be used by any calls to the other
     functions in the topo module as well as any use of the GEOPM
     runtime.  File permissions of the cache file are set to
-    "-rw-rw-rw-", i.e. 666. The path for the cache file is
-    /tmp/geopm-topo-cache.  If the file exists no operation will be
-    performed.  To force the creation of a new cache file call
-    os.unlink('/tmp/geopm-topo-cache') prior to calling this function.
+    "-rw-r--r--", i.e. 644. The path for the cache file is
+    /run/geopm-service/geopm-topo-cache.  If thie file is older than
+    the last boot it will be rengerated.  If the file exists but has
+    improper permissions is will be regenerated.  If the file has been
+    created since the last boot with the correct permissions, no
+    operation will be performed.  To force the creation of a new cache
+    file call os.unlink('/run/geopm-service/geopm-topo-cache') prior
+    to calling this function.
 
     """
     global _dl
-    err = _dl.geopm_topo_create_cache()
+    err = _dl.geopm_topo_create_cache_service()
     if err < 0:
-        raise RuntimeError("geopm_topo_create_cache() failed: {}".format(
+        raise RuntimeError("geopm_topo_create_cache_service() failed: {}".format(
             error.message(err)))
--- a/service/src/PlatformTopo.cpp
+++ b/service/src/PlatformTopo.cpp
@@ -90,9 +90,9 @@ int geopm_topo_popen(const char *cmd, FILE **fid)
         }
         g_is_popen_complete = 0;
         sigaction(SIGCHLD, &save_action, NULL);
-    }
-    if (!err && *fid == NULL) {
-        err = errno ? errno : GEOPM_ERROR_RUNTIME;
+        if (*fid == NULL) {
+            err = errno ? errno : GEOPM_ERROR_RUNTIME;
+        }
     }
     return err;
 }
@@ -508,6 +508,7 @@ namespace geopm
             }
             int err = fchmod(tmp_fd, perms);
             if (err) {
+                close(tmp_fd);
                 throw Exception("PlatformTopo::create_cache(): Could not chmod tmp_path: ",
                                 errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
             }
@@ -660,7 +661,7 @@ namespace geopm
         }
         else {
             mode_t expected_perms;
-            if (file_path.c_str() == M_SERVICE_CACHE_FILE_NAME) {
+            if (file_path == M_SERVICE_CACHE_FILE_NAME) {
                 expected_perms = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH; // 0o644
             }
             else {
--- a/service/src/PlatformTopo.cpp
+++ b/service/src/PlatformTopo.cpp
@@ -3,14 +3,21 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+#include "config.h"
+
 #include "PlatformTopoImp.hpp"
 
 #include <signal.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/sysinfo.h>
+#include <sys/time.h>
 #include <unistd.h>
 #include <cpuid.h>
 #include <string.h>
+#include <errno.h>
+#include <limits.h>
+#include <stdio.h>
 
 #include <map>
 #include <sstream>
@@ -18,11 +25,11 @@
 #include <stdexcept>
 
 #include "geopm_sched.h"
+#include "geopm_time.h"
 #include "geopm/Exception.hpp"
 #include "geopm/Helper.hpp"
 #include "AcceleratorTopo.hpp"
 
-#include "config.h"
 
 int geopm_read_cpuid(void)
 {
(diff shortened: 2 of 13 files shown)


service/src/PlatformTopo.cpp
```diff
@@ -8,17 +8,25 @@
...
 #include <stdexcept>
+#include <iostream> // debug
 
 #include "geopm_sched.h"
+#include "geopm_time.h"
 #include "geopm/Exception.hpp"
+#include "geopm/Helper.hpp"
 #include "AcceleratorTopo.hpp"
 
 #include "config.h"
```
cmcantalupo: Unrelated to this patch, but config.h should be first.

service/src/PlatformTopo.cpp
```diff
@@ -418,30 +461,80 @@ namespace geopm
...
     void PlatformTopoImp::create_cache(const std::string &cache_file_name)
     {
-        // If cache file is not present, create it
-        struct stat cache_stat;
-        if (stat(cache_file_name.c_str(), &cache_stat)) {
-            std::string cmd = "out=" + cache_file_name + ";"
-                              "lscpu -x > $out && "
-                              "chmod a+rw $out";
+        // If cache file is not present, or is too old, create it
+        bool file_ok = false;
```
cmcantalupo: file_ok -> is_file_okay?

service/src/PlatformTopo.cpp
```diff
@@ -418,30 +461,80 @@ namespace geopm
...
+            char tmp_path[NAME_MAX];
+            strncpy(tmp_path, tmp_string.c_str(), NAME_MAX);
+            int tmp_fd = mkstemp(tmp_path);
+            if (tmp_fd == -1) {
+                throw Exception("PlatformTopo::create_cache(): Could not create temp file: ",
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+            close(tmp_fd);
+
+            int err = chmod(tmp_path, perms);
```
cmcantalupo:
```suggestion

            int err = fchmod(tmp_fd, perms);
            close(tmp_fd);
```

service/src/PlatformTopo.cpp
```diff
@@ -418,30 +461,80 @@ namespace geopm
...
+            if (err) {
+                throw Exception("PlatformTopo::create_cache(): Could not chmod tmp_path: ",
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+
+            std::string cmd = "out=" + std::string(tmp_path) + ";"
+                              "lscpu -x > $out";
```
cmcantalupo:
```suggestion
            std::ostringstream cmd;
            cmd << "lscpu -x >> " << tmp_path << ";"
```

service/src/PlatformTopo.cpp
```diff
@@ -418,30 +461,80 @@ namespace geopm
...
+
             FILE *pid;
-            int err = geopm_sched_popen(cmd.c_str(), &pid);
+            err = geopm_topo_popen(cmd.c_str(), &pid);
```
cmcantalupo:
```suggestion
            err = geopm_topo_popen(cmd.str().c_str(), &pid);
```

service/src/PlatformTopo.cpp
```diff
@@ -418,30 +461,80 @@ namespace geopm
...
+                                err, __FILE__, __LINE__);
             }
             if (pclose(pid)) {
                 unlink(cache_file_name.c_str());
                 throw Exception("PlatformTopo::create_cache(): Could not pclose lscpu command: ",
-                         errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+
+            err = rename(tmp_path, cache_file_name.c_str()); // Ensure both files are on the same filesystem
```
cmcantalupo: think we can drop the comment now.

cmcantalupo:
```
In file included from src/PlatformTopo.cpp:8:
./src/PlatformTopoImp.hpp:50:18: error: private field 'm_do_fclose' is not used [-Werror,-Wunused-private-field]
            bool m_do_fclose;
```

service/src/geopm/PlatformTopo.hpp
```diff
@@ -17,7 +17,8 @@
 extern "C"
 {
     /// @brief Identify host CPU.
-    int geopm_read_cpuid(void); // Only used in examples/geopm_platform_supported.cpp
+    int geopm_read_cpuid(void);
+    int geopm_topo_popen(const char *cmd, FILE **fid);
```
cmcantalupo: Maybe it would be better to just re-declare this function in the test file that needs it.  Otherwise we are committing to the API for 2.0 since this is an installed header.  I hope we can get away from lscpu eventually.

cmcantalupo (commented):
Not sure I like having the geopm_topo_popen() in the public API.  Otherwise looks good.

service/geopmdpy/topo.py
```diff
@@ -200,14 +200,18 @@ def create_cache():
     exist.  This cache file will be used by any calls to the other
     functions in the topo module as well as any use of the GEOPM
     runtime.  File permissions of the cache file are set to
-    "-rw-rw-rw-", i.e. 666. The path for the cache file is
-    /tmp/geopm-topo-cache.  If the file exists no operation will be
-    performed.  To force the creation of a new cache file call
-    os.unlink('/tmp/geopm-topo-cache') prior to calling this function.
+    "-rw-r--r--", i.e. 644. The path for the cache file is
+    /run/geopm-service/geopm-topo-cache.  If thie file is older than
```
avilcheslopez: Minor typo: thie

service/src/PlatformTopo.cpp
```diff
@@ -55,9 +63,44 @@ int geopm_read_cpuid(void)
...
+    struct sigaction save_action;
+    g_popen_complete_signal_action.sa_handler = geopm_topo_popen_complete;
+    sigemptyset(&g_popen_complete_signal_action.sa_mask);
+    g_popen_complete_signal_action.sa_flags = 0;
+    err = sigaction(SIGCHLD, &g_popen_complete_signal_action, &save_action);
+    if (!err) {
+        *fid = popen(cmd, "r");
+        while (*fid && !g_is_popen_complete) {
+
+        }
```
avilcheslopez, line 90: It makes me a bit nervous to have this wait until the expected signal is received, especially because the interface this provides allows passing any string as the command for popen(). At the very least, input validation should be performed on cmd, but would also recommend establishing a limit on how long this loops.

This is more of a concern if this function is exposed more publicly. Otherwise, if it's only used internally and input to it is very well controlled, it's much less of a concern.

service/src/PlatformTopo.cpp
```diff
@@ -55,9 +63,44 @@ int geopm_read_cpuid(void)
...
+        g_is_popen_complete = 0;
+        sigaction(SIGCHLD, &save_action, NULL);
+    }
+    if (!err && *fid == NULL) {
+        err = errno ? errno : GEOPM_ERROR_RUNTIME;
+    }
```
avilcheslopez: Minor suggestion: you can move this `if` into the previous `if` and reduce it to:

`if (*fid == NULL) {...`

since the previous `if` already ensures `!err` is true.

service/src/PlatformTopo.cpp
```diff
@@ -418,30 +460,79 @@ namespace geopm
...
+            std::string tmp_string = cache_file_name + "XXXXXX";
+            char tmp_path[NAME_MAX];
+            strncpy(tmp_path, tmp_string.c_str(), NAME_MAX);
+            int tmp_fd = mkstemp(tmp_path);
+            if (tmp_fd == -1) {
+                throw Exception("PlatformTopo::create_cache(): Could not create temp file: ",
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+            int err = fchmod(tmp_fd, perms);
+            if (err) {
```
avilcheslopez, line 505: In case of error, before throwing: `close(tmp_fd)`

However, seeing the intention is to close it, anyway, you could simply close it right after the call to `fchmod()` regardless of return value.

service/src/PlatformTopo.cpp
```diff
@@ -544,77 +635,90 @@ namespace geopm
...
-            result = fopen(M_CACHE_FILE_NAME.c_str(), "r");
-            if (!result) {
-                int err = geopm_sched_popen("lscpu -x", &result);
-                if (err) {
-                    throw Exception("PlatformTopoImp::open_lscpu(): Could not popen lscpu command",
-                                    errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
-                }
-                m_do_fclose = false;
+            mode_t expected_perms;
+            if (file_path.c_str() == M_SERVICE_CACHE_FILE_NAME) {
```
avilcheslopez: This banks on the fact that when the service cache file name is passed in `file_path`, it will be passed as `M_SERVICE_CACHE_FILE_NAME` (which is how it should be, hence this const member's existence). Maybe I'm being too paranoid, but for full robustness, I would recommend using std::string's operator==.

If not for that, at least for consistency, seeing an earlier similar check is using std::string's operator==, as opposed to doing pointer comparison.

avilcheslopez (commented):
LGTM! Just a few minor comments...

service/src/PlatformTopo.cpp
avilcheslopez: Sorry, I'm way off. M_SERVICE_CACHE_FILE_NAME is of type std::string. Even if it had been const char* my comment is incorrect.

The operator== that takes char* and std::string is being used. I guess the only thing worth noting is that calling c_str() is not necessary and both std::strings could be passed to operator==.

service/src/PlatformTopo.cpp
bgeltz: I have moved the declaration of this out of the public header, so for the moment this is only used in the PlatformTopoImp, and the MSRIOGroupTest.  In both cases the input is very well controlled.

Moreover, this chunk of code was just moved around in this PR.  It is certainly a candidate for a refactor if we wanted to put this back into a public API.

avilcheslopez (dismissed):

avilcheslopez (approved):

PR #2286, opened 2022-04-26
Fixed LevelZero error function to handle basic errors encountered and…

### LINKED ISSUE #2277: Improper handling of oneAPI Level Zero error codes
**Describe the bug**
In the Level Zero API wrapper (LevelZero.cpp), in the error handling code for API calls, when constructing error messages the code may throw an exception when converting error codes to strings if the error code is not found, which would result in a confusing error message for the user (as it would be an error message related to an error code not found in a map, as opposed to the intended error message that was being constructed).

**GEOPM version**
1.1.0+dev1968g621ab397

**Expected behavior**
The Level Zero wrapper code in GEOPM should be able to provide a relevant/meaningful/helpful error message when a call to a Level Zero API fails, even when it encounters an error code it's not familiar with.

**Actual behavior**
Unexpected Level Zero API error codes cause the Level Zero wrapper code in GEOPM to throw a mapping error (instead of the actual error that should be reported).

lhlawson:
… provide error code in all other cases

- Relates to #2277  from github issues
- Fixes #2277  change request from github issues.

Removed uncommon/un-encountered errors, added generic error handling to throw level zero error code.


service/src/LevelZero.cpp
```diff
@@ -628,46 +620,15 @@ namespace geopm
...
-               "ZE_RESULT_ERROR_OVERLAPPING_REGIONS"},
               {ZE_RESULT_ERROR_UNKNOWN,
                "ZE_RESULT_ERROR_UNKNOWN"},
             };
-        std::string error_string = error_mapping.at(ze_result);
+
+        std::string error_string = std::to_string(ze_result);
+        if (error_mapping.find(ze_result) != error_mapping.end()) {
+            error_string = error_mapping.at(ze_result);
+        }
```
avilcheslopez, line 630: Very minor suggestion, not to stop the review:

```suggestion
        const auto it = error_mapping.find(ze_result);
        if (it != error_mapping.end()) {
            error_string = it->second;
        }
```

avilcheslopez (approved):

service/src/LevelZero.cpp
lhlawson: Seems reasonable to me!

bgeltz (approved):

PR #2288, opened 2022-04-26
GPU Compute Activity Agent updates for LevelZero GPUs supporting the gpu chip domain

### LINKED ISSUE #2287: Update the LevelZero IO Group and GPU Compute Activity Agent to support LevelZero GPUs
> estimate 1
Parent [feature/story]: #2045 

The implementation for the GPU Compute Activity agent was modified because it did not support LevelZero GPUs exposing chip level controls.

The implementation for the LevelZero IO Group was modified to provide a GPU_COMPUTE_ACTIVITY alias for the GPU Compute Activity agent.  

Done criteria:

- [x] Unit tests that cover the feature run and pass
- [x] Documentation changes are required to inform the user about CHIP vs GPU level support

lhlawson:
- Relates to #2287 from github issues
- Fixes #2287 change request from github issues.

Updates to the GPU-CA to support accelerators providing GPU CHIP level signals and controls.  

This change is intended to be fully backwards compatible with hardware previously supported by the GPU-CA.  

Testing of backwards compatibility using the integration infrastructure PARRES DGEMM and PARRES NSTREAM implementation matched expectations on systems using NVIDIA V100s.

### DIFF
--- a/integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_oneapi.py
+++ b/integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_oneapi.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python3
+#  Copyright (c) 2015 - 2022, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+'''
+Run ParRes dgemm with the gpu_activity agent.
+'''
+import argparse
+
+from experiment import machine
+from experiment.energy_efficiency import gpu_activity
+from apps.parres import parres
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_activity.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_dgemm_appconf_oneapi(mach, args)
+    gpu_activity.launch(app_conf=app_conf, args=args,
+                        experiment_cli_args=extra_args)
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -150,7 +150,7 @@ namespace geopm
 
             // If we have more than one device confirm all devices have the same
             // number of subdevices
-            for (int idx = 1; idx < m_devices.size(); ++idx) {
+            for (unsigned int idx = 1; idx < m_devices.size(); ++idx) {
                 if (m_devices.at(idx).m_num_subdevice != m_devices.at(idx - 1).m_num_subdevice) {
                     throw Exception("LevelZero::" + std::string(__func__) +
                                     ": GEOPM Requires the number of subdevices to be" +
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -151,7 +151,7 @@ namespace geopm
             // If we have more than one device confirm all devices have the same
             // number of subdevices
             for (int idx = 1; idx < m_devices.size(); ++idx) {
-                if (m_devices.at(idx).m_num_subdevice != m_devices.at(idx-1).m_num_subdevice) {
+                if (m_devices.at(idx).m_num_subdevice != m_devices.at(idx - 1).m_num_subdevice) {
                     throw Exception("LevelZero::" + std::string(__func__) +
                                     ": GEOPM Requires the number of subdevices to be" +
                                     " the same on all devices. " +
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -147,6 +147,18 @@ namespace geopm
                                 " Please check ZE_AFFINITY_MASK enviroment variable settings",
                                 GEOPM_ERROR_INVALID, __FILE__, __LINE__);
             }
+
+            // If we have more than one device confirm all devices have the same
+            // number of subdevices
+            for (int idx = 1; idx < m_devices.size(); ++idx) {
+                if (m_devices.at(idx).m_num_subdevice != m_devices.at(idx-1).m_num_subdevice) {
+                    throw Exception("LevelZero::" + std::string(__func__) +
+                                    ": GEOPM Requires the number of subdevices to be" +
+                                    " the same on all devices. " +
+                                    " Please check ZE_AFFINITY_MASK enviroment variable settings",
+                                    GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                }
+            }
         }
 
         // TODO: When additional device types such as FPGA, MCA, and Integrated GPU are supported by GEOPM
--- a/src/GPUActivityAgent.cpp
+++ b/src/GPUActivityAgent.cpp
@@ -40,8 +40,8 @@ namespace geopm
         , M_NUM_GPU(m_platform_topo.num_domain(
                     GEOPM_DOMAIN_GPU))
         , M_NUM_GPU_CHIP(m_platform_topo.num_domain(
-                    GEOPM_DOMAIN_GPU_CHIP))
-        , M_NUM_CHIP_PER_GPU(M_NUM_GPU_CHIP/M_NUM_GPU)
+                         GEOPM_DOMAIN_GPU_CHIP))
+        , M_NUM_CHIP_PER_GPU(M_NUM_GPU_CHIP / M_NUM_GPU)
         , m_do_write_batch(false)
         , m_do_send_policy(true)
     {
@@ -82,7 +82,7 @@ namespace geopm
         signal_domains.push_back(m_platform_io.signal_domain_type("GPU_UTILIZATION"));
 
         // We'll use the coarsest granularity supported by any of the controls or signals except Energy
-        // i.e. GPU if one control supports GPU and another supports GPU_CHIP
+        // i.e. If one control supports domain GPU and another supports domain GPU_CHIP
         m_agent_domain = std::min(*std::min_element(std::begin(control_domains), std::end(control_domains)),
                                   *std::min_element(std::begin(signal_domains), std::end(signal_domains)));
 
@@ -100,10 +100,9 @@ namespace geopm
 #endif
 
         if (m_agent_domain != GEOPM_DOMAIN_GPU &&
-            m_agent_domain != GEOPM_DOMAIN_GPU_CHIP &&
-            m_agent_domain != GEOPM_DOMAIN_BOARD) {
+            m_agent_domain != GEOPM_DOMAIN_GPU_CHIP) {
             throw Exception("GPUActivityAgent::" + std::string(__func__) +
-                            "(): Required signals and controls do not exist at the BOARD, " +
+                            "(): Required signals and controls do not exist at the " +
                             "GPU or GPU_CHIP domain!", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
 
@@ -271,6 +270,7 @@ namespace geopm
 
         // Per GPU freq
         std::vector<double> gpu_freq_request;
+        std::vector<double> gpu_scoped_core_activity;
 
         // Values after phi has been applied
         m_f_max = in_policy[M_POLICY_GPU_FREQ_MAX];
@@ -322,24 +322,15 @@ namespace geopm
                     f_request = m_f_efficient + m_f_range * gpu_core_activity;
                 }
 
-                // Doing energy reading per GPU, not per CHIP
+                // We're using the activity of the first
+                // GPU_CHIP per GPU as a rough estimate of total GPU activity
+                // for tracking the workload region of interest later on.
+                // This is non-ideal, but is intended to be a temporary
+                // solution to the lack of GPU region support and may be
+                // removed when that support is added to GEOPM.
                 if (domain_idx % (M_NUM_CHIP_PER_GPU) == 0) {
-                    int gpu_idx = domain_idx / M_NUM_CHIP_PER_GPU;
-                    // Tracking logic.  This is not needed for any performance reason,
-                    // but does provide useful metrics for tracking agent behavior
-                    if (gpu_core_activity >= M_GPU_ACTIVITY_CUTOFF) {
-                        m_gpu_active_region_stop.at(gpu_idx) = 0;
-                        if (m_gpu_active_region_start.at(gpu_idx) == 0) {
-                            m_gpu_active_region_start.at(gpu_idx) = m_time.value;
-                            m_gpu_active_energy_start.at(gpu_idx) = m_gpu_energy.at(gpu_idx).value;
-                        }
-                    }
-                    else {
-                        if (m_gpu_active_region_stop.at(gpu_idx) == 0) {
-                            m_gpu_active_region_stop.at(gpu_idx) = m_time.value;
-                            m_gpu_active_energy_stop.at(gpu_idx) = m_gpu_energy.at(gpu_idx).value;
-                        }
-                    }
+                    //int gpu_idx = domain_idx / M_NUM_CHIP_PER_GPU;
+                    gpu_scoped_core_activity.push_back(gpu_core_activity);
                 }
             }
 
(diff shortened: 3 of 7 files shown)


src/GPUActivityAgent.cpp
```diff
@@ -6,11 +6,9 @@
 #include "config.h"
 #include "GPUActivityAgent.hpp"
 
-#include <algorithm>
+//#include <algorithm>
```
lhlawson: remove

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,464 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "config.h"
+#include "GPUActivityAgent.hpp"
+
+//#include <algorithm>
```
lhlawson:
```suggestion
```

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,485 @@
...
+    {
+
+        std::vector<int> control_domains;
+        control_domains.push_back(m_platform_io.control_domain_type("GPU_CORE_FREQUENCY_MIN_CONTROL"));
+        control_domains.push_back(m_platform_io.control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL"));
+
+        std::vector<int> signal_domains;
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_FREQUENCY_STATUS"));
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_ACTIVITY"));
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_UTILIZATION"));
```
lhlawson:
```suggestion
        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_UTILIZATION"));
```

src/GPUActivityAgent.cpp
```diff
@@ -68,36 +68,72 @@ namespace geopm
...
+
+        // We'll use the coarsest granularity supported by any of the controls or signals except Energy
+        // i.e. GPU if one control supports GPU and another supports GPU_CHIP
+        int agent_domain = std::min(*std::min_element(std::begin(control_domains), std::end(control_domains)),
+                                    *std::min_element(std::begin(signal_domains), std::end(signal_domains)));
```
lhlawson: Should this be replaced with a series of platform_topo.isnested() calls?

src/GPUActivityAgent.cpp
```diff
@@ -0,0 +1,485 @@
...
+            throw Exception("GPUActivityAgent::" + std::string(__func__) +
+                            "(): GPU_FREQ_MAX out of range: " +
+                            std::to_string(in_policy[M_POLICY_GPU_FREQ_MAX]) +
+                            ".", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        // Not all gpus provide an 'efficient' frequency signal, and the
+        // value provided by the policy may not be valid.  In this case approximating
+        // f_efficient as midway between F_min and F_max is reasonable.
+        if (std::isnan(in_policy[M_POLICY_GPU_FREQ_EFFICIENT])) {
```
lhlawson, line 166: Check for LEVELZERO::GPU_CORE_FREQUENCY_EFFICIENT (#2662) and use it if available

integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_oneapi.py
```diff
@@ -0,0 +1,24 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_activity.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_dgemm_appconf_oneapi(mach, args)
+    gpu_activity.launch(app_conf=app_conf, args=args,
+                        experiment_cli_args=extra_args)
```
avilcheslopez, line 24: I know there's a broader experiment infrastructure refactoring pending, but I think there's opportunity to at least refactor smaller portions like this one. It's practically the same block of code as the run script below.

src/GPUActivityAgent.cpp
```diff
@@ -37,7 +37,11 @@ namespace geopm
         , M_WAIT_SEC(0.020) // 20ms wait default
         , M_POLICY_PHI_DEFAULT(0.5)
         , M_GPU_ACTIVITY_CUTOFF(0.05)
-        , M_NUM_GPU(m_platform_topo.num_domain(GEOPM_DOMAIN_GPU))
+        , M_NUM_GPU(m_platform_topo.num_domain(
+                    GEOPM_DOMAIN_GPU))
+        , M_NUM_GPU_CHIP(m_platform_topo.num_domain(
+                    GEOPM_DOMAIN_GPU_CHIP))
+        , M_NUM_CHIP_PER_GPU(M_NUM_GPU_CHIP/M_NUM_GPU)
```
avilcheslopez: This is assuming a homogeneous setup where every GPU has the same number of chips. How much can we rely on this assumption?

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,69 @@ namespace geopm
...
+
+        std::vector<int> signal_domains;
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_FREQUENCY_STATUS"));
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_ACTIVITY"));
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_UTILIZATION"));
+
+        // We'll use the coarsest granularity supported by any of the controls or signals except Energy
+        // i.e. GPU if one control supports GPU and another supports GPU_CHIP
+        int agent_domain = std::min(*std::min_element(std::begin(control_domains), std::end(control_domains)),
+                                    *std::min_element(std::begin(signal_domains), std::end(signal_domains)));
```
avilcheslopez: Are domains purposely organized by granularity in descending order (from coarse to fine) in the domain enum? I just want to make sure this is an assumption/invariant we can rely on, because otherwise, this code would be so fragile to any tweaking of that enum.

If not, then perhaps your suggestion of relying on the "nested()" methods in PlatformTopo is safer.

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,69 @@ namespace geopm
...
+#ifdef GEOPM_DEBUG
+        int max_agent_domain = std::max(*std::max_element(std::begin(control_domains), std::end(control_domains)),
+                                    *std::max_element(std::begin(signal_domains), std::end(signal_domains)));
+
+        if (agent_domain != max_agent_domain) {
+            throw Exception("GPUActivityAgent::" + std::string(__func__) +
+                            "(): Required signals and controls do not all exist at the same domain.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+#endif
```
avilcheslopez: Minor suggestion: consider encapsulating this code within its own scope (``{ ... }``). This ``max_agent_domain`` variable is only needed for the purpose of this check which is only included when ``GEOPM_DEBUG`` is defined.

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,69 @@ namespace geopm
...
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_UTILIZATION"));
+
+        // We'll use the coarsest granularity supported by any of the controls or signals except Energy
+        // i.e. GPU if one control supports GPU and another supports GPU_CHIP
+        int agent_domain = std::min(*std::min_element(std::begin(control_domains), std::end(control_domains)),
+                                    *std::min_element(std::begin(signal_domains), std::end(signal_domains)));
+
+#ifdef GEOPM_DEBUG
+        int max_agent_domain = std::max(*std::max_element(std::begin(control_domains), std::end(control_domains)),
+                                    *std::max_element(std::begin(signal_domains), std::end(signal_domains)));
```
avilcheslopez: Nit: alignment

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,69 @@ namespace geopm
...
+
+        if (agent_domain != max_agent_domain) {
```
avilcheslopez: I'm sure I'm missing or misunderstanding something. Why do you expect the max domain to be equal to the min domain? If so, then you shouldn't need to make a decision on the level of coarseness and should enforce the invariant that all signals and controls should exist at the same domain.

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,69 @@ namespace geopm
...
             m_gpu_utilization.push_back({m_platform_io.push_signal("GPU_UTILIZATION",
-                                         GEOPM_DOMAIN_GPU,
+                                         agent_domain,
                                          domain_idx), NAN});
-            m_gpu_energy.push_back({m_platform_io.push_signal("GPU_ENERGY",
-                                    GEOPM_DOMAIN_GPU,
-                                    domain_idx), NAN});
+        }
 
+        for (int domain_idx = 0; domain_idx < m_agent_domain_count; ++domain_idx) {
```
avilcheslopez: It seems like you can merge this ``for`` loop with the one above, since they iterate over the same range and it doesn't seem like the instructions executed here need to happen separately or after the prior ones.

src/GPUActivityAgent.cpp
```diff
@@ -114,8 +161,16 @@ namespace geopm
         // value provided by the policy may not be valid.  In this case approximating
         // f_efficient as midway between F_min and F_max is reasonable.
         if (std::isnan(in_policy[M_POLICY_GPU_FREQ_EFFICIENT])) {
-            in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = (in_policy[M_POLICY_GPU_FREQ_MAX]
-                                                      + gpu_min_freq) / 2;
+            auto all_names = m_platform_io.signal_names();
+            std::string fe_sig_name = "LEVELZERO::GPU_CORE_FREQUENCY_EFFICIENT";
+            if (all_names.count(fe_sig_name) != 0) {
+                in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = m_platform_io.read_signal(fe_sig_name,
+                                                                                   GEOPM_DOMAIN_BOARD, 0);
```
avilcheslopez, line 171: I'm guessing you've already run some experiments and found that taking the average Fe value is sufficient/good enough. I realize you've been managing a single/global Fe value all along, but just now it occurred to me to ask of there's any value in tracking it per GPU or "domain_idx".

src/GPUActivityAgent.cpp
lhlawson: Currently in LevelZero we check that the number of sub-devices is evenly divisible by the number of devices:
https://github.com/geopm/geopm/blob/dev/service/src/LevelZero.cpp#L143

However it doesn't fully address the issue that in several locations (including this agent) we make the assumption that ever device will have the same number of sub-devices.  
This may be worth enforcing in GPUTopo, LevlZeroDevicePool, or LevelZero.cpp

src/GPUActivityAgent.cpp
lhlawson: https://github.com/geopm/geopm/blob/dev/service/src/geopm_topo.h#L13
Yes within their own domain type.
GEOPM_DOMAIN_PACKAGE < GEOPM_DOMAIN_CPU
GEOPM_DOMAIN_GPU < GPU_DOMAIN_GPU_CHIP
The issue is that GEOPM_DOMAIN_CPU < GEOPM_DOMAIN_GPU, which is less sensible. 

A later check currently confirms the domain is either GPU or GPU_CHIP to work around that issue.

src/GPUActivityAgent.cpp
lhlawson: There is some value in tracking Fe per domain_idx, but until recently (#2635 & #2435) there was no convenient method to provide this.
The value of per domain Fe is largely based upon how much variability is exhibited in the Fe values across the domain.

src/GPUActivityAgent.cpp
lhlawson: We're only accepting two domains here - GPU & GPU_CHIP.  
If the maximum domain for all signals and controls is GPU_CHIP, and the minimum of all signals and controls is GPU then at least one signal or control is not exposed at the same domain as the others.  

In debug we want to expose this as either a warning or an error.  

In normal operation we're treating this as acceptable, using the coarser domains, and relying on GEOPM's aggregation capabilities to provide the CHIP scoped domains to us as the GPU scope.

integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_oneapi.py
lhlawson: I would still push this off until the larger refactor as it's also practically the same code as the parres_*_cuda scripts and most other scripts in the experiment directory.

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,73 @@ namespace geopm
...
+            }
+        }
+#endif
+
+        if (m_agent_domain != GEOPM_DOMAIN_GPU &&
+            m_agent_domain != GEOPM_DOMAIN_GPU_CHIP &&
+            m_agent_domain != GEOPM_DOMAIN_BOARD) {
+            throw Exception("GPUActivityAgent::" + std::string(__func__) +
+                            "(): Required signals and controls do not exist at the " +
+                            "GPU or GPU_CHIP domain!", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
avilcheslopez: The error message indicates the signals and controls need to exist at either the GPU or GPU_CHIP domain level. However, the check also includes the board domain. Should one of the two be updated?

src/GPUActivityAgent.cpp
```diff
@@ -114,8 +165,16 @@ namespace geopm
         // value provided by the policy may not be valid.  In this case approximating
         // f_efficient as midway between F_min and F_max is reasonable.
         if (std::isnan(in_policy[M_POLICY_GPU_FREQ_EFFICIENT])) {
-            in_policy[M_POLICY_GPU_FREQ_EFFICIENT] = (in_policy[M_POLICY_GPU_FREQ_MAX]
-                                                      + gpu_min_freq) / 2;
+            auto all_names = m_platform_io.signal_names();
+            std::string fe_sig_name = "LEVELZERO::GPU_CORE_FREQUENCY_EFFICIENT";
```
avilcheslopez: Make this const.

src/GPUActivityAgent.cpp
lhlawson: Added Board

integration/experiment/energy_efficiency/run_gpu_activity_parres_dgemm_oneapi.py
lhlawson: #2674 Should tackle this.

src/GPUActivityAgent.cpp
avilcheslopez: Originally, I meant ``fe_sig_name`` should be ``const``. It seems ``all_names`` could actually be: ``const std::set<std::string> &`` or ``const auto &``.

Also, use const naming convention.

service/src/LevelZero.cpp
```diff
@@ -147,6 +147,18 @@ namespace geopm
                                 " Please check ZE_AFFINITY_MASK enviroment variable settings",
                                 GEOPM_ERROR_INVALID, __FILE__, __LINE__);
             }
+
+            // If we have more than one device confirm all devices have the same
+            // number of subdevices
+            for (int idx = 1; idx < m_devices.size(); ++idx) {
+                if (m_devices.at(idx).m_num_subdevice != m_devices.at(idx-1).m_num_subdevice) {
```
```suggestion
                if (m_devices.at(idx).m_num_subdevice != m_devices.at(idx - 1).m_num_subdevice) {
```

src/GPUActivityAgent.cpp
```diff
@@ -37,7 +37,11 @@ namespace geopm
         , M_WAIT_SEC(0.020) // 20ms wait default
         , M_POLICY_PHI_DEFAULT(0.5)
         , M_GPU_ACTIVITY_CUTOFF(0.05)
-        , M_NUM_GPU(m_platform_topo.num_domain(GEOPM_DOMAIN_GPU))
+        , M_NUM_GPU(m_platform_topo.num_domain(
+                    GEOPM_DOMAIN_GPU))
+        , M_NUM_GPU_CHIP(m_platform_topo.num_domain(
+                    GEOPM_DOMAIN_GPU_CHIP))
```
```suggestion
        , M_NUM_GPU_CHIP(m_platform_topo.num_domain(
                         GEOPM_DOMAIN_GPU_CHIP))
```

src/GPUActivityAgent.cpp
```suggestion
        , M_NUM_CHIP_PER_GPU(M_NUM_GPU_CHIP / M_NUM_GPU)
```

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,73 @@ namespace geopm
...
+        control_domains.push_back(m_platform_io.control_domain_type("GPU_CORE_FREQUENCY_MIN_CONTROL"));
+        control_domains.push_back(m_platform_io.control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL"));
+
+        std::vector<int> signal_domains;
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_FREQUENCY_STATUS"));
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_CORE_ACTIVITY"));
+        signal_domains.push_back(m_platform_io.signal_domain_type("GPU_UTILIZATION"));
+
+        // We'll use the coarsest granularity supported by any of the controls or signals except Energy
+        // i.e. GPU if one control supports GPU and another supports GPU_CHIP
```
```suggestion
        // i.e. If one control supports domain GPU and another supports domain GPU_CHIP
```

bgeltz (changes requested):

src/GPUActivityAgent.cpp
```diff
@@ -263,19 +322,23 @@ namespace geopm
...
-                }
-                else {
-                    if (m_gpu_active_region_stop.at(domain_idx) == 0) {
-                        m_gpu_active_region_stop.at(domain_idx) = m_time.value;
-                        m_gpu_active_energy_stop.at(domain_idx) = m_gpu_energy.at(domain_idx).value;
+                    else {
+                        if (m_gpu_active_region_stop.at(gpu_idx) == 0) {
+                            m_gpu_active_region_stop.at(gpu_idx) = m_time.value;
+                            m_gpu_active_energy_stop.at(gpu_idx) = m_gpu_energy.at(gpu_idx).value;
+                        }
```
bgeltz: Please pull this tracking code out into its own for() loop.  It will help with the maintainability and understandability.

src/GPUActivityAgent.cpp
```diff
@@ -291,25 +354,26 @@ namespace geopm
...
-                                     gpu_freq_request.at(domain_idx);
-
-                m_platform_io.adjust(m_gpu_freq_max_control.at(domain_idx).batch_idx,
-                                     gpu_freq_request.at(domain_idx));
-                m_gpu_freq_max_control.at(domain_idx).last_setting =
-                                     gpu_freq_request.at(domain_idx);
-                ++m_gpu_frequency_requests;
-
-                m_do_write_batch = true;
+        if (!gpu_freq_request.empty()) {
```
```suggestion
```

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,73 @@ namespace geopm
...
+                throw Exception("GPUActivityAgent::" + std::string(__func__) +
+                                "(): Required signals and controls do not all exist at the same domain.",
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+#endif
+
+        if (m_agent_domain != GEOPM_DOMAIN_GPU &&
+            m_agent_domain != GEOPM_DOMAIN_GPU_CHIP &&
+            m_agent_domain != GEOPM_DOMAIN_BOARD) {
```
```suggestion
        if (m_agent_domain != GEOPM_DOMAIN_GPU &&
            m_agent_domain != GEOPM_DOMAIN_GPU_CHIP) {
```

src/GPUActivityAgent.cpp
```diff
@@ -67,26 +71,73 @@ namespace geopm
...
+            throw Exception("GPUActivityAgent::" + std::string(__func__) +
+                            "(): Required signals and controls do not exist at the BOARD, " +
+                            "GPU or GPU_CHIP domain!", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
```suggestion
                            "(): Required signals and controls do not exist at the " +
                            "GPU or GPU_CHIP domain!", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```

src/GPUActivityAgent.cpp
```diff
@@ -362,6 +428,7 @@ namespace geopm
     {
         std::vector<std::pair<std::string, std::string> > result;
 
+        result.push_back({"Agent Domain", std::to_string(m_agent_domain)});
```
bgeltz: You can pretty print this with `PlatformTopo::domain_type_to_name()`

bgeltz (changes requested):

src/GPUActivityAgent.cpp
```diff
@@ -322,24 +322,15 @@ namespace geopm
...
-                            m_gpu_active_energy_start.at(gpu_idx) = m_gpu_energy.at(gpu_idx).value;
-                        }
-                    }
-                    else {
-                        if (m_gpu_active_region_stop.at(gpu_idx) == 0) {
-                            m_gpu_active_region_stop.at(gpu_idx) = m_time.value;
-                            m_gpu_active_energy_stop.at(gpu_idx) = m_gpu_energy.at(gpu_idx).value;
-                        }
-                    }
+                    //int gpu_idx = domain_idx / M_NUM_CHIP_PER_GPU;
```
bgeltz, line 332: Remove this comment

bgeltz (changes requested):

src/GPUActivityAgent.cpp
```diff
@@ -263,9 +322,35 @@ namespace geopm
...
-                // but does provide useful metrics for tracking agent behavior
-                if (gpu_core_activity >= M_GPU_ACTIVITY_CUTOFF) {
+                // We're using the activity of the first
+                // GPU_CHIP per GPU as a rough estimate of total GPU activity
+                // for tracking the workload region of interest later on.
+                // This is non-ideal, but is intended to be a temporary
+                // solution to the lack of GPU region support and may be
+                // removed when that support is added to GEOPM.
+                if (domain_idx % (M_NUM_CHIP_PER_GPU) == 0) {
+                    //int gpu_idx = domain_idx / M_NUM_CHIP_PER_GPU;
```
```suggestion
```

bgeltz (changes requested):

bgeltz (approved):

PR #2304, opened 2022-05-10
Update geopmaccess man page and CLI

### LINKED ISSUE #2198: As a user of geopmaccess I would like a more explict way to delete names from the access lists
I'm always frustrated when I have to echo an empty string to delete an access list.  There should be a command line option that supports this.  Additionally, it would be nice to be able pass as input a list of names to remove from an access list.  

Add an option to geopmaccess to remove all signals or controls from an access list:
```
geopmaccess -D [-g GROUP] [-c]
```

Additionally add an option that removes specific names from an access list:
```
geopmaccess -d [-g GROUP] [-c] < REMOVE_LIST
```
In this can the names in REMOVE_LIST would be removed from an access list, but not be a warning or error if names occur in the REMOVE_LIST which are not in the original access list.

The current implementation requires the user to echo an empty string as the access list to delete it entirely.  There is no way to remove individual names from an access list other than to rewrite the entire list.

**Additional context**

### LINKED ISSUE #2212: As a GEOPM service administrator I would like better support for validating access
It is common to not be able to write to /etc within the same environment where the service is running.  An example would be having /etc be mounted to support multiple runtime environments.  

I'm always frustrated when I cannot modify /etc/geopm-service within the runtime environment, but have no way to validate files in /etc/geopm-service have been configured correctly, or check that a request to change an access list is valid without actually modifying /etc. 

Provide command line options to geopmaccess to support:
    a. Verify that all of the files in /etc/geopm-service are named correctly, and that all of the signals and controls listed in these files are valid.
    b. Add a "dry-run" option to geopmaccess that will not modify files in /etc/geopm-service, but will tell the administrator if the request to modify an access list is valid.

As it is now, an administrator must validate access lists created in a different environment by hand.

**Additional context**
Thanks @bensallen for the suggestion.

cmcantalupo:
- Fixes #2212
- Relates to #2198
- Could have better negative testing for incompatible command line option errors.
- Does not implement the -d option described in #2198 to remove a subset of names from an access list, added --edit mode instead
- Solution for #2212 is not exactly as described in the issue, but hopefully is sufficient.


cmcantalupo:
@bensallen 

This is currently a draft PR since the tests are broken.  It also does not provide exactly the solutions described in the associated issues.  At this time, feedback on the man page here:

https://github.com/cmcantalupo/geopm/blob/geopmaccess-cli/service/docs/source/geopmaccess.1.rst

is appreciated.  I'm hoping to get this out of draft status tomorrow or Wednesday.  Thanks!

dannosliwcd (commented):
I noticed `geopmaccess` is installed even when using `--disable-systemd`. Should it be skipped in that case?

service/docs/source/geopmaccess.1.rst
```diff
@@ -1,28 +1,181 @@
...
+*
+  ``-D``, ``--delete``:
+  Remove an access list for default user or a particular Unix Group.
+*
+  ``-n``, ``--dry-run``:
+  Do error checking on all user input, but do not modify configuration
+  files
+
+*
+  ``-E``, ``--skip-check``:
```
dannosliwcd: Maybe `-q` `--quiet`?

service/docs/source/geopmaccess.1.rst
```diff
@@ -1,28 +1,181 @@
...
+
+Options
+~~~~~~~
+
+*
+  ``-c``, ``--controls``:
+  Command applies to controls not signals.
+
+
+* ``-u``, ``--user``:
```
dannosliwcd: `-u` is missing from the synopsis. Given that there's a `-g GID`/`--group GID` it seems unconventional to have a `-u`/`--user` that doesn't accept UID. Maybe `--default` instead?

service/docs/source/geopmaccess.1.rst
```diff
@@ -1,28 +1,181 @@
...
+Access Management
+~~~~~~~~~~~~~~~~~
+
+Reading access lists may be done by an unprivileged client or by the
+system administrator, but only a process with the Linux capability
+``CAP_SYS_ADMIN`` has permission to write or delete an access list.
+The administrator may execute ``geopmaccess`` to write to an access
+list by providing the ``-w`` / ``--write`` command line option.  The
+``-D`` / ``--delete`` option will remove all signals or controls from
+the configuration.
```
dannosliwcd: The default behavior is different in reads vs writes. In reads, it defaults to the signals of the calling process. In writes/deletes it defaults to the user-default list. Maybe make the difference stand out by ending this paragraph with the reading default behavior, to mirror the end of the writing paragraph.

service/docs/source/geopmaccess.1.rst
dannosliwcd: Nevermind. Reached the description at the end of the man page, and I see this is different from a quiet mode. Maybe `-f`, `--force` fits other CLIs that allow similar validation skips.

dannosliwcd:
Regarding "remove a subset of names from an access list", one option could be to open the list-to-be-modified in an editor (e.g., `-e` `--editor <given editor or default to $EDITOR>`. If the file can't be modified in place (e.g., since it needs to be validated), we could prime a temporary buffer with the current state, then do our work on the written state (like the default behavior on git commit messages).

service/docs/source/geopmaccess.1.rst
```diff
@@ -1,28 +1,181 @@
...
+
+The access list created is derived from standard input by reading one
+name per line.  Standard input lines that begin with the ``#``
+character are ignored, and an empty line or EOF (end of file) will
```
Krebrov001:
```rst
character are ignored, and an empty line or ``EOF`` *(end of file)* will
```

Krebrov001 (dismissed):
Looks good, I like it.

service/docs/source/geopmaccess.1.rst
Krebrov001: Please add this formatting.

service/docs/source/geopmaccess.1.rst
```diff
@@ -1,28 +1,181 @@
...
 
 
 DESCRIPTION
 -----------
 
-Access management for the geopm service.
+The GEOPM service uses the ``/etc/geopm-service`` directory to store
+files that control which signals and controls may be accessed by a
+user through the service.  The purpose of the ``geopmaccess`` command
+line tool is to enable reading and writing of these access list files.
```
dannosliwcd, line 54: One of the described use cases is for config validation. Maybe add a section or paragraph on failure scenarios.

I don't think it is necessary to enumerate every failure case (though I'm not against that if you want it). It should be sufficient to say what the user should expect to happen in pass/fail cases though. E.g., does it say that an error occurred, or what input caused the error? Does it set the process return code based on pass/fail state? How do options (e.g., dry-run or skip-check) impact this behavior?

cmcantalupo:
@dannosliwcd The --disable-systemd configure option seems to be confusing for most people who use it.  Perhaps it should be something like --disable-service-io-group since this is really the only thing this option does.  We require libsystemd.so specifically to call into systemd DBus from C++ to support the ServiceIOGroup, and for no other reason.  

I'll complete the EXIT STATUS section of the man page soon.  Thanks for all the helpful suggestions.

cmcantalupo:
Still lacking unit tests for all these new features added to the command line interface, but a preliminary by hand integration testing shows geopmaccess functioning as expected (including edit mode).

cmcantalupo:
Unit tests have been added for the new features.  This included fixing the implementation to match man page.  Some negative testing could still be added to hit the exceptions at the top of the run command.

service/docs/source/geopmaccess.1.rst
```diff
@@ -170,6 +190,215 @@ Note that having signal or control names in an access list in
...
+Service is not active.
+
+Modification of access lists is typically is done by the root user or
+with the ``sudo`` command.  A process is required to have the Linux
+`capabilities(7) <http://man7.org/linux/man-pages/man7/capabilities.7.html>`_
+``CAP_SYS_ADMIN`` in order to modify any access lists, and a failure
+will occur otherwise.
+
+Some command line options cannot be used together, for example, a
+failures will occur if the user specifies both ``--all`` and
```
dannosliwcd:
```suggestion
failure will occur if the user specifies both ``--all`` and
```

service/geopmdpy/access.py
```diff
@@ -207,13 +250,46 @@ def get_group_controls(self, group):
...
+        if is_control:
+            initial_names = self.get_group_controls(group)
+        else:
+            initial_names = self.get_group_signals(group)
+        edit_path = self.edited_names_path(initial_names)
+        try:
+            with open(edit_path) as fid:
+                names = self.read_names(fid)
+        finally:
+            os.unlink(edit_path)
```
dannosliwcd: Could potentially avoid manual `os.unlink` and `os.close` calls here and in `edited_names_path` if this block is instead wrapped in a `with tempfile.NamedTemporaryFile()` context and the tempfile.name path is passed as an argument to `edited_names_path`.

cmcantalupo: I had some trouble implementing that without having the file open in python while the editor also had it open.  Because the unlink() is bound to the close() call when you use the ``__enter__()``/``__exit__()`` interfaces through ``with`` you always get both calls.  In this case we would like to close() the file in python without unlink() while the editor process modifies the file.  After the file is open() again in python we will unlink() it when we close() the second time.

cmcantalupo: Okay, no need to close the file while it is edited.  Thanks for the suggestion.

dannosliwcd (approved):

PR #2305, opened 2022-05-10
Remove event hint

### LINKED ISSUE #1808: EVENT_HINT should be removed from record.hpp
> estimate 1

I think we are now in the future according to this document.  

https://github.com/geopm/geopm/blob/abd7e6ecdac17d572cd0100dd279cd59fa0a9894/src/record.hpp#L29-L34

Done criteria:
- [ ] EVENT_HINT is removed
- [ ] Remaining enum fields are adjusted to be sequential

cmcantalupo:
- Fixes #1808


avilcheslopez (approved):
Just wanted to make a comment on something completely unrelated to this PR:

As I was trying to understand the event records and record types, I was looking at all the related code, and initially I had a hard time understanding "ValidateRecord". Later, I understood it's really meant to be used to validate a stream of event records. So just wanted to note that simply naming it something like "RecordStreamValidator" would likely help the reader save time when reading/trying to understand the code.

PR #2312, opened 2022-05-11
Move schemas into separate python file

cmcantalupo:
Unfortunately this solution does not enable displaying the json schema in the doc string.  It does avoid having a lot of python logic in the file with the .py.in suffix which is not rendered properly by most code viewers.

### DIFF
--- a/service/Makefile.am
+++ b/service/Makefile.am
@@ -74,8 +74,9 @@ EXTRA_DIST = .gitignore \
              geopmdpy/service.py \
              geopmdpy/session.py \
              geopmdpy/topo.py \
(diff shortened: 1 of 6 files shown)


avilcheslopez (commented):
I see the motivation for this change. However, the ultimate goal we were after to satisfy the task to document all JSON schemas. I like this change, but I would request that we find a way to make the schema available in the GEOPM documentation.

One approach would be to include schemas.py in the rst documentation, similar to how system_files.py is being included, and find a way to render the schema there. Another alternative would be to find another place for the schema in the rst documentation.

avilcheslopez (approved):

PR #2316, opened 2022-05-13
Avoid TOCTOU issues in PlatformTopo::create_cache()

### LINKED ISSUE #2309: Change use of stat() in PlatformTopo to use fstat() to avoid any possible TOCTOU issue
> estimate 1

Relates to #2257 
Relates to #2269 

The implementation for the PlatformTopo use of the cache file has potential for TOCTOU issues due to use of the stat() system call.  Instead these checks should be executed on the open file descriptor that is used to do the read.

Done criteria:

- [ ] fstat() replaces stat() call, and geopm::read_file() is replaced with different implementation using file descriptor doing error checking on return codes rather than relying on ifstream class.

bgeltz:
- Fixes #2309.

### DIFF
--- a/service/src/Helper.cpp
+++ b/service/src/Helper.cpp
@@ -49,6 +49,28 @@ namespace geopm
         return contents;
     }
 
+    std::string read_file(const int fd)
+    {
+        std::string contents;
+        struct stat file_stat;
+
+        int err = fstat(fd, &file_stat);
+        if (err < 0) {
+            throw Exception("Helper::" + std::string(__func__) + "(): ",
+                            errno ? errno : GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        contents.resize(file_stat.st_size);
+
+        ssize_t num_read = read(fd, (char *)(contents.data()), file_stat.st_size);
+        if (num_read == -1) {
+            throw Exception("Helper::" + std::string(__func__) + "(): ",
+                            errno ? errno : GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        contents[num_read] = '0';
+
+        return contents;
+    }
+
     double read_double_from_file(const std::string &path, const std::string &expected_units)
     {
         const std::string separators(" \t\n\0", 4);
--- a/service/src/PlatformTopo.cpp
+++ b/service/src/PlatformTopo.cpp
@@ -18,6 +18,7 @@
 #include <errno.h>
 #include <limits.h>
 #include <stdio.h>
+#include <fcntl.h>
 
 #include <map>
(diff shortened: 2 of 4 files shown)


service/src/PlatformTopo.cpp
```diff
@@ -463,28 +464,56 @@ namespace geopm
     void PlatformTopoImp::create_cache(void)
     {
         try {
-            PlatformTopoImp::create_cache(M_SERVICE_CACHE_FILE_NAME);
+            // TODO: Handle return string?
+            (void) PlatformTopoImp::create_cache(M_SERVICE_CACHE_FILE_NAME);
```
bgeltz, line 468: I wasn't sure if I should plumb this string all the way through to the CFFI wrappers, so for the moment I'm ignoring it.

If this is all OK, I could do all that plumbing which would make it so we don't have to open/read any files around this in the Python code.  The CFFI wrapper would just return the file's contents.

service/src/PlatformTopo.cpp
```diff
@@ -463,28 +464,56 @@ namespace geopm
...
+                }
+                else {
+                    expected_perms = S_IRUSR | S_IWUSR; // 0o600
+                }
+                is_file_ok = check_file(cache_fd, expected_perms);
+            }
+            catch (const geopm::Exception &ex) {
+                (void) close(cache_fd);
+                // sysinfo or stat failed; file does not exist (2) or permission denied (13)
+                throw;
```
bgeltz, line 507: Since it's assured that the file exists now (since it was opened above) all errors in check_file() should throw all the way up.  If the file does not exist, this else block is not evaluated.

service/src/Helper.cpp
```diff
@@ -49,6 +49,28 @@ namespace geopm
...
+    {
+        std::string contents;
+        struct stat file_stat;
+
+        int err = fstat(fd, &file_stat);
+        if (err < 0) {
+            throw Exception("Helper::" + std::string(__func__) + "(): ",
+                            errno ? errno : GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        contents.resize(file_stat.st_size);
```
cmcantalupo, line 62:
```suggestion
        contents.resize(file_stat.st_size + 1);
```

cmcantalupo, line 62: This should put the null char at end of array.  Seems like an error to index into contents[contents.size()] on line 69.

cmcantalupo:
Lets check with David to see if this is really necessary.    These changes pretty difficult to maintain.  Hope I didn't send you on a wild goose chase.

bgeltz:
Abandoning in favor of #2326.

PR #2313, opened 2022-05-11
Increase padding for domain names

bgeltz:
- This accommodates the recent changes to these
  strings such that they can be parsed properly.

Fixes #2311.
Broken after #2302.

PR #2324, opened 2022-05-17
Add a Third Party Programs license file to the distribution

cmcantalupo:
- Adds a list of all third party licenses
- This information is also contained in:
   https://github.com/geopm/geopm/tree/dev/copying_headers
- There is also a test that verifies copyright attribution in every file:
   https://github.com/geopm/geopm/blob/dev/copying_headers/test-license
- This test verifies the data held in copying_headers
    + Every file in the git repository must be enumerated in a copying_headers/MANIFEST.* with corresponding header
    + If the file is not enumerated in one of these manifests it is original source and must have the Intel BSD-3 license header
    + This test is run against every PR as part of our unit tests in github actions
        * e.g. https://github.com/geopm/geopm/pull/2313/checks#step:13:21


avilcheslopez (approved):
Looks good to me! I know the board hasn't replied to our latest email, but I interpret their lack of response to mean they don't have any serious concerns with our approach or understanding of the purpose of the TPP file. If we don 't hear from them, I believe this is good to go.

Issue #1676, opened 2021-06-16
As a developer of the GEOPM service I would like to ensure that all user input is properly vetted


cmcantalupo:
It is especially important to test that the service is robust in all user facing interfaces to bad or malicious input. 

We know that better error checking is required of the GEOPM_PROFILE environment variable.  This task is to look for any other potential related issues and fix them.

cmcantalupo:
Requires review of existing tests and determine if there are existing gaps.  Many if not all of our user facing interfaces have robust unit tests.

dannosliwcd:
One known case is geopm profile names. Users can input `--geopm-profile=[0, 1, 2, anything that parses as a complex yaml data structure]` and it will bubble through to our report files which will be interpreted as yaml. In that specific case, a solution could be to write the profile (or any field) to the report in a way that ensures it is interpreted as a string, e.g., by using a full yaml-aware dumper.

PR #2328, opened 2022-05-19
Sanitize GEOPM_PROFILE

bgeltz:
- Relates to #1676.

### DIFF
--- a/service/docs/source/geopmlaunch.1.rst
+++ b/service/docs/source/geopmlaunch.1.rst
@@ -64,33 +64,33 @@ command line parameter.  Available *launcher* values are
 listed below.
 
 
-* 
+*
   *srun*\ , *SrunLauncher*\ :
   Wrapper for the SLURM resource manager's ``srun`` job launcher.  The
   ``--cpu_bind`` and ``--cpu-bind`` options are reserved for use by GEOPM;
   do not specify when using ``geopmlaunch``.
 
-* 
+*
   *aprun*\ , *AlpsLauncher*\ :
   Wrapper for the Cray ALPS ``aprun`` job launcher.  The ``-cc`` and
   ``--cpu-binding`` options are reserserved for use by GEOPM; do not
   specify these when using ``geopmlaunch``.
 
-* 
+*
   *impi*\ , *IMPIExecLauncher*\ :
   Wrapper for the Intel MPI ``mpiexec`` job launcher.  The
   ``KMP_AFFINITY`` ``I_MPI_PIN_DOMAIN``\ , and ``MV2_ENABLE_AFFINITY``
   environment variables reserved for use by GEOPM and are overwritten
   when using ``geopmlaunch``.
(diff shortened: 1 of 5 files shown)


cmcantalupo:
Maybe do it in the Reporter where the requirement exists?

https://github.com/cmcantalupo/geopm/commit/06c05b635b6a54ec31c5c3b65066a758fc7d9b66

the profile name here **may not come from the environment** since the reporter API is exposed in python now.

dannosliwcd:
Use-case-wise, it would be nice if spaces were allowed. When I'm exploring a lot of ad-hoc workload configs, it is nice if I can dump the wrapped command inside the profile string. I can work around that by replacing spaces with a filler character at launch time and replacing back if I need to reproduce a run later on, so I'm just mentioning this as a nice-to-have.

bgeltz:
> Maybe do it in the Reporter where the requirement exists?
> 
> [cmcantalupo@06c05b6](https://github.com/cmcantalupo/geopm/commit/06c05b635b6a54ec31c5c3b65066a758fc7d9b66)
> 
> the profile name here **may not come from the environment** since the reporter API is exposed in python now.

As we discussed in standup, the Python API calls the other version of generate() which does not take a `TreeComm`.  I'm going back to the solution of warning in the Environment class as the `Tracer` also gets the profile string from this path through the `CSVImp`.

bgeltz:
The current implementation does print the warning about erasing bad characters once for every rank in the app.  This is an existing problem for warnings that are emitted out of the EnvironmentImp (i.e. from `parse_environment_file()`).

service/docs/source/geopmlaunch.1.rst
```diff
@@ -228,7 +228,9 @@ GEOPM OPTIONS
   default profile name is the name of the compute application
   executable.  This option is used by the launcher to set the
   GEOPM_PROFILE environment variable.  The command line option will
-  override any value currently set in the environment.  See the
+  override any value currently set in the environment.  Note that
+  the name may not contain carriage returns nor double qoutes, but
```
cmcantalupo:
```suggestion
  the name may not contain carriage returns nor double quotes, but
```

src/Environment.cpp
```diff
@@ -260,6 +260,21 @@ namespace geopm
         if (do_profile() && ret.empty()) {
             ret = std::string(program_invocation_name);
         }
+        else if (!ret.empty()) {
+            // Sanitize the input: No carriage returns nor double quotes
+            ret.erase(std::remove_if(ret.begin(), ret.end(),
```
cmcantalupo, line 266: Maybe we could limit the warning messages to once per profile() call rather than once per bad character?  Maybe cache the start value, filter out bad values, and then print a warning if the string changed?  If we updated m_name_value_map we would be able to limit the warning message to be printed once per process...

cmcantalupo (approved):

PR #2332, opened 2022-05-23
Rename SST weight to priority and scale from 0 to 1

### LINKED ISSUE #1658: SST::COREPRIORITY:n:WEIGHT should scale from 0 to 1
Re-define SST::COREPRIORITY:n:WEIGHT controls so that they has the least impact at 0 and the most impact at 1.

dannosliwcd:
- Resolves #1658


cmcantalupo (approved):
LGTM

PR #2334, opened 2022-05-23
Unlink any temporary files left around from the Batch Server

### LINKED ISSUE #2272: The ActiveSessions class should clean up any files created for batch server at session close
> estimate 3

- Related to #2248

When a batch server is ended, if the shared memory files or FIFO files are still around, the service should clean them up as part of ActiveSessions.remove_batch_server() implementation.  If these file are removed at this point, that is the result of an error.  This should be logged.

Look into solutions that enable logging of failures to unlink files by the batch server. 

Done criteria:

- [ ] Verify feature in unit test
- [ ] Documentation changes are not required since this change does

Krebrov001:
This includes shared memory files, fifo files, and the lock file.

- Relates to #2272
- Fixes #2272

### DIFF
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -18,6 +18,7 @@ abstraction for updating the contents.
 
 """
 
+import re
 import os
 import sys
 import stat
@@ -332,6 +333,8 @@ class ActiveSessions(object):
 
         """
         self._RUN_PATH = run_path
+        self._M_SHMEM_PREFIX = os.path.join(self._RUN_PATH, "geopm-service-batch-buffer-")
+        self._M_DEFAULT_FIFO_PREFIX = os.path.join(self._RUN_PATH, "batch-status-")
         self._daemon_uid = os.getuid()
         self._daemon_gid = os.getgid()
         self._sessions = dict()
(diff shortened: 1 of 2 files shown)


service/geopmdpy/system_files.py
```diff
@@ -682,7 +685,21 @@ def remove_batch_server(self, client_pid):
         self.check_client_active(client_pid, 'remove_batch_server')
         self._sessions[client_pid].pop('batch_server')
         self._update_session_file(client_pid)
-        # TODO: Unlink any shared memory files left around
+
```
cmcantalupo, line 690: Please add some tests.  This code should not do any pattern matching.  The file paths should be deterministic based on the batch server PID.  As it it written now this could delete files that are unrelated to the particular batch server that ended.

service/geopmdpy/system_files.py
```diff
@@ -679,10 +682,42 @@ def remove_batch_server(self, client_pid):
...
+        batch_pid = self._sessions[client_pid]['batch_server']
+
         self.check_client_active(client_pid, 'remove_batch_server')
         self._sessions[client_pid].pop('batch_server')
         self._update_session_file(client_pid)
-        # TODO: Unlink any shared memory files left around
+
+        # Unlink any temporary files left around from the Batch Server
+        # This includes shared memory files, fifo files, and the lock file.
+        if (os.path.exists(self._LOCK_PATH)):
```
cmcantalupo: Cleaning the lock file is done when the session ends (line 687).  That code checks if the session that obtained the lock file is still alive, and it is not it is removed.  The code here will remove the lock file regardless of which session owns it.  This may remove a lock created by a process other than the batch server.  The removal of this file must depend on the PID that has ended.

service/geopmdpy_test/TestActiveSessions.py
```diff
@@ -437,6 +437,77 @@ def test_batch_server_bad_service_restart(self):
...
+        client_pid = json_good_example['client_pid']
+        signals = json_good_example['signals']
+        controls = json_good_example['controls']
+        watch_id = json_good_example['watch_id']
+        batch_pid = 8765
+
+        sess_path = f'{self._TEMP_DIR.name}/geopm-service'
+        full_file_path = os.path.join(sess_path, f"session-{client_pid}.json")
+
+        lock_path = os.path.join(sess_path, "CONTROL_LOCK")
```
cmcantalupo: remove the test around the write lock.

cmcantalupo (commented):
Looks good.  We don't want to unlink the lock file if it exists any time a batch server ends, so please remove this feature, otherwise that should solve the issue.

cmcantalupo (approved):

PR #2353, opened 2022-05-25
Update geopm service admin guide and readme

cmcantalupo:


service/docs/source/admin.rst
```diff
@@ -24,11 +24,29 @@ command can be used to interact with the ``geopm`` Systemd Unit.
...
+All files and directories within ``/etc/geopm-service`` or
+``/run/geopm-service`` are created by the GEOPM Service with
+restricted access permissions and root ownership.  The GEOPM Service
+will not read any file or directory if they are modified to have more
+permissive access restrictions, non-root ownership, or if they are
+replaced by a symbolic link or other non-regular file.  If these
+checks fail, the file or directory will be renamed to include a UUID
+string and a warning is printed in the syslog.  These renamed files or
+directories enable an administrator to perform an investigation into
+problem, but they will not be used by the GEOPM Service in any way.
```
bgeltz, line 47: Might be worth adding a reminder here that files in /run will be removed upon reboot.

service/docs/source/admin.rst
```diff
@@ -24,11 +24,29 @@ command can be used to interact with the ``geopm`` Systemd Unit.
...
+
+It is recommended that these GEOPM Service system files are always
```
bgeltz, line 49:
```suggestion
It is recommended that these GEOPM service system files are always
```

Though, it seems it should always be capitalized?  This section presently has both.

service/README.rst
```diff
@@ -104,12 +104,11 @@ access to any subset of the privileged PlatformIO features.
...
 ------
 
-The GEOPM systemd service is a new feature.  What exists currently is
-a work in progress, and there are several outstanding issues that must
-be resolved before the GEOPM Service is ready for release with GEOPM
-version 2.0.  These
+The GEOPM systemd service is a new feature for version 2.0.  The
+``v2.0.0-rc1`` tag is the first release candidate and for 2.0 and
+provides the GEOPM Service feature.  Tagging ``v2.0.0`` will not occur
+until all issues with the "2.0" label are closed or relabeled:
```
avilcheslopez: Minor suggestion:

For some reason, the wording makes it feel like this is providing process-level details. If the intention is to give an idea of when v2.0.0 will be tagged and/or the features/changes that will be included in v2.0.0, maybe it could be worded something like:
- "The v2.0.0 tag/release will include the following changes: <link to all 2.0 issues>. For a list of remaining issues/changes, please see: <link to remaining 2.0 issues>"
- "Work for v2.0.0 is currently ongoing and will be tagged once all remaining issues/changes are completed/addressed: <link to remaining 2.0 issues>"

service/README.rst
```diff
@@ -190,25 +189,6 @@ restored to the value they had when the session was converted,
...
-
-Note that if any control adjustments are made during a session through
-the GEOPM Service then every control supported by GEOPM will be
-reverted when the session ends.  One consequence of this is that when
-a control is exposed to a user only through the GEOPM Service, then
-the geopmwrite command line tool will not be effective (the value will
-be written, but reverted when the geopmwrite process ends).  The
-geopmsession command line tool can be used to write any number of the
-GEOPM supported controls and keep a session open for a specified
-duration (or until the geopmsession process is killed).
```
avilcheslopez, line 210: Just to understand: is the reason for removing these paragraphs that this level of detail perhaps belongs in the developer's doc/guide?

service/docs/source/admin.rst
avilcheslopez, line 47: Minor suggestion: perhaps this might be a good opportunity to reference the service security guide for further details about security measures/checks we have in place.

service/README.rst
cmcantalupo: The first paragraph was deleted because it is really an implementation detail of the ServiceIOGroup, and way too specific for the service top level readme.  The second paragraph was deleted because it is no longer true.  The write lock is tied to the session leader process, which is not the geopmwrite process in almost all use cases.  The geopmsession tool also no longer supports write mode exactly for this reason.

service/README.rst
```diff
@@ -104,12 +104,11 @@ access to any subset of the privileged PlatformIO features.
 Status
 ------
 
-The GEOPM systemd service is a new feature.  What exists currently is
-a work in progress, and there are several outstanding issues that must
-be resolved before the GEOPM Service is ready for release with GEOPM
-version 2.0.  These
+The GEOPM systemd service is a new feature for version 2.0.  The
+``v2.0.0-rc1`` tag is the first release candidate and for 2.0 and
```
bgeltz:
```suggestion
``v2.0.0-rc1`` tag is the first release candidate for 2.0 which
```

bgeltz (dismissed):

bgeltz (approved):

PR #2362, opened 2022-06-01
[WIP] Add a C api to support sampling an array of requests based on a configuration file

cmcantalupo:


cmcantalupo:
Don't think we want to add this interface for now.  Maybe in the future.

PR #1977, opened 2021-10-27
Support for MSR::CPU_SCALABILITY_RATIO (a PPERF based derivative signal)

lhlawson:
Signed-off-by: Lowren Lawson <lowren.h.lawson@intel.com>

- Relates to #2295 feature request from github issues
- Fixes #2295 change request from github issues.

This PR adds a new DivisionSignal class required to build the MSR::CPU_SCALABILITY_RATIO signal, and registers the MSR::CPU_SCALABILITY_RATIO signal in the MSRIOGroup based on derivative signals of PCNT and ACNT.  Changes are made to the MSRIOGroup as required, along with a change to the format_function calls

This PR is a rework of #1578 to work with the new service branch and reduce PR size.  As such all related comments previously addressed in #1578 are resolved.  The Scalability Hint regions that were part of #1578 have been removed (to be added in a later PR).

Tasks to resolve:
- [x] Port over testing from existing PRs & Repos
- [x] Confirmation that all signals provide a format function/don't result in bad function call error - every signal available on mcfly read using geopmread

Issue #1065, opened 2020-04-23
Add control loop period as parameter to all agents that wait() on time


cmcantalupo:
Could also be specified as a rate.

dianarg:
Might be duplicate #576

cmcantalupo:
We need to get rid of the `void Agent::wait(void)` method and replace it with an `double Agent::wait_time(void)` method that specifies the default control interval time target.

cmcantalupo:
Be nice to also have a command line override for the wait_time().  In the mean time a simple solution 

```
void MonitorAgent::wait(void)
{
    double timeout = m_policy[M_POLICY_WAIT_TIME]; // Could come from Environment class in the future
    Agent::wait(timeout);
}
```

With Agent::wait() as a static method of the base class.

bgeltz:
Related to #1411

PR #2364, opened 2022-06-01
Addition of the CPU Activity Agent to steer CPU frequency based upon CPU Compute Activity 

### LINKED ISSUE #2363: As a user of CPU Hardware I would like an agent that dynamically modulates CPU frequency based on CPU Compute Activity
An agent that uses system characterization and HW metrics of CPU activity in a similar fashion to #1906 to drive energy efficient frequency decisions could provide better overall energy savings than agents that do not use CPU compute and memory activity metrics.

An agent that uses HW telemetry gathered from the CPU to  reduce energy consumption by automatically selecting a lower frequency when doing so would have minimal impact on the performance of the CPU.  

Region based agents that learn over time based on software metrics.

**Additional context**
This feature is an extension of the approach used in #1906, #1908, and #2045.

### LINKED ISSUE #2639: CPU Activity Agent fixes (address review comments)
> estimate 1
Relates to: #2363 
Fixes: #2363 

Complete review of CPU-CA agent, address review comments.

lhlawson:
- Relates to #2363 feature request from github issues
- Fixes #2363 request from github issues.
- Requires #1977
- Resolves #2639

This PR introduces and agent that modulates the CPU frequency within a set of user specified bounds based upon the MSR::SCALABILITY_RATIO signal AND modulate the CPU Uncore frequency within a set of user specified bounds based upon the Memory Bandwidth Monitor feature.  

The user may also provide an optional input 'phi' which can bias the agent frequency selection towards energy efficiency (phi of 1.0) or performance (phi of 0.0).  The default behavior is phi = 0.5 for the full range of the agent.  

Additional Tasks for this PR: 
- [x] Integration tests based upon the run_cpu_activity_*.py files
- [x] Policy generation script to find the Fe (frequency efficient) policy input based on AIB (arithmetic intensity benchmark) frequency sweeps
- [x] Update documentation of simplified policy generation script
- [x] Policy generation script to characterize relevant parts of the platform based upon AIB frequency sweeps
- [x] Updates to the cpu_activity.py experiment script to allow user specified Fe and Fmax values for phi sweep (current default is frequency min to frequency max range)
- [x] Updates to the cpu_activity.py experiment script to allow user specified system characterization for phi sweep (current default is no characterization, which disables uncore frequency control) 
- [x] Update to GEOPM 2.0 RC alias changes
- [x] Phi frequency response analysis
- [x] Agent documentation


src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+            double qm_normalized = (double) m_qm_rate.at(domain_idx).signal /
+                                            qm_max_itr->second;
+
+            // This is intended to handle dived by zero, either numerator or
+            // denominator being NAN, and the uncharacterized case
+            if (std::isnan(qm_normalized) ||
+                m_qm_max_rate.size() == 0) {
+                qm_normalized = 1.0;
+            }
```
lhlawson:
```suggestion
            double qm_normalized = 1
            
            // Handle divided by zero, either numerator or
            // denominator being NAN, and the un-characterized case
            if (!std::isnan(m_qm_rate.at(domain_idx).signal) &&
                !std::isnan(qm_max_itr->second) && 
                qm_max_itr->second != 0 &&         
                m_qm_max_rate.size() != 0) {
                qm_normalized =(double) m_qm_rate.at(domain_idx).signal /
                                            qm_max_itr->second;
            }
```

test/CPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,665 @@
...
+//                .WillRepeatedly(Return(123456789));
+//    m_agent->sample_platform(tmp);
+//
+//    //Adjust
+//    //Check frequency
+//    double expected_freq = policy[FREQ_EFFICIENT] + (m_freq_max - policy[FREQ_EFFICIENT]) * mock_active;
+//    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQUENCY_CONTROL_IDX, expected_freq)).Times(M_NUM_GPU);
+//    m_agent->adjust_platform(policy);
+//    //Check a frequency decision resulted in write batch being true
+//    EXPECT_TRUE(m_agent->do_write_batch());
```
lhlawson:
```suggestion

```

src/CPUActivityAgent.hpp
```diff
@@ -0,0 +1,116 @@
...
+            const int M_NUM_PACKAGE;
+            const int M_NUM_CORE;
+            bool m_do_write_batch;
+            bool m_update_qm_max_rate;
+
+            struct signal
+            {
+                int batch_idx;
+                double signal;
+                double sample;
```
lhlawson:
```suggestion
                int batch_idx;
                double value;
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+        for (int domain_idx = 0; domain_idx < M_NUM_PACKAGE; ++domain_idx) {
+            m_uncore_freq_min_control.push_back({m_platform_io.push_control("CPU_UNCORE_FREQUENCY_MIN_CONTROL",
+                                              GEOPM_DOMAIN_PACKAGE,
+                                              domain_idx), -1});
+            m_uncore_freq_max_control.push_back({m_platform_io.push_control("CPU_UNCORE_FREQUENCY_MAX_CONTROL",
+                                              GEOPM_DOMAIN_PACKAGE,
+                                              domain_idx), -1});
+        }
+
+        //Configuration of QM_CTR must match QM_CTR config used for training data
```
bgeltz:
```suggestion
        // Configuration of QM_CTR must match QM_CTR config used for training data
```
Also this comment should explain what is being setup here.  For instance, what is EVENT_ID 2 and why do we care?

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+                           "CPUActivityAgent::" + std::string(__func__) +
+                           "(): policy vector not correctly sized.  Expected  " +
+                            std::to_string(M_NUM_POLICY) + ", actual: " +
+                            std::to_string(in_policy.size()));
+
+        ///////////////////////
+        //CPU POLICY CHECKING//
+        ///////////////////////
+        double freq_core_min = m_platform_io.read_signal("CPU_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+        double freq_core_max = m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
```
bgeltz: ```validate_policy()``` gets called every time the Controller walks down.  Having these individual ```read_signal()``` calls in here will slow this down.  Do you need to read them every time through, or just once?

If it's just once, I'd put these values in a member variable, and cache them in ``init()`` like you do for the controls.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+        //UNCORE POLICY CHECKING//
+        //////////////////////////
+        if (std::isnan(in_policy[M_POLICY_UNCORE_FREQ_MAX])) {
+            in_policy[M_POLICY_UNCORE_FREQ_MAX] = m_freq_uncore_max;
+        }
+        if (std::isnan(in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT])) {
+            in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT] = m_freq_uncore_min;
+        }
+
+        if (in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT] > in_policy[M_POLICY_UNCORE_FREQ_MAX]) {
```
bgeltz, line 168:
```suggestion
        // When the policy is all NaNs, this check also verifies that
        // the system was not left in a bad state with regard to the
        // UNCORE_FREQUENCY_<MAX/MIN>_CONTROLs.
        if (in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT] > in_policy[M_POLICY_UNCORE_FREQ_MAX]) {
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+        // Validate all (uncore frequency, max memory bandwidth) pairs
+        std::set<double> policy_uncore_freqs;
+        for (auto it = in_policy.begin() + M_POLICY_FIRST_UNCORE_FREQ;
+             it != in_policy.end() && std::next(it) != in_policy.end(); std::advance(it, 2)) {
+            auto mapped_mem_bw = *(it + 1);
+            if (!std::isnan(*it)) {
+                // We are using a static cast rather than reinterpreting the
+                // memory so that regions can be input to this policy in the
+                // same form they are output from a report.
```
bgeltz: An example policy somewhere in the documentation would help explain this comment.  A one line example for the uncore freq policy field right here would also help.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+    bool CPUActivityAgent::do_send_sample(void) const
+    {
+        return false;
+    }
+
+    void CPUActivityAgent::adjust_platform(const std::vector<double>& in_policy)
+    {
+        m_do_write_batch = false;
+
+        if (m_update_qm_max_rate) {
```
bgeltz: The body of this if() is only evaluated on the first call to ```adjust_platform()```.  ```m_update_qm_max_rate``` is set to true in the initializer list, and false at the end of this if().  Was this intentional?

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+            m_update_qm_max_rate = false;
+        }
+
+        // Per package freq
+        std::vector<double> uncore_freq_request;
+        m_resolved_f_uncore_efficient = in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT];
+        m_resolved_f_uncore_max = in_policy[M_POLICY_UNCORE_FREQ_MAX];
+        double f_uncore_range = in_policy[M_POLICY_UNCORE_FREQ_MAX] - in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT];
+
+        //m_resolved_f_uncore_efficient = 0;
```
bgeltz:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+        for (int domain_idx = 0; domain_idx < M_NUM_PACKAGE; ++domain_idx) {
+            double uncore_freq = (double) m_uncore_freq_status.at(domain_idx).value;
+
+            /////////////////////////////////////////////
+            // L3 Total External Bandwidth Measurement //
+            /////////////////////////////////////////////
+            auto qm_max_itr = m_qm_max_rate.lower_bound(uncore_freq);
+            if(qm_max_itr != m_qm_max_rate.begin()) {
+                qm_max_itr = std::prev(qm_max_itr, 1);
+            }
```
bgeltz: Please comment on what's going on here.

You're assign ```qm_max_itr``` an iterator that 1 *before* the lower bound of the current uncore frequency?  Why?

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+            // L3 usage, Network Traffic, HBM, and PCIE (GPUs) all use the uncore.
+            // Eventually all these components should be considered when scaling
+            // the uncore frequency in the efficient - performant range.
+            // A more robust/future proof solution may be to directly query uncore
+            // counters that indicate utilization (when/if available).
+            // For now only L3 bandwith metric is used.
+            double uncore_req = m_resolved_f_uncore_efficient + f_uncore_range * scalability_uncore;
+
+            //Clip uncore request within policy limits
```
bgeltz: Debug warn when this is happening?

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,532 @@
...
+            //////////////////////////////////
+            // Core Scalability Measurement //
+            //////////////////////////////////
+            double scalability = (double) m_core_scal.at(domain_idx).value;
+            if (std::isnan(scalability)) {
+                scalability = 1.0;
+            }
+
+            double core_req = m_resolved_f_core_efficient + f_core_range * scalability;
+            //Clip core request within policy limits
```
bgeltz: Debug warn?

integration/test/test_cpu_activity_agent.py
```diff
@@ -0,0 +1,321 @@
...
+@util.skip_unless_workload_exists("apps/arithmetic_intensity/ARITHMETIC_INTENSITY/bench_avx2")
+@util.skip_unless_workload_exists("apps/arithmetic_intensity/ARITHMETIC_INTENSITY/bench_avx512")
+@util.skip_unless_workload_exists("apps/minife/miniFE_openmp-2.0-rc3/src/miniFE.x")
+class TestIntegration_cpu_activity(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """
+        Setup applications, execute, and set up class variables.
+        """
+
```
bgeltz, line 44: Add a comment explaining the setup here.

integration/test/test_cpu_activity_agent.py
```diff
@@ -0,0 +1,321 @@
...
+from apps.minife import minife
+
+@util.skip_unless_do_launch()
+@util.skip_unless_workload_exists("apps/arithmetic_intensity/ARITHMETIC_INTENSITY/bench_sse")
+@util.skip_unless_workload_exists("apps/arithmetic_intensity/ARITHMETIC_INTENSITY/bench_avx2")
+@util.skip_unless_workload_exists("apps/arithmetic_intensity/ARITHMETIC_INTENSITY/bench_avx512")
+@util.skip_unless_workload_exists("apps/minife/miniFE_openmp-2.0-rc3/src/miniFE.x")
+class TestIntegration_cpu_activity(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
```
bgeltz, line 40: This needs a refactor.

bgeltz (changes requested):

src/CPUActivityAgent.cpp
lhlawson: This was intentional.  The system characterization used by the agent in this version is based upon a single system wide characterization provided via policy and could be handled in the validate_policy section, but future updates in PR #2437 allow it to use per node signals (added in PR #2435) which necessitated moving the logic into adjust_platform to read the per node signals.

src/CPUActivityAgent.cpp
lhlawson: This is intended to handle cases where the signal does not cleanly map to entries in the qm_max_rate keys (or vice versa).  
For a map of {1.9: a, 2.0: b, 2.2: c, 2.3: d} lower_bound(2.2) & lower_bound(2.1) will both return an iterator to (2.2: c).  Similarly lower_bound(2.21) will return an iterator to (2.3: d).  
As a simplification to address this issue while targeting energy efficiency we map to one iterator below the lower bound.

Alternative solutions or updates that may be implemented
- Only use std::prev if the uncore_freq doesn't match a key in the map
- Round (or floor) uncore_freq to a key in the m_qm_max_rate map before doing lower_bound
- Require map entries be valid values for the hardware specified by checking them vs the step size (not currently possible with existing GEOPM signals)

src/CPUActivityAgent.cpp
lhlawson: Moved to a report header following the discussion on the (similar) GPU compute activity agent clipping tracking [here](https://github.com/geopm/geopm/pull/2045#discussion_r871859791) as part of #2045

src/CPUActivityAgent.cpp
lhlawson: Moved to a report header following the discussion on the (similar) GPU compute activity agent clipping tracking [here](https://github.com/geopm/geopm/pull/2045#discussion_r871859791) as part of #2045

src/CPUActivityAgent.cpp
lhlawson: "Only use std::prev if the uncore_freq doesn't match a key in the map" has been implemented in the latest set of changes.  Performance testing against 4 workloads of interest has shown no harm or better energy savings & perf deg for each workload.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+        m_resolved_f_core_efficient = 0;
+        m_resolved_f_core_max = 0;
+
+        // These are not currently guaranteed to be the system uncore min and max,
+        // just what the user/admin has previously set.
+        m_freq_uncore_min = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+        m_freq_uncore_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+        m_freq_core_min = m_platform_io.read_signal("CPU_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+        m_freq_core_max = m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+
```
bgeltz, line 65:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+            m_qm_rate.push_back({m_platform_io.push_signal("MSR::QM_CTR_SCALED_RATE",
+                                                           GEOPM_DOMAIN_PACKAGE,
+                                                           domain_idx), NAN});
+
+            m_uncore_freq_status.push_back({m_platform_io.push_signal("CPU_UNCORE_FREQUENCY_STATUS",
+                                                                      GEOPM_DOMAIN_PACKAGE,
+                                                                      domain_idx), NAN});
+        }
+
+        for (int domain_idx = 0; domain_idx < M_NUM_PACKAGE; ++domain_idx) {
```
bgeltz:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+        // Validate all (uncore frequency, max memory bandwidth) pairs
+        std::set<double> policy_uncore_freqs;
+        for (auto it = in_policy.begin() + M_POLICY_FIRST_UNCORE_FREQ;
+             it != in_policy.end() && std::next(it) != in_policy.end(); std::advance(it, 2)) {
+            auto mapped_mem_bw = *(it + 1);
+            if (!std::isnan(*it)) {
+                // We are using a static cast rather than reinterpreting the
+                // memory so that regions can be input to this policy in the
+                // same form they are output from a report.
```
bgeltz:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+                // Example policy values parsed here:
+                //
+                // UNCORE_FREQ_0": 1800000000.0,
+                // "MAX_MEMORY_BANDWIDTH_0": 108066060000.0,
+                // "CPU_UNCORE_FREQ_1": 1900000000.0,
+                // "MAX_MEMORY_BANDWIDTH_1": 116333135000.0,
+                // ...
+                // CPU_UNCORE_FREQ_<#>": 2400000000.0,
+                // "MAX_MEMORY_BANDWIDTH_<#>": 106613110000.0
+                auto uncore_freq = static_cast<uint64_t>(*it);
```
bgeltz:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+        in_policy[M_POLICY_CPU_FREQ_MAX] = f_core_max;
+        in_policy[M_POLICY_CPU_FREQ_EFFICIENT] = f_core_efficient;
+        in_policy[M_POLICY_UNCORE_FREQ_MAX] = f_uncore_max;
+        in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT] = f_uncore_efficient;
+
+        // Validate all (uncore frequency, max memory bandwidth) pairs
+        std::set<double> policy_uncore_freqs;
+        for (auto it = in_policy.begin() + M_POLICY_FIRST_UNCORE_FREQ;
+             it != in_policy.end() && std::next(it) != in_policy.end(); std::advance(it, 2)) {
+            auto mapped_mem_bw = *(it + 1);
```
bgeltz, line 237:
```suggestion
            auto uncore_freq = *it;
            auto mapped_mem_bw = *(it + 1);
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+            }
+            else if (!std::isnan(mapped_mem_bw)) {
+                throw Exception("CPUActivityAgent::" + std::string(__func__) +
+                                " policy maps a NaN CPU_UNCORE_FREQUENCY with max memory bandwidth: " +
+                                std::to_string(mapped_mem_bw),
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+
+            }
+        }
+
```
bgeltz:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+    bool CPUActivityAgent::do_send_sample(void) const
+    {
+        return false;
+    }
+
+    void CPUActivityAgent::adjust_platform(const std::vector<double>& in_policy)
+    {
+        m_do_write_batch = false;
+
+        if (m_update_qm_max_rate) {
```
bgeltz:
```suggestion
        if (m_do_qm_max_rate_update) {
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+    void CPUActivityAgent::init(int level, const std::vector<int> &fan_in, bool is_level_root)
+    {
+        m_core_frequency_requests = 0;
+        m_uncore_frequency_requests = 0;
+        m_core_frequency_clipped = 0;
+        m_uncore_frequency_clipped = 0;
+        m_resolved_f_uncore_efficient = 0;
+        m_resolved_f_uncore_max = 0;
+        m_resolved_f_core_efficient = 0;
+        m_resolved_f_core_max = 0;
```
bgeltz: Move this all to the initializer list in the constructor.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+        // These are not currently guaranteed to be the system uncore min and max,
+        // just what the user/admin has previously set.
+        m_freq_uncore_min = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+        m_freq_uncore_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+        m_freq_core_min = m_platform_io.read_signal("CPU_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+        m_freq_core_max = m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
```
bgeltz, line 64: Move this all to init_platform_io.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+
+        init_platform_io();
```
bgeltz: Wrap this in a if level == 0 check.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+                 it != in_policy.end() && std::next(it) != in_policy.end();
+                 std::advance(it, 2)) {
+
+                if (!std::isnan(*it)) {
+                    auto uncore_freq = static_cast<uint64_t>(*it);
+                    auto max_mem_bw = *(it + 1);
+                    // Not valid to have NAN max mem bw for uncore freq.
+                    GEOPM_DEBUG_ASSERT(!std::isnan(max_mem_bw),
+                                       "mapped CPU_UNCORE_FREQUENCY with no max memory bandwidth assigned.");
+                    m_qm_max_rate[uncore_freq] = max_mem_bw;
```
bgeltz, line 305: This seems like it can be moved into validate_policy(), then the if() block on 298 down to 320 can be removed from here.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+            // L3 usage, Network Traffic, HBM, and PCIE (GPUs) all use the uncore.
+            // Eventually all these components should be considered when scaling
+            // the uncore frequency in the efficient - performant range.
+            // A more robust/future proof solution may be to directly query uncore
+            // counters that indicate utilization (when/if available).
+            // For now only L3 bandwith metric is used.
+            double uncore_req = m_resolved_f_uncore_efficient + f_uncore_range * scalability_uncore;
+
+            //Clip uncore request within policy limits
```
bgeltz:
```suggestion
            // Clip uncore request within policy limits
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+        for (int domain_idx = 0; domain_idx < M_NUM_PACKAGE; ++domain_idx) {
+            if(std::isnan(uncore_freq_request.at(domain_idx))) {
+                uncore_freq_request.at(domain_idx) = in_policy[M_POLICY_UNCORE_FREQ_MAX];
+            }
+
+            if (uncore_freq_request.at(domain_idx) !=
+                m_uncore_freq_min_control.at(domain_idx).last_setting ||
+                uncore_freq_request.at(domain_idx) !=
+                m_uncore_freq_max_control.at(domain_idx).last_setting) {
+                //Adjust
```
bgeltz:
```suggestion
                // Adjust
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+                m_platform_io.adjust(m_uncore_freq_min_control.at(domain_idx).batch_idx,
+                                    uncore_freq_request.at(domain_idx));
+
+                m_platform_io.adjust(m_uncore_freq_max_control.at(domain_idx).batch_idx,
+                                    uncore_freq_request.at(domain_idx));
+
+                //save the value for future comparison
```
bgeltz:
```suggestion
                // save the value for future comparison
```

bgeltz (changes requested):

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,567 @@
...
+                    m_qm_max_rate[uncore_freq] = max_mem_bw;
+                }
+            }
+            // Warn the user if they have not entered a policy with memory bandwidth characterization.
+            // In this case the maximum uncore frequency will be used.  This can easily be converted
+            // to an assertion if it should be considered a failure case.
+            if (m_qm_max_rate.size() == 0) {
+                std::cerr << "Warning: <geopm> CPUActivityAgent did not receive a policy containing memory " <<
+                             "bandwidth characterization.  This may negatively impact agent performance." << std::endl;
+            }
```
bgeltz: Make this an error, then use the .size() to determine whether or not we need to build the map.

lhlawson:
A follow on PR that refactorsthe performance model used in this agent into its own helper class (ActivityPerformanceModel.cpp) is underway on a fork [here](https://github.com/lhlawson/geopm/tree/public-lhlawson-node-ca) for the second version of the agent.  
The helper class is [here](https://github.com/lhlawson/geopm/blob/public-lhlawson-node-ca/src/ActivityPerformanceModel.cpp), and an agent using it [here](https://github.com/lhlawson/geopm/blob/public-lhlawson-node-ca/src/NodeActivityAgent.cpp#L474). 

This refactor does not require any user visible changes (such as modification to the policy) but has not been fully tested against a wide range of workloads at this time.

test/CPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,623 @@
...
+                .WillRepeatedly(Return(mock_active));
+    EXPECT_CALL(*m_platform_io, sample(QM_CTR_SCALED_RATE_IDX))
+                .WillRepeatedly(Return(mock_active));
+    EXPECT_CALL(*m_platform_io, sample(CPU_UNCORE_FREQUENCY_IDX))
+                .WillRepeatedly(Return(m_cpu_uncore_freq_max));
+    m_agent->sample_platform(tmp);
+
+    //Adjust
+    //Check frequency
+    GEOPM_EXPECT_THROW_MESSAGE(m_agent->adjust_platform(policy), GEOPM_ERROR_INVALID,
```
lhlawson, line 581:
```suggestion
    GEOPM_EXPECT_THROW_MESSAGE(m_agent->adjust_platform(policy), GEOPM_ERROR_LOGIC,
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+                            std::to_string(in_policy[M_POLICY_CPU_FREQ_MAX]) +
+                            ".", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        // Check for NAN to set default values for policy
+        if (std::isnan(in_policy[M_POLICY_CPU_FREQ_EFFICIENT])) {
+            in_policy[M_POLICY_CPU_FREQ_EFFICIENT] = m_freq_core_min;
+        }
+
+        if (in_policy[M_POLICY_CPU_FREQ_EFFICIENT] > m_freq_core_max ||
```
avilcheslopez, line 138: Very minor note: it seems this check is a bit redundant with the one below. In other words, it seems like the "if" below could be merged with this "if" taking advantage of the fact that a prior check ensures in_policy[M_POLICY_CPU_FREQ_MAX] is <= m_freq_core_max.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+    bool CPUActivityAgent::do_send_sample(void) const
+    {
+        return false;
+    }
+
+    void CPUActivityAgent::adjust_platform(const std::vector<double>& in_policy)
+    {
+        m_do_write_batch = false;
+
+        if (m_qm_max_rate.size() == 0) {
```
avilcheslopez: Minor suggestion: ``m_qm_max_rate.empty()``

Same suggestion applies below.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+    // If controls have a valid updated value write them.
+    bool CPUActivityAgent::do_write_batch(void) const
+    {
+        return m_do_write_batch;
+    }
+
+    // Read signals from the platform and calculate samples to be sent up
+    void CPUActivityAgent::sample_platform(std::vector<double> &out_sample)
+    {
+        assert(out_sample.size() == M_NUM_SAMPLE);
```
avilcheslopez: GEOPM_DEBUG_ASSERT

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,238 @@
...
+characterization using the GEOPM experiment infrastructure.  This can be
+accomplished by running::
+
+    gen_slurm.sh 1 arithmetic_intensity_benchmark uncore_frequency_sweep
+
+The generated test.sbatch should be modified to include the following
+options, where ``MAX_UNCORE_FREQ``, ``MIN_UNCORE_FREQ``,
+``MAX_CORE_FREQ``, and ``MIN_CORE_FREQ`` are replaced with the relevant
+system (or administrator chosen) values::
+
```
lhlawson, line 151: RMID values need to be set as well.

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+      The maximum possible memory bandwidth associated with the
+      same numbered uncore frequency.
+      Used to build a mapping of uncore frequencies to maximum
+      memory bandwidths for frequency steering.
+
+Policy Generation
+-----------------
+
+Generating a policy for the CPU Compute Activity agent requires system
+characterization using the GEOPM experiment infrastructure.  This can be
```
lhlawson:
```suggestion
characterization using the GEOPM experiment infrastructure located in $GEOPM_SOURCE/integration/experiment/.  This can be
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+accomplished by running::
+
+    gen_slurm.sh 1 arithmetic_intensity_benchmark uncore_frequency_sweep
+
+The generated test.sbatch should be modified to enable Memory Bandwidth
+Monitoring using the following::
```
lhlawson:
```suggestion
Monitoring by adding the following to the test.sbatch file above the experiment script::
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+will not be able to accurately determine the maximum memory bandwidth at each
+uncore frequency.
+
+Additionally the test.sbatch should be modified to include the following
+experiment options, where ``MAX_UNCORE_FREQ``, ``MIN_UNCORE_FREQ``,
+``MAX_CORE_FREQ``, and ``MIN_CORE_FREQ`` are replaced with the relevant
+system (or administrator chosen) values::
+
+    --geopm-report-signals="MSR::QM_CTR_SCALED_RATE@package,CPU_UNCORE_FREQUENCY_STATUS@package,MSR::CPU_SCALABILITY_RATIO@package,CPU_FREQUENCY_MAX_CONTROL@package,CPU_UNCORE_FREQUENCY_MIN_CONTROL@package,CPU_UNCORE_FREQUENCY_MAX_CONTROL@package" \
+    --min-frequency=MIN_CORE_FREQ \
```
avilcheslopez: Suggestion (instead of using MIN_CORE_FREQ, ...): ``--min-frequency=<min system frequency>`` \

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+    --max-frequency=MAX_CORE_FREQ \
+    --step-frequency=100000000 \
+    --max-uncore-frequency=MAX_UNCORE_FREQ \
+    --min-uncore-frequency=MIN_UNCORE_FREQ \
```
lhlawson:
```suggestion
    --min-uncore-frequency=MIN_UNCORE_FREQ \
    --max-uncore-frequency=MAX_UNCORE_FREQ \
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+    --step-uncore-frequency=100000000 \
+    --trial-count=5 \
+
```
lhlawson, line 160: Add note about increasing runtime in SBATCH file to an appropriately large value

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+Then the test.sbatch script should be run on the node of interest.  This
```
lhlawson:
```suggestion
Then the test.sbatch script should be run on the node of interest.  Using::

sbatch -w <NODE OF INTEREST> test.sbatch

This
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+      memory bandwidths for frequency steering.
+
+Policy Generation
+-----------------
+
+Generating a policy for the CPU Compute Activity agent requires system
+characterization using the GEOPM experiment infrastructure.  This can be
+accomplished by running::
+
+    gen_slurm.sh 1 arithmetic_intensity_benchmark uncore_frequency_sweep
```
lhlawson:
```suggestion
    gen_slurm.sh 1 arithmetic_intensity uncore_frequency_sweep
```

avilcheslopez: NOTE: need to mention that AIB needs to be built

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+    --max-uncore-frequency=MAX_UNCORE_FREQ \
+    --min-uncore-frequency=MIN_UNCORE_FREQ \
+    --step-uncore-frequency=100000000 \
+    --trial-count=5 \
+
+Then the test.sbatch script should be run on the node of interest.  This
+will run multiple kernels of varying intensity that stress the core and
+uncore to help with system characterization.
+
+The CPU compute activity agent policy can then be generated by running::
```
lhlawson:
```suggestion
After sourcing the $GEOPM_SOURCE/integration/config/run_env.sh file
the CPU compute activity agent policy can then be generated by running::
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+
+
+
+
+
+
+Description
+-----------
+
+This agent is an experimental agent that is expected change significantly between
```
avilcheslopez: ...that is expected ^to change...

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+GEOPM 2.0 and 2.1.  Some examples of changes that may occur include agent policy
+changes, algorithmic chagnes, and folding the agent into other agents that achieve
```
avilcheslopez: typo: chagnes

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+high activity regions (uncore activity of 1.0) run at the ``Fumax`` frequency,
+and regions in between the extremes run at a frequency (``Fureq``) selected using
+the equation:
+
+``Fureq = Fue + (Fumax - Fue) * CPU_UNCORE_ACTIVITY``
+
+The ``CPU_UNCORE_ACTIVITY`` is defined as a simple ratio of the current memory bandwdith
+divided by the maximum possible bandwidth at the current ``CPU_UNCORE_FREQUENCY_STATUS`` value.
+
+The ``Fe`` value for all domains isintended to be an energy efficient frequency
```
avilcheslopez: typo: isintended

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,256 @@
...
+that is selected via system characterization.  The recommended approach to selecting
+``Fe`` for a given domain is to perform a frequency sweep of that domain (CPU or UNCORE)
+using a workload that scales strongly with the domain frequency.
+Using this approach ``Fe`` will be the frequency that provides the lowest
+energy consumption for the workload.
+
+``Fmax`` is intended to be the maximum allowable frequency for the domain,
+and may be set as the domain's default  maximum frequency, or limited based
```
avilcheslopez: Extra space between "default" and "maximum".

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+            in_policy[M_POLICY_CPU_FREQ_MAX] < m_freq_core_min ) {
+            throw Exception("CPUActivityAgent::" + std::string(__func__) +
+                            "():CPU_FREQ_MAX out of range: " +
+                            std::to_string(in_policy[M_POLICY_CPU_FREQ_MAX]) +
+                            ".", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        // Check for NAN to set default values for policy
+        if (std::isnan(in_policy[M_POLICY_CPU_FREQ_EFFICIENT])) {
+            in_policy[M_POLICY_CPU_FREQ_EFFICIENT] = m_freq_core_min;
```
avilcheslopez, line 135: Q: why not take the midpoint between the min and the max as the default?

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+        }
+
+        //////////////////////////
+        //UNCORE POLICY CHECKING//
+        //////////////////////////
+        if (std::isnan(in_policy[M_POLICY_UNCORE_FREQ_MAX])) {
+            in_policy[M_POLICY_UNCORE_FREQ_MAX] = m_freq_uncore_max;
+        }
+        if (std::isnan(in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT])) {
+            in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT] = m_freq_uncore_min;
```
avilcheslopez, line 162: Same question as before

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+    {
+        for (auto &child_pol : out_policy) {
+            child_pol = in_policy;
+        }
+    }
+
+    // Indicate whether to send the policy down to children
+    bool CPUActivityAgent::do_send_policy(void) const
+    {
+        return true;
```
avilcheslopez: Q: always return true or only send policy to children once?

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,560 @@
...
+            // less than the uncore_freq, which may cause issues when
+            // the uncore_freq signal does not exactly match the keys in the
+            // map.
+            //
+            // To handle this case we grab an entry one less than the
+            // lower bound if and only if the qm_max_itr isn't the first key
+            // and the uncore_freq is not an exact match.
+            if(qm_max_itr != m_qm_max_rate.begin() &&
+               m_qm_max_rate.find(uncore_freq) == m_qm_max_rate.end()) {
+                qm_max_itr = std::prev(qm_max_itr, 1);
```
avilcheslopez: qm_max_itr--?

Makefile.am
```diff
@@ -457,7 +457,9 @@ libgeopmpolicy_la_SOURCES = src/Accumulator.cpp \
                             src/record.hpp \
                             # end
 
-beta_source_files = src/Daemon.cpp \
+beta_source_files = src/CPUActivityAgent.cpp \
```
dannosliwcd, line 460: We should probably add a [.. note::](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) block to the .rst files for these files that specifies they are behind the beta flag.

integration/experiment/energy_efficiency/cpu_activity.py
```diff
@@ -0,0 +1,110 @@
...
+from experiment import launch_util
+from experiment import common_args
+from experiment import machine
+
+def setup_run_args(parser):
+    common_args.setup_run_args(parser)
+    parser.add_argument('--cpu-frequency-efficient', dest='cpu_fe',
+                        action='store', default='nan',
+                        help='The efficient frequency of the cpu (a.k.a. Fce) for this experiment')
+    parser.add_argument('--cpu-frequency-max', dest='cpu_fmax',
```
dannosliwcd, line 25: May be confusing vs `--max-frequency` in `common_args.py` (which is a max used in a sweep range). Maybe "--cpu-ca-max-frequency'? Maybe it doesn't matter.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+    void CPUActivityAgent::init_platform_io(void)
+    {
+        for (int domain_idx = 0; domain_idx < M_NUM_CORE; ++domain_idx) {
+            m_core_scal.push_back({m_platform_io.push_signal("MSR::CPU_SCALABILITY_RATIO",
+                                                             GEOPM_DOMAIN_CORE,
+                                                             domain_idx), NAN});
+            m_core_freq_control.push_back({m_platform_io.push_control("CPU_FREQUENCY_MAX_CONTROL",
+                                                                      GEOPM_DOMAIN_CORE,
+                                                                      domain_idx), -1});
```
dannosliwcd: Why -1 instead of NaN?

Makefile.am
avilcheslopez: Yep! Will do.

integration/experiment/energy_efficiency/cpu_activity.py
avilcheslopez: You know what... Looking at ``common_args.py``, I'm thinking it deserves a second look and perhaps some refactoring is in order (after renaming some of the arguments). At the moment, the meaning of "frequency" in ``common_args.py`` is dependent on the experiment script that uses it (could be core or uncore, for example).

That being the case, I'm tempted to leave these (``cpu_activity``) arguments as they are because this agent needs both: a core frequency limit and an uncore frequency limit. Then we can revisit this once we take a second look at ``common_args.py``.

src/CPUActivityAgent.cpp
avilcheslopez: In ``CPUActivityAgent::adjust_platform()``, when it determines what the frequency should be (``core_request``), it needs to compare it against the last setting to decide whether the setting actually needs to be updated. Using -1 allows that check to work properly the first time a frequency decision is made.

dannosliwcd: I think `NaN` would be a more defensive way to meet that goal.  This placeholder value is not expected to be written. If we use NaN instead, then `PlatformIOImp::adjust` will throw an exception if that ever does happen, instead of silently writing the placeholder value.

With the no-fast-math change that just went in, we can assume `NaN != <anything>`, so the `CPUActivityAgent::adjust_platform` check should still work with NaN.

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,564 @@
...
+                    GEOPM_DEBUG_ASSERT(!std::isnan(max_mem_bw),
+                                       "mapped CPU_UNCORE_FREQUENCY with no max memory bandwidth assigned.");
+                    m_qm_max_rate[uncore_freq] = max_mem_bw;
+                }
+            }
+
+            if (m_qm_max_rate.empty()) {
+                throw Exception("CPUActivityAgent::" + std::string(__func__) +
+                                "(): CPUActivityAgent policy did not contain" +
+                                " memory bandwidth characteriztaion.",
```
dannosliwcd:
```suggestion
                                " memory bandwidth characterization.",
```

src/CPUActivityAgent.hpp
```diff
@@ -0,0 +1,118 @@
...
+                M_NUM_POLICY = 63,
+            };
+
+            // Sample indices; must match sample_names()
+            enum m_sample_e {
+                M_NUM_SAMPLE
+            };
+
+            std::map<std::string, double> m_policy_available;
+            std::map<double,double> m_qm_max_rate;
```
dannosliwcd: Maybe add a comment like: This maps (uncore frequency) -> (maximum memory bandwidth) specified in the input policy

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,564 @@
...
+            // the uncore_freq signal does not exactly match the keys in the
+            // map.
+            //
+            // To handle this case we grab an entry one less than the
+            // lower bound if and only if the qm_max_itr isn't the first key
+            // and the uncore_freq is not an exact match.
+            if(qm_max_itr != m_qm_max_rate.begin() &&
+               m_qm_max_rate.find(uncore_freq) == m_qm_max_rate.end()) {
+                qm_max_itr--;
+            }
```
dannosliwcd: I think this means we want the last element <= uncore_freq.

Could probably simplify with something like:
```
qm_max_itr = upper_bound(uncore_freq)
if (not the beginning) --qm_max_itr;
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,564 @@
...
+                // Adjust
+                m_platform_io.adjust(m_core_freq_control.at(domain_idx).batch_idx,
+                                     core_freq_request.at(domain_idx));
+
+                // Save the value for future comparison
+                m_core_freq_control.at(domain_idx).last_setting = core_freq_request.at(domain_idx);
+                ++m_core_frequency_requests;
+                m_do_write_batch = true;
+            }
+        }
```
dannosliwcd: Bounds-clamping, sending to PIO, and flagging whether write_batch is needed could be delegated to `geopm::FrequencyGovernor`

src/CPUActivityAgent.cpp
dannosliwcd: Actually, I wonder if this should be delegated to `FrequencyGovernor` instead.

src/CPUActivityAgent.cpp
dannosliwcd: Though I guess that would only work if the governor runs on a system that natively uses core domain for frequency controls. Otherwise, all of the M_NUM_CORE loops here need to change to ask the governor which  `FrequencyGovernor::frequency_domain_type()` to use, or we need to tell the governor which domain we want (latter option doesn't currently exist, but is in one of my upcoming PRs for the balancer agent).

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,564 @@
...
+    }
+
+    // Wait for the remaining cycle time to keep Controller loop cadence
+    void CPUActivityAgent::wait(void)
+    {
+        geopm_time_s current_time;
+        do {
+            geopm_time(&current_time);
+        }
```
dannosliwcd, line 474: Do we really need a busy wait? It seems like this might be done to reduce variation in time between controller steps. Does this agent actually need that? All of the sampled values are read as rates anyways (vs point-in-time snapshots of monotonic counters).

avilcheslopez: Yeah, this is just the standard wait implementation all agents are using. We've talked about standardizing the implementation and factoring it out so that it doesn't have to be implemented in every agent. We've talked about a smarter implementation that leverages nanosleep() to minimize the amount of time we rely on a busy wait.

I would prefer to defer addressing this to when this refactoring work is done, as it will address it for all the agents.

dannosliwcd: Sounds good to defer this comment to #1065

src/CPUActivityAgent.cpp
dannosliwcd: From our offline discussion, let's defer this comment thread because it also depends on other work:

 - Optional: #2642 would enable you to use FrequencyGovernor without modifying your M_NUM_CORE loops to ask the governor what to use.
 - #2643 to resolve this issue, and to resolve the discovery that FrequencyMapAgent also does not use the helper

src/CPUActivityAgent.cpp
```diff
@@ -324,19 +324,14 @@ namespace geopm
...
-            if(qm_max_itr != m_qm_max_rate.begin() &&
-               m_qm_max_rate.find(uncore_freq) == m_qm_max_rate.end()) {
+            // Get max mem. bandwidth for uncore_freq. There may be uncore
+            // frequencies for which an exact match doesn't exist. To handle
+            // this case, we grab the entry prior to upper_bound() (as long as
+            // it's not the first entry), in other words, the last entry that
+            // is <= uncore_freq.
+            auto qm_max_itr = m_qm_max_rate.upper_bound(uncore_freq);
+            if (qm_max_itr != m_qm_max_rate.begin())
                 qm_max_itr--;
```
dannosliwcd:
```suggestion
            if (qm_max_itr != m_qm_max_rate.begin()) {
                --qm_max_itr;
            }
```

or on one line: `if (qm_max_itr != m_qm_max_rate.begin()) --qm_max_itr;`

Not a required change, since this feels like a style preference issue and I don't know if the team has any rules about this. My personal motivation is that I've seen bugs appear on projects from moving code into or out the body of this style of if-statement during later changes.

integration/experiment/uncore_frequency_sweep/gen_cpu_activity_policy_recommendation.py
```diff
@@ -0,0 +1,169 @@
...
+    Extract the columns of interest from the full report collection
+    dataframe.
+    """
+    df_filtered = df
+    if region_list:
+        df_filtered = df[df['region'].isin(region_list)]
+
+    if ('MSR::QM_CTR_SCALED_RATE' not in df_filtered.columns):
+        df_filtered['MSR::QM_CTR_SCALED_RATE'] = (df_filtered['MSR::QM_CTR_SCALED_RATE@package-0'] +
+                                             df_filtered['MSR::QM_CTR_SCALED_RATE@package-1'])/2
```
dannosliwcd: Does this print a warning about setting a value in a copied slice (vs setting in a slice of the original).

If it does, then maybe modify the start of the selection to:
```suggestion
        df_filtered = df.loc[df['region'].isin(region_list)]

    if ('MSR::QM_CTR_SCALED_RATE' not in df_filtered.columns):
        df_filtered['MSR::QM_CTR_SCALED_RATE'] = (df_filtered['MSR::QM_CTR_SCALED_RATE@package-0'] +
                                             df_filtered['MSR::QM_CTR_SCALED_RATE@package-1'])/2
```

integration/experiment/uncore_frequency_sweep/gen_cpu_activity_policy_recommendation.py
```diff
@@ -0,0 +1,169 @@
...
+    Perform characterization of the system memory bandwidth.  This
+    is used by the agent to help determine the appropriate uncore frequency
+    """
+    df = df_region_group.get_group(region)
+    uncore_freq_set = sorted(set(df['uncore-frequency (Hz)'].to_list()))
+
+    mem_bw_dict = {}
+    for k in uncore_freq_set:
+        mem_df = df.groupby('uncore-frequency (Hz)').get_group(k)
+        #TODO: use both packages!  This will skew in favor of package-0
```
dannosliwcd, line 73: Is this still needed? Should we open an issue?

integration/experiment/uncore_frequency_sweep/gen_cpu_activity_policy_recommendation.py
```diff
@@ -0,0 +1,169 @@
...
+def frequency_recommendation(df_region_group, region, domain):
+    if domain == "UNCORE":
+        freq_col = 'uncore-frequency (Hz)'
+        fixed_freq_col = 'frequency (Hz)'
+    elif domain == "CORE":
+        freq_col = 'frequency (Hz)'
+        fixed_freq_col = 'uncore-frequency (Hz)'
+    else:
+        sys.stderr.write('<geopm> Error: unsupported domain ' + domain + \
+                         'proovided\n')
```
dannosliwcd:
```suggestion
                         'provided\n')
```

src/CPUActivityAgent.cpp
avilcheslopez: Yeah, I don't actually need post-increment semantics here (was mainly following common style). In the rest of the codebase, for similar operations, we have a bit of a mix, but in the vast majority of cases we use pre-increment. I'll update this accordingly.

integration/experiment/uncore_frequency_sweep/gen_cpu_activity_policy_recommendation.py
avilcheslopez: Will fix!

integration/experiment/uncore_frequency_sweep/gen_cpu_activity_policy_recommendation.py
lhlawson: This shouldn't be needed currently, but is a future feature request if we find there is a significant amount of package to package variability that we can capitalize on for better energy efficiency.

dannosliwcd (dismissed):

service/docs/source/GEOPM_CXX_MAN_CPUActivityAgent.3.rst
```diff
@@ -0,0 +1,34 @@
+
```
bgeltz:
```suggestion
```

service/docs/source/GEOPM_CXX_MAN_CPUActivityAgent.3.rst
```diff
@@ -0,0 +1,34 @@
...
+geopm::CPUActivityAgent(3) -- agent for selecting CPU frequency based on CPU compute activity
+=============================================================================================
+
```
bgeltz, line 3:
```suggestion
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,270 @@
+
```
bgeltz:
```suggestion
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,270 @@
...
+geopm_agent_cpu_activity(7) -- agent for selecting CPU frequency based on CPU compute activity
+=================================================================================================
+
+
+
+
+
```
bgeltz:
```suggestion
```

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,270 @@
...
+-----------
+
+.. note::
+    This is currently an experimental agent and is only available when
+    building GEOPM with the ``--enable-beta`` flag. Some areas or aspects that
+    are subject to change include its interface (e.g. the policy) and
+    algorithm. It is also possible that this agent may be refactored and
+    combined with other agents.
+
+The goal of this Agent is to save CPU energy by scaling CPU frequency based upon
```
bgeltz: Lowercase unless you're talking about the Agent class/interface.  This seems to be the pattern we use in other files AFIACT.

```suggestion
The goal of this agent is to save CPU energy by scaling CPU frequency based upon
```

Or you could spell out the full agent name as you do elseware.

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,270 @@
...
+Without this, the uncore bandwidth characteriztaion analysis scripts will not
+be able to accurately determine the maximum memory bandwidth at each uncore
+frequency.
+
+Additionally the ``test.sbatch`` should be modified to include the following
+experiment options, where the text within angle brackets (``<>``) needs to be
+replaced with relevant system (or administrator chosen) values::
+
+    --geopm-report-signals="MSR::QM_CTR_SCALED_RATE@package,CPU_UNCORE_FREQUENCY_STATUS@package,MSR::CPU_SCALABILITY_RATIO@package,CPU_FREQUENCY_MAX_CONTROL@package,CPU_UNCORE_FREQUENCY_MIN_CONTROL@package,CPU_UNCORE_FREQUENCY_MAX_CONTROL@package" \
+    --min-frequency=<min. core frequency> \
```
bgeltz, line 153:
```suggestion
    --min-frequency=<min core frequency> \
```

Either min or minimum.  Do this for all max as well.

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,270 @@
...
+benchmark located in ``$GEOPM_SOURCE/integration/apps/arithmetic_intensity``.
+Prior to starting, the arithmetic intensity benchmark needs to be built (use
+the ``build.sh`` script provided in the benchmark's folder).
+
+
+.. note::
+    Before performing the system characterization, please ensure the
+    system is quiesced (i.e. not running other heavy processes/workloads).
+
+The first step is to generate the execution script by running::
```
bgeltz, line 133:
```suggestion
The first step is to generate the execution script by running:
```

Does this need to be a double colon?

service/docs/source/geopm_agent_cpu_activity.7.rst
```diff
@@ -0,0 +1,270 @@
...
+    srun -N ${SLURM_NNODES} geopmwrite MSR::QM_EVTSEL:RMID board 0 0
+    srun -N ${SLURM_NNODES} geopmwrite MSR::QM_EVTSEL:EVENT_ID board 0 2
+
+Without this, the uncore bandwidth characteriztaion analysis scripts will not
+be able to accurately determine the maximum memory bandwidth at each uncore
+frequency.
+
+Additionally the ``test.sbatch`` should be modified to include the following
+experiment options, where the text within angle brackets (``<>``) needs to be
+replaced with relevant system (or administrator chosen) values::
```
bgeltz, line 150: Include examples of how to derive this information with geopmread.

src/CPUActivityAgent.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+            // Sample indices; must match sample_names()
+            enum m_sample_e {
+                M_NUM_SAMPLE
+            };
+
+            std::map<std::string, double> m_policy_available;
+            // Maps uncore frequency -> maximum memory bandwidth achieved by
+            // that frequency (determined by system characterization)
+            std::map<double,double> m_qm_max_rate;
```
bgeltz:
```suggestion
            std::map<double, double> m_qm_max_rate;
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,559 @@
...
+                                " policy maps a NaN CPU_UNCORE_FREQUENCY with max memory bandwidth: " +
+                                std::to_string(mapped_mem_bw),
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+    }
+
+    // Distribute incoming policy to children
+    void CPUActivityAgent::split_policy(const std::vector<double>& in_policy,
+                                    std::vector<std::vector<double> >& out_policy)
```
bgeltz:
```suggestion
    void CPUActivityAgent::split_policy(const std::vector<double>& in_policy,
                                        std::vector<std::vector<double> >& out_policy)
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,559 @@
...
+            }
+
+            core_req = std::max(in_policy[M_POLICY_CPU_FREQ_EFFICIENT], core_req);
+            core_req = std::min(in_policy[M_POLICY_CPU_FREQ_MAX], core_req);
+            core_freq_request.push_back(core_req);
+        }
+
+        // Set per core controls
+        for (int domain_idx = 0; domain_idx < M_NUM_CORE; ++domain_idx) {
+            if(std::isnan(core_freq_request.at(domain_idx))) {
```
bgeltz:
```suggestion
            if (std::isnan(core_freq_request.at(domain_idx))) {
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,559 @@
...
+                // Save the value for future comparison
+                m_core_freq_control.at(domain_idx).last_setting = core_freq_request.at(domain_idx);
+                ++m_core_frequency_requests;
+                m_do_write_batch = true;
+            }
+        }
+
+        // Set per package controls
+        for (int domain_idx = 0; domain_idx < M_NUM_PACKAGE; ++domain_idx) {
+            if(std::isnan(uncore_freq_request.at(domain_idx))) {
```
bgeltz:
```suggestion
            if (std::isnan(uncore_freq_request.at(domain_idx))) {
```

src/CPUActivityAgent.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+    }
+
+    // Indicate whether to send samples up to the parent
+    bool CPUActivityAgent::do_send_sample(void) const
+    {
+        return false;
+    }
+
+    void CPUActivityAgent::adjust_platform(const std::vector<double>& in_policy)
```
bgeltz, line 289: This method is great candidate for a refactor.  It's pretty long, and can be broken up into helper functions.  Hopefully we can clean this up with the PerformanceModel work in the near future.

integration/test/test_cpu_activity_agent.py
```diff
@@ -0,0 +1,269 @@
...
+        ca_experiment_args.output_dir = cls._minife_agent_dir
+        ca_experiment_args.trial_count = 1
+
+        launch_helper(cpu_activity, ca_experiment_args, minife.MinifeAppConf(node_count), [])
+
+    def tearDown(self):
+        if sys.exc_info() != (None, None, None):
+            TestIntegration_cpu_activity._keep_files = True
+
+    @util.skip_unless_config_enable('beta')
```
bgeltz: This should be added up at the class level (and removed from here) to ensure that setUpClass() does not run if beta is not enabled.  Seems like it would take a long time given how many combinations of agents and runs there are.

test/CPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,583 @@
...
+    m_mbm_max = {45414967307.69231, 64326515384.61539, 72956528846.15384,
+                 77349315384.61539, 82345998076.92308, 87738286538.46153,
+                 91966364814.81482, 96728174074.07408, 100648379629.62962,
+                 102409246296.2963, 103624103703.7037, 104268944444.44444,
+                 104748888888.88889};
+    ASSERT_EQ(m_cpu_uncore_freqs.size(), m_mbm_max.size());
+    ASSERT_EQ(m_mbm_max.size(), M_NUM_UNCORE_MBM_READINGS);
+
+    for (size_t i = 0; i < M_NUM_UNCORE_MBM_READINGS; ++i) {
+        m_default_policy.push_back(static_cast<double>(m_cpu_uncore_freqs[i]));
```
bgeltz: Why is the static_cast necessary? m_cpu_uncore_freqs is already a vector of doubles.

test/CPUActivityAgentTest.cpp
```diff
@@ -0,0 +1,583 @@
...
+    EXPECT_NE("bad_string", m_agent->plugin_name());
+}
+
+TEST_F(CPUActivityAgentTest, validate_policy)
+{
+    const std::vector<double> policy_nan(m_num_policy, NAN);
+    std::vector<double> policy;
+
+    // default policy with 1.2-2.4GHz MBM
+    // max rates defined is accepted
```
bgeltz:
```suggestion
    // max rates defined are accepted
```

bgeltz (changes requested):

service/docs/source/geopm_agent_cpu_activity.7.rst
dannosliwcd: The colons are not rendered, but the indented block following the double colon is treated as preformatted text. I think without the `::`, the indented text would be rendered as a block quote.

src/CPUActivityAgent.cpp
lhlawson: That approach may also work but was found in testing to be less beneficial than either:
1) Sticker - 2*freq step
2) Min freq

integration/experiment/energy_efficiency/cpu_activity.py
avilcheslopez: Filed an issue to look into this: #2648

service/docs/source/geopm_agent_cpu_activity.7.rst
avilcheslopez: Opted to spell out the name of the agent in the beginning (first reference) and simply refer to it as "the agent" afterwards.

service/docs/source/geopm_agent_cpu_activity.7.rst
avilcheslopez: I tested it out: without the double colon the following block of text appears as normal text (i.e. not preformatted text). At least it still appears indented.

Leaving the double colon.

bgeltz (approved):

PR #2366, opened 2022-06-01
Updated LevelZero IOGroup to per signal behavior 

### LINKED ISSUE #2365: LevelZero IOGroup signals are set to variable behavior
https://github.com/geopm/geopm/blob/9d6f1d73f23eafa93d2be554807915b78d822ed5/service/src/LevelZeroIOGroup.cpp#L963

All signals should not use the variable behavior, most notably energy.  This will impact the gpu-energy report values.  
Similar issues were addressed for NVML in https://github.com/geopm/geopm/pull/1945

lhlawson:

- Relates to #2365 from github issues
- Fixes #2365 change request from github issues.

Updates the Level Zero IO Group to allow signals to have a different behavior than variable.  This is needed for proper GPU reporting, and mirrors the change made in the NVMLIOGroup as part of #1945

Tasks required to move this from Draft
- [ ] Testing on a live system.

### DIFF
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -49,6 +49,7 @@ namespace geopm
                                   "Compute/GPU chip domain current frequency in hertz",
                                   GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP,
                                   Agg::average,
+                                  IOGroup::M_SIGNAL_BEHAVIOR_VARIABLE,
                                   string_format_double,
                                   {},
                                   [this](unsigned int domain_idx) -> double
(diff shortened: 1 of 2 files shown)


service/src/LevelZeroIOGroup.cpp
```diff
@@ -345,6 +364,7 @@ namespace geopm
                                   "\nReadings are valid only after writing to this control",
                                   GEOPM_DOMAIN_BOARD_ACCELERATOR_CHIP,
                                   Agg::average,
+                                  IOGroup::M_SIGNAL_BEHAVIOR_MONOTONE,
```
lhlawson:
```suggestion
                                  IOGroup::M_SIGNAL_BEHAVIOR_VARIABLE,
```

bgeltz (approved):

PR #2369, opened 2022-06-03
Added Level Zero Throttle Reasons signal

### LINKED ISSUE #2368: As a GPU researcher I would like throttle reasons for Intel Xe Hardware available as a signal
I'm always frustrated when hardware or firmware throttles below the requested frequency and I don't know why.

LevelZeroIOGroup should expose the throttle reasons as discussed in the Level Zero spec here:
https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#frequency

Tracking frequency and power using LevelZeroIOGroup throughout a run and correlating high power draw and frequency excursions below requested.  

**Additional context**
Future updates should be made for throttle time or % of time throttled as discussed in the Level Zero Sysman public spec.

### LINKED ISSUE #2638: As a user of the GEOPM integration infrastructure I would like for application configurations to be automatically detected by experiment scripts
**What problem does the story solve? Please describe.**
Whenever I add a new application configuration generator to `integration/experiment/apps/`, I also need to add a launcher for that application in each `integration/experiment/` subdirectory. Most of these experiment files are nearly duplicates of each other, with the only differences in their contents being the application names.

**Describe the tasks that will be required to complete the story**
Each application launcher should be able to search for compatible applications and make it possible to launch those applications in the experiment.

For example instead of a collection of `run_monitor_<appname>.py` scripts, just have a single `run_monitor.py` script like:

```
$ run_monitor.py help
usage: run_monitor.py [-h] {list,run,help} ...

positional arguments:
  {list,run,help}
    list           List the applications can run with this experiment
    run            Run an application with the monitor experiment
    help           Show help for an application with the monitor experiment
$ run_monitor.py list
arithmetic_intensity
<omitted for brevity>
pennant
$ run_monitor.py help arithmetic_intensity
<prints what the current run_monitor_arithmetic_intensity.py --help prints>
$ run_monitor.py run arithmetic_intensity <args that normally go to run_monitor_arithmetic_intensity.py>
```

This could probably have an additional layer of separation from the experiment types. In that case, it should be possible to separately list experiments or apps, and specify both experiment and app in the `run` subcommand.

lhlawson:

- Relates to #2368 feature request from github issues
- Fixes #2368 change request from github issues.

Provides GPU core domain throttle reasons as a LevelZeroIOGroup signal at a per chip level as covered in the Level Zero Sysman API - https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#frequency

Future PRs for percentage of time throttled should be considered.  

Tasks before merge:
- [x] Add bitwise_or aggregration function for throttle reasons
- [x] Test bitwise_or agg on a live system
- [x] Retest after data type change (double --> uint32_t) on live system.  Complete on system at idle (reports 0) and hitting power cap (reports [ZES_FREQ_THROTTLE_REASON_FLAG_AVE_PWR_CAP ](https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes__api_8h_1a03134de1353e4ecaa94d5ec76db12651))

### DIFF
--- a/service/src/Agg.cpp
+++ b/service/src/Agg.cpp
@@ -66,6 +66,21 @@ namespace geopm
         return result;
     }
 
+    double Agg::bitwise_or(const std::vector<double> &operand)
+    {
+        auto filtered = nan_filter(operand);
+        uint32_t result = 0;
+        if (filtered.size()) {
+            result = 0;
+            for (auto it : filtered) {
+                if (it >= 0) {
+                    result |= (uint32_t) it;
+                }
+            }
+        }
+        return result;
+    }
+
     double Agg::logical_and(const std::vector<double> &operand)
     {
         auto filtered = nan_filter(operand);
--- a/service/src/Agg.cpp
+++ b/service/src/Agg.cpp
@@ -69,14 +69,15 @@ namespace geopm
     double Agg::bitwise_or(const std::vector<double> &operand)
     {
         auto filtered = nan_filter(operand);
-        uint32_t result = 0;
+        double result = NAN;
+        uint64_t agg_tmp = 0;
         if (filtered.size()) {
-            result = 0;
             for (auto it : filtered) {
                 if (it >= 0) {
-                    result |= (uint32_t) it;
+                    agg_tmp |= (uint64_t) it;
                 }
             }
+            result = (double) agg_tmp;
         }
         return result;
     }
--- a/service/src/Agg.cpp
+++ b/service/src/Agg.cpp
@@ -66,16 +66,14 @@ namespace geopm
         return result;
     }
 
-    double Agg::bitwise_or(const std::vector<double> &operand)
+    double Agg::integer_bitwise_or(const std::vector<double> &operand)
     {
         auto filtered = nan_filter(operand);
         double result = NAN;
-        uint64_t agg_tmp = 0;
         if (filtered.size()) {
+            int64_t agg_tmp = 0;
             for (const auto &it : filtered) {
-                if (it >= 0) {
-                    agg_tmp |= (uint64_t) it;
-                }
+                agg_tmp |= (int64_t) it;
             }
             result = (double) agg_tmp;
         }
@@ -202,6 +200,7 @@ namespace geopm
             {"sum", sum},
             {"average", average},
             {"median", median},
+            {"integer_bitwise_or", integer_bitwise_or},
             {"logical_and", logical_and},
             {"logical_or", logical_or},
             {"region_hash", region_hash},
@@ -227,6 +226,7 @@ namespace geopm
             {sum, "sum"},
             {average, "average"},
             {median, "median"},
+            {integer_bitwise_or, "integer_bitwise_or"},
             {logical_and, "logical_and"},
             {logical_or, "logical_or"},
             {region_hash, "region_hash"},
@@ -253,6 +253,7 @@ namespace geopm
             {sum, M_SUM},
             {average, M_AVERAGE},
             {median, M_MEDIAN},
+            {integer_bitwise_or, M_INTEGER_BITWISE_OR},
             {logical_and, M_LOGICAL_AND},
             {logical_or, M_LOGICAL_OR},
             {region_hash, M_REGION_HASH},
@@ -279,6 +280,7 @@ namespace geopm
             {M_SUM, sum},
             {M_AVERAGE, average},
             {M_MEDIAN, median},
+            {M_INTEGER_BITWISE_OR, integer_bitwise_or},
             {M_LOGICAL_AND, logical_and},
             {M_LOGICAL_OR, logical_or},
             {M_REGION_HASH, region_hash},
@@ -303,6 +305,7 @@ namespace geopm
             {M_SUM, "sum"},
             {M_AVERAGE, "average"},
             {M_MEDIAN, "median"},
+            {M_INTEGER_BITWISE_OR, "integer_bitwise_or"},
             {M_LOGICAL_AND, "logical_and"},
             {M_LOGICAL_OR, "logical_or"},
             {M_REGION_HASH, "region_hash"},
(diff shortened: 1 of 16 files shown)


service/src/LevelZeroIOGroup.cpp
```diff
@@ -115,6 +115,22 @@ namespace geopm
                                   },
                                   1e6
                                   }},
+                              {M_NAME_PREFIX + "GPU_CORE_THROTTLE_REASONS", {
+                                  "GPU Compute/Core domain throttle reasons.  See oneAPI Level Zero Sysman Spec for decoding",
+                                  GEOPM_DOMAIN_GPU_CHIP,
+                                  Agg::average,
```
lhlawson: A bitwise OR makes more sense here for aggregation as it would represent all the throttling across the device compute domains

service/src/LevelZeroDevicePool.cpp
```diff
@@ -142,6 +142,25 @@ namespace geopm
                                          dev_subdev_idx_pair.second);
     }
 
+    uint32_t LevelZeroDevicePoolImp::frequency_throttle_reasons(int domain, unsigned int domain_idx,
+                                                              int l0_domain) const
+    {
+        if (domain != GEOPM_DOMAIN_GPU_CHIP) {
+            throw Exception("LevelZeroDevicePool::" + std::string(__func__) +
+                             ": domain " + std::to_string(domain) +
+                            " is not supported for the frequency domain.",
```
cmcantalupo:
```suggestion
                            " is not supported for reading the \"frequency throttle reason\"",
```

cmcantalupo: Perhaps something more descriptive is possible, but "domain" was overused here.

service/src/Agg.cpp
```diff
@@ -66,6 +66,22 @@ namespace geopm
         return result;
     }
 
+    double Agg::bitwise_or(const std::vector<double> &operand)
+    {
+        auto filtered = nan_filter(operand);
+        double result = NAN;
+        uint64_t agg_tmp = 0;
+        if (filtered.size()) {
+            for (auto it : filtered) {
```
cmcantalupo:
```suggestion
            for (const auto &it : filtered) {
```

service/src/geopm/Agg.hpp
```diff
@@ -38,6 +38,10 @@ namespace geopm
             static double average(const std::vector<double> &operand);
             /// @brief Returns the median of the input operands.
             static double median(const std::vector<double> &operand);
+            /// @brief Returns the output of bitwise OR over all the
```
cmcantalupo, line 42: The fact that the double is cast to a uint64_t before the bit comparison is not documented here.  I'm a little confused about what the user would expect if they passed a bit-field MSR here (one that ends in #).  In that case, it would be surprising that the cast is done, since the double precision representation of the register is not intended to be interpreted as a number.  I think this should be somehow identified in the name, perhaps: ``uint64_bitwise_or()``?

A double precision number cannot represent the bit reserved for the sign-bit in a int64_t as an integer value (the mantissa has fewer than 64 bits).  For this reason it is also a little surprising that the unsigned cast is done.  Why not just cast to a int64_t, or a long?  There is no logical difference here.

Perhaps we can have:
```
// Do a bitwise-or of the memory representing the double, this useful for bit field MSRs that end in #.
field_bitswise_or()
// Cast the double to a int64_t and then do bitwise-or.
integer_bitwise_or()
```

service/test/AggTest.cpp
```diff
@@ -41,6 +41,20 @@ TEST(AggTest, agg_function)
...
     EXPECT_EQ(0.0, Agg::logical_or({0.0, 0.0}));
 
+    EXPECT_EQ(1, Agg::bitwise_or({0.0, 0.0, 1.0, 0.0}));
+    EXPECT_EQ(0, Agg::bitwise_or({0.0, 0.0}));
+    EXPECT_EQ(7, Agg::bitwise_or({5.0, 2.0}));
+    EXPECT_EQ(3, Agg::bitwise_or({3.0, 1.0}));
+    EXPECT_EQ(6, Agg::bitwise_or({4.0, 2.0}));
+    EXPECT_EQ(0, Agg::bitwise_or({0.0, 0.0, NAN}));
+    EXPECT_EQ(0, Agg::bitwise_or({0.1, 0}));
+    EXPECT_EQ(0, Agg::bitwise_or({-1, 0}));
```
cmcantalupo: Surprising really....

cmcantalupo (changes requested):
See comment about adding another aggregator.  The unsigned cast is a little confusing.

service/src/Agg.cpp
```diff
@@ -66,6 +66,22 @@ namespace geopm
...
     }
 
+    double Agg::bitwise_or(const std::vector<double> &operand)
+    {
+        auto filtered = nan_filter(operand);
+        double result = NAN;
+        uint64_t agg_tmp = 0;
+        if (filtered.size()) {
+            for (const auto &it : filtered) {
+                if (it >= 0) {
```
cmcantalupo: Lets not support -1 as a special value in the this implementation, rather do it on the LevelZeroIOGroup side.

service/src/Agg.cpp
```diff
@@ -66,6 +66,22 @@ namespace geopm
         return result;
     }
 
+    double Agg::bitwise_or(const std::vector<double> &operand)
```
cmcantalupo: Needs to be mapped to a string and an enum in the bottom of the Agg.cpp imp.

service/src/geopm/Agg.hpp
lhlawson: Updated to be integer_bitwise_or(), issue #2454 filed to track feature request for field_bitwise_or() and identify signals that may use it by default (such as raw MSRs).

cmcantalupo (approved):

Issue #2139, opened 2022-03-09
As a GEOPM researcher I would like to be able to use Pytorch models to control GPU frequency in a distributed fashion


lhlawson:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to run ML based agents in a distributed fashion, or in a single node fashion using a C++ agent.  

**Describe the solution you'd like**
Inclusion of a C++ agent that loads a pytorch model and uses its output to control frequency based on GPU metrics

**Describe alternatives you've considered**
Using a python agent to do the same (see #2044).  This does not currently provide a multi-node solution.  

**Additional context**
Pytorch provides easier C++ interfacing than tensorflow for this task, hence the shift from TF in #2044 to pytorch.

PR #2370, opened 2022-06-06
Pytorch GPU-NN Phi frequency response linearization

lhlawson:
- Relates to #2139 feature request from github issues
- Based on  #2268

This PR is based on several in-flight PRs. The relevant commit is [312d532](https://github.com/geopm/geopm/pull/2370/commits/312d5328c40549af5f36d7b5341ddc2a89a5bc1d)

This PR provides an additional -force-linear-phi-freq option for GPU frequency sweep processing that modifies the existing perf-energy tradeoff frequency selection by forcing non repeating linearly spaced frequencies for every phi value between the initial perf-energy tradeoff maximum and minimum frequency identified. 

In practice this should help phi have a greater dynamic range.  When running the  process_gpu_frequency_sweep.py without this option on a frequency sweep of dgemm and stream the following phi-freq values were chosen:
image

Note that stream effectively has 3 responses for all 10 phi values, with phi 0.2-1.0 being identical.  Similarly dgemm shows multiple repeated frequency selections.  The new option attempts to correct this by shaping the frequency selection within the per workload range selected.
image

Future Tasks:
- [ ] Confirm linearization of all regions of a workload when workload regions are available (such as in the CPU-NN processing case)


lhlawson:
Closing, functionality has been folded into #2268

PR #2381, opened 2022-06-15
Update region hint enum values to make them sequential

### LINKED ISSUE #1547: Region hint enum can be sequential
> estimate 2

As a GEOPM developer, I'd like to refactor how the hint enum is defined and utilized.

The hints associated with regions were originally designed as powers of 2 such that a bitfield could be used to set multiple hints.  The code is here:
https://github.com/geopm/geopm/blob/7f7e398333bcace834c3882542d4ec1e313f9d4d/src/geopm.h#L55-L73

The runtime no longer supports setting multiple hints.  This enum should be updated to simply be sequential, and any/all map lookups on the old enum should be converted to simple array operations.  This should result in an efficiency improvement since the current map lookups are in the hot path.

avilcheslopez:

- Fixes #1547 

Updated geopm_region_hint_e to make the enum values sequential, updated all code using mappings involving this enum to use array indexing, instead (where appropriate).


service/src/Helper.cpp
```diff
@@ -210,17 +210,27 @@ namespace geopm
...
+        unsigned int hint_idx = hint_to_index(hint);
+        if (hint_idx >= GEOPM_SENTINEL_REGION_HINT) {
+            throw Exception("Helper::" + std::string(__func__) +
+                            "(): hint out of range: " +
                             geopm::string_format_hex(hint),
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
     }
 
+    unsigned int hint_to_index(uint64_t hint)
```
avilcheslopez: This is perhaps not the best file for this function, but placed it here for now. Wanted to get a draft PR out there for now. Will revisit this in the meantime.

avilcheslopez: A bit of context: I decided to put it there for now to follow the current pattern, as I noticed another hint enum helper function was placed there as well (``check_hint()``).

However, thinking about it a bit more, I would propose these functions (``check_hint()``, ``hint_to_index()``) be moved to ``geopm_hint.h`` for cohesion. The functions in ``Helper.hpp`` are more general-purpose, while the hint functions are directly tied to the hint enum. Also, keeping them together with the enum will make them easier to maintain, and you can reduce header dependencies as code that needs to use the enum will only need to include one header (as opposed to two).

service/src/geopm_hint.h
```diff
@@ -17,14 +17,15 @@ extern "C" {
...
     GEOPM_REGION_HINT_UNKNOWN =   1ULL << 32, /* Region with unknown or varying characteristics */
-    GEOPM_REGION_HINT_COMPUTE =   1ULL << 33, /* Region dominated by compute */
-    GEOPM_REGION_HINT_MEMORY =    1ULL << 34, /* Region dominated by memory access */
-    GEOPM_REGION_HINT_NETWORK =   1ULL << 35, /* Region dominated by network traffic */
-    GEOPM_REGION_HINT_IO =        1ULL << 36, /* Region dominated by disk access */
-    GEOPM_REGION_HINT_SERIAL =    1ULL << 37, /* Single threaded region */
-    GEOPM_REGION_HINT_PARALLEL =  1ULL << 38, /* Region is threaded */
-    GEOPM_REGION_HINT_IGNORE =    1ULL << 39, /* Do not add region time to epoch */
-    GEOPM_REGION_HINT_INACTIVE =  1ULL << 40, /* Used to marke CPUs that are not running application */
+    GEOPM_REGION_HINT_COMPUTE =   2ULL << 32, /* Region dominated by compute */
```
cmcantalupo: I think hint can actually range from 0-N (i.e. I don't think we need the 32 bit shift)? There should no longer be a "hint" embedded in the upper 32 bits of the "region_id" but maybe I have forgotten something.

service/src/Helper.cpp
cmcantalupo: sounds good!

service/src/geopm_hint.h
avilcheslopez: I opted to still do the shifting because it is still being embedded in region_id. See: ``Profile.cpp``. If the plan is to move away from that, I can look into it.

service/src/GEOPMHint.cpp
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#include "geopm_hint.h"
+
+#include <string>
+
+#include "geopm/Helper.hpp"
+#include "geopm/Exception.hpp"
+
+extern "C" {
```
cmcantalupo:
```suggestion
namespace geopm
{
```

service/src/geopm_hint.h
```diff
@@ -15,28 +15,31 @@ extern "C" {
...
+                                GEOPM_REGION_HINT_IGNORE |
+                                GEOPM_REGION_HINT_INACTIVE,
 };
 
+/// @brief Verify a region's hint value is legal for use.
+void check_hint(uint64_t hint);
+
 #ifdef __cplusplus
 }
 #endif
```
cmcantalupo:
```suggestion
#ifdef __cplusplus
}
namespace geopm
{
    /// @brief Verify a region's hint value is legal for use.
    void check_hint(uint64_t hint);
}
#endif
```

cmcantalupo: The sole purpose of check_hint is to throw exceptions, so lets make sure it is only called in C++ contexts.

src/ApplicationStatus.cpp
```diff
@@ -61,10 +60,10 @@ namespace geopm
             throw Exception("ApplicationStatusImp::set_hint(): invalid CPU index: " + std::to_string(cpu_idx),
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        geopm::check_hint(hint);
```
cmcantalupo, line 64: lets leave the namespace in there...other similar changes below.

service/src/geopm_hint.h
```diff
@@ -15,28 +15,31 @@ extern "C" {
...
+    GEOPM_SENTINEL_REGION_HINT,
+    GEOPM_MASK_REGION_HINT =    GEOPM_REGION_HINT_UNKNOWN |
+                                GEOPM_REGION_HINT_COMPUTE |
+                                GEOPM_REGION_HINT_MEMORY |
+                                GEOPM_REGION_HINT_NETWORK |
+                                GEOPM_REGION_HINT_IO |
+                                GEOPM_REGION_HINT_SERIAL |
+                                GEOPM_REGION_HINT_PARALLEL |
+                                GEOPM_REGION_HINT_IGNORE |
+                                GEOPM_REGION_HINT_INACTIVE,
```
cmcantalupo: This mask is no longer valid now that we aren't using bit hits as bitmasks.

src/Profile.hpp
```diff
@@ -53,7 +53,7 @@ static inline uint64_t geopm_region_id_hint(uint64_t region_id)
         ret = GEOPM_REGION_HINT_NETWORK;
     }
     else {
-        ret = region_id & GEOPM_MASK_REGION_HINT;
+        ret = (region_id >> 32) & GEOPM_MASK_REGION_HINT;
```
cmcantalupo: no need for masking anymore.

cmcantalupo (commented):
Mostly pretty good.  Just want to make sure check_hint is only called in C++, and we don't really have a use for the hint mask anymore.

service/src/geopm_hint.h
avilcheslopez: Yep, my bad, forgot to circle back to delete the mask enum.

service/src/geopm_hint.h
avilcheslopez: Yep, it's only being used in C++ contexts.

service/src/GEOPMHint.cpp
avilcheslopez: Considering the header is a C header, were you thinking of converting it to a C++ header, or is it that you just want to put ``check_hint()`` in the geopm namespace?

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #2387, opened 2022-06-21
Skip permission error test when running as root user

### LINKED ISSUE #2382: POSIXSignalTest.sig_queue_EPERM test failing
**Describe the bug**
I tried to run the tests of geopmd and I expected them to pass instead one failed

**GEOPM version**
2.0.0+rc1

**Expected behavior**
All tests passing

**Actual behavior**
1 test is failing

**Error messages produced**
FAIL: test/gtest_links/POSIXSignalTest.sig_queue_EPERM
Was debug messaging was enabled at build time? No

**Additional context**

see https://bugs.gentoo.org/852260
log https://852260.bugs.gentoo.org/attachment.cgi?id=785567

cmcantalupo:
- The test POSIXSignalTest.sig_queue_EPERM sends a
  SIGCONT to a root owned process, and checks that
  an exception is raised.
- This is not a valid test when run as the root user
  so this patch skips the test and warns the user in
  that case.
- Fixes #2382

### DIFF
--- a/service/test/POSIXSignalTest.cpp
+++ b/service/test/POSIXSignalTest.cpp
@@ -201,10 +201,15 @@ TEST_F(POSIXSignalTest, sig_queue_ESRCH)
  */
 TEST_F(POSIXSignalTest, sig_queue_EPERM)
 {
-    std::string errmsg_expect = "Operation not permitted: POSIXSignal(): POSIX signal function call sigqueue() returned an error";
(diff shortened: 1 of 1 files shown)


bgeltz (approved):

Alessandro-Barbieri:
It still fail for me, also I'm runnig the tests as the user `portage`

PR #2404, opened 2022-06-24
Allow resetting/cleaning PlatformIO resources

### LINKED ISSUE #2403: Allow resetting PlatformIO resources
> estimate 1

Parent: #2421 

Done criteria:
- PlatformIO interface has been extended to allow resetting
- Integration test added?
- GEOPM Sampler uses new API to reset PlatformIO

avilcheslopez:
- Fixes #2403

### DIFF
--- a/service/src/PlatformIO.cpp
+++ b/service/src/PlatformIO.cpp
@@ -13,6 +13,7 @@
 #include <cmath>
 #include <iostream>
 #include <tuple>
+#include <memory>
 
 #include "geopm/Agg.hpp"
 #include "geopm/Exception.hpp"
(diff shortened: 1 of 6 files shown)


service/src/PlatformIO.cpp
```diff
@@ -27,10 +28,24 @@
 
 namespace geopm
 {
+
+    static PlatformIO &platform_io_helper(bool do_reset)
+    {
+        static std::unique_ptr<PlatformIOImp> instance;
```
```suggestion
        static std::unique_ptr<PlatformIOImp> instance = geopm::make_unique<PlatformIOImp>();
```

cmcantalupo:
There are documentation changes required for all the PIO interfaces for push_signal(), and push_control().  These have language bindings for C, C++, python.  For all of these there is user documentation and developer documentation (man pages, doxygen strings, and python docstrings).  The C++ interface register_iogroup() should also be updated about the effect of reset.

Some examples that should reflect the effect of reset:

After a reset the stack will be gone, and the user will be able to create a new stack:
https://github.com/geopm/geopm/blob/dev/service/src/geopm_pio.h#L144-L153

Any IOGroup that was registered by a user (outside of the PlatformIO constructor) prior to a reset will need to be registered again:
https://github.com/geopm/geopm/blob/dev/service/src/geopm/PlatformIO.hpp#L36-L44

cmcantalupo (approved):

PR #2416, opened 2022-06-28
Automatic removal of unsupported GPU signals and controls

### LINKED ISSUE #2386: GPU signals are exposed even if they are not supported
**Describe the bug**
I tried to read a signal exposed by a GPU IO Group and I expected it to work instead I received a underlying library error.

**GEOPM version**
1.1.0+dev2160gbf0e92f8

**Expected behavior**
geopmread & geopmwrite should only list signals available to the user

**Actual behavior**
geopmread & geopmwrite for GPU IO Groups will list all possible signals, including those that are not available to a given user.

**Error messages produced**
None, by inspection

**Additional context**
The MSRIOGroup now provides a similar feature for pruning signals that are not valid/available (around the HWP vs legacy P-state switch)

lhlawson:
- Relates to #2386 bug report from github issues
- Fixes #2386 change request from github issues.

This PR checks for LevelZero signal and control availability at boot time and removes them from the available list if they cannot be accessed.

Live tested on a system with basic level zero support, unit tests for checking removal of signals also added.

### DIFF
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -467,6 +467,7 @@ namespace geopm
                         throw;
                     }
                     unsupported_signal_names.push_back(sv.first);
+                    break;
                 }
             }
             sv.second.m_signals = result;
@@ -522,10 +523,9 @@ namespace geopm
                                                            control_domain_type(sv.first), domain_idx);
                             }
                         }
-                        else {
-                            //Try to write the signals
-                            write_control(sv.first, control_domain_type(sv.first), domain_idx, init_setting);
-                        }
+
+                        //Try to write the signals
+                        write_control(sv.first, control_domain_type(sv.first), domain_idx, init_setting);
                     }
                 }
             }
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -521,7 +521,8 @@ namespace geopm
                                 init_setting = read_signal(M_NAME_PREFIX + "GPU_CORE_FREQUENCY_MAX_AVAIL",
                                                            control_domain_type(sv.first), domain_idx);
                             }
-                        } else {
+                        }
+                        else {
                             //Try to write the signals
                             write_control(sv.first, control_domain_type(sv.first), domain_idx, init_setting);
                         }
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -499,6 +499,49 @@ namespace geopm
 
         // Cache the initial min and max frequencies
         save_control();
+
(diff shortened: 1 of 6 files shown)


service/src/LevelZeroIOGroup.cpp
```diff
@@ -484,6 +499,49 @@ namespace geopm
...
+
+                        if(std::isnan(init_setting)) {
+                            // Specialized handling for signals that may be NAN
+                            if(sv.first == "GPU_CORE_FREQUENCY_CONTROL" ||
+                               sv.first == M_NAME_PREFIX + "GPU_CORE_FREQUENCY_CONTROL") {
+
+                                init_setting = read_signal(M_NAME_PREFIX + "GPU_CORE_FREQUENCY_MAX_AVAIL",
+                                                           control_domain_type(sv.first), domain_idx);
+                            }
+                        } else {
```
bgeltz:
```suggestion
                        } 
                        else {
```

bgeltz (changes requested):

bgeltz (dismissed):

service/src/LevelZeroIOGroup.cpp
```diff
@@ -484,6 +499,50 @@ namespace geopm
...
+                               sv.first == M_NAME_PREFIX + "GPU_CORE_FREQUENCY_CONTROL") {
+
+                                init_setting = read_signal(M_NAME_PREFIX + "GPU_CORE_FREQUENCY_MAX_AVAIL",
+                                                           control_domain_type(sv.first), domain_idx);
+                            }
+                        }
+                        else {
+                            //Try to write the signals
+                            write_control(sv.first, control_domain_type(sv.first), domain_idx, init_setting);
+                        }
```
lhlawson:
```suggestion

                            //Try to write the signals
                            write_control(sv.first, control_domain_type(sv.first), domain_idx, init_setting);

```

service/src/NVMLIOGroup.cpp
```diff
@@ -236,26 +275,58 @@ namespace geopm
...
+                         ++domain_idx) {
+                        init_setting = read_signal(sv.first, control_domain_type(sv.first), domain_idx);
+
+                        if(std::isnan(init_setting)) {
+                            // Specialized handling for signals that may be NAN
+                            if (sv.first == M_NAME_PREFIX + "GPU_CORE_FREQUENCY_RESET_CONTROL") {
+                                init_setting = 0;
+                            }
+                        }
+                        else if (init_setting == 0 &&
```
lhlawson: It's unclear to me why the default GPU_CORE_FREQUENCY_CONTROL value is 0 instead of NAN (defined in initializer list), but either way this is handled as a special case by using the max frequency for the write/read check then calling restore_control.

cmcantalupo (approved):

PR #2427, opened 2022-07-01
Make `POSIXSignalTest.siq_queue_EPERM` consider elevated permissions

### LINKED ISSUE #2382: POSIXSignalTest.sig_queue_EPERM test failing

Krebrov001:
Add a function `has_elevated_permissions()` to check if the current user
has elevated permissions by examining the `CapEff` field in the status
file of the running process's proc entry.

- Relates to #2382

### DIFF
--- a/service/test/POSIXSignalTest.cpp
+++ b/service/test/POSIXSignalTest.cpp
@@ -31,7 +31,7 @@ class POSIXSignalTest : public :: testing :: Test
         void SetUp(void);
         void TearDown(void);
     protected:
-        bool has_elevated_permissions(void) const;
+        bool has_cap_kill(void) const;
 
         std::shared_ptr<POSIXSignalImp> m_posix_sig;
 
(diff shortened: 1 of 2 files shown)


service/test/POSIXSignalTest.cpp
```diff
@@ -61,6 +69,40 @@ void POSIXSignalTest::TearDown(void)
     sigprocmask(SIG_SETMASK, &m_backup_sigset, nullptr);
 }
 
+bool POSIXSignalTest::has_elevated_permissions(void) const
+{
+    uint64_t cap = 0;
+    uint64_t cap_sys_admin = 0x00200000;  // TODO: magic number, is there a #define for it?
```
Krebrov001, line 75: magic number, is there a `#define` for it?

service/test/POSIXSignalTest.cpp
```diff
@@ -201,10 +243,13 @@ TEST_F(POSIXSignalTest, sig_queue_ESRCH)
  */
 TEST_F(POSIXSignalTest, sig_queue_EPERM)
 {
-    if (geteuid() == 0) {
+    if (geteuid() == 0) {  // the root user
         std::cerr << "Warning: <geopm> Skipping POSIXSignalTest.sig_queue_EPERM cannot be run by user \"root\"\n";
     }
-    else {
+    else if (has_elevated_permissions()) {  // the non root user with elevated permissions
+        m_posix_sig->sig_queue(1, SIGCONT, 2);  // TODO: how to do EXPECT_NO_THROW?
```
Krebrov001: How to do `EXPECT_NO_THROW`? Like for any other non root user, we used a `GEOPM_EXPECT_THROW_MESSAGE` to say that we are expecting `sig_queue()` to necessarily fail. So if it doesn't throw an exception, the test would fail.

For the non root user with elevated permissions, can we have a similar assertion that excepts the call `sig_queue()` to necessarily **not** fail, or necessarily **not** throw an exception. So if it throws an exception, the test would fail, and if it doesn't throw, the test would pass. How can we do that?

Krebrov001 (commented):
Please answer my questions.

service/test/POSIXSignalTest.cpp
cmcantalupo: If the test throws an unhandled exception, gtest considers it a failure, that is why there isn't an "EXPECT_NO_THROW" macro.  Everything is expected not to throw :)

service/test/POSIXSignalTest.cpp
cmcantalupo: We probably just want to check for CAP_KILL which is 0x20, but perhaps renaming the function as well "has_cap_kill()"

service/test/POSIXSignalTest.cpp
Krebrov001: > If the test throws an unhandled exception, gtest considers it a failure, that is why there isn't an "EXPECT_NO_THROW" macro. Everything is expected not to throw :)

So nothing else needs to be done here?

service/test/POSIXSignalTest.cpp
```diff
@@ -61,6 +69,40 @@ void POSIXSignalTest::TearDown(void)
...
+bool POSIXSignalTest::has_elevated_permissions(void) const
+{
+    uint64_t cap = 0;
+    uint64_t cap_sys_admin = 0x00200000;  // TODO: magic number, is there a #define for it?
+    pid_t pid = getpid();
+
+    std::string status_path = "/proc/" + std::to_string(pid) + "/status";
+
+    // TODO: This is a very memory inefficient algorithm and should be replaced some day.
+    try {
```
cmcantalupo: Why the catch all?  Under what conditions would we expect this function to throw?

cmcantalupo (changes requested):
Looks pretty good.  Just a couple of comments.

cmcantalupo (approved):
LGTM

Issue #1766, opened 2021-07-23
record.cpp exception with unsupported event type when using --geopm-trace-profile


fkeceli:
Seems like following files are missing the new EVENT_SHORT_REGION:

- record.cpp
- ProfileTracerImp.hpp
- ProfileTracer.cpp

fkeceli:
For now fixing record.cpp is the most essential part.

Note that there is future work to be done per:

https://github.com/geopm/geopm/blob/120209a29cf50e531adb67ddfd785aac84a4e3a2/src/record.hpp#L63

I am not sure if there is already an issue opened for this.

PR #2433, opened 2022-07-08
Add support for `EVENT_SHORT_REGION`

Krebrov001:
- Relates to #1766

### DIFF
--- a/src/ProfileTracer.cpp
+++ b/src/ProfileTracer.cpp
@@ -24,6 +24,9 @@
 
 namespace geopm
 {
+    // defintion of the static data member
+    ApplicationSampler* ProfileTracerImp::m_application_sampler = nullptr;
+
     ProfileTracerImp::ProfileTracerImp(const std::string &start_time)
         : ProfileTracerImp(start_time,
                            1024 * 1024,
--- a/src/ProfileTracer.cpp
+++ b/src/ProfileTracer.cpp
@@ -79,8 +79,7 @@ namespace geopm
                     result = string_format_integer(value);
                     break;
                 case EVENT_SHORT_REGION:
-                    // integer value of number of short regions
-                    result = string_format_integer(value);
+                    result = string_format_hex(value);
                     break;
                 default:
                     result = "INVALID";
--- a/src/ProfileTracer.cpp
+++ b/src/ProfileTracer.cpp
@@ -78,6 +78,10 @@ namespace geopm
                 case EVENT_EPOCH_COUNT:
                     result = string_format_integer(value);
                     break;
+                case EVENT_SHORT_REGION:
+                    // integer value of number of short regions
+                    result = string_format_integer(value);
+                    break;
                 default:
                     result = "INVALID";
                     GEOPM_DEBUG_ASSERT(false, "ProfileTracer::event_format(): event out of range");
--- a/src/ProfileTracer.cpp
+++ b/src/ProfileTracer.cpp
@@ -29,7 +29,8 @@ namespace geopm
                            1024 * 1024,
                            environment().do_trace_profile(),
                            environment().trace_profile(),
-                           hostname())
+                           hostname(),
+                           ApplicationSampler::application_sampler())
     {
 
     }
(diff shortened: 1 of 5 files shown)


src/ProfileTracer.cpp
```diff
@@ -78,6 +78,10 @@ namespace geopm
                 case EVENT_EPOCH_COUNT:
                     result = string_format_integer(value);
                     break;
+                case EVENT_SHORT_REGION:
+                    // integer value of number of short regions
+                    result = string_format_integer(value);
```
lhlawson:
```suggestion
                    result = string_format_hex(value);
```
This should provide the hex value of region as it is written.  
We may want to add an EVENT_SHORT_REGION count (which would use the string_format_integer), but for #1766 only the region addition is needed.

lhlawson:
We will want to run this against a series of workloads to see if we can hit the case called out in #1766 where an exception would occur.  We've confirmed that without this fix PENNANT still encounters the error:
`Error: <geopm> Invalid argument: unsupported event type: 3: at src/record.cpp:28`
when run using --geopm-profile-trace

If this fix does not resolve that error further work is needed.  Otherwise LGTM.

src/ProfileTracerImp.hpp
```diff
@@ -36,7 +37,13 @@ namespace geopm
             };
             bool m_is_trace_enabled;
             std::unique_ptr<CSV> m_csv;
+            static ApplicationSampler* m_application_sampler;
+            static std::string event_format(double value);
     };
+
+    // defintion of the static data member
```
cmcantalupo: Please move to ProfileTracer.cpp otherwise it could potentially end up in multiple compilation units.

test/ProfileTracerTest.cpp
```diff
@@ -44,6 +46,9 @@ void ProfileTracerTest::SetUp(void)
         m_data.push_back({time, rank, event, region_hash});
         time += 1.0;
     }
+
+     m_data.push_back({40, 0, geopm::EVENT_SHORT_REGION, 0xdeadbeef});
```
cmcantalupo:
```suggestion
     m_data.push_back({40, 0, geopm::EVENT_SHORT_REGION, 88});
```
in general the event number will not match the region hash

test/ProfileTracerTest.cpp
```diff
@@ -44,6 +46,9 @@ void ProfileTracerTest::SetUp(void)
...
+     m_data.push_back({41, 1, geopm::EVENT_EPOCH_COUNT, region_hash});
```
cmcantalupo:
```suggestion
     m_data.push_back({41, 1, geopm::EVENT_EPOCH_COUNT, 1});
```
The epoch count event should reflect a number times the epoch event has occurred.

test/ProfileTracerTest.cpp
```diff
@@ -60,10 +65,17 @@ TEST_F(ProfileTracerTest, construct_update_destruct)
 
 TEST_F(ProfileTracerTest, format)
 {
+    EXPECT_CALL(m_application_sampler, get_short_region(0xdeadbeef))
```
cmcantalupo:
```suggestion
    EXPECT_CALL(m_application_sampler, get_short_region(88))
```

test/ProfileTracerTest.cpp
```diff
@@ -77,6 +89,8 @@ TEST_F(ProfileTracerTest, format)
         "35|2|REGION_EXIT|0xfa5920d6",
         "36|1|REGION_EXIT|0xfa5920d6",
         "37|0|REGION_EXIT|0xfa5920d6",
+        "40|0|EVENT_SHORT_REGION|0xdeadbeef",
+        "41|1|EPOCH_COUNT|4200145110"
```
cmcantalupo:
```suggestion
        "41|1|EPOCH_COUNT|1"
```

cmcantalupo (changes requested):
Just a couple of minor comments.

Looks like some issue with clang 12:
```
 FAIL: test/gtest_links/ProfileTracerTest.construct_update_destruct
 ```
 showed up in CI

cmcantalupo (approved):

PR #2437, opened 2022-07-08
CPU-CA update to use per node characterization information from PlatformCharacterizationIOGroup

### LINKED ISSUE #2436: Modify the CPU-CA to use the ConstConfigIOGroup
> estimate 1
Parent feature: #2698 

The implementation for the CPU-CA feature was modified because the policy values may be acquired from the ConstConfigIOGroup on a per node basis

Done criteria:

- [x] Documentation changes since this change does
      modify interfaces or features.

lhlawson:
- Relates to #2436 change request from github issues & #2364 feature request from github issues
- Fixes #2436 change request from github issues.

The CPU-CA has been updated to use the characterization signals provided by the PlatformCharacterization IOGroup.  
If the user provides a policy with NAN for the values required by the CPU-CA, it will query the PlatformCharacterization IOGroup to determine if per node values are available.  If so these are used.  

If the user provides a policy with non-NAN values they are treated as a job level override and the per-node values available from the IOGroup will not be used.  

CPU-CA policy recommendation scripts have been updated to provide both a policy value (unchanged from #2364) and the geopmwrite commands that may be used to write the associated signals for the node characterized.  

Future tasks:
- [ ] Creation of integration tests that handle end to end characterization: Run AIB, parse AIB with existing policy recommendation scripts, write PlatCharIO values from recommendation script


lhlawson:
Superseded by #2692

PR #2452, opened 2022-07-14
Make SST::*ENABLE controls trigger prerequisite controls

### LINKED ISSUE #2451: SST control dependencies might not work in save/restore depending on initial state
**Describe the bug**
`SST::TURBO_ENABLE:ENABLE` and `SST::COREPRIORITY_ENABLE:ENABLE` controls have cross-dependencies. If a user attempts to enable SST-TF while SST-CP is disabled, or to disable SST-CP while SST-TF is enabled, then the request will be rejected by the isst driver. This can cause problems in save/restore because there isn't a guarantee about the order of restore operations that checks the value that is being restored.

**GEOPM version**
cacfa817b

**Expected behavior**
GEOPM should run without errors with an agent that modifies SST state.

**Actual behavior**
GEOPM runs the application and correctly executes the agent, but throws an exception while restoring controls.

**Error messages produced**
This may manifest at the end of even a successful application run once the restore logic is triggered
```
Error: <geopm> No such device or address: sstioimp::write_mbox_once() mbox write failed: at src/SSTIO.cpp:431
  what():  <geopm> No such device or address: sstioimp::write_mbox_once() mbox write failed: at src/SSTIO.cpp:431
```

dannosliwcd:
- `SST::TURBO_ENABLE:ENABLE` and `SST::COREPRIORITY_ENABLE:ENABLE` controls
  have cross-dependencies. If a user attempts to enable SST-TF while
  SST-CP is disabled, or to disable SST-CP while SST-TF is enabled, then
  the request will be rejected by the isst driver.
- This change makes turbo enable imply corepriority enable, and makes
  corepriority disable imply turbo disable for writes.
- This change removes batch write capabilities for SST::*ENABLE controls.
- Resolves #2451

### DIFF
--- a/service/docs/source/geopm_pio_sst.7.rst
+++ b/service/docs/source/geopm_pio_sst.7.rst
@@ -174,7 +174,7 @@ Controls
 --------
 
 ``SST::TURBO_ENABLE``
-    Enable SST-TF feature.
+    Enable SST-TF feature. Enabling SST-TF also causes SST-CP to be enabled.
 
     * **Aggregation**: select_first
     * **Domain**: package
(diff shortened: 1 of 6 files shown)


cmcantalupo (approved):
Fancy use of the weak_ptr.

PR #2462, opened 2022-07-18
Set missing MSR aggregation functions

### LINKED ISSUE #1126: Add aggregation functions for DRAM signals, and other uncommon signals
> estimate 1

For example, DRAM min/max/TDP limits do not use "sum" the way the package limit signals do.

avilcheslopez:

- Fixes #1126

### DIFF
--- a/service/src/msr_data_arch.json
+++ b/service/src/msr_data_arch.json
@@ -1304,8 +1304,8 @@
                     "scalar":    1.0,
                     "behavior":  "label",
                     "writeable": true,
-                    "description": "Write 1 to clear the global status bit for PMC0 overflow. When reading at a higher level domain than its native domain, it aggregates as the count of all such bits that have been set.",
-                    "aggregation": "sum"
+                    "description": "Write 1 to clear the global status bit for PMC0 overflow.",
+                    "aggregation": "expect_same"
                 },
                 "CLEAR_OVF_PMC1": {
                     "begin_bit": 1,
@@ -1315,8 +1315,8 @@
                     "scalar":    1.0,
                     "behavior":  "label",
                     "writeable": true,
-                    "description": "Write 1 to clear the global status bit for PMC1 overflow. When reading at a higher level domain than its native domain, it aggregates as the count of all such bits that have been set.",
-                    "aggregation": "sum"
+                    "description": "Write 1 to clear the global status bit for PMC1 overflow.",
+                    "aggregation": "expect_same"
                 },
                 "CLEAR_OVF_PMC2": {
                     "begin_bit": 1,
@@ -1326,8 +1326,8 @@
                     "scalar":    1.0,
                     "behavior":  "label",
                     "writeable": true,
-                    "description": "Write 1 to clear the global status bit for PMC2 overflow. When reading at a higher level domain than its native domain, it aggregates as the count of all such bits that have been set.",
-                    "aggregation": "sum"
+                    "description": "Write 1 to clear the global status bit for PMC2 overflow.",
+                    "aggregation": "expect_same"
                 },
                 "CLEAR_OVF_PMC3": {
                     "begin_bit": 1,
@@ -1337,8 +1337,8 @@
                     "scalar":    1.0,
                     "behavior":  "label",
                     "writeable": true,
-                    "description": "Write 1 to clear the global status bit for PMC3 overflow. When reading at a higher level domain than its native domain, it aggregates as the count of all such bits that have been set.",
-                    "aggregation": "sum"
+                    "description": "Write 1 to clear the global status bit for PMC3 overflow.",
+                    "aggregation": "expect_same"
                 },
                 "CLEAR_OVF_FIXED_CTR0": {
                     "begin_bit": 32,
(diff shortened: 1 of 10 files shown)


dannosliwcd:
FYI (as far as I can tell, your work causes no conflicts / required changes -- just pointing out since it's related):


Here's where the rst documentation gets the aggregation type from these json files.
https://github.com/geopm/geopm/blob/561c59c7698b640c1d929fe53c09eabf57618032/service/docs/source/_ext/geopm_rst_extensions.py#L133-L138

You could optionally add something like below if you want to ensure future modifiers of these json files fail the documentation build if they forget to add an aggregation type:

```python
try:
    aggregation_type = msr_field_data['aggregation']
except KeyError:
    aggregation_type = 'select_first'
    logger.error('Missing an aggregation function for %s in %s',
                 geopm_msr_name,
                 json_path)
...
description_items = [
                    ('Aggregation', aggregation_type),
                    ...
```

avilcheslopez:
I think this is a great idea and we should incorporate it as part of this PR. I have this implemented locally and tested it out. I can add it to the PR once we transition it from draft.

However, I'm trying to think if there's a more opportune place where we should make the build fail if one of these MSR JSON files is misconfigured. Besides checking for missing fields, I'd like to validate the file against the schema. I'll think about this for a bit...

service/src/msr_data_arch.json
```diff
@@ -1209,7 +1304,8 @@
                     "scalar":    1.0,
                     "behavior":  "label",
                     "writeable": true,
-                    "description": "Write 1 to clear the global status bit for PMC0 overflow."
+                    "description": "Write 1 to clear the global status bit for PMC0 overflow. When reading at a higher level domain than its native domain, it aggregates as the count of all such bits that have been set.",
```
dannosliwcd: These types of controls (names like `CLEAR_...` or descriptions like `Write 1 to <do xyz>` typically act like a trigger signal and have no meaningful read-back value.

This reads back as 0 when I try it on our server right after writing 1 to it. The SDM only specifies the write behavior of this MSR, so I think we should not add any specific read expectations to it.

service/src/msr_data_skx.json
```diff
@@ -762,7 +815,7 @@
                     "behavior":  "variable",
                     "aggregation": "average",
                     "writeable": true,
-                    "description": "Disable for the L2 hardware prefetcher."
+                    "description": "Disable for the L2 hardware prefetcher. When reading at a higher level domain than its native domain, it aggregates as the count of all such bits that have been set."
```
dannosliwcd, line 818: Some of these don't have the explicit `expect_same` type. Not a functional problem, just checking whether that's intentional.

src/PowerBalancerAgent.cpp
```diff
@@ -749,9 +749,7 @@ namespace geopm
             throw Exception("PowerBalancerAgent::enforce_policy(): policy vector incorrectly sized.",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        int control_domain = m_platform_io.control_domain_type("CPU_POWER_LIMIT_CONTROL");
-        double pkg_policy = policy[M_POLICY_CPU_POWER_LIMIT] / m_platform_topo.num_domain(control_domain);
-        m_platform_io.write_control("CPU_POWER_LIMIT_CONTROL", GEOPM_DOMAIN_BOARD, 0, pkg_policy);
+        m_platform_io.write_control("CPU_POWER_LIMIT_CONTROL", GEOPM_DOMAIN_BOARD, 0, policy[M_POLICY_CPU_POWER_LIMIT]);
```
dannosliwcd, line 752: Should this be part of #2479?

avilcheslopez: Nah. Chris held off on doing it there and opted to wait for the aggregation function change for these power settings to incorporate these changes.

service/src/msr_data_arch.json
avilcheslopez: I see. Hmmm, that being the case, I'm thinking "expect_same" makes sense for these. That would avoid any confusion from the inconsistent behavior that would result with the "sum" aggregation function and would also get rid of the added "reading" note in the description.

Does this make sense?

service/src/msr_data_skx.json
avilcheslopez: Nope! This was set by mistake by my script. Thanks for catching this!

avilcheslopez: Actually, why were these set as "average"? It seems like we can set them to "sum" (as we're doing with the other bool fields).

dannosliwcd: Yeah, this should work like those other indicators using sum.

dannosliwcd (approved):
LGTM. You mentioned you wanted to add aggregators to the skx ones that have no explicit aggregation. I think those are still missing, but either way is fine since the default aggregation matches what you wanted for those ones.

dannosliwcd:
To correct my previous comment: all agg functions do appear to be present.

cmcantalupo (approved):

PR #2490, opened 2022-07-28
Retry SST operations when they fail due to a busy mailbox

### LINKED ISSUE #2489: Some SSTIOGroup signals/controls throw exceptions for transient errors
**Describe the bug**
Some signals and controls in SSTIOGroup go through the OS/BIOS mailbox. If the mailbox is busy during a read/write, the driver sets the EBUSY error code. It is generally sufficient to retry at the IOGroup layer as soon as the busy code is read back, and the access conflict will be over, so the IOGroup should attempt to retry once before failing.

**GEOPM version**
61e413c

**Expected behavior**
SSTIOGroup writes and reads either succeed or are able to fail in a non-fatal way.

**Actual behavior**
An agent that performs a lot of SSTIOGroup reads/writes. The GEOPM controller occasionally fails due to a busy mailbox.

**Error messages produced**
`Error: <geopm> Device or resource busy: SSTIOImp::read_batch() mbox read failed: at src/SSTIO.cpp:248`

dannosliwcd:
- Retry mbox operations one time if EBUSY is returned from the operation. If a failure happens during the retry, then throw an exception.
- Resolves #2489

### DIFF
--- a/service/src/SSTIO.cpp
+++ b/service/src/SSTIO.cpp
@@ -242,8 +242,10 @@ namespace geopm
                 m_mbox_read_interfaces);
 
             for (auto &batch : m_mbox_read_batch) {
+                errno = 0;
                 int err = m_ioctl->mbox(batch.get());
                 if (err == -1 && errno == EBUSY) {
+                    errno = 0;
                     err = m_ioctl->mbox(batch.get());
                 }
                 if (err == -1) {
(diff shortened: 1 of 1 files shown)


service/src/SSTIO.cpp
```diff
@@ -243,6 +243,9 @@ namespace geopm
 
             for (auto &batch : m_mbox_read_batch) {
                 int err = m_ioctl->mbox(batch.get());
+                if (err == -1 && errno == EBUSY) {
+                    err = m_ioctl->mbox(batch.get());
```
cmcantalupo:
```suggestion
                if (err == -1 && errno == EBUSY) {
                    errno = 0;
                    err = m_ioctl->mbox(batch.get());
```

cmcantalupo: This change should be made throughout.  If errno is handled, it should be reset.

cmcantalupo (changes requested):
Should reset errno when the error is handled.

cmcantalupo (approved):

Issue #2356, opened 2022-05-26
Batch server process has signal handlers put in place by geopmd after the fork that do not terminate with SIGTERM


cmcantalupo:
**Describe the bug**
I tried to send SIGTERM to the batch server process and I expected the process to terminate instead the parent process (geopmd) was sent a SIGCHLD.  This is due to some signal handing that is inserted by our software.  This should not only send a signal to the parent process, but also terminate.  Currently it only sends the SIGCHILD, but then continues to run.

**GEOPM version**
Output from `geopmread --version`
1017903

**Expected behavior**
The forked process should reset all signal handlers to the defaults after the fork occurs.

**Actual behavior**
Signal handlers for geopmd prevent the termination of a batch server process.

**Error messages produced**
```
May 26 15:09:47 mcfly1 geopmd[60055]: The call io.github.geopm.PlatformStartBatch has failed with an exception:
May 26 15:09:47 mcfly1 geopmd[60055]: Traceback (most recent call last):
May 26 15:09:47 mcfly1 geopmd[60055]:   File "/usr/lib/python3.6/site-packages/dasbus/server/handler.py", line 459, in _method_callback
May 26 15:09:47 mcfly1 geopmd[60055]:     **additional_args
May 26 15:09:47 mcfly1 geopmd[60055]:   File "/usr/lib/python3.6/site-packages/dasbus/server/handler.py", line 265, in _handle_call
May 26 15:09:47 mcfly1 geopmd[60055]:     return handler(*parameters, **additional_args)
May 26 15:09:47 mcfly1 geopmd[60055]:   File "/usr/lib/python3.6/site-packages/geopmdpy/service.py", line 793, in PlatformStartBatch
May 26 15:09:47 mcfly1 geopmd[60055]:     return self._platform.start_batch(self._get_pid(**call_info), signal_config, control_config)
May 26 15:09:47 mcfly1 geopmd[60055]:   File "/usr/lib/python3.6/site-packages/geopmdpy/service.py", line 519, in start_batch
May 26 15:09:47 mcfly1 geopmd[60055]:     batch_pid , batch_key = self._pio.start_batch_server(client_pid, signal_config, control_config)
May 26 15:09:47 mcfly1 geopmd[60055]:   File "/usr/lib/python3.6/site-packages/geopmdpy/pio.py", line 631, in start_batch_server
[...]

PR #2494, opened 2022-07-28
Add session info printing to geopmaccess

bgeltz:
- Fix issues stemming from #2356.
- Add `geopmaccess -s` option to print information about active sessions.
- Fix integration tests that are related to killing the batch session (user owned) and the handling of the batch server (root owned).


cmcantalupo (dismissed):

service/geopmdpy/access.py
```diff
@@ -314,9 +314,24 @@ def run(self, is_write, is_all, is_control, group, is_default, is_delete,
...
+                with open(file_path) as fid:
+                    # Early return if session info is requested
+                    return fid.read()
+            except PermissionError as ex:
+                sys.stderr.write(f'Error: {ex}\n\n')
+                return
+            except FileNotFoundError as ex:
+                sys.stderr.write(f'Error: {session_pid} is invalid\n\n')
+                return
+
```
cmcantalupo:
```suggestion
            file_path = f'/run/geopm-service/session-{session_pid}.json'
            result = None
            try:
                with open(file_path) as fid:
                    result = fid.read()
            except PermissionError as ex:
                sys.stderr.write(f'Error: {ex}\n\n')
            except FileNotFoundError as ex:
                sys.stderr.write(f'Error: {session_pid} is invalid\n\n
            # Early return if session info is requested
            return result

```

cmcantalupo: maybe just one early return :)

cmcantalupo (approved):

PR #2495, opened 2022-07-29
Add proper support for controls like power that are summed

### LINKED ISSUE #2479: As a user of PlatformIO, when writing controls configured with a sum aggregation function at a higher domain than their native domain, I would like the value to be distributed evenly across the sub elements
**What problem does the story solve? Please describe.**
For example, when setting power at the board level, we would like for that setting to be distributed evenly across the board packages such that the sum of the power setting for all packages equals the desired power set at the board level.

**Describe the tasks that will be required to complete the story**
TODO

With the current behavior, when setting a control at a higher domain the same value is simply taken and applied to all sub elements in the lower level domain. This behavior makes sense for controls that have an average aggregation function, but not for controls for which a sum is applied. The current interface requires the user of PlatformIO to figure out the value to apply at the higher level to achieve the desired total among all the sub elements, which makes the interface less intuitive/easy to use.

**Additional context**

cmcantalupo:
- Fixes #2479

### DIFF
--- a/service/src/CombinedControl.hpp
+++ b/service/src/CombinedControl.hpp
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef COMBINEDCONTROL_HPP_INCLUDE
+#define COMBINEDCONTROL_HPP_INCLUDE
+
+
+namespace geopm
+{
+    /// @brief Used by PlatformIO to define a control as a function of
+    ///        other controls.
+    class CombinedControl
+    {
+        public:
+            CombinedControl();
+            CombinedControl(double factor);
+            virtual ~CombinedControl() = default;
+            /// @brief Multiply setting by factor and return
+            virtual double adjust(double setting);
+        private:
+            double m_factor;
+    };
+}
+
+#endif
--- a/service/src/PlatformIO.cpp
+++ b/service/src/PlatformIO.cpp
@@ -23,6 +23,7 @@
 
 #include "geopm_pio.h"
 #include "BatchServer.hpp"
+#include "CombinedControl.hpp"
 #include "CombinedSignal.hpp"
 #include "ServiceIOGroup.hpp"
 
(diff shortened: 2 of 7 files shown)


service/src/PlatformIOImp.hpp
```diff
@@ -117,6 +125,8 @@ namespace geopm
             std::vector<std::shared_ptr<IOGroup> > find_signal_iogroup(const std::string &signal_name) const;
             /// @brief Look up the IOGroup that provides the given control.
             std::vector<std::shared_ptr<IOGroup> > find_control_iogroup(const std::string &control_name) const;
+            bool is_control_split_uniform(const std::string &control_name) const;
+            void adjust_combined(int control_idx, double setting);
```
cmcantalupo: If we are going to reorder the vtable, why not put adjust_combined next to sample_combined.

service/src/PlatformIO.cpp
```diff
@@ -475,6 +492,21 @@ namespace geopm
...
 
+    int PlatformIOImp::push_combined_control(const std::string &control_name,
+                                             int domain_type,
+                                             int domain_idx,
+                                             const std::vector<int> &sub_control_idx)
+    {
+        int result = m_active_control.size();
+        double factor = 1.0;
+        if (!is_control_split_uniform(control_name)) {
+            factor = 1.0 / sub_control_idx.size();
```
cmcantalupo, line 504: divide by zero check required.

service/src/PlatformIO.cpp
```diff
@@ -797,6 +843,9 @@ namespace geopm
         if (m_platform_topo.is_nested_domain(base_domain_type, domain_type)) {
             std::set<int> base_domain_idx = m_platform_topo.domain_nested(base_domain_type,
                                                                           domain_type, domain_idx);
+            if (!is_control_split_uniform(control_name)) {
+                setting /= base_domain_idx.size();
```
cmcantalupo, line 849: divide by zero check

service/src/CombinedControl.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// @brief Used by PlatformIO to define a control as a function of
+    ///        other controls.
+    class CombinedControl
+    {
+        public:
+            CombinedControl();
+            CombinedControl(double factor);
+            virtual ~CombinedControl() = default;
+            /// @brief Sample all required controls and aggregate
```
cmcantalupo: Copy paste error, also dox strings should be added.

bgeltz (approved):

PR #2518, opened 2022-08-05
Created a new integration `test_su_term_batch_write`

### LINKED ISSUE #2510: Improve integration testing for batch server when used to write
> estimate 1

Parent: #2014

Create a test that combines the test_batch_term with the check_write_session implementation so that the batch server is used to write, and the client of the batch server is killed.  Check that the save restore of the control settings happens in this case.

Krebrov001:
This test:
 - Prior to setting the control to a custom value,
   check that it is != to that custom value, and
   save the old value of the control.
 - Run a python script to overwrite the old value
   of the control with a custom value
 - Wait for the session to terminate.
 - Check that the control retains it's old value
   after the session has shut down (ie. the batch
   server restores the values of all signals and
   controls when it shuts down).

- Relates to #2510
- Fixes #2510

### DIFF
--- a/service/integration/test/test_su_term_batch_write.sh
+++ b/service/integration/test/test_su_term_batch_write.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+#  Copyright (c) 2015 - 2022, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+if [[ $# -gt 0 ]] && [[ $1 == '--help' ]]; then
+    echo "
+    Terminating Batch Server
+    ------------------------
+
+    As a user, send term signal to a geopmsession process owned by the
+    test user, and make sure that server (owned by root) shuts down
+    cleanly.
+
+"
+    exit 0
+fi
+
+# PARAMETERS
+CONTROL=SERVICE::MSR::PERF_CTL:FREQ
+STICKER=CPU_FREQUENCY_STICKER
+DOMAIN=core
+DOMAIN_IDX=0
+REQUEST="${CONTROL} ${DOMAIN} ${DOMAIN_IDX}"
+REQUEST_STICKER="${STICKER} board ${DOMAIN_IDX}"
+
+test_error() {
+    echo "Error: $1" 1>&2
+    exit -1
+}
+
+TEST_DIR=$(dirname $(readlink -f $0))
+TEST_USER=test
+
+set -x
+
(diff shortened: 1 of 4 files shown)


cmcantalupo (dismissed):

service/integration/test/test_su_term_batch_write.sh
```diff
@@ -0,0 +1,99 @@
...
+
+if [[ $# -gt 0 ]] && [[ $1 == '--help' ]]; then
+    echo "
+    Terminating Batch Server
+    ------------------------
+
+    As a user, send term signal to a geopmsession process owned by the
+    test user, and make sure that server (owned by root) shuts down
+    cleanly.
+
```
cmcantalupo: geopmsession is not used actually, this is a copy paste error

service/integration/test/test_su_term_batch_write.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+
+if [[ $# -gt 0 ]] && [[ $1 == '--help' ]]; then
+    echo "
+    Terminating Batch Server
+    ------------------------
+
+    As a user, send term signal to a geopmsession process owned by the
+    test user, and make sure that server (owned by root) shuts down
+    cleanly.
```
cmcantalupo:
```suggestion
    As a user, send term signal a python script that creates a writing batch server owned by the
    test user, and make sure that server (owned by root) shuts down cleanly.
```

service/integration/test/test_su_term_batch_write.sh
```diff
@@ -0,0 +1,98 @@
...
+    test_error "Writing the new value has failed"
+
+# TERMINATE THE CLIENT PROCESS, VERIFY THAT THE BATCH SERVER IS GONE TOO
+sleep 5
+# geopmaccess returns a JSON object containing the PID of the batch server
+BATCH_PID=$(sudo geopmaccess -s ${SESSION_PID} | \
+            python3 -c 'import sys,json; print(json.loads(sys.stdin.read())["batch_server"])')
+
+# BATCH_PID=$(sudo geopmaccess -s ${SESSION_PID} | \
+#             python3 -c 'import sys; print(sys.stdin.read())')
```
bgeltz:
```suggestion
```

Needed?

bgeltz (dismissed):

bgeltz (approved):

PR #2519, opened 2022-08-05
Output from clang-format based on Daniel's style file

cmcantalupo:


examples/geopm_platform_supported.cpp
```diff
@@ -3,62 +3,61 @@
...
     else {
         int cpu_id = geopm_read_cpuid();
-        if (cpu_id != 0x63F &&
-            cpu_id != 0x63F &&
-            cpu_id != 0x63E &&
-            cpu_id != 0x62D &&
-            cpu_id != 0x64F &&
-            cpu_id != 0x657) {
+        if (cpu_id != 0x63F && cpu_id != 0x63F && cpu_id != 0x63E &&
+            cpu_id != 0x62D && cpu_id != 0x64F && cpu_id != 0x657) {
```
bgeltz: I liked it the old way personally.  it made maintenance easy.

examples/geopmhash.c
```diff
@@ -3,19 +3,19 @@
...
-                        "    this is useful for determining the region name\n"
-                        "    from the region_id printed in the trace.\n";
-    if (argc == 1 ||
-        strncmp("-h", argv[1], strlen("-h")) == 0 ||
+    const char *usage =
+        "%s string\n\n"
+        "    Returns the geopm_hash() of the input string.\n"
+        "    this is useful for determining the region name\n"
+        "    from the region_id printed in the trace.\n";
+    if (argc == 1 || strncmp("-h", argv[1], strlen("-h")) == 0 ||
```
bgeltz: Not a fan of the unindent.

examples/print_affinity.cpp
```diff
@@ -53,22 +53,22 @@ int main(int argc, char** argv)
...
-        if (i) {
-            thread_list << ",";
+        {
+            if (i) {
+                thread_list << ",";
+            }
+            thread_list << cpu;
         }
-        thread_list << cpu;
-}
```
bgeltz: At least the `#pragma` is in the right spot now.  The curly braces are erroneously indented imo

examples/print_affinity.cpp
```diff
@@ -53,22 +53,22 @@ int main(int argc, char** argv)
...
         if (rank == world_rank) {
             // Print off affinity message
-            std::cout << "Rank: " << world_rank
-                      << " cgroup CPUs: [" << cpu_list.str() << "]"
-                      << " omp CPUs: [" << thread_list.str() << "]"
-                      << std::endl << std::flush;
+            std::cout << "Rank: " << world_rank << " cgroup CPUs: ["
+                      << cpu_list.str() << "]"
+                      << " omp CPUs: [" << thread_list.str() << "]" << std::endl
+                      << std::flush;
```
bgeltz: I liked it better the way it was.  This makes it difficult to see what's being printed a bit.  Maintenance issue?

examples/simple_pio_example.c
```diff
@@ -33,33 +32,28 @@ if [ ! -e $HOME/build/geopm/lib/libgeopmpolicy.so ]; then
     make install
     cd ..
 fi
-gcc -I$HOME/build/geopm/include -L$HOME/build/geopm/lib -lgeopmpolicy simple_pio_example.c
-LD_LIBRARY_PATH=$HOME/build/geopm/lib:$LD_LIBRARY_PATH ./a.out
-# Total energy for package 0: 518.16 (joules)
-# END SCRIPT
+gcc -I$HOME/build/geopm/include -L$HOME/build/geopm/lib -lgeopmpolicy
+simple_pio_example.c LD_LIBRARY_PATH=$HOME/build/geopm/lib:$LD_LIBRARY_PATH
+./a.out # Total energy for package 0: 518.16 (joules) # END SCRIPT
```
bgeltz: I think those comments were on a line by themself for a reason.  It's hard to see now.

examples/geopm_platform_supported.cpp
dannosliwcd: Maybe related this and other line-breaking-related comments:

Setting `ColumnLimit` to 0 will tell `clang-format` to "respect the input’s line breaking decisions within statements unless they contradict other rules." Otherwise, messing with the weights of some of the line-breaking penalties might help get closer to the desired behavior. The penalties allow us to prioritize how a line is wrapped when wrapping is needed.

Maybe that'll resolve some of these cases.

examples/geopmhash.c
dannosliwcd: Setting `AlwaysBreakBeforeMultilineStrings` to false may help achieve the desired behavior.

examples/geopm_platform_supported.cpp
cmcantalupo: I'm not sure I'm willing to do away with the column limit feature of clang for this particular use case.  The output looks okay to me.

examples/geopmhash.c
cmcantalupo: I have updated the rule, in this case I think we are still bound by the 80 char limit, which I'm okay with.

examples/simple_pio_example.c
cmcantalupo: fixed with special comment

examples/print_affinity.cpp
cmcantalupo: eh, it isn't that easy to read either way

cmcantalupo:
Maybe we want to bring in the .clang-format file, and document that future changes which conform to this style will not be rejected due to style issues?  For now sifting through 10k lines of changes is out of scope for the v2.0.0 release.

.clang-format
```diff
@@ -0,0 +1,128 @@
...
+  SplitEmptyNamespace: true
+IndentWidth:     4
+UseTab:          Never
+BinPackArguments:  true
+BinPackParameters: true
+ExperimentalAutoDetectBinPacking: false
+# IncludeIsMainRegex determines the first group. The rest of the groups are
+# ordered by increasing priority. They are processed in the order they are
+# listed below (it may be different from the priority order)
+SortIncludes:    true
```
dannosliwcd, line 30:
```suggestion
IncludeBlocks: Regroup
SortIncludes:    true
```
Automatically groups each section of `#include` statements with an empty line between sections

PR #2529, opened 2022-08-09
Fixup RawReportCollection handling of reports with no regions

### LINKED ISSUE #1517: Invalid report when application has no regions
When an application has no regions (user marked, MPI, or otherwise) the `Regions:` section of the report contains nothing.  It should be an empty list, e.g. `Regions: [] `.  The report parsing and caching in io.py also needs to be updated to support an empty regions dataframe.

A workaround is to add at least one MPI call to any application with no regions.

See #1513

bgeltz:
- Fixes #1517.


cmcantalupo (approved):
LGTM

PR #2531, opened 2022-08-09
Add a pre-commit hook that performs GEOPM license checks.

### LINKED ISSUE #2530: As a GEOPM contributor I would like simple/fast CI failure reasons to be available as pre-commit hooks
We have GitHub Actions that prevent merging a PR if they fail. These actions typically take around 5-10 minutes to indicate common failure reasons. Some of those reasons, such as license header checks, can be detected very quickly. It would be nice if some of the common and fast-to-check errors were available as pre-commit hooks.

I would like pre-commit hooks that apply GEOPM-specific checks I would like these to be easily configurable. That is, if multiple checks are present, it should be possible to enable/disable them independently. E.g., that can be managed by through git config flags (demonstrated in the official git samples for pre-commit hooks). Alternatively, we could define hooks that are compatible with the [pre-commit framework](https://pre-commit.com/#repository-local-hooks).

Ideas for pre-commit checks that could quickly detect test or review issues:

- Check the header include order. Reject or auto-fix commits with the wrong order.
- Check for license headers / license manifests. Reject commits that fail the license checks.
- Check for whitespace issues (trailing whitespace, mixed tabs/spaces, missing final line)
- Check for generally-valid YAML/JSON files
- Check that specific YAML/JSON files implement their expected schemas
- Check for common misspellings (especially if we can add a dictionary of domain-specific candidates). Go for a package that focuses on low false-positive rates rather than high detection rate

Pre-work to get existing files in a passing state for checkers we might use:

- [x] #2671
- [ ] Add pre-commit hooks to check for rst file issues
- [x] #2664
- [x] #2672

dannosliwcd:
- Resolves #2530

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -174,6 +174,7 @@ EXTRA_DIST = $(TUTORIAL_DIST) \
              .github/ISSUE_TEMPLATE/change.md \
              .github/ISSUE_TEMPLATE/tech_debt.md \
              .gitignore \
+	     git_hooks/README.rst \
              BLURB \
              CODE_OF_CONDUCT.md \
              CONTRIBUTING.rst \
--- a/copying_headers/MANIFEST.EXEMPT
+++ b/copying_headers/MANIFEST.EXEMPT
@@ -38,6 +38,7 @@ examples/README.md
 .github/ISSUE_TEMPLATE/change_test.md
 .github/ISSUE_TEMPLATE/tech_debt.md
 .gitignore
+git_hooks/README.rst
 integration/apps/amg/README.md
 integration/apps/arithmetic_intensity/README.md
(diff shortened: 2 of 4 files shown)


dannosliwcd:
It might be more flexible to design it to tie in with https://pre-commit.com/#repository-local-hooks

I guess the main change would be to rename the file to something more specific to this single check, then add the .yaml config file that indicates what the GEOPM-specific checks are. Interested users could either manually invoke these hooks as we do in this current PR revision, or they could use that pre-commit framework to enable whatever GEOPM asks them to enable, plus whatever those users have in their personal pre-commit configurations.

dannosliwcd:
Other suggested changes in #2530 more flexibly use pre-commit.com configurations.

Issue #2372, opened 2022-06-07
As a GEOPM researcher I would like agent.py to consider agents in the GEOPM_PLUGIN_PATH


lhlawson:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I try to write experiment scripts for tutorial agents and the agent.py does not recognize agents outside the supported agent list:
https://github.com/geopm/geopm/blob/fc0efcd72d0b92e9c69bc9eb5267e9f1c40efb10/scripts/geopmpy/agent.py#L175-L180

**Describe the solution you'd like**
Ideally agent.py should pull in GEOPM_PLUGIN_PATH and append any agent names to the supported_agents list

**Describe alternatives you've considered**
Simply removing 
https://github.com/geopm/geopm/blob/fc0efcd72d0b92e9c69bc9eb5267e9f1c40efb10/scripts/geopmpy/agent.py#L178-L179

Should be safe as any agent that does not exist will fail on:
https://github.com/geopm/geopm/blob/fc0efcd72d0b92e9c69bc9eb5267e9f1c40efb10/scripts/geopmpy/agent.py#L62-L65
This was seen in initial testing (see Additional Context) but not formally tested.

**Additional context**
The alternative solution is take in #2371 commit [f8a8bd8](https://github.com/geopm/geopm/pull/2371/commits/f8a8bd871b031b3ce0b9829f9fdd110f3b2a28eb)

PR #2532, opened 2022-08-09
Adjusted agent.py to grab the list of supported agents at runtime.

asmaalrawi:
Signed-off-by: Ali H Mohammad <ali.mohammad@intel.com>
Signed-off-by: Asma H Al-Rawi <asma.h.al-rawi@intel.com>

- Relates to #2372 from github issues

agent.py grabs the list of agents from gffi and determines whether the passed-in agent name is supported by the current  geopm environment.

### DIFF
--- a/scripts/geopmpy/agent.py
+++ b/scripts/geopmpy/agent.py
@@ -171,12 +171,8 @@ class AgentConf(object):
 
     """
     def __init__(self, path, agent='monitor', options=dict()):
-        agent_name_cstr = gffi.new("char[]", 1024)
-        agents = []
-        while _dl.geopm_agent_name(len(agents), 1024, agent_name_cstr) >= 0:
-            agents.append(gffi.string(agent_name_cstr).decode())
         self._path = path
-        if agent not in agents:
+        if agent not in names():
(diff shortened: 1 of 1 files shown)


scripts/geopmpy/agent.py
```diff
@@ -171,10 +171,12 @@ class AgentConf(object):
 
     """
     def __init__(self, path, agent='monitor', options=dict()):
-        supported_agents = {'monitor', 'power_governor', 'power_balancer',
-                            'frequency_map'}
+        agent_name_cstr = gffi.new("char[]", 1024)
+        agents = []
+        while _dl.geopm_agent_name(len(agents), 1024, agent_name_cstr) >= 0:
+            agents.append(gffi.string(agent_name_cstr).decode())
```
bgeltz: Use the `names()` function defined on line 129 instead of calling into gffi directly.  I think it'll look like:
```agents = names()```
That should be it.

bgeltz (changes requested):

alawibaba:
@bgeltz PTAL

bgeltz (approved):

PR #2551, opened 2022-08-11
Add check against cgroup cpuset when client makes request through service

### LINKED ISSUE #1685: As a system administrator, I would like to use Linux cgroup restrictions to configure the GEOPM service
Use Linux cgroups(7) to filter signal and control requests based on cgroup restrictions on the client PID, in particular the "cpuset" should be pretty straight forward to implement.  We may also want to consider accounting for other restrictions such as "rdma", "io", or "perf_event".

cmcantalupo:
- Fixes #1685

This is a WIP PR.  The unit tests need to be updated to mock sysfs queries.  Some by hand testing looks okay.

### DIFF
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -514,6 +514,10 @@ class PlatformService(object):
         elif not cont_req.issubset(supported_controls):
             raise RuntimeError('Requested controls that are not in allowed list: {}' \
                                .format(sorted(cont_req.difference(supported_controls))))
+        for domain, domain_idx in [(cc[0], cc[1]) for cc in signal_config]:
+            self._check_domain_permissions(client_pid, domain, domain_idx)
+        for domain, domain_idx in [(cc[0], cc[1]) for cc in control_config]:
+            self._check_domain_permissions(client_pid, domain, domain_idx)
         if len(control_config) != 0:
             self._write_mode(client_pid)
         batch_pid = self._active_sessions.get_batch_server(client_pid)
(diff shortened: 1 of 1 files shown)


service/geopmdpy/service.py
```diff
@@ -685,6 +691,63 @@ def check_client(self, client_pid):
...
+        result = []
+        for ent in cpu_str.split(','):
+            if '-' in ent:
+                ent = [int(ee) for ee in ent.split('-')]
+                result.extend(range(ent[0], ent[1] + 1))
+            else:
+                result.append(int(ent))
+        return result
+
+    def _get_cgroup(self, pid):
```
cmcantalupo, line 715: parameter is `pid` is called `client_pid` elsewhere

service/geopmdpy/service.py
```diff
@@ -514,6 +514,10 @@ def start_batch(self, client_pid, signal_config, control_config):
         elif not cont_req.issubset(supported_controls):
             raise RuntimeError('Requested controls that are not in allowed list: {}' \
                                .format(sorted(cont_req.difference(supported_controls))))
+        for domain, domain_idx in [(cc[0], cc[1]) for cc in signal_config]:
+            self._check_domain_permissions(client_pid, domain, domain_idx)
+        for domain, domain_idx in [(cc[0], cc[1]) for cc in control_config]:
```
Krebrov001, line 519: I would refactor these loops. They look somewhat hard to read and understand, in such a compact format. Looks like a very cool code trick, but somewhat obfuscated.

cmcantalupo:
```
        for domain, domain_idx in [(cc[0], cc[1]) for cc in signal_config]:
            self._check_domain_permissions(client_pid, domain, domain_idx)
```

The other simpler way to write this is:

```
        for cc in signal_config:
            self._check_domain_permissions(client_pid, cc[0], cc[1])
```

I'm not sure which is less confusing for the reader, but the original implementation is just to enable the values to be named in the function call.

perhaps just:

```
        for cc in signal_config:
            domain = cc[0]
            domain_idx = cc[1]
            self._check_domain_permissions(client_pid, domain, domain_idx)
```
?

PR #2562, opened 2022-08-15
Use CPUID for querying sticker frequency

### LINKED ISSUE #2336: CpuInfoGroup will fail to load if Model name is missing a space between @ <Sticker Freq>
**Describe the bug**
If the model name from lscpu is:
`Model name:          Intel(R) Xeon(R) Gold 6152 CPU @2.10GHz`
instead of 
`Model name:          Intel(R) Xeon(R) Gold 6152 CPU @ 2.10GHz`
The CpuInfoIOGroup will fail to parse the sticker frequency and throw an exception, despite a fallback option of reading sticker from cpuid existing.

**GEOPM version**
1.1.0+dev2063g37095df0

**Expected behavior**
CpuInfoIOGroup should either strip the @ symbol at the start of the string and get the frequency, or not throw and set result=NAN so that the fallback behavior of reading cpuid is hit.

**Actual behavior**
CpuInfoIOGroup errors and fails to load.  

**Error messages produced**
The error was: <geopm> Invalid argument: Invalid frequency: stod: at src/CpuinfoIOGroup.cpp:109

**Additional context**
read_cpu_freq_sticker already has a cpuid fallback option as of #1587, it simply requires an update to not throw.  A better overall approach may be to flip the order of operations: first try reading sticker via cpuid, and if that fails use lscpu parsing.

lhlawson:

- Relates to #2336 bug report from github issues
- Fixes #2336 change request from github issues.

Removes throw on lscpu parsing error in favor of falling back to cpuid calls and parsing.  
It may be preferable to have the entire function throw if the end result is NAN (not implemented in this PR)

### DIFF
--- a/service/src/CpuinfoIOGroup.cpp
+++ b/service/src/CpuinfoIOGroup.cpp
@@ -61,79 +61,28 @@ namespace geopm
         return result;
     }
 
-    static double read_cpu_freq_sticker(const std::string &read_str)
+    static double read_cpu_freq_sticker(void)
     {
-        double result = NAN;
-        const std::string key = "model name";
-        std::ifstream cpuinfo_file(read_str);
-        if (!cpuinfo_file.good()) {
-            throw Exception("Failed to open " + read_str + ": " + strerror(errno),
-                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
-        }
-        while (std::isnan(result) && cpuinfo_file.good()) {
-            std::string line;
-            std::getline(cpuinfo_file, line);
-            if (line.find(key) == 0 && line.find(':') != std::string::npos) {
-                size_t colon_pos = line.find(':');
-                bool match = true;
-                for (size_t pos = key.size(); pos != colon_pos; ++pos) {
-                    if (!std::isspace(line[pos])) {
-                        match = false;
-                    }
-                }
-                if (!match) {
-                    continue;
-                }
-                std::transform(line.begin(), line.end(), line.begin(),
-                               [](unsigned char c) {
-                                   return std::tolower(c);
-                               });
-                std::string unit_str[3] = {"ghz", "mhz", "khz"};
-                double unit_factor[3] = {1e9, 1e6, 1e3};
-                for (int unit_idx = 0; unit_idx != 3; ++unit_idx) {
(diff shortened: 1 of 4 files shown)


lhlawson:
Latest changes LGTM, tested on system with known lscpu sticker freq issues.

service/src/CpuinfoIOGroup.cpp
```diff
@@ -61,88 +63,32 @@ namespace geopm
...
-            std::cerr << "Warning: <geopm> Invalid brand string.  Using cpuid fallback." << std::endl;
-#endif
-            result = read_cpuid_freq_sticker();
-            if (result == 0) {
-                result = NAN;
-            }
+        double result = read_cpuid_freq_sticker();
+        if (result == 0) {
+            throw Exception("Sticker frequency not supported by CPUID",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
cmcantalupo: INVALID is usually for invalid user input, this is an invalid system configuration, so maybe a GEOPM_ERROR_PLATFORM_UNSUPPORTED.

service/src/CpuinfoIOGroup.cpp
```diff
@@ -157,6 +103,22 @@ namespace geopm
                                    Agg::expect_same,
                                    "Step size between processor frequency settings"}}})
     {
+        if (read_signal("CPUINFO::FREQ_MAX", GEOPM_DOMAIN_BOARD, 0) <=
+            read_signal("CPUINFO::FREQ_MIN", GEOPM_DOMAIN_BOARD, 0)) {
+            throw Exception("CpuinfoIOGroup::CpuinfoIOGroup(): Max frequency less than min",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
```
cmcantalupo: maybe GEOPM_ERROR_PLATFORM_UNSUPPORTED here as well

cmcantalupo (changes requested):
Just a couple of suggestions for the ERROR types otherwise LGTM

cmcantalupo (dismissed):

bgeltz (dismissed):

bgeltz (approved):

PR #2565, opened 2022-08-15
Fix issue in python agents for reports

cmcantalupo:
- Python runtime would not detect when application ended
- Add signal hander for SIGCHLD and stop iteration when this
  signal is detected.
- Modified TimedLoop to end if a SIGCHLD is raised


cmcantalupo:
May want to look into a similar solution in the future because pid.wait() may have higher CPU overhead than time.sleep.

PR #2567, opened 2022-08-16
Use pid.wait() in timed loop

cmcantalupo:
- This enables the runtime to terminate when the process ends
- Using pid.wait() may incur a higher CPU overhead than time.sleep
- May consider a solution like #2567 to enable the direct use of time.sleep if this performance overhead is an issue.

### DIFF
--- a/service/geopmdpy/runtime.py
+++ b/service/geopmdpy/runtime.py
@@ -58,9 +58,10 @@ class TimedLoop:
                             units of seconds.
 
             num_period (int): Number of time periods spanned by the
-                              loop.  The total loop time is num_periods * period,
-                              but since there is no delay in the first loop, there
-                              will be num_period + 1 loop iterations.
+                              loop.  The total loop time is
+                              num_periods * period, but since there is
+                              no delay in the first loop, there will
+                              be num_period + 1 loop iterations.
 
         """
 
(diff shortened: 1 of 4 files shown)


service/geopmdpy/runtime.py
```diff
@@ -100,11 +100,29 @@ def __next__(self):
...
         self._loop_idx += 1
         return result
 
+    def wait(self, timeout):
+        time.sleep(timeout)
+
+
+class PIDTimedLoop(TimedLoop):
+    def __init__(self, pid, period, num_period=None):
+        super(PIDTimedLoop, self).__init__(period, num_period)
```
cmcantalupo, line 146: Some doc strings would be nice, but maybe not a blocker for the RC

dannosliwcd (commented):
Changes to the runtime look good. I think the amount of measurement error is too wide for the current test implementation though. This solution still over-measures by around 0.02-0.06 seconds, and stress-ng over-runs by up 0.05 seconds. If both extreme cases intermittently hit at the same time, then the `5.0 +- 0.1 s` cutoff in test_tutorial_python_agents.py is too strict.

```
for i in {1..12}; do stresstime=$(./static_agent.py --geopm-report short.report -- stress-ng --cpu 5 --taskset 1-5 --timeout 5 2>&1 | awk -e '/successful run/ { print $8 }') && awk -v stresstime=$stresstime -e '/ runtime / { printf "geopm-stress = %f-%f = %f\n", $3, stresstime, $3-stresstime }' short.report ; done
geopm-stress = 5.064630-5.040000 = 0.024630
geopm-stress = 5.064450-5.040000 = 0.024450
geopm-stress = 5.064650-5.050000 = 0.014650
geopm-stress = 5.064610-5.040000 = 0.024610
geopm-stress = 5.064590-5.050000 = 0.014590
geopm-stress = 5.064490-5.040000 = 0.024490
geopm-stress = 5.064440-5.040000 = 0.024440
geopm-stress = 5.064300-5.050000 = 0.014300
geopm-stress = 5.064220-5.040000 = 0.024220
geopm-stress = 5.064270-5.040000 = 0.024270
geopm-stress = 5.064310-5.040000 = 0.024310
geopm-stress = 5.114490-5.050000 = 0.064490 <--- The test would be fail on this one (5.11)
```

One resolution would be to spy on the output from stress-ng and diff against that instead of the ideal 5.0s time, but that'll be a messy fit in to the current test implementation. I don't think it's worth implementing that strict test for this tutorial-oriented test suite (it would be much easier to implement in a performance-oriented test).

Another resolution is just to expand the time delta allowance, e.g., to 0.2 s.

`test_tutorial_python_agents.py` has 3 checks like 
```python
self.assertAlmostEqual(
    self._expected_stress_ng_timeout_seconds, host_data['Application Totals']['runtime (s)'],
    places=1)  
```

Those could either be relaxed to 0 places, or relaxed to `+- 0.2 s` as in:

```python
self.assertLess(
    abs(host_data['Application Totals']['runtime (s)'] - self._expected_stress_ng_timeout_seconds),
    0.2)  
```

service/geopmdpy/runtime.py
```diff
@@ -100,11 +101,70 @@ def __next__(self):
...
             if sleep_time > 0:
-                time.sleep(sleep_time)
+                self.wait(sleep_time)
         self._target_time += self._period
         self._loop_idx += 1
         return result
 
+    def wait(self, timeout):
+        """Pass-through to time.sleep()
+
```
cmcantalupo, line 111: Args: is missing

bgeltz (dismissed):

service/geopmdpy/runtime.py
```diff
@@ -318,6 +379,4 @@ def run(self, argv, policy=None, profile=None):
             self._agent.run_end()
```
cmcantalupo: swap order with restore_control()

dannosliwcd (dismissed):

bgeltz (approved):

PR #2574, opened 2022-08-18
Fix configure check

### LINKED ISSUE #2573: Configure check for LevelZero header does not include directory
**Describe the bug**
I tried to build the geopm service enabling level zero based on the OBS packaging and I expected configure to find the level zero headers instead the configure check failed.

**GEOPM version**
2.0.0+rc2

cmcantalupo:
- Include check for level zero header to include "level_zero" directory prefix
- Fixes #2573


lhlawson (approved):
LGTM, tested on surrogate system with oneAPI support.

bgeltz (approved):

PR #2581, opened 2022-08-22
Add SST-TF and TRL frequency limit estimators

### LINKED ISSUE #2580: Add a helper that estimates maximum-achievable frequency of a CPU core.

dannosliwcd:
- Classes estimate the peak achievable CPU core frequency based on current
  achieved frequencies and SST or TRL configuration state.
- Resolves #2580

### DIFF
--- a/src/SSTFrequencyLimitDetector.cpp
+++ b/src/SSTFrequencyLimitDetector.cpp
@@ -1,0 +1,211 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include "config.h"
+
+#include "SSTFrequencyLimitDetector.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <iterator>
+
+#include "geopm/Exception.hpp"
+#include "geopm/Helper.hpp"
+#include "geopm/PlatformIO.hpp"
+#include "geopm/PlatformTopo.hpp"
+#include "geopm_debug.hpp"
+
+namespace geopm
+{
+    enum Priorities_e {
+        HIGH_PRIORITY = 0,
+        MEDIUM_HIGH_PRIORITY = 1,
+        MEDIUM_LOW_PRIORITY = 2,
+        LOW_PRIORITY = 3,
+    };
+
+    SSTFrequencyLimitDetector::SSTFrequencyLimitDetector(
+        PlatformIO &platform_io,
+        const PlatformTopo &platform_topo)
+        : m_platform_io(platform_io)
+        , m_package_count(platform_topo.num_domain(GEOPM_DOMAIN_PACKAGE))
+        , m_core_count(platform_topo.num_domain(GEOPM_DOMAIN_CORE))
+        , m_clos_association_signals()
+        , m_frequency_limit_signals()
+        , m_sst_tf_enable_signals()
+        , M_CPU_FREQUENCY_MAX(m_platform_io.read_signal(
+              "CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0))
+        , M_CPU_FREQUENCY_STICKER(m_platform_io.read_signal(
+              "CPU_FREQUENCY_STICKER", GEOPM_DOMAIN_BOARD, 0))
+        , M_CPU_FREQUENCY_STEP(m_platform_io.read_signal(
+              "CPU_FREQUENCY_STEP", GEOPM_DOMAIN_BOARD, 0))
+        , M_ALL_CORE_TURBO_FREQUENCY(m_platform_io.read_signal(
+              "MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_7", GEOPM_DOMAIN_BOARD, 0))
+        , m_bucket_hp_cores{
+              static_cast<unsigned int>(m_platform_io.read_signal(
+                  "SST::HIGHPRIORITY_NCORES:0", GEOPM_DOMAIN_BOARD, 0)),
+              static_cast<unsigned int>(m_platform_io.read_signal(
+                  "SST::HIGHPRIORITY_NCORES:1", GEOPM_DOMAIN_BOARD, 0)),
+              static_cast<unsigned int>(m_platform_io.read_signal(
+                  "SST::HIGHPRIORITY_NCORES:2", GEOPM_DOMAIN_BOARD, 0))}
+        , M_LOW_PRIORITY_SSE_FREQUENCY(m_platform_io.read_signal(
+                    "SST::LOWPRIORITY_FREQUENCY:SSE", GEOPM_DOMAIN_BOARD, 0))
+        , M_LOW_PRIORITY_AVX2_FREQUENCY(m_platform_io.read_signal(
+                    "SST::LOWPRIORITY_FREQUENCY:AVX2", GEOPM_DOMAIN_BOARD, 0))
+        , M_LOW_PRIORITY_AVX512_FREQUENCY(m_platform_io.read_signal(
+                    "SST::LOWPRIORITY_FREQUENCY:AVX512", GEOPM_DOMAIN_BOARD, 0))
+        , m_bucket_sse_frequency{
+            m_platform_io.read_signal(
+                    "SST::HIGHPRIORITY_FREQUENCY_SSE:0", GEOPM_DOMAIN_BOARD, 0),
+            m_platform_io.read_signal(
+                    "SST::HIGHPRIORITY_FREQUENCY_SSE:1", GEOPM_DOMAIN_BOARD, 0),
+            m_platform_io.read_signal(
+                    "SST::HIGHPRIORITY_FREQUENCY_SSE:2", GEOPM_DOMAIN_BOARD, 0)}
+        , m_bucket_avx2_frequency{
+            m_platform_io.read_signal(
+                    "SST::HIGHPRIORITY_FREQUENCY_AVX2:0", GEOPM_DOMAIN_BOARD, 0),
+            m_platform_io.read_signal(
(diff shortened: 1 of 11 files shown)


src/SSTFrequencyLimitDetector.cpp
```diff
@@ -0,0 +1,209 @@
...
+        , m_clos_association_signals()
+        , m_sst_tf_enable_signals()
+        , m_cpu_frequency_max(m_platform_io.read_signal(
+              "CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0))
+        , m_cpu_frequency_sticker(m_platform_io.read_signal(
+              "CPU_FREQUENCY_STICKER", GEOPM_DOMAIN_BOARD, 0))
+        , m_cpu_frequency_step(m_platform_io.read_signal(
+              "CPU_FREQUENCY_STEP", GEOPM_DOMAIN_BOARD, 0))
+        , m_all_core_turbo_frequency(m_platform_io.read_signal(
+              "MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_7", GEOPM_DOMAIN_BOARD, 0))
```
avilcheslopez, line 44: Q: if this signal is not found, I understand an exception would be raised; is that what you expect for this class' constructor? Another way to ask: are future classes that will be using this class expected and prepared to handle this exception?

The same question would apply to the SST signals below, but I do see the ``FrequencyLimitDetector::make_unique()`` method will only choose the ``SSTFrequencyLimitDetector`` if support for SST is available. So, not urgent/important for these signals, but worth considering as the logic that protects the concern raised by the question is away from here (in case a future maintainer breaks the ``make_unique`` method).

src/TRLFrequencyLimitDetector.hpp
```diff
@@ -0,0 +1,42 @@
...
+                const std::vector<double> &observed_core_frequencies) override;
+            std::vector<std::pair<unsigned int, double> > get_core_frequency_limits(
+                unsigned int core_idx) const override;
+            double get_core_low_priority_frequency(unsigned int core_idx) const override;
+
+        private:
+            unsigned int m_package_count;
+            size_t m_core_count;
+            const double m_cpu_frequency_max;
+            const double m_cpu_frequency_sticker;
```
avilcheslopez: Same nit as before (regarding naming convention)

test/SSTFrequencyLimitDetectorTest.cpp
```diff
@@ -0,0 +1,222 @@
...
+static const double LP_FREQ_AVX512 = 1.7e9;
+
+static const int CORE_COUNT = 4;
+static const std::vector<double> HP_CORES = {2, 3, 4};
+static const std::vector<double> HP_FREQS_SSE = {3.6e9, 3.3e9, 3.0e9};
+static const std::vector<double> HP_FREQS_AVX2 = {3.5e9, 3.2e9, 2.9e9};
+static const std::vector<double> HP_FREQS_AVX512 = {3.4e9, 3.1e9, 2.8e9};
+
+static const int CLOS_SIGNAL_INDEX_OFFSET = 100;
+static const int SST_ENABLE_SIGNAL_INDEX_OFFSET = 1000;
```
avilcheslopez: Not important, but these could go in the test fixture class (just to have everything related to the tests neatly encapsulated).

src/TRLFrequencyLimitDetector.cpp
```diff
@@ -0,0 +1,88 @@
...
+    {
+#ifdef GEOPM_DEBUG
+        if (core_idx >= m_core_frequency_limits.size()) {
+            throw Exception("TRLFrequencyLimitDetector::" + std::string(__func__) +
+                                "Encountered invalid core index: " +
+                                std::to_string(core_idx),
+                            GEOPM_ERROR_INVALID,
+                            __FILE__, __LINE__);
+        }
+#endif
```
avilcheslopez: Same note as before

src/SSTFrequencyLimitDetector.hpp
```diff
@@ -0,0 +1,57 @@
...
+            std::vector<int> m_clos_association_signals;
+            std::vector<int> m_sst_tf_enable_signals;
+            const double m_cpu_frequency_max;
+            const double m_cpu_frequency_sticker;
+            const double m_cpu_frequency_step;
+            const double m_all_core_turbo_frequency;
+            std::vector<unsigned int> m_bucket_hp_cores;
+            const double m_low_priority_sse_frequency;
+            const double m_low_priority_avx2_frequency;
+            const double m_low_priority_avx512_frequency;
```
avilcheslopez: Nit: const naming convention.

src/SSTFrequencyLimitDetector.cpp
```diff
@@ -0,0 +1,209 @@
...
+    {
+#ifdef GEOPM_DEBUG
+        if (core_idx >= m_core_frequency_limits.size()) {
+            throw Exception("SSTFrequencyLimitDetector::" + std::string(__func__) +
+                                "Encountered invalid core index: " +
+                                std::to_string(core_idx),
+                            GEOPM_ERROR_INVALID,
+                            __FILE__, __LINE__);
+        }
+#endif
```
avilcheslopez: I'm a bit worried about this check being hidden behind a ``GEOPM_DEBUG`` "flag". I'm tempted to say this should be turned into an assertion for two reasons: (1) it seems like passing an invalid ``core_idx`` would be a programmer error, (2) the behavior of ``std::vector`` when passing an index out of bounds is undefined, which would result in an issue that's difficult to detect and debug.

If there are scenarios that could result in invalid input (i.e. not programmer error), I would say to simply go ahead and detect the situation and throw an exception.

avilcheslopez (commented):
Looks good to me! Just minor comments.

src/SSTFrequencyLimitDetector.cpp
dannosliwcd: The `use_sst_tf_signals` thing is a best-effort approach to use SST-TF. It asks platformIO if the platform reports that it supports the feature so it can choose whether to use the SST detector instead of the TRL detector. "Unsupported" could mean "the IOGroup exists and reports that this feature is unsupported", or it could mean "this is so unsupported that the IOGroup doesn't even exist." The try-catch deals with that second case.

In either the SST or the TRL case, I still want exceptions from the detector itself to bubble up past `make_unique`. If the limit detector can't read these signals then it cannot make its limit-detection decisions, so there's no point finishing construction. Ultimately, construction of the balancer agent will fail due to this exception.

avilcheslopez: Thanks for clarifying. Mainly wanted to make sure you were expecting these exceptions to propagate up.

src/SSTFrequencyLimitDetector.cpp
```diff
@@ -0,0 +1,209 @@
...
+                m_bucket_hp_cores[bucket], m_bucket_avx2_frequency[bucket]);
+            m_avx512_hp_tradeoffs.emplace_back(
+                m_bucket_hp_cores[bucket], m_bucket_avx512_frequency[bucket]);
+        }
+    }
+
+    void SSTFrequencyLimitDetector::update_max_frequency_estimates(
+        const std::vector<double> &observed_core_frequencies)
+    {
+        for (size_t package_idx = 0; package_idx < m_package_count; ++package_idx) {
```
dannosliwcd, line 115: Need to evaluate the direction of this search: Optimistic vs pessimistic peak freq estimation.

dannosliwcd: Does not impact the agent, which avoids depending on scenarios where other throttling points are the main factor. However, the interface allows other users of this class to interact in that way. Need to add the following:

If the core is limited by a given cap, start at the highest license level whose max freq is <= the cap (instead of always starting at avx512)

src/SSTFrequencyLimitDetector.cpp
```diff
@@ -184,26 +194,18 @@ namespace geopm
...
-                                "Encountered invalid core index: " +
-                                std::to_string(core_idx),
-                            GEOPM_ERROR_INVALID,
-                            __FILE__, __LINE__);
-        }
-#endif
-        std::vector<std::map<unsigned int, double> > tradeoffs;
-
-        return m_core_frequency_limits[core_idx];
+        return m_core_frequency_limits.at(core_idx);
```
avilcheslopez, line 197: I do like that this is cleaner. Question, though: what is our policy for managing exceptions? Is a class responsible for catching all internal exceptions and raising geopm::Exception (to provide a more standardized interface and hide internal implementation details) or do we allow underlying exceptions to propagate up?

src/TRLFrequencyLimitDetector.cpp
```diff
@@ -0,0 +1,74 @@
...
+            const auto core_with_max_frequency = *std::max_element(
+                cores_in_package.begin(), cores_in_package.end(),
+                [&observed_core_frequencies](int lhs, int rhs) {
+                    return observed_core_frequencies[lhs] < observed_core_frequencies[rhs];
+                });
+            const auto max_frequency = observed_core_frequencies[core_with_max_frequency];
+            for (const auto core_idx : cores_in_package) {
+                m_core_frequency_limits[core_idx] = {
+                    {cores_in_package.size(), max_frequency}};
+                m_core_lp_frequencies[core_idx] = M_CPU_FREQUENCY_STICKER;
```
dannosliwcd: Actually, I think this should be
```suggestion
                m_core_lp_frequencies[core_idx] = max_frequency;
```
because the TRLFrequencyLimitDetector case is meant to be used when we are assuming uniform max-achievable frequencies.

This change seems to help energy savings in one case on our non-sst-tf-capable cluster. I'll look into that bit more.

avilcheslopez (approved):

PR #2586, opened 2022-08-24
Add side effect when creating topo cache to delete shared memory keys

### LINKED ISSUE #2579: The GEOPM HPC runtime occasionally leaves behind files in /dev/shm when an application abnormally exits during launch
**Describe the bug**
I tried to launch an application with the GEOPM HPC runtime. 

**GEOPM version**
eb5137a60

**Expected behavior**
`geopmlaunch` successfully launches an application and generates a geopm report after the application finishes running.

**Actual behavior**
Sometimes `geopmlaunch` fails to launch with a message about `SharedMemoryImp: Could not open shared memory with key`, stating that the key cannot be created because it already exists. This seems to correlate with application launches that follow an abnormally-exiting application launch. For example, I have encountered this while learning how to launch new benchmarks, where I attempt to launch a benchmark config after previously trying an incompatible configuration that terminates during initialization.

The suspected cause is that this may be a result of an unhandled failure case in the GEOPM HPC runtime's handshake.

cmcantalupo:
- Resolves #2579.

### DIFF
--- a/service/src/geopmread_main.cpp
+++ b/service/src/geopmread_main.cpp
@@ -21,6 +21,7 @@
 #include "geopm/PlatformIO.hpp"
 #include "geopm/PlatformTopo.hpp"
 #include "geopm/Exception.hpp"
+#include "geopm/SharedMemory.hpp"
 
 #include "config.h"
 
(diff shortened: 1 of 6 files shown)


service/src/geopmread_main.cpp
```diff
@@ -43,7 +44,7 @@ int main(int argc, char **argv)
                         "  -d, --domain                     print domains detected\n"
                         "  -i, --info                       print longer description of a signal\n"
                         "  -I, --info-all                   print longer description of all signals\n"
-                        "  -c, --cache                      create geopm topo cache if it does not exist\n"
+                        "  -c, --cache                      create geopm topo cach and clean up /dev/shm\n"
```
bgeltz:
```suggestion
                        "  -c, --cache                      create geopm topo cache and clean up /dev/shm\n"
```

bgeltz (changes requested):

bgeltz (approved):

PR #2598, opened 2022-08-27
Fix the overflow in `CircularBuffer::set_capacity`

### LINKED ISSUE #2507: Heap buffer overflow exposed by CircularBufferTest.buffer_capacity
**Describe the bug**
I tried to run the service test suite under ASAN+UBSAN. The `CircularBufferTest.buffer_capacity` test case failed with a heap-buffer-overflow error.

**GEOPM version**

**Expected behavior**
I expect all unit tests to pass under UBSAN+ASAN sanitizers.

**Actual behavior**
The `CircularBufferTest.buffer_capacity` test case fails with a heap-buffer-overflow error. When the test case attempts to shrink an already-full buffer, it attempts to copy data from beyond the end of the original buffer into the new buffer.

The overflow occurs in an 8-byte read at: 
```
#4 0x979e6e in geopm::CircularBuffer<double>::set_capacity(unsigned int) .../service/./src/geopm/CircularBuffer.hpp:158:22
    # Which is: temp.push_back(m_buffer[i]);
    #                                   ^-- i == 2 (previous capacity)
    #                                         Loop's stopping condition is i != ((m_head + m_count) % m_max_size)
    #                                                                      2 != ((0 + 2) % 2)
#5 0x96d6fb in CircularBufferTest_buffer_capacity_Test::TestBody() .../service/test/CircularBufferTest.cpp:89:15
    # Which is: m_buffer->set_capacity(0);
```

The read occurs in a buffer allocated at:
```
#6 0x979e6e in geopm::CircularBuffer<double>::set_capacity(unsigned int) .../service/./src/geopm/CircularBuffer.hpp:158:22
#7 0x96cc4c in CircularBufferTest_buffer_capacity_Test::TestBody() .../service/test/CircularBufferTest.cpp:84:15  
```
i.e., the `push_back()` from the preceding `set_capacity(2)`

Krebrov001:
- Relates to #2507 
- Fixes #2507 

 - Previously the loop that used to copy the newest elements had a wrong offset.
 - It didn't correctly take into account looping over the `m_head` in the modulo.
 - I refactored the code to have a simplified loop using a proven class method.

### DIFF
--- a/service/src/geopm/CircularBuffer.hpp
+++ b/service/src/geopm/CircularBuffer.hpp
@@ -176,11 +176,11 @@ namespace geopm
             throw Exception("CircularBuffer::insert(): Cannot insert into a buffer of 0 size", GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
         if (m_count < m_max_size) {
-            m_buffer[m_count] = value;
+            m_buffer.at(m_count) = value;
             m_count++;
         }
         else {
-            m_buffer[m_head] = value;
+            m_buffer.at(m_head) = value;
             m_head = ((m_head + 1) % m_max_size);
         }
     }
@@ -191,7 +191,7 @@ namespace geopm
         if (index >= m_count) {
             throw Exception("CircularBuffer::value(): index is out of bounds", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        return m_buffer[(m_head + index) % m_max_size];
+        return m_buffer.at((m_head + index) % m_max_size);
     }
 
     template <class type>
--- a/service/src/geopm/CircularBuffer.hpp
+++ b/service/src/geopm/CircularBuffer.hpp
@@ -73,7 +73,6 @@ namespace geopm
             ///
             /// Accesses the contents of the circular buffer
             /// at a particular index.
-            /// There are two kinds of valid indices: positive and negative ones.
             /// Valid positive indices range from 0 to [size-1].
             /// Valid negative indices range from -1 to -size.
             /// Negative indices work just like python indices,
@@ -197,9 +196,9 @@ namespace geopm
     template <class type>
     const type& CircularBuffer<type>::value(const int index) const
     {
-        if (index >= 0 && index >= m_count ||
-            index <  0 && index < -m_count) {
-            throw Exception("CircularBuffer::value(): index is out of bounds", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        if (index >= static_cast<int>(m_count) || index < -static_cast<int>(m_count)) {
+            throw Exception(std::string("CircularBuffer::value(): index [") + std::to_string(index) + "] is out of bounds",
+                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
         if (index < 0) {
             const int new_index = m_count + index;
--- a/service/src/geopm/CircularBuffer.hpp
+++ b/service/src/geopm/CircularBuffer.hpp
@@ -32,10 +32,13 @@ namespace geopm
             virtual ~CircularBuffer();
             /// @brief Re-size the circular buffer.
             ///
-            /// Resets the capacity of the circular buffer without
-            /// modifying its current contents.
+            /// Grows or shrinks the capacity of the circular buffer.
+            /// If the new capacity is greater than the size, then the current contents
+            /// will not be modified.
+            /// If the new capacity is smaller than the size, then only the newest elements
+            /// of the difference will be retained.
             ///
-            /// @param [in] size Requested capacity for the buffer.
+            /// @param [in] size Requested new capacity for the buffer.
             void set_capacity(const unsigned int size);
             /// @brief Clears all entries from the buffer.
             ///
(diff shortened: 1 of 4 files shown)


dannosliwcd (changes requested):
The implementation change looks good. Please update the doxygen strings and add a unit test case. 

* Document that `set_capacity()` retains the newest elements when it is set to a value smaller than the current size.
* Document that `value()` accepts negative indices to represent nth-most-recent insertions. Our documentation currently calls that an error, but this change explicitly depends on that behavior not being treated as an error.
* Add a unit test that verifies that `value()` correctly wraps negative inputs.

service/src/geopm/CircularBuffer.hpp
```diff
@@ -69,17 +72,23 @@ namespace geopm
...
+            /// An out of bounds index is > [size-1] OR < [-size].
+            /// An attempt to retrieve a value for an out of bound index
             /// will throw a geopm::Exception with an
             /// error_value() of GEOPM_ERROR_INVALID.
             ///
             /// @param [in] index Buffer index to retrieve.
             ///
             /// @return Value from the specified buffer index.
-            const type& value(const unsigned int index) const;
+            const type& value(const int index) const;
```
dannosliwcd, line 90: @cmcantalupo I think this type change is allowed for 2.1 since this is a header-only interface (users always recompile with it so we don't need to worry about caller/callee type mismatches). Is that true?

service/src/geopm/CircularBuffer.hpp
```diff
@@ -69,17 +72,23 @@ namespace geopm
             /// @brief Returns a constant reference to the value from the buffer.
             ///
             /// Accesses the contents of the circular buffer
-            /// at a particular index. Valid indices range
-            /// from 0 to [size-1]. Where size is the number
-            /// of valid entries in the buffer. An attempt to
-            /// retrieve a value for an out of bound index
+            /// at a particular index.
+            /// There are two kinds of valid indices: positive and negative ones.
```
dannosliwcd:
```suggestion
```
Positive,  negative, _and zero_ are allowed. Since `const int index` already limits to that domain, you can omit this line.

service/src/geopm/CircularBuffer.hpp
```diff
@@ -69,17 +72,23 @@ namespace geopm
...
+            /// Valid positive indices range from 0 to [size-1].
+            /// Valid negative indices range from -1 to -size.
+            /// Negative indices work just like python indices,
+            /// to represent nth-most-recent insertions. For example,
+            /// -1 is the last element, -2 is the second to last element, and so on.
+            /// Where size is the number of valid entries in the buffer.
+            /// An out of bounds index is > [size-1] OR < [-size].
+            /// An attempt to retrieve a value for an out of bound index
             /// will throw a geopm::Exception with an
```
dannosliwcd, line 84: This level of detail is great! I think a user will know exactly what should/shouldn't go in to this function, and what will happen for each class of good and bad inputs.

service/src/geopm/CircularBuffer.hpp
```diff
@@ -176,22 +185,28 @@ namespace geopm
...
         }
     }
 
     template <class type>
-    const type& CircularBuffer<type>::value(const unsigned int index) const
+    const type& CircularBuffer<type>::value(const int index) const
     {
-        if (index >= m_count) {
+        if (index >= 0 && index >= m_count ||
+            index <  0 && index < -m_count) {
```
dannosliwcd:
```suggestion
        if (index >= m_count || index < -m_count) {
```

`unsigned long m_count` cannot be negative (by data type and by nature of being a non-borrowable count), so `index >= m_count` implies `index >= 0` and `index < -m_count` implies `index < 0`.

service/src/geopm/CircularBuffer.hpp
```diff
@@ -176,22 +185,28 @@ namespace geopm
...
             throw Exception("CircularBuffer::value(): index is out of bounds", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        return m_buffer[(m_head + index) % m_max_size];
+        if (index < 0) {
+            const int new_index = m_count + index;
+            return m_buffer.at((m_head + new_index) % m_max_size);
+        } else {
+            return m_buffer.at((m_head + index) % m_max_size);
+        }
```
dannosliwcd: Would it be sufficient to only do the single following case?

`return m_buffer.at((m_head + index) % m_count)`

I think that would work as long as we guarantee `m_count` is always `<= m_max_size`. Thoughts?

Krebrov001: @dannosliwcd That doesn't work. I already tried it.

service/src/geopm/CircularBuffer.hpp
Krebrov001: It's ok to make the change in this case.

service/src/geopm/CircularBuffer.hpp
Krebrov001: That change causes a lot of tests to fail, so it cannot be implemented.

dannosliwcd (approved):

PR #2602, opened 2022-08-31
Use SKX MSRs for everything newer than SKX

bgeltz:


service/src/MSRIOGroup.cpp
```diff
@@ -1193,25 +1190,23 @@ namespace geopm
...
+        }
+        else if (cpu_id == MSRIOGroup::M_CPUID_HSX ||
+                 cpu_id == MSRIOGroup::M_CPUID_BDX) {
+            platform_msrs = hsx_msr_json();
+        }
+        else if (cpu_id == MSRIOGroup::M_CPUID_SNB ||
+                 cpu_id == MSRIOGroup::M_CPUID_IVT) {
+            platform_msrs = snb_msr_json();
+        }
+        else if (cpu_id >= MSRIOGroup::M_CPUID_SKX) {
```
lhlawson: Do we want to add an ifdef GEOPM_DEBUG warning if cpu_id > M_CPUID_SKX that indicates we're using the skx msr because a newer MSR json isn't specified?

lhlawson (commented):
Other than a potential debug message LGTM.  Tested on a live system with a non-SKX CPUID and worked as expected.

service/src/MSRIOGroup.cpp
cmcantalupo: I agree here.  We should maintain a list of known cpu_id values that are tested, and print a debug warning if we find a newer cpu_id value than the ones we have validated.

cmcantalupo (approved):

PR #2613, opened 2022-09-07
Handle nans via `is_valid_value()`

### LINKED ISSUE #2252: NAN checks are wrong under the default build configuration for icpx
**Describe the bug**
There are several places in GEOPM code that inspect NAN values (e.g., by inequalities, isnan(), etc.) with an implicit assumption that they follow [IEEE 754 rules for NaN comparison](https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN). That assumption does not hold true when a compiler is allowed to follow non-strict floating point semantics. The default floating point behavior in `icc` and `icx` families of compilers is to use `-fp-model=fast=1`, which allows optimizations that break those assumptions.

In practice, the default behavior has been working okay for us with icpc. But the compiler's documentation indicates that it is allowed to break our assumptions under the default configuration. icpx does use non-precise FP optimizations in a way that invalidates our assumptions.

**GEOPM version**
15c59596

**Expected behavior**
All unit tests pass under both debug and non-debug GEOPM builds across Intel compilers.

**Actual behavior**
Multiple unit tests fail under non-debug GEOPM builds under the icpx family of compilers.

**Error messages produced**
27 tests fail in non-debug builds with icpx 2021.2.0, and they all pass in icpx debug builds and both types of icpc builds. Single failure excerpt:

```
[ RUN      ] PowerGovernorTest.govern_min
test/PowerGovernorTest.cpp:124: Failure
Expected equality of these values:
  node_power_min
    Which is: 100
  node_power_set
    Which is: 0
test/PowerGovernorTest.cpp:143: Failure
Expected equality of these values:
  new_node_power_min
    Which is: 102
  node_power_set
    Which is: 0
test/PowerGovernorTest.cpp:117: Failure
Actual function call count doesn't match EXPECT_CALL(m_platform_io, adjust(_, M_PKG_POWER_MIN))...
         Expected: to be called twice
           Actual: never called - unsatisfied and active
test/PowerGovernorTest.cpp:137: Failure
Actual function call count doesn't match EXPECT_CALL(m_platform_io, adjust(_, new_pkg_power_min))...
         Expected: to be called twice
           Actual: never called - unsatisfied and active
[  FAILED  ] PowerGovernorTest.govern_min (0 ms)
```

In this case, there are multiple `==`, `!=` and `isnan` checks in `PowerGovernor` that may not work as expected.

For example, in `PowerGovernor::adjust_platform`:
```
if (m_last_pkg_power_setting != target_pkg_power) {
```

Evaluates as
```
nan != 300  -->  0
```

Krebrov001:
- Relates to #2252
- Fixes #2252 

Handle nans via `is_valid_value()`


service/docs/source/GEOPM_CXX_MAN_PlatformIO.3.rst
```diff
@@ -132,6 +134,16 @@ Singleton Accessor
   all of the `CLASS METHODS <INSPECTION CLASS METHODS_>`_ described below.  See `EXAMPLE <EXAMPLE_>`_ section
   below.
 
+Static Class Methods
+--------------------
+
+
+``is_valid_value()``
+  In geopm, the values for signals and controls can be invalid due to errors.
```
dannosliwcd:
```suggestion
  In geopm, the values for signals and controls can be invalid due to errors or absent sampling data.
```

service/src/PlatformIO.cpp
```diff
@@ -1419,4 +1424,8 @@ extern "C" {
     {
         geopm::platform_io_reset();
     }
+
+    int geopm_pio_check_valid_value(double value) {
+        return geopm::platform_io().is_valid_value(value) ? 0 : -1;
```
dannosliwcd: Should this return a value of `enum geopm_error_e`?

dannosliwcd (commented):
This relates to #2252, but it does not fix #2252. The fix for that issue would either be to guarantee that `nan` never occurs in our floating point code, or to use a more strict floating point policy with our compiler.

The second option is less intrusive to us, but potentially impacts our users (who may rely on nan-related fast-math optimizations for their best-case performance). The hope is that the change you posted here is sufficient for those users that depend on fast math optimizations and also wish to use `geopm_pio` interfaces in their applications.

Run the unit test suite with everything built under a recent version of the `icpx` compiler to reproduce #2252. It would also be nice to see whether our assumption about this change being sufficient for users is true. In that case, we'd need to compile a version of GEOPM PIO with the fix, and link to it from a test application that is built with nan-related fast math optimizations enabled.

service/src/PlatformIO.cpp
Krebrov001: `GEOPM_ERROR_INVALID`

m4/ax_check_compile_flag.m4
```diff
@@ -0,0 +1,53 @@
...
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Guido U. Draheim <guidod@gmx.de>
+#   Copyright (c) 2011 Maarten Bosmans <mkbosmans@gmail.com>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved.  This file is offered as-is, without any
+#   warranty.
```
cmcantalupo, line 35: We will have to add this header to geopm/copying_headers and list this file in the associated MANIFEST.

service/docs/source/GEOPM_CXX_MAN_PlatformIO.3.rst
```diff
@@ -132,6 +134,16 @@ Singleton Accessor
   all of the `CLASS METHODS <INSPECTION CLASS METHODS_>`_ described below.  See `EXAMPLE <EXAMPLE_>`_ section
   below.
 
+Static Class Methods
+--------------------
+
+
```
cmcantalupo:
```suggestion
```

service/docs/source/GEOPM_CXX_MAN_PlatformIO.3.rst
```diff
@@ -132,6 +134,16 @@ Singleton Accessor
...
+``is_valid_value()``
+  In geopm, the values for signals and controls can be invalid due to errors or absent sampling data.
+  This function determines if the current *value* of the signal or control is
+  a valid value. It returns true if the *value* is valid, and false if the *value*
+  is not valid.
+
```
cmcantalupo, line 145:
```suggestion


```

service/docs/source/geopm_pio.3.rst
```diff
@@ -96,6 +96,8 @@ Link with ``-lgeopm`` **(MPI)** or ``-lgeopmpolicy`` **(non-MPI)**
 
        int geopm_pio_reset(void);
 
+       int geopm_pio_check_valid_value(double value);
+
```
cmcantalupo, line 100:
```suggestion


```

service/docs/source/geopm_pio.3.rst
```diff
@@ -210,14 +212,20 @@ Inspection Functions
...
-PlatformIO instance and does not affect the PlatformIO instance
-managed by the GEOPM HPC runtime.**
+  Reset the GEOPM platform interface causing resources to be freed.
+  This will cause the internal PlatormIO instance to be
+  released/deleted and reconstructed.  As a result, any signals and
+  controls that had been pushed will be cleared, any batch servers
+  that had been started will be stopped, and all registered IOGroups
+  will be reset.  **NOTE: the reset only applies to the Service
+  PlatformIO instance and does not affect the PlatformIO instance
+  managed by the GEOPM HPC runtime.**
```
cmcantalupo: This comment is completely unrelated to this pull request, but this Note is misleading.  The reset applies to all use cases of PlatformIO with the one exception of the GEOPM HPC Runtime use case.  In the GEOPM HPC Runtime usecase, a different singleton access function is used, so this reset does not apply to the HPC Runtime PlatformIO instance.

service/docs/source/geopm_pio.3.rst
```diff
@@ -210,14 +212,20 @@ Inspection Functions
...
+
+``geopm_pio_check_valid_value()``
+  In geopm, the values for signals and controls can be invalid due to errors.
```
cmcantalupo:
```suggestion
  In geopm, signals may be unavailable due to missing information.  In these cases
  where the requested signal temporarily unavailable an invalid value is returned.
 ```

File: service/docs/source/geopm_pio.3.rst [hunk 9]
```diff
@@ -210,14 +212,20 @@ Inspection Functions
...
+  This function determines if the current *value* of the signal or control is
```
cmcantalupo:
```suggestion
  This function determines if the current *value* of the signal is
```

service/src/PlatformIO.cpp
```diff
@@ -1419,4 +1424,8 @@ extern "C" {
     {
         geopm::platform_io_reset();
     }
+
+    int geopm_pio_check_valid_value(double value) {
```
```suggestion
    int geopm_pio_check_valid_value(double value)
    {
```

avilcheslopez:
Awesome! I see this is passing the build now. Let me look into getting approval for the license being added before we merge this in.

dannosliwcd:
BTW, this also fixes the linked issue now that you also added the new build flags

service/configure.ac
```diff
@@ -321,8 +321,12 @@ AC_COMPILE_IFELSE([
...
 AC_LANG_POP([C++])
 
-[EXTRA_CFLAGS="$EXTRA_CFLAGS -Wall -Werror"]
-[EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS -Wall -Werror"]
+m4_include([m4/ax_check_compile_flag.m4])
+AX_CHECK_COMPILE_FLAG([-fp-model="fast"], [FASTMATH="-fp-model=fast"], [FASTMATH="-ffast-math"], [], [])
+AX_CHECK_COMPILE_FLAG([-fp-model="precise"], [NO_FASTMATH="-fp-model=precise"], [NO_FASTMATH="-fno-fast-math"], [], [])
+
+[EXTRA_CFLAGS="$EXTRA_CFLAGS -Wall -Werror $NO_FASTMATH"]
+[EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS -Wall -Werror $NO_FASTMATH"]
```
dannosliwcd, line 329: Can you verify that [this case](https://github.com/geopm/geopm/issues/2252#issuecomment-1213135326) from #2252 (monitor agent + geopmbench + icpx) produces a valid `geopm.report` file with this PR's current revision?

If that test passes, then everything LGTM. If that test fails, then I think these changes may be needed in the HPC Runtime's configure.ac as well.

avilcheslopez:
> This relates to #2252, but it does not fix #2252. The fix for that issue would either be to guarantee that `nan` never occurs in our floating point code, or to use a more strict floating point policy with our compiler.
> 
> The second option is less intrusive to us, but potentially impacts our users (who may rely on nan-related fast-math optimizations for their best-case performance). The hope is that the change you posted here is sufficient for those users that depend on fast math optimizations and also wish to use `geopm_pio` interfaces in their applications.
> 
> Run the unit test suite with everything built under a recent version of the `icpx` compiler to reproduce #2252. It would also be nice to see whether our assumption about this change being sufficient for users is true. In that case, we'd need to compile a version of GEOPM PIO with the fix, and link to it from a test application that is built with nan-related fast math optimizations enabled.

Helpful steps provided by Daniel:
- [ ] Build & run the GEOPM HPC runtime from dev, using icpx and icx compilers. geopmbench should be an executable output of this build.
- [ ] Confirm that you can run geopmlaunch with geopmbench (e.g., single server, two MPI processes, default geopmbench configuration with no input args)
- [ ] Confirm that the generated geopm.report file replicates the zeroed app totals shown in the previous comment on this issue.
- [ ] Re-build GEOPM with the fix in #2613, still using icpx and icx compilers. Re-run geopmbench under geopmlaunch.
- [ ] Confirm that the new geopm.report correctly displays non-zero totals.

dannosliwcd (approved):

Issue #2617, opened 2022-09-14
As a GEOPM user, I would like to be able to use a neural net for live workload region classification and frequency controls


asmaalrawi:
**Is your feature request related to a problem? Please describe.**
Making frequency decisions in a dynamic workload with the goal of energy efficiency is difficult without understanding the workload behavior a priori.

**Describe the solution you'd like**
A feed-forward agent can ingest a neural net and output region classification information, which can then be used to determine an optimal frequency modulo a user-input parameter for tuning energy efficiency vs performance degradation.

**Describe alternatives you've considered**
-  Hand-tuned frequency decisions with prior knowledge of every workload
-  Hard-coded neural net as a constant in a feed-forward agent
- Bandit algorithm that relies on knowledge of per-epoch behavior and converges to an optimal frequency online

Tasks:
- [x] #2637 
- [x] #2693 
- [x] #2694
- [x] #2679 
- [x] #3364
- [x] #3365
- [x] #3366

PR #2618, opened 2022-09-14
Added TensorOneD class and tests.

### LINKED ISSUE #2677: Introduce new C++ class: TensorOneD
> estimate 1
Parent feature: #2617 

Introduce new C++ class [...] with the following features:
- [...]
- [...]

Done criteria:

- [ ] ~Pure virtual interface class has been added with factory method~
      NVMLDevicePool)
- [ ] [...]
- [ ] [...]

asmaalrawi:
- Relates to #2617  
- Fixes #2677 
- Adds TensorOneD class which defines 1-dimensional tensor operations and is used by the Feed Forward agent

### DIFF
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -18,8 +18,6 @@
 
 namespace geopm
 {
-
-
     TensorOneD::TensorOneD(std::size_t dim)
     {
         set_dim(dim);
@@ -57,7 +55,6 @@ namespace geopm
             m_vec[idx] = input[idx].number_value();
         }
     }
-
 
     TensorOneD TensorOneD::operator+(const TensorOneD& other)
     {
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -43,27 +43,14 @@ namespace geopm
         return m_vec.size();
     }
 
-    TensorOneD::TensorOneD(json11::Json input)
+    TensorOneD::TensorOneD(std::vector<float> input)
     {
-        if (!input.is_array()) {
-            throw geopm::Exception("Neural network weights is non-array-type.\n",
-                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-        }
-        if (input.array_items().empty()) {
-            throw geopm::Exception("Empty array is invalid for neural network weights.\n",
+        if (input.size() == 0) {
+            throw geopm::Exception("TensorOneD cannot be initialized with empty vector.\n",
                                    GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
 
-        set_dim(input.array_items().size());
-        std::size_t vec_size = m_vec.size();
-
-        for (std::size_t idx = 0; idx < vec_size; ++idx) {
-            if (!input[idx].is_number()) {
-                throw geopm::Exception("Non-numeric type found in neural network weights.\n",
-                                       GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-            }
-            m_vec[idx] = input[idx].number_value();
-        }
+        m_vec = input;
     }
 
     TensorOneD TensorOneD::operator+(const TensorOneD& other)
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -59,7 +59,6 @@ namespace geopm
     }
 
 
-    //TODO: Question: What happens if other and *this are the same?
     TensorOneD TensorOneD::operator+(const TensorOneD& other)
     {
         if (get_dim() != other.get_dim()) {
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -104,6 +104,10 @@ namespace geopm
 
     TensorOneD& TensorOneD::operator=(TensorOneD &&other)
     {
+        if (&other == this) {
+            return *this;
+        }
+
         m_vec = std::move(other.m_vec);
         return *this;
     }
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -3,11 +3,13 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+
 #include <math.h>
 #include <algorithm>
 #include <functional>
 #include <numeric>
 #include "geopm/Exception.hpp"
+#include "config.h"
 
 #include "TensorOneD.hpp"
 
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -90,18 +90,23 @@ namespace geopm
         return m_vec;
     }
 
-
-    // TODO put these in the right places
-    TensorOneD TensorOneD::operator+(const TensorOneD &other) const {
+    TensorOneD TensorOneD::operator+(const TensorOneD &other) const
+    {
         return m_math->add(*this, other);
     }
-    TensorOneD TensorOneD::operator-(const TensorOneD &other) const {
+
+    TensorOneD TensorOneD::operator-(const TensorOneD &other) const
+    {
         return m_math->subtract(*this, other);
     }
-    float TensorOneD::operator*(const TensorOneD &other) const {
+
+    float TensorOneD::operator*(const TensorOneD &other) const
+    {
         return m_math->inner_product(*this, other);
     }
-    TensorOneD TensorOneD::sigmoid() const {
+
+    TensorOneD TensorOneD::sigmoid() const
+    {
         return m_math->sigmoid(*this);
     }
 }
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -1,93 +1,57 @@
 /*
- * Copyright (c) 2015 - 2022, Intel Corporation
+ * Copyright (c) 2015 - 2023, Intel Corporation
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
 
 #include "config.h"
 
+#include "TensorMath.hpp"
 #include "TensorOneD.hpp"
-
-#include <cmath>
-#include <algorithm>
-#include <functional>
-#include <numeric>
-#include <utility>
 
 #include "geopm/Exception.hpp"
 
 namespace geopm
 {
-    TensorOneD::TensorOneD(std::size_t dim)
+    TensorOneD::TensorOneD()
+        : TensorOneD(0)
     {
-        set_dim(dim);
+    }
+
+    TensorOneD::TensorOneD(size_t dim)
+        : TensorOneD(std::vector<float>(dim))
+    {
     }
 
     TensorOneD::TensorOneD(const TensorOneD &other)
-        : m_vec(other.m_vec)
+        : TensorOneD(other.m_vec)
     {
     }
 
     TensorOneD::TensorOneD(TensorOneD &&other)
-        : m_vec(std::move(other.m_vec))
+        : TensorOneD(std::move(other.m_vec))
     {
-    }
-
-    void TensorOneD::set_dim(std::size_t dim)
-    {
-        m_vec.resize(dim);
-    }
-
-    std::size_t TensorOneD::get_dim() const
-    {
-        return m_vec.size();
     }
 
     TensorOneD::TensorOneD(std::vector<float> input)
+        : TensorOneD(input, std::make_shared<TensorMathImp>())
     {
-        if (input.size() == 0) {
-            throw geopm::Exception("TensorOneD cannot be initialized with empty vector.\n",
-                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-        }
+    }
 
+    TensorOneD::TensorOneD(std::vector<float> input, std::shared_ptr<TensorMath> math)
+    {
         m_vec = input;
+        m_math = math;
     }
 
-    TensorOneD TensorOneD::operator+(const TensorOneD& other)
+    void TensorOneD::set_dim(size_t dim)
     {
-        if (get_dim() != other.get_dim()) {
-            throw geopm::Exception("Adding vectors of mismatched dimensions.",
-                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-        }
-
-        TensorOneD rval(m_vec.size());
-        std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.m_vec.begin(), std::plus<float>());
-
-        return rval;
+        m_vec.resize(dim, 0);
     }
 
-    TensorOneD TensorOneD::operator-(const TensorOneD& other)
+    size_t TensorOneD::get_dim() const
     {
-        if (get_dim() != other.get_dim()) {
-            throw geopm::Exception("Subtracting vectors of mismatched dimensions.",
-                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-        }
-
-        TensorOneD rval(m_vec.size());
-        std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.m_vec.begin(), std::minus<float>());
-
-        return rval;
-    }
-
-
-    float TensorOneD::operator*(const TensorOneD& other)
-    {
-        if (get_dim() != other.get_dim()) {
-            throw geopm::Exception("Inner product of vectors of mismatched dimensions.",
-                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-        }
-
-        return std::inner_product(m_vec.begin(), m_vec.end(), other.m_vec.begin(), 0);
+        return m_vec.size();
     }
 
     TensorOneD& TensorOneD::operator=(const TensorOneD &other)
@@ -106,22 +70,38 @@ namespace geopm
         return *this;
     }
 
-    float& TensorOneD::operator[] (std::size_t idx)
+    bool TensorOneD::operator==(const TensorOneD &other) const
+    {
+        return m_vec == other.m_vec;
+    }
+
+    float& TensorOneD::operator[] (size_t idx)
     {
         return m_vec.at(idx);
     }
 
-    float TensorOneD::operator[] (std::size_t idx) const
+    float TensorOneD::operator[] (size_t idx) const
     {
         return m_vec.at(idx);
     }
 
-    TensorOneD TensorOneD::sigmoid() const
+    const std::vector<float> &TensorOneD::get_data() const
     {
-        TensorOneD rval(m_vec.size());
-        for(std::size_t idx = 0; idx < m_vec.size(); idx++) {
-            rval[idx] = 1/(1 + expf(-(m_vec.at(idx))));
-        }
-        return rval;
+        return m_vec;
+    }
+
+
+    // TODO put these in the right places
+    TensorOneD TensorOneD::operator+(const TensorOneD &other) const {
+        return m_math->add(*this, other);
+    }
+    TensorOneD TensorOneD::operator-(const TensorOneD &other) const {
+        return m_math->subtract(*this, other);
+    }
+    float TensorOneD::operator*(const TensorOneD &other) const {
+        return m_math->inner_product(*this, other);
+    }
+    TensorOneD TensorOneD::sigmoid() const {
+        return m_math->sigmoid(*this);
     }
 }
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -33,6 +33,16 @@ namespace geopm
     {
     }
 
+    void TensorOneD::set_dim(std::size_t dim)
+    {
+        m_vec.resize(dim);
+    }
+
+    std::size_t TensorOneD::get_dim() const
+    {
+        return m_vec.size();
+    }
+
     TensorOneD::TensorOneD(json11::Json input)
     {
         if (!input.is_array()) {
--- a/src/TensorOneD.cpp
+++ b/src/TensorOneD.cpp
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <math.h>
+#include <algorithm>
+#include <functional>
+#include <numeric>
+#include "geopm/Exception.hpp"
+
+#include "TensorOneD.hpp"
+
+namespace geopm
+{
+    TensorOneD::TensorOneD()
+    {
+    }
+
+    TensorOneD::TensorOneD(std::size_t dim)
+    {
+        set_dim(dim);
+    }
+
+    TensorOneD::TensorOneD(const TensorOneD &other)
+    {
+        m_vec = other.m_vec;
+    }
+
+    TensorOneD::TensorOneD(json11::Json input)
+    {
+        if (!input.is_array()) {
+            throw geopm::Exception("Neural network weights is non-array-type.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        if (input.array_items().size() == 0) {
+            throw geopm::Exception("Empty array is invalid for neural network weights.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        set_dim(input.array_items().size());
+
+        for (std::size_t idx = 0; idx < m_vec.size(); idx++) {
+            if (!input[idx].is_number()) {
+                throw geopm::Exception("Non-numeric type found in neural network weights.\n",
+                                       GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+            m_vec[idx] = input[idx].number_value();
+        }
+    }
+
+    std::size_t TensorOneD::get_dim() const
+    {
+        return m_vec.size();
+    }
+
+    void TensorOneD::set_dim(std::size_t dim)
+    {
+        m_vec.resize(dim);
+    }
+
+    TensorOneD TensorOneD::operator+(const TensorOneD& other)
+    {
+        if (get_dim() != other.get_dim()) {
+            throw geopm::Exception("Adding vectors of mismatched dimensions.",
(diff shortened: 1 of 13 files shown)


src/TensorOneD.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+TensorOneD::TensorOneD(const TensorOneD &other)
+{
+    set_dim(other.get_dim());
+    for (int idx = 0; idx < (int)m_vec.size(); idx++) {
+        m_vec[idx] = other[idx];
+    }
+}
+
+TensorOneD::TensorOneD(json11::Json input)
```
avilcheslopez: Q: what is your plan for validating the JSON input?

src/TensorOneD.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    set_dim(dim);
+}
+
+TensorOneD::TensorOneD(const TensorOneD &other)
+{
+    set_dim(other.get_dim());
+    for (int idx = 0; idx < (int)m_vec.size(); idx++) {
+        m_vec[idx] = other[idx];
+    }
```
avilcheslopez: I believe you can replace this with: ``m_vec = other.m_vec`` (considering ``std::vector`` implements ``operator=()`` and considering the vector is of a primitive type).

However, also considering the default implementations of the copy constructor and assignment operator, I believe you should be able to get away with not defining them all together.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,128 @@
...
+TensorOneD
+TensorOneD::operator+(const TensorOneD& other)
+{
+    if (get_dim() != other.get_dim()) {
+        throw geopm::Exception("Adding vectors of mismatched dimensions.",
+                               GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+    }
+
+    TensorOneD rval(m_vec.size());
+    for (int idx = 0; idx < (int)m_vec.size(); idx++) {
```
avilcheslopez: Consider taking advantage of C++'s ``<algorithm>`` header. In this case, you could replace your ``for`` loop with:
``std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.begin(), std::plus<float>())``

If you go that route, the following notes don't apply, but still worth mentioning for future reference:

- I would recommend avoiding the use of C-style casts. You could declare ``idx`` of your vector's ``size_type``. Alternatively, you could use ``size_t`` (technically, not necessarily the same type, but it should be a safe enough choice).
- Another approach could be to use iterators for the vectors, moving away from having to worry about indexing (considering no special indexing operations are being performed and you're iterating over both vectors in parallel).

Same comment applies to other ``for`` loops below.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // assert(this->n == other.n);
+    for (int idx = 0; idx < (int)m_vec.size(); idx++) {
+        rval += m_vec[idx] * other.m_vec[idx];
+    }
+
+    return rval;
+}
+
+TensorOneD&
+TensorOneD::operator=(const TensorOneD &other)
```
avilcheslopez: See comment above in copy constructor.

src/TensorOneD.cpp
asmaalrawi: Discussed at standup: schema will be used to ensure the format. We can add some basic sanity checks as needed wherever json is ingested. We can't guarantee that there will be good behavior for an arbitrary neural net json input. Do we need to manage this expectation?

avilcheslopez: At what point will the input be validated against a schema?

Yes, regardless, some deeper checking will be needed as there are some semantic rules that a schema validator won't be able to check. As for the extent of checking, I agree: ensuring the neural net will yield good results/behavior is beyond the scope of this checking.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    TensorOneD::TensorOneD(const TensorOneD &other)
+    {
+        m_vec = other.m_vec;
+    }
+
+    TensorOneD::TensorOneD(json11::Json input)
+    {
+        set_dim(input.array_items().size());
+        for (int idx = 0; idx < (int)m_vec.size(); idx++) {
```
avilcheslopez: Earlier recommendation about avoiding C-style casts applies here as well. Declare ``idx`` as the vector's ``size_type``, or use ``size_t``.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,109 @@
...
+            return std::inner_product(m_vec.begin(), m_vec.end(), other.m_vec.begin(), 0);
+        }
+
+    TensorOneD& TensorOneD::operator=(const TensorOneD &other)
+    {
+        m_vec = other.m_vec;
+        return *this;
+    }
+
+    float& TensorOneD::operator[] (int idx)
```
avilcheslopez: Sorry, I should've noticed this before:

A similar comment about leveraging the vector's ``size_type`` applies here as well (instead of using int). See: https://cplusplus.com/reference/vector/vector/operator[]/

The same applies to other places in the code (e.g. the specialized constructor that takes the dim).

src/TensorOneD.cpp
```diff
@@ -0,0 +1,113 @@
...
+                        GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+            m_vec[idx] = input[idx].number_value();
+        }
+    }
+
+    std::size_t TensorOneD::get_dim() const
+        {
+            return m_vec.size();
+        }
```
avilcheslopez: The indentation is a bit off here. Same applies to the following few functions.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    TensorOneD::TensorOneD(const TensorOneD &other)
+    {
+        m_vec = other.m_vec;
+    }
+
+    TensorOneD::TensorOneD(json11::Json input)
+    {
+        set_dim(input.array_items().size());
```
avilcheslopez: This assumes the Json object is an array. I recommend you validate that (I believe Json has an ``is_array()`` function).

test/TensorOneDTest.cpp
```diff
@@ -0,0 +1,105 @@
...
+    EXPECT_THROW(one - three, geopm::Exception);
+    EXPECT_THROW(one * three, geopm::Exception);
+}
+
+TEST_F(TensorOneDTest, test_bad_input){
+    TensorOneD x(3);
+    x.set_dim(4);
+    std::vector<std::string> vals = {"soup", "16"};
+    EXPECT_THROW(TensorOneD(json11::Json(vals)), geopm::Exception);
+}
```
avilcheslopez: As far as testing JSON input, what happens if you pass an empty array JSON array "[]"? Is the expected behavior that you'll end up with an empty TensorOneD object (i.e. ``get_dim()`` should return 0)?

src/TensorOneD.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    TensorOneD& TensorOneD::operator=(const TensorOneD &other)
+    {
+        m_vec = other.m_vec;
+        return *this;
+    }
+
+    float& TensorOneD::operator[] (std::size_t idx)
+    {
+        return m_vec[idx];
```
avilcheslopez: Note that passing an index that is out of bounds to a ``std::vector`` results in undefined behavior. I would recommend bounds checking. You could use the ``at()`` method instead, which will throw an exception in that case. 

Same applies to the function below.

asmaalrawi: Used at() method

avilcheslopez (dismissed):

alawibaba:
+1

src/TensorOneD.cpp
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <math.h>
```
bgeltz:
```suggestion
#include "config.h"

#include <math.h>
```

bgeltz (changes requested):
Other than what I called out, this looks good.  I'm ready to approve once these changes are made.

src/TensorOneD.hpp
```diff
@@ -0,0 +1,35 @@
...
+namespace geopm
+{
+    class TensorOneD
+    {
+        public:
+            TensorOneD();
+            TensorOneD(std::size_t n);
+            TensorOneD(const TensorOneD&);
+            TensorOneD(json11::Json input);
+            void set_dim(std::size_t n);
```
bgeltz: Please add docstrings to the hpp as per the done criteria in #2677:

> All public methods have doxygen comments describing behavior, inputs, outputs and exceptions raised

bgeltz (changes requested):

src/TensorOneD.hpp
```diff
@@ -0,0 +1,100 @@
...
+            ///
+            ///@param [in] input json11::Json instance
+            TensorOneD(json11::Json input);
+            ///@brief Set length of 1D tensor
+            ///
+            ///If the instance has more than n elements, it will
+            ///be truncated. If it has fewer than n elements, the
+            ///tensor will be expanded to a total size of n, uninitialized.
+            ///A geopm::Exception will be rasied if input is not an array,
+            ///if input is empty, or if a non-numeric type is found in input.
```
avilcheslopez: It seems this comment applies to the constructor above.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,125 @@
...
+    {
+        if (get_dim() != other.get_dim()) {
+            throw geopm::Exception("Adding vectors of mismatched dimensions.",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        TensorOneD rval(m_vec.size());
+        std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.m_vec.begin(), std::plus<float>());
+
+        return rval;
```
avilcheslopez: I'm just realizing, move semantics is a worthwhile optimization for this class that should be implemented. This will optimize passing TensorOneD rvalues around (by avoiding having to copy the internal vector every time). This would be especially helpful if TensorOneD objects will be used in more complex operations, e.g.: ``float x = (a + b) * c`` (where ``a``, ``b`` and ``c`` are TensorOneD objects).

src/TensorOneD.hpp
```diff
@@ -0,0 +1,104 @@
...
+namespace geopm
+{
+    /// @brief Class to store and perform operations on 1D Tensors,
+    ///        aka vectors, suitable for use in feed-forward neural
+    ///        networks.
+    class TensorOneD
+    {
+        public:
+            ///@brief Constructor
+            TensorOneD();
```
bgeltz, line 23:
```suggestion
            TensorOneD() = default;
```
Then remove the empty implementation in the cpp.

OTOH, If you never intend on using this constructor, delete it with ```TensorOneD() = delete;``` instead.

src/TensorOneD.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace geopm
+{
+    /// @brief Class to store and perform operations on 1D Tensors,
+    ///        aka vectors, suitable for use in feed-forward neural
+    ///        networks.
+    class TensorOneD
+    {
+        public:
+            ///@brief Constructor
```
bgeltz:
```suggestion
```
For the standard constructors at least, we usually don't include docstrings.  Only for the constructors/methods that are non-standard.

src/TensorOneD.hpp
```diff
@@ -0,0 +1,104 @@
...
+            TensorOneD();
+            ///@brief Constructor with size specified 
+            ///
+            ///@param [in] n Size of 1D tensor
```
bgeltz:
```suggestion
            /// @brief Constructor with size specified 
            ///
            /// @param [in] n Size of 1D tensor
```
Please add a space after ```///```.  Do this for all docstrings.

src/TensorOneD.hpp
```diff
@@ -0,0 +1,104 @@
...
+            ///
+            ///@param [in] TensorOneD Tensor to copy
+            TensorOneD(const TensorOneD&);
+
+            TensorOneD(TensorOneD &&other);
+            ///@brief Constructor inputting from external JSON
+            ///
+            ///This function expects a Json array.
+            ///
+            ///@param [in] input json11::Json instance
```
bgeltz:
```suggestion
            /// @param [in] input json11::Json input The JSON array
```

src/TensorOneD.hpp
```diff
@@ -0,0 +1,104 @@
...
+            ///will be raised if the lengths do not match.
+            ///
+            ///@param [in] other The summand
+            ///
+            ///@return Returns a 1D tensor, the sum of two 1D tensors
+            TensorOneD operator+(const TensorOneD& other);
+            ///@brief Subtract two 1D tensors, element-wise
+            ///
+            ///The tensors need to be the same length. A geopm::Exception
+            ///will be raised if the lengths do not match.
```
bgeltz:
```suggestion
            /// @throws geopm::Exception if the lengths do not match.
```
Take a look at CNLIOGroup.hpp and/or PlatformIO.hpp for more examples of the docstring directives.

src/TensorOneD.hpp
```diff
@@ -0,0 +1,104 @@
...
+            ///
+            ///@param [in] other The subtrahend
+            ///
+            ///@return Returns a 1D tensor, the difference of two 1D tensors
```
bgeltz:
```suggestion
            /// @return A 1D tensor, the difference of two 1D tensors.
```

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+    float TensorOneD::operator[] (std::size_t idx) const
+    {
+        return m_vec.at(idx);
+    }
+
+    TensorOneD TensorOneD::sigmoid() const
+    {
+        TensorOneD rval(m_vec.size());
+        for(std::size_t idx = 0; idx < m_vec.size(); idx++) {
+            rval[idx] = 1/(1 + expf(-(m_vec.at(idx))));
```
bgeltz: Seems like there's potential for a divide-by-zero error here if the denominator evaluates to 0.  Is this possible?  Should error checks be added here?

bgeltz (changes requested):
Looks like CI is failing because this will not build as-is.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+    }
+
+    TensorOneD::TensorOneD(json11::Json input)
+    {
+        if (!input.is_array()) {
+            throw geopm::Exception("Neural network weights is non-array-type.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        if (input.array_items().size() == 0) {
```
Krebrov001: You could shorten this even more by doing
```cpp
if (input.array_items().empty()) {}
```

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+            throw geopm::Exception("Empty array is invalid for neural network weights.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        set_dim(input.array_items().size());
+
+        for (std::size_t idx = 0; idx < m_vec.size(); idx++) {
```
Krebrov001: You can pull the `.size()` out of the loop to avoid it being called upon each iteration. Also `++idx` is considered as more efficient.

```cpp
std::size_t m_vec_size = m_vec.size();
for (std::size_t idx = 0; idx < m_vec_size; ++idx) {}
```

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+        return m_vec.size();
+    }
+
+    void TensorOneD::set_dim(std::size_t dim)
+    {
+        m_vec.resize(dim);
+    }
+
+    TensorOneD TensorOneD::operator+(const TensorOneD& other)
```
Krebrov001: What happens if `other` and `*this` are the same object?

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        TensorOneD rval(m_vec.size());
+        std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.m_vec.begin(), std::plus<float>());
+
+        return std::move(rval);
+    }
+
+    TensorOneD TensorOneD::operator-(const TensorOneD& other)
```
Krebrov001: What happens if `other` and `*this` are the same object?
Can we add a check?
```cpp
if (&other != this) {}
```

test/TensorOneDTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    one[1] = 2;
+    two[0] = 3;
+    two[1] = 4;
+    three[0] = 0;
+    three[1] = 1;
+    three[2] = 1;
+}
+
+TEST_F(TensorOneDTest, test_sum)
+{
```
Krebrov001, line 23: Can we have a test when you try to add or subtract two instances of the same object?

```cpp
TensorOneD another_two = one + one;
```

test/TensorOneDTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    EXPECT_EQ(1, one[0]);
+    EXPECT_EQ(9, two[0]);
+}
+
+TEST_F(TensorOneDTest, test_input)
+{
+    TensorOneD x(3);
+    x.set_dim(4);
+    std::vector<float> vals = {8, 16};
+    x = TensorOneD(json11::Json(vals));
```
Krebrov001: A suggestion for an interface upgrade.
How would you like to add a constructor for `TensorOneD` which takes that vector in directly as a parameter?
```cpp
x = TensorOneD(vals);
```
We could have the json constructor call the vector constructor internally, since all the underlying logic is the same. What do you think?

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+
+#include "config.h"
+
+#include "TensorOneD.hpp"
+
+#include <math.h>
```
Krebrov001: In the C programming language, we add `.h` to the end of the name of standard headers.
In the C++ programming language, we add `c` to the front of the name of standard headers.
```cpp
#include <cmath>
```

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    void TensorOneD::set_dim(std::size_t dim)
+    {
+        m_vec.resize(dim);
+    }
+
+    TensorOneD TensorOneD::operator+(const TensorOneD& other)
+    {
+        if (get_dim() != other.get_dim()) {
```
Krebrov001: You can avoid an additional function call stack frame by getting the vector's size directly. You can do this, since you're inside the member function of a class, you have access to all the fields, of both `*this` object, and the `other` object as well.
```cpp
if (m_vec.size() != other.m_vec.size()) {}
```
If you think that `get_dim()` is more readable, then that's perfectly fine.

src/TensorOneD.cpp
```diff
@@ -0,0 +1,139 @@
...
+                throw geopm::Exception("Non-numeric type found in neural network weights.\n",
+                                       GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+            m_vec[idx] = input[idx].number_value();
+        }
+    }
+
+    std::size_t TensorOneD::get_dim() const
+    {
+        return m_vec.size();
```
Krebrov001, line 54: Both `set_dim()` and `get_dim()` are one liners. They can be converted into *inline* functions by specifying the body in the `*.h` file, instead of the `*.cpp` file. It will save you an extra function call stack frame.  :)

src/TensorOneD.cpp
asmaalrawi: I don't think this is possible. e^-x is always positive (zero if underflow) so the 1+e^-x in denominator should be fine.

src/TensorOneD.cpp
asmaalrawi: Does get_dim() being inline resolve this?

src/TensorOneD.cpp
asmaalrawi: Added the test for this and subtraction. It looks like it operates as expected and desired.

test/TensorOneDTest.cpp
asmaalrawi: Would this just be for tests?

src/TensorOneD.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <functional>
+#include <numeric>
+#include <utility>
+
+#include "geopm/Exception.hpp"
+
+namespace geopm
+{
+
+
```
avilcheslopez: Nit: delete this white space

src/TensorOneD.hpp
```diff
@@ -0,0 +1,106 @@
...
+            /// @param [in] input json11::Json input The JSON array
+            TensorOneD(json11::Json input);
+            ///@brief Set length of 1D tensor
+            ///
+            ///If the instance has more than n elements, it will
+            ///be truncated. If it has fewer than n elements, the
+            ///tensor will be expanded to a total size of n, uninitialized.
+            ///
+            ///@throws geopm::Exception if input is not an array,
+            ///if input is empty, or if a non-numeric type is found in input.
```
avilcheslopez: It seems this comment applies to the constructor above.

test/TensorOneDTest.cpp
asmaalrawi: Added!

test/TensorOneDTest.cpp
avilcheslopez: If possible, I think it would be a good idea to decouple these Tensor classes from json11, so that they're not dependent on a particular JSON library (since the concept they're encapsulating is independent of JSON parsing). I figure these classes take json11 objects as that's what might be convenient based on how the higher-level classes use them. I'm fine with keeping it as-is for now and revisiting this when reviewing the higher-level classes, to see if there's opportunity to isolate the JSON parsing to a single centralized place and decouple these classes from those details.

avilcheslopez (dismissed):

src/TensorOneD.hpp
asmaalrawi: Done

src/TensorOneD.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "geopm/json11.hpp"
+#include <cstddef>
+
+namespace geopm
+{
+    /// @brief Class to store and perform operations on 1D Tensors,
+    ///        aka vectors, suitable for use in feed-forward neural
+    ///        networks.
+    class TensorOneD
```
bgeltz, line 20: I'm just now realizing we need to have a ```TensorOneD``` abstract base class that ```TensorOneDImp``` implements.

bgeltz (changes requested):

src/TensorOneD.cpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+        m_vec.resize(dim);
+    }
+
+    std::size_t TensorOneD::get_dim() const
+    {
+        return m_vec.size();
+    }
+
+    TensorOneD::TensorOneD(json11::Json input)
```
Krebrov001: It doesn't appear that the `input` parameter is changed anywhere in the constructor. Let us change it to a const reference, to avoid copying the entire object as a parameter.
```cpp
TensorOneD::TensorOneD(const json11::Json &input)
```

src/TensorOneD.hpp
```diff
@@ -0,0 +1,99 @@
...
+            ///
+            /// @param [in] TensorOneD Tensor to copy
+            TensorOneD(const TensorOneD&);
+
+            TensorOneD(TensorOneD &&other);
+            /// @param [in] input json11::Json input The JSON array
+            ///
+            /// @throws geopm::Exception if input is not an array,
+            /// if input is empty, or if a non-numeric type is found in input.
+            TensorOneD(json11::Json input);
```
Krebrov001: Avoid making a copy every time we pass in the object.
```cpp
TensorOneD(const json11::Json &input);
```

src/TensorOneD.cpp
```diff
@@ -0,0 +1,140 @@
...
+        return *this;
+    }
+
+    TensorOneD& TensorOneD::operator=(TensorOneD &&other)
+    {
+        if (&other == this) {
+            return *this;
+        }
+
+        m_vec = std::move(other.m_vec);
```
Krebrov001, line 69: Is there any particular reason why we want to use `std::move()`? Please correct me if I'm wrong, but doesn't `std::move()` transfer the information from the `other.m_vec` to the `m_vec`? Would the object/variable `other.m_vec` still be accessible after the `std::move()`? Does the `std::move()` modify the `other` object or it's internal data members in any way?

I guess what I'm trying to say is, what is the rationale for using
```cpp
m_vec = std::move(other.m_vec);
```
over using
```cpp
m_vec = other.m_vec;
```
as in the other `operator=()`?

src/TensorOneD.cpp
```diff
@@ -0,0 +1,140 @@
...
+            throw geopm::Exception("Inner product of vectors of mismatched dimensions.",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        return std::inner_product(m_vec.begin(), m_vec.end(), other.m_vec.begin(), 0);
+    }
+
+    TensorOneD& TensorOneD::operator=(const TensorOneD &other)
+    {
+        m_vec = other.m_vec;
```
Krebrov001, line 59: Do we need to also include the check above the body of this function as well?
```cpp
        if (&other == this) {
            return *this;
        }
```

Issue #2625, opened 2022-09-14
Add chapter 7 man page geopm_pio_const_config.7


cmcantalupo:
> estimate 1

Define how signals and controls are configured, about the required iogroup name prefix and how the environment variable works to locate file path.

Issue #2624, opened 2022-09-14
Write tests that assert that all signals provided by the IOGroup are prefixed with `CONST_CONFIG::` regardless of user input.


cmcantalupo:
> estimate 1

This important feature of the IOGroup must be tested because without this safeguard, users would be able to overwrite high level alias signals with any value desired.

Issue #2623, opened 2022-09-14
Write unit tests that cover the non-empty base IOGroup methods.


cmcantalupo:
> estimate 1

Cover all the implementations from issue #2622 with unit tests.  Be sure to include all appropriate negative testing

Issue #2622, opened 2022-09-14
Implement other IOGroup methods including `read_signal()`, `push_signal()`, and `sample()`


cmcantalupo:
> estimate 2

The parser should be called in the constructor, and covered in issue #2620.  This task is about using the result of that parsed data to implement other IOGroup requirements.

PR #2635, opened 2022-09-20
Implement ConstConfigIOGroup

### LINKED ISSUE #2619: Define JSON schema for configuration file
> estimate 1

Done criteria:
- Define the schema for the configuration file and commit the schema to the schema's directory.
- Add documentation about the schema to the README and geopm_pio_const_config.7.rst man page.

avilcheslopez:

- Relates to #2616 
- Fixes #2619, #2625, #2620, #2621, #2622, #2623, #2624, #2660

### DIFF
--- a/copying_headers/MANIFEST.EXEMPT
+++ b/copying_headers/MANIFEST.EXEMPT
@@ -130,8 +130,8 @@ service/docs/source/geopm_imbalancer.3.rst
 service/docs/source/geopmlaunch.1.rst
 service/docs/source/geopm_pio.3.rst
 service/docs/source/geopm_pio.7.rst
-service/docs/source/geopm_pio_const_config.7.rst
 service/docs/source/geopm_pio_cnl.7.rst
+service/docs/source/geopm_pio_const_config.7.rst
 service/docs/source/geopm_pio_cpuinfo.7.rst
 service/docs/source/geopm_pio_dcgm.7.rst
 service/docs/source/geopm_pio_levelzero.7.rst
--- a/copying_headers/MANIFEST.EXEMPT
+++ b/copying_headers/MANIFEST.EXEMPT
@@ -130,6 +130,7 @@ service/docs/source/geopm_imbalancer.3.rst
 service/docs/source/geopmlaunch.1.rst
 service/docs/source/geopm_pio.3.rst
 service/docs/source/geopm_pio.7.rst
+service/docs/source/geopm_pio_const_config.7.rst
 service/docs/source/geopm_pio_cnl.7.rst
 service/docs/source/geopm_pio_cpuinfo.7.rst
 service/docs/source/geopm_pio_dcgm.7.rst
--- a/copying_headers/MANIFEST.EXEMPT
+++ b/copying_headers/MANIFEST.EXEMPT
@@ -63,6 +63,8 @@ integration/experiment/uncore_frequency_sweep/README.md
 integration/README.md
 integration/requirements.txt
 integration/test/README.md
+json_schemas/geopmbench_config.schema.json
+json_schemas/geopmagent_policy.schema.json
 MANIFEST
 pull_request_template.md
 README
@@ -174,6 +176,10 @@ service/m4
 service/README.rst
 service/requirements.txt
 service/integration/README.md
+service/json_schemas/active_sessions.schema.json
+service/json_schemas/const_config_io.schema.json
+service/json_schemas/msrs.schema.json
+service/json_schemas/saved_controls.schema.json
 shell_completion/README.md
 specs/geopm-dudley.spec.in
 specs/geopm-ohpc.spec.in
@@ -200,8 +206,3 @@ tutorial/README.md
 tutorial/python_agents/README.rst
 tutorial/tutorial_6_config.json
 tutorial/tutorial_power_policy.json
-json_schemas/geopmbench_config.schema.json
-json_schemas/geopmagent_policy.schema.json
-service/json_schemas/saved_controls.schema.json
-service/json_schemas/msrs.schema.json
-service/json_schemas/active_sessions.schema.json
--- a/service/docs/source/geopm_pio_const_config.7.rst
+++ b/service/docs/source/geopm_pio_const_config.7.rst
@@ -6,17 +6,19 @@ Description
 
 ConstConfigIOGroup implements the
 :doc:`geopm::IOGroup(3) <GEOPM_CXX_MAN_IOGroup.3>` interface to provide a
-predefined set of signals by the user with constant values. This is mainly
-useful to provide configuration information needed by some agents that's
-constant in nature, for example, system characterization information.
+predefined set of signals with constant values. This is mainly useful to
+provide configuration information needed by some agents (e.g. system
+characterization information).
 
 Signals
 -------
 
 This IOGroup doesn't provide any signals by default. In order to define
 signals, a configuration file in JSON format needs to be provided to the
-IOGroup via the ``GEOPM_CONST_CONFIG_PATH`` environment variable and it must
-comply with the following schema:
+IOGroup via the ``GEOPM_CONST_CONFIG_PATH`` environment variable. If not
+provided, it will default to looking for a file named ``const_config_io.json``
+in the GEOPM configuration path (``/etc/geopm``). In either case, the
+configuration file must comply with the following schema:
 
 .. literalinclude:: ../../json_schemas/const_config_io.schema.json
     :language: json
--- a/service/docs/source/geopm_pio_const_config.7.rst
+++ b/service/docs/source/geopm_pio_const_config.7.rst
@@ -1,0 +1,80 @@
+geopm_pio_const_config(7) -- Signals for ConstConfigIOGroup
+===========================================================
+
+Description
+-----------
+
+ConstConfigIOGroup implements the
+:doc:`geopm::IOGroup(3) <GEOPM_CXX_MAN_IOGroup.3>` interface to provide a
+predefined set of signals by the user with constant values. This is mainly
+useful to provide configuration information needed by some agents that's
+constant in nature, for example, system characterization information.
+
+Signals
+-------
+
+This IOGroup doesn't provide any signals by default. In order to define
+signals, a configuration file in JSON format needs to be provided to the
+IOGroup via the ``GEOPM_CONST_CONFIG_PATH`` environment variable and it must
+comply with the following schema:
+
+.. literalinclude:: ../../json_schemas/const_config_io.schema.json
+    :language: json
+
+Notice how all signal fields are required. The ``values`` field is an array
+that should contain a value per domain item. The name of the signal will be
+derived from the object's name and prefixing it with ``CONST_CONFIG::``.
+
+The following example helps illustrate the schema:
+
+.. code-block:: JSON
+
+    {
+        "GPU_CORE_FREQUENCY_MAX": {
+            "domain": "gpu",
+            "description": "Defines the max core frequency to use for available GPUs",
+            "units": "hertz",
+            "aggregation": "average",
+            "values": [1200, 1300, 1500]
+        }
+    }
+
+This basic example only defines one signal. Its name, once processed by the
+IOGroup, will be: ``CONST_CONFIG::GPU_CORE_FREQUENCY_MAX``. Also, it can be
+inferred that for the system being characterized, there are three available
+GPUs, each with the specified max frequency.
+
+Controls
+--------
+
+Due to the nature of this IOGroup, it doesn't provide any controls.
+
+Aliases
+-------
+
+To avoid a configuration from introducing signals that may collide with
+signals provided by other IOGroups, this IOGroup does not provide any
+high-level aliases.
+
+Example
+-------
+
+Following the example configuration provided earlier, ``geopmread`` could be
+used to read configuration values for the defined signal, for example:
+
+``geopmread CONST_CONFIG::GPU_CORE_FREQUENCY_MAX gpu 1``
+
+This example would return the signal value at domain index 1 (1300).
+
+Using ``geopmread`` like this would be useful for manually checking the loaded
+configuration values from a configuration file, but the main use case for this
+will be to provide agents with configuration information.
+
+See Also
+--------
+
+:doc:`geopm_pio(7) <geopm_pio.7>`\ ,
+:doc:`geopm(7) <geopm.7>`\ ,
+:doc:`geopm::IOGroup(3) <GEOPM_CXX_MAN_IOGroup.3>`\ ,
+:doc:`geopmwrite(1) <geopmwrite.1>`\ ,
+:doc:`geopmread(1) <geopmread.1>`
--- a/service/json_schemas/const_config_io.schema.json
+++ b/service/json_schemas/const_config_io.schema.json
@@ -3,51 +3,37 @@
     "id": "https://geopm.github.io/const_config_io.schema.json",
     "title": "ConstConfigIOGroup config file specification",
     "definitions": {
-        "signal_def": {
+        "signal_info": {
             "type": "object",
             "properties": {
                 "description": {
                     "type": "string"
+                },
+                "units": {
+                    "enum": ["none", "seconds", "hertz", "watts", "joules", "celsius"]
                 },
                 "domain": {
                     "enum": ["board", "package", "core", "cpu", "memory", "package_integrated_memory", "nic", "package_integrated_nic", "gpu", "package_integrated_gpu", "gpu_chip"]
                 },
                 "aggregation": {
                     "enum": ["sum", "average", "median", "logical_and", "logical_or", "region_hash", "region_hint", "min", "max", "stddev", "select_first", "expect_same"]
-                }
-            },
-            "required": ["description", "domain", "aggregation"],
-            "additionalProperties": false
-        },
-        "config_def": {
-            "type": "object",
-            "properties": {
-                "domain_idx": {
-                    "type": "integer"
                 },
-                "value": {
-                    "type": "number"
+                "values": {
+                    "type": "array",
+                    "items": {
+                        "type": "number"
+                    },
+                    "minItems": 1
                 }
             },
-            "required": ["value"],
+            "required": ["description", "units", "domain", "aggregation", "values"],
             "additionalProperties": false
         }
     },
     "type": "object",
     "properties": {
-        "signals": {
-            "type": "object",
-            "additionalProperties": {
-                "$ref": "#/definitions/signal_def"
-            }
-        },
-        "config": {
-            "type": "object",
-            "additionalProperties": {
-                "$ref": "#/definitions/config_def"
-            }
+        "additionalProperties": {
+            "$ref": "#/definitions/signal_info"
         }
-    },
-    "required": ["signals", "config"],
-    "additionalProperties": false
+    }
 }
--- a/service/json_schemas/const_config_io.schema.json
+++ b/service/json_schemas/const_config_io.schema.json
@@ -1,0 +1,53 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id": "https://geopm.github.io/const_config_io.schema.json",
+    "title": "ConstConfigIOGroup config file specification",
+    "definitions": {
+        "signal_def": {
+            "type": "object",
+            "properties": {
+                "description": {
+                    "type": "string"
+                },
+                "domain": {
+                    "enum": ["board", "package", "core", "cpu", "memory", "package_integrated_memory", "nic", "package_integrated_nic", "gpu", "package_integrated_gpu", "gpu_chip"]
+                },
+                "aggregation": {
+                    "enum": ["sum", "average", "median", "logical_and", "logical_or", "region_hash", "region_hint", "min", "max", "stddev", "select_first", "expect_same"]
+                }
+            },
+            "required": ["description", "domain", "aggregation"],
+            "additionalProperties": false
+        },
+        "config_def": {
+            "type": "object",
+            "properties": {
+                "domain_idx": {
+                    "type": "integer"
+                },
+                "value": {
+                    "type": "number"
+                }
+            },
+            "required": ["value"],
+            "additionalProperties": false
+        }
+    },
+    "type": "object",
+    "properties": {
+        "signals": {
+            "type": "object",
+            "additionalProperties": {
+                "$ref": "#/definitions/signal_def"
+            }
+        },
+        "config": {
+            "type": "object",
+            "additionalProperties": {
+                "$ref": "#/definitions/config_def"
+            }
+        }
+    },
+    "required": ["signals", "config"],
+    "additionalProperties": false
+}
--- a/service/src/ConstConfigIOGroup.cpp
+++ b/service/src/ConstConfigIOGroup.cpp
@@ -1,0 +1,206 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "ConstConfigIOGroup.hpp"
+
+#include <cmath>
+
+#include <iostream>
+#include <iterator>
+
+#include "geopm/Helper.hpp"
+#include "geopm/PlatformTopo.hpp"
+#include "geopm/Exception.hpp"
+#include "geopm/Agg.hpp"
+#include "geopm_topo.h"
+#include "config.h"
+
+#define GEOPM_CONSTCONFIG_IO_GROUP_PLUGIN_NAME "CONSTCONFIG"
+
+namespace geopm
+{
+    ConstConfigIOGroup::ConstConfigIOGroup()
+    {
+        // TODO
+    }
+
+    std::set<std::string> ConstConfigIOGroup::signal_names(void) const
+    {
+        std::set<std::string> result;
+        for (const auto &sv : m_signal_available) {
+            result.insert(sv.first);
+        }
+        return result;
+    }
+
+    std::set<std::string> ConstConfigIOGroup::control_names(void) const
+    {
+        return {};
+    }
+
+    bool ConstConfigIOGroup::is_valid_signal(const std::string &signal_name) const
+    {
+        return m_signal_available.find(signal_name) != m_signal_available.end();
+    }
+
+    bool ConstConfigIOGroup::is_valid_control(const std::string &control_name) const
+    {
+        return false;
+    }
+
+    int ConstConfigIOGroup::signal_domain_type(const std::string &signal_name) const
+    {
+        int result = GEOPM_DOMAIN_INVALID;
+        // TODO
+        return result;
+    }
+
+    int ConstConfigIOGroup::control_domain_type(const std::string &control_name) const
+    {
+        return GEOPM_DOMAIN_INVALID;
+    }
+
+    int ConstConfigIOGroup::push_signal(const std::string &signal_name, int domain_type, int domain_idx)
+    {
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("ConstConfigIOGroup::push_signal(): " + signal_name +
+                            "not valid for ConstConfigIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        // TODO
+        return 0;
+    }
+
+    int ConstConfigIOGroup::push_control(const std::string &control_name, int domain_type, int domain_idx)
+    {
+        throw Exception("ConstConfigIOGroup::push_control(): there are no controls supported by the ConstConfigIOGroup",
+                        GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+    }
+
+    void ConstConfigIOGroup::read_batch(void)
+    {
+        // TODO: enforce proper behavior/sequence of calls: push_signal(),
+        // read_batch(), sample()?
+    }
+
+    void ConstConfigIOGroup::write_batch(void)
+    {
+    }
+
+    double ConstConfigIOGroup::sample(int batch_idx)
+    {
+        double result = NAN;
+        // TODO
+        return result;
+    }
+
+    void ConstConfigIOGroup::adjust(int batch_idx, double setting)
+    {
+        throw Exception("ConstConfigIOGroup::adjust(): there are no controls supported by the ConstConfigIOGroup",
+                        GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+    }
+
+    double ConstConfigIOGroup::read_signal(const std::string &signal_name, int domain_type, int domain_idx)
+    {
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("ConstConfigIOGroup::read_signal(): " + signal_name +
+                            "not valid for ConstConfigIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        // TODO
+        return NAN;
+    }
+
+    void ConstConfigIOGroup::write_control(const std::string &control_name, int domain_type, int domain_idx, double setting)
+    {
+        throw Exception("ConstConfigIOGroup::write_control(): there are no controls supported by the ConstConfigIOGroup",
+                        GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+    }
+
+    void ConstConfigIOGroup::save_control(void)
+    {
+    }
+
+    void ConstConfigIOGroup::restore_control(void)
+    {
+    }
+
+    std::function<double(const std::vector<double> &)> ConstConfigIOGroup::agg_function(const std::string &signal_name) const
+    {
+        auto it = m_signal_available.find(signal_name);
+        if (it == m_signal_available.end()) {
+            throw Exception("ConstConfigIOGroup::agg_function(): unknown how to aggregate \"" + signal_name + "\"",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        return it->second.agg_function;
+    }
+
+    std::function<std::string(double)> ConstConfigIOGroup::format_function(const std::string &signal_name) const
+    {
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("ConstConfigIOGroup::format_function(): unknown how to format \"" + signal_name + "\"",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        // TODO?
+        return string_format_double;
+    }
+
+    std::string ConstConfigIOGroup::signal_description(const std::string &signal_name) const
+    {
+        auto it = m_signal_available.find(signal_name);
+        if (it == m_signal_available.end()) {
+            throw Exception("ConstConfigIOGroup::signal_description(): signal_name " + signal_name +
+                            " not valid for ConstConfigIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        std::string result;
+        result =  "    description: " + it->second.description + '\n'; // Includes alias_for if applicable
+        result += "    units: " + IOGroup::units_to_string(it->second.units) + '\n';
+        result += "    aggregation: " + Agg::function_to_name(it->second.agg_function) + '\n';
+        result += "    domain: " + platform_topo().domain_type_to_name(GEOPM_DOMAIN_BOARD) + '\n';
+        result += "    iogroup: ConstConfigIOGroup";
+        return result;
+    }
+
+    std::string ConstConfigIOGroup::control_description(const std::string &control_name) const
+    {
+        throw Exception("ConstConfigIOGroup::control_description: there are no controls supported by the ConstConfigIOGroup",
+                        GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+    }
+
+    int ConstConfigIOGroup::signal_behavior(const std::string &signal_name) const
+    {
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("ConstConfigIOGroup::signal_behavior(): signal_name " + signal_name +
+                            " not valid for ConstConfigIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        return IOGroup::M_SIGNAL_BEHAVIOR_CONSTANT;
+    }
+
+    std::string ConstConfigIOGroup::name(void) const
+    {
+        return plugin_name();
+    }
+
+    std::string ConstConfigIOGroup::plugin_name(void)
+    {
+        return GEOPM_CONSTCONFIG_IO_GROUP_PLUGIN_NAME;
+    }
+
+    std::unique_ptr<IOGroup> ConstConfigIOGroup::make_plugin(void)
+    {
+        return geopm::make_unique<ConstConfigIOGroup>();
+    }
+
+    void ConstConfigIOGroup::save_control(const std::string &save_path)
+    {
+    }
+
+    void ConstConfigIOGroup::restore_control(const std::string &save_path)
+    {
+    }
+}
--- a/service/src/ConstConfigIOGroup.cpp
+++ b/service/src/ConstConfigIOGroup.cpp
@@ -343,20 +343,16 @@ namespace geopm
             }
 
             std::string description = properties["description"].string_value();
-            if (m_signal_available.find(name) == m_signal_available.end()) {
-                m_signal_available[name] =
-                    std::make_shared<m_signal_info_s>(
-                        units,
-                        domain_type,
-                        agg_func,
-                        description,
-                        values);
-            }
-            else {
-                throw Exception("ConstConfigIOGroup::parse_config_json(): "
-                                "duplicate signal found: " + name,
-                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-            }
+            // TODO: check for duplicate signals. At the moment, we're using
+            // json11 to parse JSON strings, which handles duplicate entries
+            // by taking the latest entry encountered.
+            m_signal_available[name] =
+                std::make_shared<m_signal_info_s>(
+                    units,
+                    domain_type,
+                    agg_func,
+                    description,
+                    values);
         }
     }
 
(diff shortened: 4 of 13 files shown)


service/json_schemas/const_config_io.schema.json
```diff
@@ -0,0 +1,39 @@
...
+                "domain": {
+                    "enum": ["board", "package", "core", "cpu", "memory", "package_integrated_memory", "nic", "package_integrated_nic", "gpu", "package_integrated_gpu", "gpu_chip"]
+                },
+                "aggregation": {
+                    "enum": ["sum", "average", "median", "logical_and", "logical_or", "region_hash", "region_hint", "min", "max", "stddev", "select_first", "expect_same"]
+                },
+                "values": {
+                    "type": "array",
+                    "items": {
+                        "type": "number"
```
dannosliwcd, line 24: Consider making it `["number", "null"]` or `["number", "string"]` if you want to allow some parts of a domain to hold no value. I think the latter is more consistent with what we allow for policies. Bonus if you use a pattern-matched string type that accepts variations of `"nan"`.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+            {"description", {Json::STRING, true}},
+            {"units", {Json::STRING, true}},
+            {"domain", {Json::STRING, true}},
+            {"aggregation", {Json::STRING, true}},
+            {"values", {Json::ARRAY, true}}
+        };
+
+    ConstConfigIOGroup::ConstConfigIOGroup()
+    {
+        // TODO
```
dannosliwcd: I think this is sufficient as an empty constructor, since it just means we'll expose no signals, right?

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+    std::string ConstConfigIOGroup::plugin_name(void)
+    {
+        return M_PLUGIN_NAME;
+    }
+
+    std::unique_ptr<IOGroup> ConstConfigIOGroup::make_plugin(void)
+    {
+        return geopm::make_unique<ConstConfigIOGroup>();
+    }
```
dannosliwcd, line 305: How will this change for non-empty const configs? Maybe managed similar to other geopm config files?

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+        
+        if (batch_idx < 0 || batch_idx >= m_pushed_signals.size()) {
+            throw Exception("ConstConfigIOGroup::sample(): batch_idx " +
+                            std::to_string(batch_idx) + " out of range.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        m_signal_ref_s signal = m_pushed_signals[batch_idx];
+        auto it = m_signal_available.begin();
+        std::advance(it, signal.signal_idx);
+        result = it->second.values[signal.domain_idx];
```
dannosliwcd: I think this only works if available signals never changes in between calls to this (which would potentially change the offsets as new keys are inserted/removed). Looks your current implementation does avoid that.

Should we enforce that by making `m_signal_available` a `const` member? You'd need to initialize it in the constructor instead of post-construction. E.g., `ConstConfigIOGroup(const std::string &config) : m_signal_available(something_that_returns_the_map(config))`

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+#include <iostream>
+#include <iterator>
+#include <sstream>
+
+#include "geopm/Helper.hpp"
+#include "geopm/PlatformTopo.hpp"
+#include "geopm/Exception.hpp"
+#include "geopm/Agg.hpp"
+#include "geopm_topo.h"
+#include "config.h"
```
bgeltz:
```suggestion
#include "config.h"

#include "ConstConfigIOGroup.hpp"

#include <cmath>
#include <iostream>
#include <iterator>
#include <sstream>

#include "geopm_topo.h"
#include "geopm/Helper.hpp"
#include "geopm/PlatformTopo.hpp"
#include "geopm/Exception.hpp"
#include "geopm/Agg.hpp"
```

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+        ConstConfigIOGroup::M_SIGNAL_FIELDS = {
+            {"description", {Json::STRING, true}},
+            {"units", {Json::STRING, true}},
+            {"domain", {Json::STRING, true}},
+            {"aggregation", {Json::STRING, true}},
+            {"values", {Json::ARRAY, true}}
+        };
+
+    ConstConfigIOGroup::ConstConfigIOGroup()
+    {
```
bgeltz: I thought this would look something like:
```suggestion
    ConstConfigIOGroup::ConstConfigIOGroup()
        : ConstConfigIOGroup(M_DEFAULT_CONFIG_FILE)
    {
```

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+        // TODO
+    }
+
+    ConstConfigIOGroup::ConstConfigIOGroup(const std::string &config)
+    {
+        parse_config_json(config);
```
bgeltz: Where is the checking to see if the config file exists?  As it is, it seems if you pass a string to the parse method, and the file does not exist that this will throw, and construction will fail.

service/src/ConstConfigIOGroup.cpp
bgeltz: I'm in favor of this suggestion.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+        size_t properties_found = 0;
+        for (const auto &prop : properties) {
+            auto it = M_SIGNAL_FIELDS.find(prop.first);
+            if (it == M_SIGNAL_FIELDS.end()) {
+                throw Exception("ConstConfigIOGroup::parse_config_json():"
+                                " unexpected propety: \"" + prop.first +
+                                "\"", GEOPM_ERROR_INVALID, __FILE__,
+                                __LINE__);
+            } else if (prop.second.type() != it->second.type) {
```
bgeltz:
```suggestion
            }
            else if (prop.second.type() != it->second.type) {
```

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,424 @@
...
+                throw Exception("ConstConfigIOGroup::parse_config_json():"
+                                " incorrect type for property: \"" +
+                                prop.first + "\"", GEOPM_ERROR_INVALID,
+                                __FILE__, __LINE__);
+            } else {
```
bgeltz:
```suggestion
            }
            else {
```

service/src/ConstConfigIOGroup.cpp
bgeltz: Is Helper.hpp needed for anything?  I thought read_file() would be called somewhere, but I don't see it.

service/src/ConstConfigIOGroup.cpp
bgeltz: Also, where is the file reading code?

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+TEST_F(ConstConfigIOGroupTest, is_valid_control)
+{
+    ConstConfigIOGroup iogroup("");
+    EXPECT_EQ(iogroup.is_valid_control("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"), false);
+}
+
+void ConstConfigIOGroupTest::TearDown()
+{
+}
```
bgeltz: Move this under SetUp

bgeltz (changes requested):

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+TEST_F(ConstConfigIOGroupTest, valid_json)
+{
+    std::string json_string = "{"
+    "    \"GPU_CORE_FREQUENCY\": {"
+    "        \"domain\": \"gpu\","
+    "        \"description\": \"Provides GPU core frequency\","
+    "        \"units\": \"hertz\","
+    "        \"aggregation\": \"sum\","
+    "        \"values\": { 1500, 1600, 1700 }"
```
avilcheslopez: Note: don't forget to push the fix for this: use square brackets for an array.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+    "    }"
+    "}";
+    ConstConfigIOGroup iogroup(json_string);
+    // EXPECT_EQ(iogroup.control_names(), std::set<std::string>{});
+}
+
+TEST_F(ConstConfigIOGroupTest, control_names)
+{
+    ConstConfigIOGroup iogroup("");
```
avilcheslopez: For this test, you'll need to pass a valid JSON string to ensure the class loads correctly, and then ensure the ``control_names()`` method returns an empty set. In other words, the behavior we're validating is: given a valid config, ConstConfigIOGroup always returns an empty set for the control names (as controls are not supported).

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+    EXPECT_EQ(iogroup.control_names(), std::set<std::string>{});
+}
+
+TEST_F(ConstConfigIOGroupTest, is_valid_control)
+{
+    ConstConfigIOGroup iogroup("");
```
avilcheslopez: Same comment as before.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+    GEOPM_EXPECT_THROW_MESSAGE(
+        ConstConfigIOGroup iogroup(""),
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::parse_config_json(): "
+        "detected a malformed JSON string"
+    );
+}
+
+TEST_F(ConstConfigIOGroupTest, input_gibberish)
```
avilcheslopez, line 69: This is a good start for the negative tests. In case you don't have these in mind, make sure you cover cases like:
- Signals with missing required fields
- Fields with values of an incorrect type
- Fields with a value outside of their domain/enum
- Duplicate signals
- A signal with an empty "values" array

service/test/ConstConfigIOGroupTest.cpp
avilcheslopez: Make sure to cover more interesting configs with multiple signals.

avilcheslopez (commented):
Tests are looking good! I'm sure you're already working on it, but just in case, make sure to cover all the other methods of the class.

If while writing a test you see the class is not behaving as you'd expect, please let me know, as this might be a bug I need to fix.

service/src/ConstConfigIOGroup.cpp
avilcheslopez: Will reorder the includes, thanks!

The code is currently using ``make_unique()`` from ``Helper.hpp``, but yes: the intention is to add code to load the config from a file (i.e. ``read_file()`` will be called).

service/src/ConstConfigIOGroup.cpp
avilcheslopez: Addressing both (this and Daniel's) comments here:

The feature description calls for accepting a config file path via an environment variable. So the default constructor (called by the ``make_plugin()`` method) will read the config file (perhaps a default path could be used if none provided) and then proceed with parsing. I provided the specialized constructor as a convenience for the unit tests. I might refactor this a bit if I can think of a better solution.

service/src/ConstConfigIOGroup.cpp
avilcheslopez: See my reply to Brad's comment.

service/src/ConstConfigIOGroup.cpp
avilcheslopez: It's coming :). I first wanted to enable Konstantin to start writing unit tests.

service/src/ConstConfigIOGroup.cpp
avilcheslopez: Yeah, I'm not fully convinced with my initial implementation. I'll revisit this soon. If I do stick with this implementation, I'll make sure to follow your suggestion of making ``m_signal_available`` a const member.

service/src/ConstConfigIOGroup.cpp
avilcheslopez: I think I'll be able to get away with still invoking the default constructor once I update it to load the config from a file (which could be provided via an environment variable according to the feature description).

service/src/ConstConfigIOGroup.cpp
avilcheslopez: Thanks!

service/json_schemas/const_config_io.schema.json
avilcheslopez: After reviewing with Brad and Lowren, they were suggesting to simply not allow nulls or nans (i.e. enforce all values are provided). If there's a valid use case to allow nulls/nans, let's revisit.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,466 @@
...
+    "        \"domain\": \"gpu\","
+    "        \"units\": \"hertz\","
+    "        \"values\": [ 1500, 1600, 1700 ]"
+    "    }"
+    "}";
+    GEOPM_EXPECT_THROW_MESSAGE(
+        ConstConfigIOGroup iogroup(json_string),
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::parse_config_json(): "
+        "missing properties for signal \"" "GPU_CORE_FREQUENCY" "\": "
```
avilcheslopez, line 129: Question: does the error message include the missing properties? If not, that's a bug in ConstConfigIOGroup.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::parse_config_json(): "
+        "detected a malformed JSON string"
+    );
+}
+
+TEST_F(ConstConfigIOGroupTest, input_empty_json)
+{
+    ConstConfigIOGroup iogroup("{}");
+    EXPECT_EQ(iogroup.signal_names(), std::set<std::string>{});
```
avilcheslopez, line 65: Check ``control_names()`` also returns an empty set.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+    "        \"units\": \"hertz\","
+    "        \"aggregation\": \"sum\","
+    "        \"values\": [ 1500, 1600, 1700 ]"
+    "    }"
+    "}";
+    ConstConfigIOGroup iogroup(json_string);
+
+    EXPECT_EQ(iogroup.control_names(), std::set<std::string>{});
+    EXPECT_EQ(iogroup.is_valid_signal("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"), false);
+    EXPECT_EQ(iogroup.is_valid_control("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"), false);
```
avilcheslopez, line 498: Also check: ``is_valid_control("CONST_CONFIG::GPU_CORE_FREQUENCY")``

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+    EXPECT_EQ(iogroup.signal_domain_type("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"), GEOPM_DOMAIN_INVALID);
+    EXPECT_EQ(iogroup.control_domain_type("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"), GEOPM_DOMAIN_INVALID);
```
avilcheslopez, line 501: Also check: ``control_domain_type("CONST_CONFIG::GPU_CORE_FREQUENCY")``

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+        "-1" " out of range."
+    );
+    GEOPM_EXPECT_THROW_MESSAGE(
+        iogroup.push_signal("CONST_CONFIG::GPU_CORE_FREQUENCY", GEOPM_DOMAIN_GPU, 3),
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::push_signal(): domain_idx "
+        "3" " out of range."
+    );
+    GEOPM_EXPECT_THROW_MESSAGE(
+        iogroup.push_control("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO", 0, 0),
```
avilcheslopez, line 528: Also check: ``push_control("CONST_CONFIG::GPU_CORE_FREQUENCY")``

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+        "-1" " out of range."
+    );
+    GEOPM_EXPECT_THROW_MESSAGE(
+        iogroup.read_signal("CONST_CONFIG::GPU_CORE_FREQUENCY", GEOPM_DOMAIN_GPU, 3),
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::read_signal(): domain_idx "
+        "3" " out of range."
+    );
+    GEOPM_EXPECT_THROW_MESSAGE(
+        iogroup.write_control("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO", 0, 0, 3.14),
```
avilcheslopez, line 577: Also check: ``write_control("CONST_CONFIG::GPU_CORE_FREQUENCY")``

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+    );
+    GEOPM_EXPECT_THROW_MESSAGE(
+        iogroup.signal_description("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"),
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::signal_description(): "
+        "signal_name " "CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"
+        " not valid for ConstConfigIOGroup"
+    );
+    GEOPM_EXPECT_THROW_MESSAGE(
+        iogroup.control_description("CONST_CONFIG::UNCORE_RATIO_LIMIT:MIN_RATIO"),
```
avilcheslopez, line 608: Also check: ``control_description("CONST_CONFIG::GPU_CORE_FREQUENCY")``

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,565 @@
...
+        GEOPM_ERROR_INVALID,
+        "ConstConfigIOGroup::parse_config_json(): "
+        "detected a malformed JSON string"
+    );
+}
+
+TEST_F(ConstConfigIOGroupTest, input_duplicate_signal)
+{
+    /* When one signal is provided more than once,
+       apparently the latter entry overwrites the data of the former. */
```
avilcheslopez: Suggested change in comment: When one signal is provided more than once, the latest entry will take precedence. This is due to how the underlying JSON library handles duplicate entries.

copying_headers/MANIFEST.EXEMPT
```diff
@@ -128,6 +130,7 @@ service/docs/source/geopm_imbalancer.3.rst
 service/docs/source/geopmlaunch.1.rst
 service/docs/source/geopm_pio.3.rst
 service/docs/source/geopm_pio.7.rst
+service/docs/source/geopm_pio_const_config.7.rst
 service/docs/source/geopm_pio_cnl.7.rst
```
bgeltz:
```suggestion
service/docs/source/geopm_pio_cnl.7.rst
service/docs/source/geopm_pio_const_config.7.rst
```

service/docs/source/geopm_pio_const_config.7.rst
```diff
@@ -0,0 +1,80 @@
...
+===========================================================
+
+Description
+-----------
+
+ConstConfigIOGroup implements the
+:doc:`geopm::IOGroup(3) <GEOPM_CXX_MAN_IOGroup.3>` interface to provide a
+predefined set of signals by the user with constant values. This is mainly
+useful to provide configuration information needed by some agents that's
+constant in nature, for example, system characterization information.
```
bgeltz:
```suggestion
:doc:`geopm::IOGroup(3) <GEOPM_CXX_MAN_IOGroup.3>` interface to provide a
predefined set of signals with constant values to the user. This is mainly
useful to provide configuration information needed by some agents (e.g., system characterization information).
```

service/json_schemas/const_config_io.schema.json
```diff
@@ -0,0 +1,39 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id": "https://geopm.github.io/const_config_io.schema.json",
```
bgeltz, line 3: Reminder to update the website to include this after this PR is merged.

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,174 @@
...
+#include "geopm/IOGroup.hpp"
+
+namespace geopm
+{
+    class ConstConfigIOGroup : public IOGroup
+    {
+        public:
+            ConstConfigIOGroup();
+            ConstConfigIOGroup(const std::string &config);
+            /// @return the list of signal names provided by this IOGroup.
```
bgeltz:
```suggestion
```
It looks like most of these docstrings were boiler plate comments from either the CNLIOgroup or the CpuinfoIOGroup.  Most other IOGroups do not add these docstrings.  IMO they're not adding any benefit here.

If there are any special details about this specific class, please preserve them.  Otherwise I'd remove all the docstrings that were copy/pasted from the mentioned IOGroups.

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,174 @@
...
+                    description(desc),
+                    values(values)
+                {}
+
+                int units;
+                int domain;
+                std::function<double(const std::vector<double> &)> agg_function;
+                std::string description;
+                std::vector<double> values;
+            };
```
bgeltz: I think this can be simplified a bit.  I think there are some redundant/unused fields with the current implementation.  Will this work?

```suggestion            
            struct m_signal_info_s
            {                
                int domain;
                int units;
                std::function<double(const std::vector<double> &)> agg_function;
                std::string description;
                std::vector<double> values;
            };
```

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+            struct m_signal_ref_s {
+                std::shared_ptr<m_signal_info_s> signal_info;
+                int domain_idx;
+            };
+
+            void parse_config_json(const std::string &config);
+            static json11::Json construct_config_json_obj(
+                const std::string &config);
```
bgeltz: If you've going to have lines earlier in the file that go beyond column ~75, please keep that style consistent throughout the file.  I'd still prefer to see a line break if you get up to column 100 or so.

```suggestion
            static json11::Json construct_config_json_obj(const std::string &config);
```

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,174 @@
...
+            static void check_json_root(const json11::Json &root);
+            static void check_json_signal(
+                const json11::Json::object::value_type &signal_obj);
```
bgeltz:
```suggestion
            static void check_json_signal(const json11::Json::object::value_type &signal_obj);
```

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,174 @@
...
+            static void check_json_signal_properties(
+                const std::string &signal_name,
+                const json11::Json::object& properties);
```
bgeltz:
```suggestion
            static void check_json_signal_properties(const std::string &signal_name,
                                                     const json11::Json::object& properties);
```

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,174 @@
...
+            static void check_json_array_value(const json11::Json& val);
+
+            static const std::string M_PLUGIN_NAME;
+            static const std::string M_SIGNAL_PREFIX;
+            static const std::map<std::string, m_signal_desc_s> M_SIGNAL_FIELDS;
+            static const std::string M_CONFIG_PATH_ENV;
+            static const std::string M_DEFAULT_CONFIG_FILE_PATH;
+
+            std::map<std::string, std::shared_ptr<m_signal_info_s> >
+                m_signal_available;
```
bgeltz:
```suggestion
            std::map<std::string, std::shared_ptr<m_signal_info_s> > m_signal_available;
```

service/src/ConstConfigIOGroup.cpp
bgeltz: I think we need to refactor the constructor a bit.  There are some code paths that we cannot test in CI the way it is.  I put a sketch together with a few TODOs.  It's here:
https://github.com/bgeltz/geopm/commit/f51e3e84f4c8d5f19b0c86c6c8c4005b3b470a0c
https://github.com/bgeltz/geopm/tree/issue-2619

With a refactor like I suggested, this is easier to test and maintain.  Take a look and let me know if you have questions.  The unit tests should create real/on-disk input files that are parsed here.  Both paths should be tested (env specified file path and default path).

bgeltz (changes requested):

service/docs/source/geopm_pio_const_config.7.rst
avilcheslopez: Just to clarify, the reason for saying the signals are provided by the user is to explain that this is a user-defined set of signals. Maybe there's a better way of wording that. Maybe it should be mentioned that the administrator is meant to generate the configuration.

The concern I have with the suggestion is that technically the signals are provided to the IOGroup, not the user (unless I'm misunderstanding what you meant).

service/src/ConstConfigIOGroup.hpp
avilcheslopez: Will do! I'll keep the ones that document behavior that's particular to this IOGroup.

service/src/ConstConfigIOGroup.hpp
avilcheslopez: Wait, it seems your suggestion is to get rid of the constructor (i.e. I'm not seeing changes related to potentially redundant fields). The constructor is needed as a convenience when calling ``std::make_shared<m_signal_info_s>(...)``.

service/src/ConstConfigIOGroup.hpp
avilcheslopez: Yeah, let me see if I can clean it up for consistency.

service/json_schemas/const_config_io.schema.json
Krebrov001: It's not clear to me what needs to be done here.

service/docs/source/geopm_pio_const_config.7.rst
bgeltz: I think we need to be careful how we document this.  This IOGroup allows for signals to be either: user-defined (i.e., the user wrote a config file and set GEOPM_CONST_CONFIG_PATH to point to the file they wrote) or admin defined (i.e., GEOPM_CONST_CONFIG_PATH unset but the default file exists and has entries), or neither, or both user and admin defined.

It should be clear in the docu that:
   a. This could be leveraged by the user, the sysadmin, neither, or both.
   b. How we decide who wins if both files exist, and the env is set properly, or if only one of the files exists, etc.

bgeltz (changes requested):

service/docs/source/geopm_pio_const_config.7.rst
avilcheslopez: Yes. That needs to be addressed (realized later that the documentation needed to be updated to reflect the latest changes in the code). Just wasn't sure about your original suggestion about a wording change.

service/src/ConstConfigIOGroup.cpp
Krebrov001: @bgeltz Do you want me to write the unit tests that create real/on-disk input files?

service/src/ConstConfigIOGroup.hpp
bgeltz: Is there a reason that ```m_signal_available``` is defined that way (having a shared pointer to the signal info struct?  In all other IOGroups I looked at, the definition is just:
```
std::map<std::string, m_signal_info_s> m_signal_available;
```

I don't see the need to make ```m_signal_info_s``` a shared_ptr.  It is never passed outside this class.

Take a look at how ```m_signal_available``` is defined and used in the ```MSRIOGroup```.

bgeltz (changes requested):

service/src/ConstConfigIOGroup.cpp
bgeltz: Yes

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,451 @@
...
+    ConstConfigIOGroup::ConstConfigIOGroup(const std::string &user_file_path,
+                                           const std::string &default_file_path)
+    {
+        std::string config_json;
+        bool load_default = true;
+        if (!user_file_path.empty()) {
+            try {
+                config_json = geopm::read_file(user_file_path);
+                load_default = false;
+            } catch (...) {
```
bgeltz: Catch the specific type of exception you expect here: ```const geopm::Exception &ex```

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,451 @@
...
+    void ConstConfigIOGroup::write_batch(void)
+    {
+    }
+
+    double ConstConfigIOGroup::sample(int batch_idx)
+    {
+        double result = NAN;
+        
+        if (batch_idx < 0 ||
+            static_cast<std::size_t>(batch_idx) >= m_pushed_signals.size()) {
```
bgeltz, line 171:
```suggestion
            batch_idx >= static_cast<int>(m_pushed_signals.size())) {
```
AFAICT, all the other IOGroups do it this way.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,451 @@
...
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        else if (domain_type != it->second->domain) {
+            throw Exception("ConstConfigIOGroup::push_signal(): domain_type " +
+                            std::to_string(domain_type) +
+                            " not valid for ConstConfigIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        else if (domain_idx < 0 ||
+            static_cast<std::size_t>(domain_idx) >= it->second->values.size()) {
```
bgeltz:
```suggestion
            batch_idx >= static_cast<int>(it->second->values.size()) {
```

bgeltz: AFIACT the other IOGroups all do it this way.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,451 @@
...
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        else if (domain_type != it->second->domain) {
+            throw Exception("ConstConfigIOGroup::read_signal(): domain_type " +
+                            std::to_string(domain_type) +
+                            " not valid for ConstConfigIOGroup",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        else if (domain_idx < 0 ||
+            static_cast<std::size_t>(domain_idx) >= it->second->values.size()) {
```
bgeltz:
```suggestion
            domain_idx >= static_cast<int>(it->second->values.size()) {
```

bgeltz: Also, this line should be indented a bit more, 1 space beyond the open paren on the line above.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,451 @@
...
+        if (!is_valid_signal(signal_name)) {
+            throw Exception("ConstConfigIOGroup::format_function(): unknown "
+                            "how to format \"" + signal_name + "\"",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        return string_format_double;
+    }
+
+    std::string ConstConfigIOGroup::signal_description(
+        const std::string &signal_name) const
```
bgeltz:
```suggestion
    std::string ConstConfigIOGroup::signal_description(const std::string &signal_name) const
```

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    void ConstConfigIOGroup::check_json_root(const Json &root)
+    {
+        if (!root.is_object()) {
+            throw Exception("ConstConfigIOGroup::parse_config_json(): "
+                            "the root must be an object",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+    }
```
bgeltz: I don't see the benefit of having these methods separated out from ```parse_config_json()```.  Take a look at some of the other places in the code base we parse json:
    ```FilePolicy::parse_json()```
    ```SaveControlImp::settings()```

Those both have code of the form:
```if (!err.empty() || !root.is_object()) {```

If you agree, please combine all of these error checks into a more condensed form, and put the code in ```parse_config_json()```.

File: service/src/ConstConfigIOGroup.cpp [hunk 38]
```diff
@@ -0,0 +1,451 @@
...
+            throw Exception("ConstConfigIOGroup::parse_config_json(): "
+                            "invalid signal: \"" + signal_obj.first + "\" ("
+                            "object expected)", GEOPM_ERROR_INVALID,
+                            __FILE__, __LINE__);
+        }
+    }
+
+    void ConstConfigIOGroup::check_json_signal_properties(
+        const std::string &signal_name,
+        const Json::object& properties)
```
bgeltz:
```suggestion
    void ConstConfigIOGroup::check_json_signal_properties(const std::string &signal_name,
                                                          const Json::object &properties)
```

service/src/ConstConfigIOGroup.hpp
```diff
@@ -0,0 +1,115 @@
...
+                int domain_idx;
+            };
+
+            void parse_config_json(const std::string &config);
+            static json11::Json construct_config_json_obj(const std::string &config);
+            static void check_json_root(const json11::Json &root);
+            static void check_json_signal(const json11::Json::object::value_type &signal_obj);
+            static void check_json_signal_properties(const std::string &signal_name,
+                                                     const json11::Json::object& properties);
+            static void check_json_array_value(const json11::Json& val);
```
bgeltz: Let's talk about what all we really need to expose as a static function at this time.

bgeltz (changes requested):

service/src/ConstConfigIOGroup.cpp
avilcheslopez: Is there a reason why this would be preferred? I'm thinking casting ``int`` to ``unsigned int`` (after we've ensured it's >= 0) is safer than casting ``unsigned int`` to ``int``. I highly doubt the ``values`` vector would ever be so large for that to matter (under normal use cases), but I would still prefer to do the safer cast (in case of malicious input).

service/src/ConstConfigIOGroup.cpp
avilcheslopez: I wanted to separate the error checking from the actual parsing to make the parsing code easier to read (by removing the clutter of all the error checking, and keeping the function short and focused on the parsing).

I originally looked at MSRIOGroup's parse method and had a bit of a hard time following it because it's somewhat lengthy, and has some clutter as a result of the checks. A bit more distracting is that it has some of the checks inline while it farms out some checks to other functions.

I looked at FilePolicy's parse method and I do see it has its error checking inline. However, its parsing logic is simple enough that the method is pretty short and easy to read, in which case the error checking doesn't cause a distraction.

I generally try to keep functions reasonably short and focused on one purpose to keep them easy to read, but I know some of this is subjective. In this case I would prefer to keep the parse method as-is, but if you feel strongly that it's a deviation on an important pattern being followed in the codebase and feel it makes the function hard to read, then I'm happy to make the suggested change.

service/src/ConstConfigIOGroup.hpp
avilcheslopez: What do you mean? These functions are private and not being exposed.

service/src/ConstConfigIOGroup.cpp
avilcheslopez: Update: Daniel has suggested looking into using json11's ``has_shape()`` method to simplify the checking. Looking at it, I see I can leverage it to at least check required fields and their types. I'll do a pass at improving the checking with that.

avilcheslopez: Update: simplified the parsing method to an extent by reducing the number of "check" functions by leveraging the ``has_shape()`` method.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,397 @@
...
+    ConstConfigIOGroup::ConstConfigIOGroup(const std::string &user_file_path,
+                                           const std::string &default_file_path)
+    {
+        std::string config_json;
+        bool load_default = true;
+        if (!user_file_path.empty()) {
+            try {
+                config_json = geopm::read_file(user_file_path);
+                load_default = false;
+            } catch (const Exception &ex) {
```
bgeltz:
```suggestion
            }
            catch (const geopm::Exception &ex) {
```

bgeltz (changes requested):
I'll take a better look on Monday

service/src/ConstConfigIOGroup.cpp
bgeltz: Casting a signed int to an unsigned int can result in weird values if the initial value was negative (e.g. 2^32 - 1 or similar for an initial value of -1).  It's why we avoid this sort of thing in other spots in the code.

Casting an unsigned int to a signed int should always be safe unless there's a difference in precision that results in data truncation.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+class ConstConfigIOGroupTest : public::testing::Test
+{
+    protected:
+        static const std::string M_CONFIG_FILE_PATH;
+        static void create_config_file(const std::string &config);
+};
+
+const std::string ConstConfigIOGroupTest::M_CONFIG_FILE_PATH =
+        "/tmp/const_config_test.json";
```
bgeltz:
```suggestion
        "const_config_test.json";
```
Is there a need to write these to tmp as opposed to the CWD?  If we put them in tmp, we may have other GEOPM users who are also writing to the same spot.  If 2 people are running make check at the same time, this is a problem.

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+    EXPECT_EQ(iogroup.sample(0), 1060);
+    EXPECT_EQ(iogroup.sample(1), 1700);
+    EXPECT_EQ(iogroup.read_signal("CONST_CONFIG::CPU_CORE_FREQUENCY", GEOPM_DOMAIN_CPU, 2), 1070);
+    EXPECT_EQ(iogroup.read_signal("CONST_CONFIG::GPU_CORE_FREQUENCY", GEOPM_DOMAIN_GPU, 0), 1500);
+    {
+        auto expected_function = geopm::Agg::average;
+        auto actual_function = *iogroup.agg_function("CONST_CONFIG::CPU_CORE_FREQUENCY")
+            .target<double (*) (const std::vector<double>&)>();
+        EXPECT_EQ(expected_function, actual_function);
+    }
```
bgeltz: Take a look at ```service/test/geopm_test.hpp```.  It has helper functions for this purpose so you can write this in the form:
```
std::function<double(const std::vector<double> &)> func = iogroup.agg_function("CONST:CONFIG::CPU_CORE_FREQUENCY");
// You could also use: auto func = iogroup.agg_function("CONST:CONFIG::CPU_CORE_FREQUENCY");
EXPECT_TRUE(is_agg_average(func));
```

Please write these all this way.

bgeltz (changes requested):

service/test/ConstConfigIOGroupTest.cpp
avilcheslopez: Thanks, will fix.

service/test/ConstConfigIOGroupTest.cpp
avilcheslopez: Ah, cool. Thanks!

service/json_schemas/const_config_io.schema.json
bgeltz: When this PR is merged, this new json file will need to be added to the repo that drives the website.

bgeltz (approved):

service/src/ConstConfigIOGroup.cpp
```diff
@@ -0,0 +1,435 @@
...
+            for (const auto &val : json_values) {
+                if (!val.is_number()) {
+                    throw Exception("ConstConfigIOGroup::parse_config_json():"
+                                    " incorrect type for property: "
+                                    "\"values\"", GEOPM_ERROR_INVALID,
+                                    __FILE__, __LINE__);
+                }
+                values.push_back(val.number_value());
+            }
+
```
cmcantalupo, line 343: Add check that length of array matches topo.nun_domain() for the system.  This will avoid a late failure that would occur when a user tries to read the value.

PR #2649, opened 2022-10-13
Move power_domain error check outside inspection loop

### LINKED ISSUE #2644: LevelZero Power Domain check will not throw if zero power domains exist
https://github.com/geopm/geopm/blob/c809a07f9474745149a16bbcb057e91227fb51ff/service/src/LevelZero.cpp#L260

line 260 should be outside the loop to catch instances where no power domains exist on a GPU, as GEOPM does not support configurations without energy/power domains.

bgeltz:
- Resolves #2644.

### DIFF
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -257,12 +257,13 @@ namespace geopm
                 }
 #endif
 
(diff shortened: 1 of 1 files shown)


lhlawson (approved):
LGTM

Issue #2642, opened 2022-10-10
Update the existing FrequencyGovernor to accept its user's desired control domain.


dannosliwcd:
The planned work for the FrequencyBalancerAgent runs into a case where the agent queries one governor for its domain type, then aims to match controls across both that governor and the FrequencyGovernor. Add a function to the FrequencyGovernor that lets the user specify which domain is desired, so the user doesn't need to manually translate every request through platform topo.

PR #2652, opened 2022-10-14
Update CPU-CA to use Frequency Governor

### LINKED ISSUE #2645: Make CPUActivityAgent use FrequencyGovernor for frequency control (NOTE: needs to be addressed before moving agent out of the beta flag)

### LINKED ISSUE #2651: Modify CPU CA to use the Frequency Governor
> estimate 1
Parent [feature/story]: #2363 

The implementation for the CPU CA feature was modified because it did not use the frequency governor to write frequency controls.

Done criteria:

- [x] Documentation changes are not required since this change does

lhlawson:
- Relates to #2645 feature change from github issues
- Fixes #2645 change request from github issues.

Update to the CPU-CA v1 to use the frequency governor for CPU frequency controls.
Uses #2658

### DIFF
--- a/src/CPUActivityAgent.cpp
+++ b/src/CPUActivityAgent.cpp
@@ -10,7 +10,6 @@
 #include <cmath>
 #include <cassert>
 #include <algorithm>
-#include <iostream>
 #include <string>
 
 #include "geopm/Agg.hpp"
@@ -64,26 +63,37 @@ namespace geopm
         // just what the user/admin has previously set.
         m_freq_uncore_min = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
         m_freq_uncore_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
-        m_freq_core_min = m_freq_governor->get_frequency_min();
-        m_freq_core_max = m_freq_governor->get_frequency_max();
 
         if (level == 0) {
-            m_freq_governor->init_platform_io();
             init_platform_io();
         }
     }
 
     void CPUActivityAgent::init_platform_io(void)
     {
-        if (m_platform_io.signal_domain_type("MSR::CPU_SCALABILITY_RATIO") <
-            m_freq_ctl_domain_type) {
+        int scalability_signal_domain = m_platform_io.signal_domain_type("MSR::CPU_SCALABILITY_RATIO");
+        if (scalability_signal_domain < m_freq_ctl_domain_type) {
+            // Set Freq gov domain
+            m_freq_governor->set_domain_type(scalability_signal_domain);
+
+            // update member vars
+            m_freq_ctl_domain_type = m_freq_governor->frequency_domain_type();
+            m_num_freq_ctl_domain = m_platform_topo.num_domain(m_freq_ctl_domain_type);
+
+#ifdef GEOPM_DEBUG
             throw Exception("CPUActivityAgent::" + std::string(__func__) +
-                            "():MSR::CPU_SCALABILITY RATIO domain (" +
-                            std::to_string(m_platform_io.signal_domain_type("MSR::CPU_SCALABILITY_RATIO")) +
-                            ") is a coarser granularity than the CPU Frequency control domain (" +
+                            "():MSR::CPU_SCALABILITY_RATIO domain (" +
+                            std::to_string(scalability_signal_domain) +
+                            ") is a coarser granularity than the CPU frequency control domain (" +
                             std::to_string(m_freq_ctl_domain_type) +
                             ").", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+#endif
         }
+
+        m_freq_governor->init_platform_io();
+
+        m_freq_core_min = m_freq_governor->get_frequency_min();
+        m_freq_core_max = m_freq_governor->get_frequency_max();
 
         for (int domain_idx = 0; domain_idx < m_num_freq_ctl_domain; ++domain_idx) {
             m_core_scal.push_back({m_platform_io.push_signal("MSR::CPU_SCALABILITY_RATIO",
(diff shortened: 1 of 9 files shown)


src/CPUActivityAgent.cpp
```diff
@@ -60,22 +65,23 @@ namespace geopm
...
+        m_freq_core_max = m_freq_governor->get_frequency_max();
 
         if (level == 0) {
+            m_freq_governor->init_platform_io();
             init_platform_io();
         }
     }
 
     void CPUActivityAgent::init_platform_io(void)
     {
```
bgeltz, line 74: Pull in Daniel's change for the the FrequencyGovernor to be able to tell the governor the preferred domain, #2642.

src/CPUActivityAgent.cpp
```diff
@@ -380,34 +366,26 @@ namespace geopm
 
             double core_req = m_resolved_f_core_efficient + f_core_range * scalability;
 
-            // Clip core request within policy limits
-            if (core_req > m_resolved_f_core_max || core_req < m_resolved_f_core_efficient) {
+            if (std::isnan(core_req)) {
+                core_req = in_policy[M_POLICY_CPU_FREQ_MAX];
+            }
```
bgeltz:
```suggestion
```

src/CPUActivityAgent.cpp
```diff
@@ -380,34 +366,26 @@ namespace geopm
...
-                m_core_freq_control.at(domain_idx).last_setting) {
-                // Adjust
-                m_platform_io.adjust(m_core_freq_control.at(domain_idx).batch_idx,
-                                     core_freq_request.at(domain_idx));
-
-                // Save the value for future comparison
-                m_core_freq_control.at(domain_idx).last_setting = core_freq_request.at(domain_idx);
+            // Track number of core requests
+            if (std::isnan(m_core_freq_control.at(domain_idx).last_setting) ||
+                core_req != m_core_freq_control.at(domain_idx).last_setting) {
```
bgeltz: Check the batch write boolean to determine frequency request count.  If you do this, you can remove the push_control for this.

bgeltz (changes requested):

src/CPUActivityAgent.cpp
```diff
@@ -70,13 +71,34 @@ namespace geopm
 
     void CPUActivityAgent::init_platform_io(void)
     {
-        for (int domain_idx = 0; domain_idx < M_NUM_CORE; ++domain_idx) {
+        int scalability_signal_domain = m_platform_io.signal_domain_type("MSR::CPU_SCALABILITY_RATIO");
+        if (scalability_signal_domain < m_freq_ctl_domain_type) {
```
dannosliwcd: I'm not sure whether this type of comparison is always expected to hold. `m_platform_topo.is_nested_domain` exposes that relationship at the interface level.

dannosliwcd: Could probably do away with the `if(){}`. e.g.,

```
m_freq_ctl_domain_type = m_platform_io.signal_domain_type("MSR::CPU_SCALABILITY_RATIO");
m_freq_governor->set_domain_type(m_freq_ctl_domain_type);
m_num_freq_ctl_domain = m_platform_topo.num_domain(m_freq_ctl_domain_type);
```

src/CPUActivityAgent.cpp
```diff
@@ -70,13 +71,34 @@ namespace geopm
...
+            // Set Freq gov domain
+            m_freq_governor->set_domain_type(scalability_signal_domain);
+
+            // update member vars
+            m_freq_ctl_domain_type = m_freq_governor->frequency_domain_type();
+            m_num_freq_ctl_domain = m_platform_topo.num_domain(m_freq_ctl_domain_type);
+
+#ifdef GEOPM_DEBUG
```
dannosliwcd: Is this really an exception? If it is fine to work with, but an unexpected scenario, could it be demoted to a debug-only warning on stderr?

If actually is problematic, should we always throw, instead of only in debug builds?

src/CPUActivityAgent.cpp
dannosliwcd: Or another alternative:
```
set my domain <-- outermost of governor domain and signal domain
set governor domain <-- my domain
push signal domain <-- my domain
```

lhlawson: Doing away with the if entirely could land us in a situation where scalability is per CPU but frequency control is per Core, ending in a different error from the FrequencyGovernor so the second approach is what was used via the if block, but it should be switched to use isnested and remove the if as suggested.

src/CPUActivityAgent.cpp
lhlawson: It's not problematic for a normal run, this was following the example of other agents https://github.com/geopm/geopm/blob/dev/src/PowerBalancer.cpp#L159 have similar throws in debug error cases to inform developers.  

It could be converted to a stderr without issue.

bgeltz (dismissed):

bgeltz:
Go ahead and rebase on dev.  Then one more round of approvals and we're good.

src/CPUActivityAgent.cpp
```diff
@@ -487,16 +475,26 @@ namespace geopm
...
-        result.push_back({"Resolved Maximum Core Frequency", std::to_string(m_resolved_f_core_max)});
-        result.push_back({"Resolved Efficient Core Frequency", std::to_string(m_resolved_f_core_efficient)});
-        result.push_back({"Resolved Core Frequency Range", std::to_string(m_resolved_f_core_max - m_resolved_f_core_efficient)});
-        result.push_back({"Resolved Maximum Uncore Frequency", std::to_string(m_resolved_f_uncore_max)});
-        result.push_back({"Resolved Efficient Uncore Frequency", std::to_string(m_resolved_f_uncore_efficient)});
-        result.push_back({"Resolved Uncore Frequency Range", std::to_string(m_resolved_f_uncore_max - m_resolved_f_uncore_efficient)});
+        result.push_back({"Core Batch Writes",
+                          std::to_string(m_core_batch_writes)});
+        result.push_back({"Core Frequency Requests Clipped",
+                          std::to_string(m_freq_governor->get_clamp_count())});
```
bgeltz: Make this "clamped" everywhere to align with what's in the FrequencyGovernor.  When we go to refactor these later, if we want to make it clipped, we can do it in all the spots.

bgeltz (changes requested):

bgeltz (approved):

dannosliwcd (approved):

PR #2655, opened 2022-10-14
Properly initialize LevelZero struct fields

### LINKED ISSUE #2654: Remove deprecated Sysman functions
> estimate 1
Parent [feature/story]: #XXXX

The implementation for the LevelZero feature was modified because of deprecation of the interface used. 

Done criteria:

- [ ] Documentation changes are not required since this change does

lhlawson:

- Relates to #2654 feature request from github issues

### DIFF
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -64,7 +64,7 @@ namespace geopm
                             ": LevelZero Device acquisition failed.", __LINE__);
 
             for (unsigned int device_idx = 0; device_idx < num_device; ++device_idx) {
-                ze_device_properties_t property;
+                ze_device_properties_t property = {};
                 ze_result = zeDeviceGetProperties(device_handle.at(device_idx), &property);
                 check_ze_result(ze_result, GEOPM_ERROR_RUNTIME, "LevelZero::"
                                 + std::string(__func__) +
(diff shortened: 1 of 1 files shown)


service/src/LevelZero.cpp
```diff
@@ -187,7 +187,7 @@ namespace geopm
                             geopm::LevelZero::M_DOMAIN_SIZE);
 
             for (auto handle : freq_domain) {
-                zes_freq_properties_t property;
+                zes_freq_properties_t property = {};
```
bgeltz, line 190: I see other instances of this pattern for the other property structs.
E.g. on line 67:
```
ze_device_properties_t property;
```
I think we should initialize all the property structs to zero.

service/src/LevelZero.cpp
```diff
@@ -557,18 +557,20 @@ namespace geopm
...
                         ": Sysman failed to get domain power properties", __LINE__);
-        result_power.tdp = property.defaultLimit;
-        result_power.min = property.minLimit;
-        result_power.max = property.maxLimit;
+
+        // Forcing to -1 due to API deprecation, other API calls should be used
+        // to replace this functionality.
+        result_power.tdp = -1;
+        result_power.min = -1;
+        result_power.max = -1;
```
```suggestion
```

If we can rely on the underlying API to return -1, we should do so.

bgeltz (changes requested):

bgeltz (approved):

PR #2658, opened 2022-10-19
Make FrequencyGovernor accept desired domain

### LINKED ISSUE #2642: Update the existing FrequencyGovernor to accept its user's desired control domain.

dannosliwcd:
- Add a FrequencyGovernor function that sets a new control domain to use in the governor.
- Users are allowed to specify the native domain or a more coarse domain.
- Requests to use a more fine-grained domain result in an exception.
- Resolves #2642

### DIFF
--- a/src/FrequencyGovernor.hpp
+++ b/src/FrequencyGovernor.hpp
@@ -28,6 +28,14 @@ namespace geopm
             ///        the vector to pass to adjust_platform().
             /// @return The domain with which frequency will be governed.
             virtual int frequency_domain_type(void) const = 0;
+            /// @brief Set the domain type of frequency control that will
+            ///        be used in adjust_platform(). Must be called before
+            ///        init_platform_io().
+            /// @throw Exception the requested domain does not contain the
+            ///        frequency control's native domain.
+            /// @throw Exception The caller attemped to set the domain type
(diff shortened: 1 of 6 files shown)


src/FrequencyGovernor.hpp
```diff
@@ -28,6 +28,14 @@ namespace geopm
             ///        the vector to pass to adjust_platform().
             /// @return The domain with which frequency will be governed.
             virtual int frequency_domain_type(void) const = 0;
+            /// @brief Set the domain type of frequency control that will
+            ///        be used in adjust_platform(). Must be called before
+            ///        init_platform_io().
```
lhlawson, line 33: There doesn't seem to be any issue with a user calling init_platform_io(), checking the domain, calling set_domain_type, and then re-calling init_platform_io().  As written this implies you have to call set_domain_type() before any call to init_platform_io().

Rewording for clarity might help.

lhlawson (approved):
LGTM

bgeltz (approved):

PR #2659, opened 2022-10-19
Add a FrequencyTimeBalancer class

### LINKED ISSUE #2253: Add a helper that estimates relationships between per-core frequency controls and epoch performance

dannosliwcd:
- Selects frequencies that should be set to balance execution times
- Resolves #2253

### DIFF
--- a/src/FrequencyTimeBalancer.cpp
+++ b/src/FrequencyTimeBalancer.cpp
@@ -1,0 +1,287 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "config.h"
+
+#include "FrequencyTimeBalancer.hpp"
+
+#include "geopm/Helper.hpp"
+#include "geopm_debug.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <numeric>
+
+namespace geopm
+{
+    class FrequencyTimeBalancerImp : public FrequencyTimeBalancer
+    {
+        public:
+            FrequencyTimeBalancerImp() = delete;
+            FrequencyTimeBalancerImp(
+                double minimum_frequency,
+                double maximum_frequency);
+            virtual ~FrequencyTimeBalancerImp() = default;
+
+            std::vector<double> balance_frequencies_by_time(
+                const std::vector<double> &previous_times,
+                const std::vector<double> &previous_control_frequencies,
+                const std::vector<double> &previous_achieved_frequencies,
+                const std::vector<std::pair<unsigned int, double> > &frequency_limits_by_high_priority_count,
+                double low_priority_frequency) override;
+
+            double get_target_time() const override;
+
+            static std::unique_ptr<FrequencyTimeBalancer> make_unique(
+                double minimum_frequency,
+                double maximum_frequency);
+            static std::shared_ptr<FrequencyTimeBalancer> make_shared(
+                double minimum_frequency,
+                double maximum_frequency);
+
+        private:
+            void update_balance_targets(
+                const std::vector<double> &previous_times,
+                const std::vector<double> &previous_control_frequencies,
+                const std::vector<double> &previous_achieved_frequencies,
+                const std::vector<std::pair<unsigned int, double> > &frequency_limits_by_high_priority_count,
+                double low_priority_frequency);
+
+            std::vector<double> get_balanced_frequencies(
+                const std::vector<double> &previous_times,
+                const std::vector<double> &previous_control_frequencies,
+                const std::vector<double> &previous_achieved_frequencies);
+
+            std::vector<std::vector<int> > m_domain_index_groups;
+            double m_minimum_frequency;
+            double m_maximum_frequency;
+            // Target balancing time
+            double m_target_time;
+            // High/Low priority cutoff frequenciy
+            double m_cutoff_frequency;
+            // Index order for argsorting each monitored domain by application
+            // lagginess of processes running in that domain.
+            std::vector<size_t> m_lagginess_idx;
+    };
+
+    FrequencyTimeBalancerImp::FrequencyTimeBalancerImp(
+        double minimum_frequency,
+        double maximum_frequency)
+        : m_minimum_frequency(minimum_frequency)
+        , m_maximum_frequency(maximum_frequency)
+        , m_target_time(NAN)
+        , m_cutoff_frequency(NAN)
+    {
+    }
+
+    std::vector<double> FrequencyTimeBalancerImp::balance_frequencies_by_time(
+        const std::vector<double> &previous_times,
+        const std::vector<double> &previous_control_frequencies,
+        const std::vector<double> &previous_achieved_frequencies,
+        const std::vector<std::pair<unsigned int, double> > &frequency_limits_by_high_priority_count,
+        double low_priority_frequency)
+    {
+        GEOPM_DEBUG_ASSERT(
+            (previous_times.size() == previous_control_frequencies.size() &&
+             previous_times.size() == previous_achieved_frequencies.size()),
+            "FrequencyTimeBalancerImp::balance_frequencies_by_time(): "
+            "input vectors must be the same size.");
+
+        // Estimate the target time we should balance against
+        update_balance_targets(
+            previous_times, previous_control_frequencies,
+            previous_achieved_frequencies,
+            frequency_limits_by_high_priority_count, low_priority_frequency);
+
+        // Estimate the frequency controls that would achieve our target balancing time
+        return get_balanced_frequencies(
+            previous_times, previous_control_frequencies,
+            previous_achieved_frequencies);
+    }
+
+    void FrequencyTimeBalancerImp::update_balance_targets(
+        const std::vector<double> &previous_times,
+        const std::vector<double> &previous_control_frequencies,
+        const std::vector<double> &previous_achieved_frequencies,
+        const std::vector<std::pair<unsigned int, double> > &frequency_limits_by_high_priority_count,
+        double low_priority_frequency)
+    {
+        // Perform an argsort, ordered by decreasing lagginess of a domain
+        m_lagginess_idx.resize(previous_times.size());
+        std::iota(m_lagginess_idx.begin(), m_lagginess_idx.end(), 0);
+
+        auto is_laggier_core_idx =
+            [&previous_times, &previous_achieved_frequencies](size_t lhs, size_t rhs) {
+                // Place unrecorded times at the end of the sorted collection
+                if (std::isnan(previous_times[lhs])) {
+                    return false;
+                }
+                if (std::isnan(previous_times[rhs])) {
+                    return true;
+                }
+
+                // Sort by cycles in region of interest, rather than by
+                // time. Keep the most sensitive one at the front.
+                return (previous_times[lhs] * previous_achieved_frequencies[lhs]) >
+                       (previous_times[rhs] * previous_achieved_frequencies[rhs]);
+            };
+
+        std::sort(m_lagginess_idx.begin(), m_lagginess_idx.end(), is_laggier_core_idx);
+
+        const auto lagger_time = previous_times[m_lagginess_idx[0]];
+
+        // Set the frequency of each index to match the last-observed
+        // non-network time of the slowest recently-frequency-unlimited index.
+        // We don't want to balance against frequency-limited indices in case
+        // a previous frequency limit was set too low, which could set our
+        // target performance too low.
+        auto target_it = std::find_if(
+            m_lagginess_idx.begin(),
+            m_lagginess_idx.end(),
+            [&previous_control_frequencies, &previous_times, this](size_t frequency_idx) {
+                return !std::isnan(previous_times[frequency_idx]) &&
+                       previous_control_frequencies[frequency_idx] >= m_maximum_frequency;
+            });
+        size_t reference_core_idx;
+        if (target_it == m_lagginess_idx.end()) {
+            reference_core_idx = m_lagginess_idx[0];
+        }
+        else {
+            reference_core_idx = *target_it;
+        }
+
+        // From previously-unthrottled cores, match the core with the most
+        // time, scaled to the estimated time at its estimated
+        // maximum-achievable frequency.
+        double balance_target_time = previous_times[reference_core_idx];
+        auto reference_core_hp_cutoff = m_minimum_frequency;
+
+        // See if we can opt for an even lower desired time based on
+        // prioritized frequency tradeoffs.
+        for (const auto &hp_count_frequency : frequency_limits_by_high_priority_count) {
+            const size_t hp_count = hp_count_frequency.first;
+            const double hp_frequency = hp_count_frequency.second;
+
+            // Estimate achieved vs achievable impacts
+            const auto laggiest_high_priority_time =
+                lagger_time *
+                previous_achieved_frequencies[m_lagginess_idx[0]] /
+                hp_frequency;
+            if (hp_count < m_lagginess_idx.size()) {
+                const auto laggiest_lp_idx = m_lagginess_idx[hp_count];
+                const auto laggiest_low_priority_time =
+                    previous_times[laggiest_lp_idx] *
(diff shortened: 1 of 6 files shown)


src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,307 @@
...
+        const std::vector<double> &previous_max_frequencies)
+    {
+        GEOPM_DEBUG_ASSERT(previous_times.size() == previous_control_frequencies.size(),
+                           "FrequencyTimeBalancerImp::balance_frequencies_by_time(): "
+                           "time vector must be the same size as the frequency vector.");
+
+        m_frequency_limit_detector->update_max_frequency_estimates(previous_max_frequencies);
+
+        const size_t domain_count_per_group =
+            previous_control_frequencies.size() / m_subdomain_group_count;
```
avilcheslopez: Q: what guarantee do you have that the control frequencies vector's size is divisible by the group count?

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,307 @@
...
+        std::vector<double> desired_frequencies = previous_control_frequencies;
+
+        std::vector<size_t> idx(previous_times.size());
```
avilcheslopez: What does ``idx`` represent? I can look at the code below to try to infer it, but after looking for a bit, I see it would make it easier for the reader to understand the code if it had a more descriptive name. I do see that it's used as an indexer of some sort, so a longer name might be inconvenient, so if it ends up being too much trouble consider adding a comment that describes what it represents or its purpose.

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    std::vector<double> FrequencyTimeBalancerImp::balance_frequencies_by_time(
+        const std::vector<double> &previous_times,
+        const std::vector<double> &previous_control_frequencies,
+        const std::vector<double> &previous_achieved_frequencies,
+        const std::vector<double> &previous_max_frequencies)
+    {
+        GEOPM_DEBUG_ASSERT(previous_times.size() == previous_control_frequencies.size(),
+                           "FrequencyTimeBalancerImp::balance_frequencies_by_time(): "
+                           "time vector must be the same size as the frequency vector.");
```
avilcheslopez: Are ``previous_achieved_frequencies`` and ``previous_max_frequencies`` also expected to be parallel with ``previous_times`` and ``previous_control_frequencies``? Looking below, it seems like they are, in which case that should be asserted as well.

test/FrequencyTimeBalancerTest.cpp
```diff
@@ -0,0 +1,392 @@
...
+        double m_lp_frequency;
+        std::vector<double> m_observed_core_frequencies;
+};
+
+static const double UNCERTAINTY_WINDOW = 0.000;
+static const int PACKAGE_COUNT = 1;
+
+class FrequencyTimeBalancerTest : public ::testing::Test
+{
+};
```
avilcheslopez: Is there a need for a test fixture class for these tests? If tests don't need to share common set-up or tear-down routines, nor share any resources, we could get away with regular tests without a test fixture.

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,307 @@
...
+        , m_do_ignore_domain_idx(do_ignore_domain_idx)
+        , m_minimum_frequency(minimum_frequency)
+        , m_maximum_frequency(maximum_frequency)
+        , m_target_times(subdomain_group_count, NAN)
+        , m_cutoff_frequencies(subdomain_group_count, NAN)
+        , m_frequency_limit_detector(std::move(frequency_limit_detector))
+    {
+    }
+
+    std::vector<double> FrequencyTimeBalancerImp::balance_frequencies_by_time(
```
avilcheslopez, line 79: Suggestion: I feel this function is a bit lengthy and there's quite a bit going on, so I think it could be broken down into smaller functions to improve readability, allowing the reader to focus on more digestible pieces at a time.

Note: this is mostly coming from personal preference, as I'm used to smaller functions. To be fair, I've noticed our level of comfort regarding length of function is higher going by other functions I've seen in the codebase. So feel free to disregard the suggestion.

src/FrequencyTimeBalancer.cpp
dannosliwcd: None right now.

Simple fix: I could throw an exception if that is not true here.

More complicated fix, but avoids questions of divisibility: Generate sets of cores grouped by a caller-provided GEOPM domain. Modify other parts of the interface to translate between vectors-of-board-domain things and the sub-groups of cores.

Thoughts on those approaches?

src/FrequencyTimeBalancer.hpp
```diff
@@ -0,0 +1,95 @@
...
+                double uncertainty_window_seconds,
+                int subdomain_group_count,
+                std::function<bool(int)> do_ignore_domain_idx,
+                double minimum_frequency,
+                double maximum_frequency,
+                std::unique_ptr<FrequencyLimitDetector> frequency_limit_detector);
+    };
+}
+
+#endif
```
Krebrov001:
```cpp
#endif  /* FREQUENCYTIMEBALANCER_HPP_INCLUDE */
```

test/MockFrequencyTimeBalancer.hpp
```diff
@@ -0,0 +1,26 @@
...
+                    (const std::vector<double>& previous_times,
+                     const std::vector<double>& previous_control_frequencies,
+                     const std::vector<double>& previous_achieved_frequencies,
+                     const std::vector<double>& previous_max_frequencies),
+                    (override));
+        MOCK_METHOD(std::vector<double>, get_target_times, (), (const, override));
+        MOCK_METHOD(double, get_cutoff_frequency, (unsigned int core), (const, override));
+};
+
+#endif
```
Krebrov001:
```cpp
#endif  /* MOCKFREQUENCYTIMEBALANCER_HPP_INCLUDE */
```

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    if (std::isnan(previous_times[lhs]) || m_do_ignore_domain_idx(lhs)) {
+                        return false;
+                    }
+                    if (std::isnan(previous_times[rhs]) || m_do_ignore_domain_idx(rhs)) {
+                        return true;
+                    }
+
+                    // Sort by cycles in region of interest, rather than by
+                    // time. Keep the most sensitive one at the front.
+                    return (previous_times[lhs] * previous_achieved_frequencies[lhs]) >
```
Krebrov001: Can we save the value into a temporary variable, and then return that variable, for improved readability?
```cpp
auto result = /* code here */;
return result;
```

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,360 @@
...
+                           (previous_times[rhs] * previous_achieved_frequencies[rhs]);
+                });
+        }
+
+        for (size_t group_idx = 0;
+             group_idx < static_cast<size_t>(m_subdomain_group_count); ++group_idx) {
+            const auto group_idx_offset = group_idx * domain_count_per_group;
+            const auto next_group_idx_offset = (group_idx + 1) * domain_count_per_group;
```
Krebrov001: We can use the distributive property to optimize this computation.
```cpp
const auto group_idx_offset = group_idx * domain_count_per_group;
const auto next_group_idx_offset = group_idx_offset + domain_count_per_group;
```

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,360 @@
...
+        std::vector<double> desired_frequencies = previous_control_frequencies;
+
+        // Perform an argsort, ordered by decreasing lagginess of a domain
+        m_lagginess_idx.resize(previous_times.size());
+        std::iota(m_lagginess_idx.begin(), m_lagginess_idx.end(), 0);
+        for (size_t group_idx = 0; group_idx < static_cast<size_t>(m_subdomain_group_count); ++group_idx) {
+            std::sort(
+                m_lagginess_idx.begin() + group_idx * domain_count_per_group,
+                m_lagginess_idx.begin() + (group_idx + 1) * domain_count_per_group,
+                [this, &previous_times, &previous_achieved_frequencies](size_t lhs, size_t rhs) {
```
Krebrov001: Perhaps it would be better to move the lambda into it's own separate variable, and then pass it into the `std::sort()` for readability purposes?

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+        const size_t domain_count_per_group =
+            previous_control_frequencies.size() / m_subdomain_group_count;
+        std::vector<double> desired_frequencies = previous_control_frequencies;
+
+        // Perform an argsort, ordered by decreasing lagginess of a domain
+        m_lagginess_idx.resize(previous_times.size());
+        std::iota(m_lagginess_idx.begin(), m_lagginess_idx.end(), 0);
+        for (size_t group_idx = 0; group_idx < static_cast<size_t>(m_subdomain_group_count); ++group_idx) {
+            std::sort(
```
Krebrov001:
```cpp
const auto group_idx_offset = group_idx * domain_count_per_group;
const auto next_group_idx_offset = group_idx_offset + domain_count_per_group;
std::sort(m_lagginess_idx.begin() + group_idx_offset,
          m_lagginess_idx.begin() + next_group_idx_offset,
```

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+    }
+
+    std::vector<double> FrequencyTimeBalancerImp::balance_frequencies_by_time(
+        const std::vector<double> &previous_times,
+        const std::vector<double> &previous_control_frequencies,
+        const std::vector<double> &previous_achieved_frequencies,
+        const std::vector<double> &previous_max_frequencies)
+    {
+        GEOPM_DEBUG_ASSERT(
```
Krebrov001, line 86: Caching repeatedly used functions to avoid duplicated calls.
```cpp
std::size_t previous_times_size = previous_times.size();
GEOPM_DEBUG_ASSERT(
    (previous_times_size == previous_control_frequencies.size() &&
     previous_times_size == previous_achieved_frequencies.size() &&
     previous_times_size == previous_max_frequencies.size()),
    "FrequencyTimeBalancerImp::balance_frequencies_by_time(): "
    "input vectors must be the same size.");
if (previous_times_size % m_subdomain_group_count != 0) {}
```

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,360 @@
...
+        const std::vector<double> &previous_achieved_frequencies)
+    {
+        const size_t domain_count_per_group =
+            previous_achieved_frequencies.size() / m_subdomain_group_count;
+        std::vector<double> desired_frequencies = previous_control_frequencies;
+
+        for (size_t group_idx = 0;
+             group_idx < static_cast<size_t>(m_subdomain_group_count); ++group_idx) {
+            const auto group_idx_offset = group_idx * domain_count_per_group;
+            const auto next_group_idx_offset = (group_idx + 1) * domain_count_per_group;
```
Krebrov001:
```cpp
const auto group_idx_offset = group_idx * domain_count_per_group;
const auto next_group_idx_offset = group_idx_offset + domain_count_per_group;
```

src/FrequencyTimeBalancer.cpp
dannosliwcd: This macro only results in executable code on debug builds.

src/FrequencyTimeBalancer.hpp
```diff
@@ -0,0 +1,95 @@
...
+    /// percent. Workloads less frequency-sensitive than that should be able to
+    /// go lower than the recommended frequencies. This is expected to converge
+    /// toward those lower frequencies if it is repeatedly re-evaluated some
+    /// time after applying the recommended frequency controls.
+    class FrequencyTimeBalancer
+    {
+        public:
+            virtual ~FrequencyTimeBalancer() = default;
+
+            /// @brief Return the recommended frequency controls given observed
```
dannosliwcd, line 26: Make sure all `std::vector` comments explain what the vector index represents.

src/FrequencyTimeBalancer.cpp
lhlawson: If this is holding up the feature for now my preference is for the simple fix and file an issue on the more complex solution if we hit it in initial system testing.

test/FrequencyTimeBalancerTest.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+using geopm::FrequencyTimeBalancer;
+using ::testing::_;
+using ::testing::DoubleEq;
+using ::testing::ElementsAre;
+using ::testing::Ge;
+using ::testing::Pointwise;
+using ::testing::Return;
+
+// This CPU core filter always returns false, indicating that no cores should
```
lhlawson: Are other core filters going to be added, or will this change in the future?

dannosliwcd: In this commit, `can_ignore_control_domain_indices` tests the filter. In future commits, the agent ignores cores that are not running the application.

src/FrequencyTimeBalancer.cpp
```diff
@@ -0,0 +1,370 @@
...
+            std::vector<double> m_cutoff_frequencies;
+            std::unique_ptr<FrequencyLimitDetector> m_frequency_limit_detector;
+            // Index order for argsorting each monitored domain by application
+            // lagginess of processes running in that domain.
+            std::vector<size_t> m_lagginess_idx;
+    };
+
+    FrequencyTimeBalancerImp::FrequencyTimeBalancerImp(
+        double uncertainty_window_seconds,
+        int subdomain_group_count,
```
cmcantalupo: Rather than counts and division, use GEOPM domain types.  Specify an inner_domain type and outer_domain type and use PlatformTopo::domain_nested() api to determine the indexing vector.

dannosliwcd:
Force-push was a rebase on top of latest dev. No other changes were added in the force-push. Will add commits for Chris's review comments

src/FrequencyTimeBalancer.hpp
```diff
@@ -0,0 +1,68 @@
...
+    /// Select frequency control settings that are expected to balance measured
+    /// execution times. Assumes time impact of up to (frequency_old/frequency_new)
+    /// percent. Workloads less frequency-sensitive than that should be able to
+    /// go lower than the recommended frequencies. This is expected to converge
+    /// toward those lower frequencies if it is repeatedly re-evaluated some
+    /// time after applying the recommended frequency controls.
+    class FrequencyTimeBalancer
+    {
+        public:
+            virtual ~FrequencyTimeBalancer() = default;
```
cmcantalupo, line 24: Add default constructor.

cmcantalupo (commented):
Drop uncertainty window from interface, not needed.

cmcantalupo (approved):
LGTM

PR #2667, opened 2022-10-25
Add a CLOS Governor

### LINKED ISSUE #2665: Add a governor that helps set SST-TF controls to influence achievable core frequency.
> estimate 1
> Parent feature: #2575

Introduce new C++ class that acts as an abstraction to SST-TF, similar to how the frequency governor and power governor act as abstractions to frequency and power controls.

Done criteria:

- [ ] `is_supported` returns true if SST-TF is supported. Returns false if unsupported (both if the platform reports that it is unsupported and if the platform fails to report anything about SST-TF)
- [ ] `govern` modifies the class of service for cores, similar to how frequency governor modifies frequency controls for cores
- [ ] `enable` enables SST-TF and SST-CP
- [ ] `disable` disables SST-TF and SST-CP

dannosliwcd:
- Resolves #2665
- Offers a control abstraction similar to power and frequency governors, but interacts with class of service controls from SSTIOGroup.

### DIFF
--- a/src/SSTClosGovernor.cpp
+++ b/src/SSTClosGovernor.cpp
@@ -1,0 +1,197 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "SSTClosGovernor.hpp"
+
+#include <unistd.h>
+
+#include <algorithm>
+#include <cmath>
+#include <iostream>
+#include <vector>
+
+#include "SSTClosGovernorImp.hpp"
+#include "geopm/Exception.hpp"
+#include "geopm/Helper.hpp"
+#include "geopm/PlatformIO.hpp"
+#include "geopm/PlatformTopo.hpp"
+
+#include "config.h"
+
+namespace geopm
+{
+    std::unique_ptr<SSTClosGovernor> SSTClosGovernor::make_unique(void)
+    {
+        return geopm::make_unique<SSTClosGovernorImp>();
+    }
+
+    std::shared_ptr<SSTClosGovernor> SSTClosGovernor::make_shared(void)
+    {
+        return std::make_shared<SSTClosGovernorImp>();
+    }
+
+    SSTClosGovernorImp::SSTClosGovernorImp()
+        : SSTClosGovernorImp(platform_io(), platform_topo())
+    {
+    }
+
+    SSTClosGovernorImp::SSTClosGovernorImp(PlatformIO &platform_io, const PlatformTopo &platform_topo)
+        : m_platform_io(platform_io)
+        , m_platform_topo(platform_topo)
+        , m_do_write_batch(false)
+        , m_is_enabled(true)
+        , m_clos_assoc_ctl_domain_type(m_platform_io.control_domain_type("SST::COREPRIORITY:ASSOCIATION"))
+        , m_num_clos_assoc_ctl_domain(m_platform_topo.num_domain(m_clos_assoc_ctl_domain_type))
+        , m_clos_config_ctl_domain_type(m_platform_io.control_domain_type("SST::COREPRIORITY:0:FREQUENCY_MIN"))
+        , m_num_clos_config_ctl_domain(m_platform_topo.num_domain(m_clos_config_ctl_domain_type))
+        , m_frequency_min(m_platform_io.read_signal("CPU_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0))
+        , m_frequency_sticker(m_platform_io.read_signal("CPU_FREQUENCY_STICKER", GEOPM_DOMAIN_BOARD, 0))
+        , m_frequency_max(m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0))
+        , m_last_clos(m_num_clos_assoc_ctl_domain, HIGH_PRIORITY)
+    {
+    }
+
+    bool SSTClosGovernor::is_supported(PlatformIO &platform_io)
+    {
+        bool result = false;
+
+        try {
+            result = platform_io.read_signal("SST::COREPRIORITY_SUPPORT:CAPABILITIES",
+                                             GEOPM_DOMAIN_BOARD, 0) > 0;
+            result &= platform_io.read_signal("SST::TURBOFREQ_SUPPORT:SUPPORTED",
+                                              GEOPM_DOMAIN_BOARD, 0) > 0;
+        }
+        catch (...) {
+            result = false;
+        }
+
+        return result;
+    }
+
+    void SSTClosGovernorImp::init_platform_io(void)
+    {
+        for (size_t ctl_idx = 0; ctl_idx < m_num_clos_assoc_ctl_domain; ++ctl_idx) {
+            m_clos_control_idx.push_back(m_platform_io.push_control(
+                "SST::COREPRIORITY:ASSOCIATION",
+                m_clos_assoc_ctl_domain_type,
+                static_cast<int>(ctl_idx)));
+        }
+        // Increase the turbo ratio limits so we can take advantage of the
+        // increased range offered by SST-TF
+        m_platform_io.write_control("MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_0",
+                                    GEOPM_DOMAIN_BOARD, 0, 255e8);
+        m_platform_io.write_control("MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_1",
+                                    GEOPM_DOMAIN_BOARD, 0, 255e8);
+        m_platform_io.write_control("MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_2",
+                                    GEOPM_DOMAIN_BOARD, 0, 255e8);
+        m_platform_io.write_control("MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_3",
+                                    GEOPM_DOMAIN_BOARD, 0, 255e8);
+        m_platform_io.write_control("MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_4",
+                                    GEOPM_DOMAIN_BOARD, 0, 255e8);
+        m_platform_io.write_control("MSR::TURBO_RATIO_LIMIT:MAX_RATIO_LIMIT_5",
+                                    GEOPM_DOMAIN_BOARD, 0, 255e8);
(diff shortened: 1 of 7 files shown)


src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,196 @@
...
+        , m_frequency_min(m_platform_io.read_signal("CPU_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0))
+        , m_frequency_sticker(m_platform_io.read_signal("CPU_FREQUENCY_STICKER", GEOPM_DOMAIN_BOARD, 0))
+        , m_frequency_max(m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0))
+        , m_last_clos(m_num_clos_assoc_ctl_domain, HIGH_PRIORITY)
+    {
+    }
+
+    bool SSTClosGovernor::is_supported(PlatformIO &platform_io)
+    {
+        bool result;
```
Krebrov001:
```cpp
bool result = false;
```

src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,196 @@
...
+        catch (...) {
+            result = false;
+        }
+
+        return result;
+    }
+
+    void SSTClosGovernorImp::init_platform_io(void)
+    {
+        for (size_t ctl_idx = 0; ctl_idx != m_num_clos_assoc_ctl_domain; ++ctl_idx) {
```
Krebrov001: I would prefer the `<` operator. I know that this is `size_t`, which can never be signed negative, but still I think that `<` is more "readable". Usually `!=` is used for iterators. This is an index or counter, not an iterator.
```cpp
for (size_t ctl_idx = 0; ctl_idx < m_num_clos_assoc_ctl_domain; ++ctl_idx) {}
```

src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,196 @@
...
+            m_clos_control_idx.push_back(m_platform_io.push_control(
+                "SST::COREPRIORITY:ASSOCIATION",
+                m_clos_assoc_ctl_domain_type,
+                static_cast<int>(ctl_idx)));
```
Krebrov001, line 79: Since that `size_t` is usually 64 bits wide, and `int` is usually 32 bits wide, is there ever a possibility where this conversion would overflow, for example when `ctl_idx > INT_MAX`? What is the maximum value that is stored in `ctl_idx`?

src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,196 @@
...
+    {
+        if (clos_by_core.size() != m_clos_control_idx.size()) {
+            throw Exception("SSTClosGovernorImp::" + std::string(__func__) +
+                                "(): size of request vector does not match size of control domain.",
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        m_do_write_batch = m_is_enabled && clos_by_core != m_last_clos;
+        if (m_is_enabled) {
+            for (size_t idx = 0; idx < m_clos_control_idx.size(); ++idx) {
```
Krebrov001: What if you move the `size()` call out of the loop, to avoid it getting called during every single iteration? You have one call instead of **N** of them.
```cpp
szie_t m_clos_control_idx_size = m_clos_control_idx.size();
for (size_t idx = 0; idx < m_clos_control_idx_size; ++idx) {}
```

src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,196 @@
...
+        // Enable prioritized turbo by cores
+        m_platform_io.write_control("SST::COREPRIORITY_ENABLE:ENABLE",
+                                    GEOPM_DOMAIN_BOARD, 0, 1);
+
+        // Enable the ability to extend the turbo range of high priority cores
+        // by decreasing the turbo range of low priority cores
+        m_platform_io.write_control("SST::TURBO_ENABLE:ENABLE",
+                                    GEOPM_DOMAIN_BOARD, 0, 1);
+
+        m_is_enabled = true;
```
Krebrov001, line 181: Why do we have `m_is_enabled = true;` at the end of `enable_sst_turbo_prioritization()` function, whereas the analogous statement it at the start of the `disable_sst_turbo_prioritization()` function? I think that we should probably be consistent for readability purposes, unless there is some kind of meaning of semantic difference between the enableing process and the disableing process. Is the property enabled only explicitly *after* performing those two `write_control()` calls?

src/SSTClosGovernorImp.hpp
```diff
@@ -0,0 +1,45 @@
...
+            bool do_write_batch(void) const override;
+
+            // No-op if sst-tf is not supported. Mods both tf and cp
+            void enable_sst_turbo_prioritization() override;
+            void disable_sst_turbo_prioritization() override;
+
+        private:
+            PlatformIO &m_platform_io;
+            const PlatformTopo &m_platform_topo;
+            bool m_do_write_batch;
```
Krebrov001, line 30: Have you considered putting the larger data types to the top of the list, for making the object take up less space in memory? Perhaps we should rearrange the fields in such a way that there is less padding in between them, hence the object doesn't take up as much space in memory? We could save like 12 bytes worth of memory.

src/SSTClosGovernorImp.hpp
```diff
@@ -0,0 +1,45 @@
...
+            double m_frequency_min;
+            double m_frequency_sticker;
+            double m_frequency_max;
+            std::vector<int> m_clos_control_idx;
+            std::vector<int> m_frequency_control_idx;
+            std::vector<double> m_last_clos;
+    };
+}
+
+#endif
```
Krebrov001: I like to put the name of the class at the end of the file too, to figure out what the endif is covering.

```cpp
#endif /* SSTCLOSGOVERNORIMP_HPP_INCLUDE */
```

src/SSTClosGovernor.hpp
```diff
@@ -0,0 +1,72 @@
...
+            /// Indicate whether this platform supports core priority and
+            /// prioritized turbo frequency limits.
+            static bool is_supported(PlatformIO &platform_io);
+
+            static std::unique_ptr<SSTClosGovernor> make_unique(void);
+            static std::shared_ptr<SSTClosGovernor> make_shared(void);
+    };
+}
+
+#endif
```
Krebrov001: I like to put the name of the class at the end of the file too, to figure out what the endif is covering.
```cpp
#endif /* SSTCLOSGOVERNOR_HPP_INCLUDE */
```

test/MockSSTClosGovernor.hpp
```diff
@@ -0,0 +1,25 @@
...
+        MOCK_METHOD(void, init_platform_io, (), (override));
+        MOCK_METHOD(int, clos_domain_type, (), (const, override));
+        MOCK_METHOD(void, adjust_platform,
+                    (const std::vector<double> &clos_by_core), (override));
+        MOCK_METHOD(bool, do_write_batch, (), (const, override));
+        MOCK_METHOD(void, enable_sst_turbo_prioritization, (), (override));
+        MOCK_METHOD(void, disable_sst_turbo_prioritization, (), (override));
+};
+
+#endif
```
Krebrov001:
```cpp
#endif /* MOCKSSTCLOSGOVERNOR_HPP_INCLUDE */
```

test/SSTClosGovernorTest.cpp
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <memory>
```
Krebrov001, line 7: Perhaps it would be good to put all the std includes together, and all the google includes together?
```cpp
#include <memory>
#include <stdexcept>

#include "gmock/gmock-matchers.h"
#include "gmock/gmock.h"
#include "gtest/gtest.h"
```

src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,196 @@
...
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "SSTClosGovernor.hpp"
+
+#include <unistd.h>
+#include <algorithm>
+#include <cmath>
+#include <iostream>
```
Krebrov001, line 12: Just nitpicking, but I want iostream and vector to be next to each other.
```cpp
#include <iostream>
#include <vector>
```

Krebrov001 (dismissed):

src/SSTClosGovernor.cpp
dannosliwcd: `ctl_idx` has `m_num_clos_assoc_ctl_domain` as an upper-bound. `m_num_clos_assoc_ctl_domain` is initialized by a return value from `int PlatformTopo::num_domain(int)`

src/SSTClosGovernor.cpp
dannosliwcd: This class offers a high-level enable/disable abstraction over two dependencies' low-level enable/disable state. The high-level wrapper is only enabled while all of its low-level dependencies are enabled. The high-level wrapper is no longer fully enabled once we start disabling its dependencies.

If for some reason any error occurs part-way through enable/disable, the `m_is_enabled` state indicates whether the high-level wrapper is enabled.

src/SSTClosGovernorImp.hpp
dannosliwcd: Some member initializations depend on the outcome of other member initializations in this class. With this order, we can avoid initializing some members twice (once to create a throwaway instance, and again when we have the prerequisites ready).

This class typically gets one instance, so I don't think the 12B savings is a worthwhile trade-off of either initializing some of its members twice or adding more code complexity to work around it.

src/SSTClosGovernor.cpp
```diff
@@ -0,0 +1,197 @@
...
+    bool SSTClosGovernorImp::do_write_batch(void) const
+    {
+        return m_do_write_batch;
+    }
+
+    void SSTClosGovernorImp::enable_sst_turbo_prioritization()
+    {
+        // Enable prioritized turbo by cores
+        m_platform_io.write_control("SST::COREPRIORITY_ENABLE:ENABLE",
+                                    GEOPM_DOMAIN_BOARD, 0, 1);
```
dannosliwcd, line 174: Should these operations be done at domain `m_clos_config_ctl_domain_type` inside a loop up to `m_num_clos_config_ctl_domain`? As written, this may fail with an exception due to #2707

cmcantalupo (dismissed):
LGTM

dannosliwcd:
Rebased to squash the fixup commits from review.

cmcantalupo (approved):

PR #2692, opened 2022-11-06
Update the CPU-CA to use the ConstConfigIOGroup

### LINKED ISSUE #2436: Modify the CPU-CA to use the ConstConfigIOGroup

lhlawson:
- Relates to #2698 feature request from github issues
- Fixes #2436  change request from github issues.

Updates the CPU-CA to use the information in the ConstConfigIOGroup during runs, reduce the CPU-CA policy to only the phi value, and allows for per-node characterization runs.

### DIFF
--- a/integration/test/test_cpu_activity_agent.py
+++ b/integration/test/test_cpu_activity_agent.py
@@ -171,7 +171,9 @@ class TestIntegration_cpu_activity(unittest.TestCase):
         json_config = json.dumps(uncore_config, indent=4)
         with open("const_config_io-ca.json", "w") as outfile:
             outfile.write(json_config)
-        os.environ["GEOPM_CONST_CONFIG_PATH"] = "const_config_io-ca.json"
+
+        const_path = Path('const_config_io-ca.json').resolve()
+        os.environ["GEOPM_CONST_CONFIG_PATH"] = str(const_path)
 
         ################################
         # CPU Activity Agent phi sweep #
--- a/integration/test/test_cpu_activity_agent.py
+++ b/integration/test/test_cpu_activity_agent.py
@@ -9,6 +9,7 @@ This integration test verifies that the cpu_activity agent can improve
 the energy efficiency of an application.
 """
 
+import json
 import sys
 import unittest
 import os
@@ -25,7 +26,7 @@ from integration.test import geopm_test_launcher
 from experiment.energy_efficiency import cpu_activity
 from experiment.monitor import monitor
 from experiment.uncore_frequency_sweep import uncore_frequency_sweep
-from experiment.uncore_frequency_sweep import gen_cpu_activity_policy_recommendation
+from experiment.uncore_frequency_sweep import gen_cpu_activity_constconfig_recommendation
 from apps.arithmetic_intensity import arithmetic_intensity
 from apps.minife import minife
 
@@ -137,14 +138,10 @@ class TestIntegration_cpu_activity(unittest.TestCase):
         # Parse data #
         ##############
         df_frequency_sweep = geopmpy.io.RawReportCollection('*report', dir_name=cls._aib_uncore_freq_sweep_dir).get_df()
-        policy = gen_cpu_activity_policy_recommendation.main(df_frequency_sweep, ['intensity_1', 'intensity_16'])
-
-        uncore_efficient_freq = policy['CPU_UNCORE_FREQ_EFFICIENT']
-        uncore_mbm_list = []
-
-        for key,value in policy.items():
-            if key not in ("CPU_FREQ_MAX", "CPU_FREQ_EFFICIENT", "CPU_UNCORE_FREQ_MAX", "CPU_UNCORE_FREQ_EFFICIENT", "CPU_PHI"):
-                uncore_mbm_list.append(value)
+        uncore_config = gen_cpu_activity_constconfig_recommendation.get_config_from_frequency_sweep(df_frequency_sweep,
+                                                                                                    ['intensity_1',
+                                                                                                     'intensity_16'])
+        uncore_efficient_freq = uncore_config['CPU_UNCORE_FREQUENCY_EFFICIENT_HIGH_INTENSITY']['values'][0]
 
         #######################################################
         # Core frequency sweep at fixed uncore_efficient_freq #
@@ -162,8 +159,19 @@ class TestIntegration_cpu_activity(unittest.TestCase):
         # Parse data #
         ##############
         df_frequency_sweep = geopmpy.io.RawReportCollection('*report', dir_name=cls._aib_core_freq_sweep_dir).get_df()
-        policy = gen_cpu_activity_policy_recommendation.main(df_frequency_sweep, ['intensity_1', 'intensity_16'])
-        cpu_efficient_freq = policy['CPU_FREQ_EFFICIENT']
+
+        core_config = gen_cpu_activity_constconfig_recommendation.get_config_from_frequency_sweep(df_frequency_sweep,
+                                                                                                  ['intensity_1',
+                                                                                                   'intensity_16'])
+        # The core config has the updated CPU Fe value,
+        # but none of the memory bandwidth info so we
+        # combine them into a full const config file
+        uncore_config['CPU_FREQUENCY_EFFICIENT_HIGH_INTENSITY']['values'][0] = core_config['CPU_FREQUENCY_EFFICIENT_HIGH_INTENSITY']['values'][0]
+
+        json_config = json.dumps(uncore_config, indent=4)
+        with open("const_config_io-ca.json", "w") as outfile:
+            outfile.write(json_config)
+        os.environ["GEOPM_CONST_CONFIG_PATH"] = "const_config_io-ca.json"
 
         ################################
         # CPU Activity Agent phi sweep #
@@ -183,11 +191,6 @@ class TestIntegration_cpu_activity(unittest.TestCase):
             enable_traces=False,
             enable_profile_traces=False,
             verbose=False,
-            cpu_fe=cpu_efficient_freq,
-            cpu_fmax=cpu_max_freq,
-            uncore_fe=uncore_efficient_freq,
-            uncore_fmax=cpu_max_freq,
-            uncore_mbm_list=uncore_mbm_list,
             phi_list=[0.2, 0.5, 0.7],
         )
 
--- a/src/CPUActivityAgent.cpp
+++ b/src/CPUActivityAgent.cpp
@@ -179,7 +179,7 @@ namespace geopm
 
         if (m_qm_max_rate.empty()) {
             throw Exception("CPUActivityAgent::" + std::string(__func__) +
-                            "(): ConstConfigIO file did not contain" +
+                            "(): ConstConfigIOGroup configuration file does not contain" +
                             " memory bandwidth information.",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
--- a/src/CPUActivityAgent.cpp
+++ b/src/CPUActivityAgent.cpp
@@ -64,9 +64,11 @@ namespace geopm
         // just what the user/admin has previously set.
         m_freq_uncore_min = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
         m_freq_uncore_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+        m_resolved_f_uncore_max = m_freq_uncore_max;
 
         if (level == 0) {
             init_platform_io();
+            init_constconfig_io();
         }
     }
 
@@ -100,6 +102,9 @@ namespace geopm
 
         m_freq_core_min = m_freq_governor->get_frequency_min();
         m_freq_core_max = m_freq_governor->get_frequency_max();
+
+        // init to system max
+        m_resolved_f_core_max = m_freq_core_max;
 
         for (int domain_idx = 0; domain_idx < m_num_freq_ctl_domain; ++domain_idx) {
             m_core_scal.push_back({m_platform_io.push_signal("MSR::CPU_SCALABILITY_RATIO",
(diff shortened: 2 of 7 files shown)


integration/test/test_cpu_activity_agent.py
```diff
@@ -162,8 +159,19 @@ def launch_helper(experiment_type, experiment_args, app_conf, experiment_cli_arg
...
+                                                                                                   'intensity_16'])
+        # The core config has the updated CPU Fe value,
+        # but none of the memory bandwidth info so we
+        # combine them into a full const config file
+        uncore_config['CPU_FREQUENCY_EFFICIENT_HIGH_INTENSITY']['values'][0] = core_config['CPU_FREQUENCY_EFFICIENT_HIGH_INTENSITY']['values'][0]
+
+        json_config = json.dumps(uncore_config, indent=4)
+        with open("const_config_io-ca.json", "w") as outfile:
+            outfile.write(json_config)
+        os.environ["GEOPM_CONST_CONFIG_PATH"] = "const_config_io-ca.json"
```
dannosliwcd: Is it possible for an app-under-evaluation to change its working directory before this variable is read? If so, maybe convert to an absolute path.

src/CPUActivityAgent.cpp
```diff
@@ -299,34 +249,47 @@ namespace geopm
...
+        double f_core_range = m_freq_core_max - m_freq_core_efficient;
+        double f_uncore_range = m_freq_uncore_max - m_freq_uncore_efficient;
 
-            if (m_qm_max_rate.empty()) {
-                throw Exception("CPUActivityAgent::" + std::string(__func__) +
-                                "(): CPUActivityAgent policy did not contain" +
-                                " memory bandwidth characterization.",
-                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-            }
+        double phi = in_policy[M_POLICY_CPU_PHI];
```
dannosliwcd, line 248: This function always checks `in_policy` and re-evaluates what the frequency settings should be, which suggests that you might be considering cases where the incoming policy changes over time (e.g., if running with the endpoint).

However, the target frequencies are only set for `phi != 0.5` cases, which suggests that you expect the target frequencies to change between initialization and the first adjust.

dannosliwcd (commented):
Partial review pass. Stopping here for planning meetings.

integration/test/test_cpu_activity_agent.py
lhlawson: The app under evaluation shouldn't change the working directory before the variable is read, but moving to absolute path seems safer in either case.

src/CPUActivityAgent.cpp
lhlawson: Great catch.  
The default case of phi=0.5 is intended to be the full frequency range, but was not being set in the adjust platform.

Previously the freq efficient & max setting was part of the policy, and adjusted in each validate_policy loop based upon the phi value.  Now that those are pulled in via signals we need to handle this in the adjust_platform loop.

dannosliwcd (dismissed):

integration/test/test_cpu_activity_agent.py
```diff
@@ -137,14 +138,10 @@ def launch_helper(experiment_type, experiment_args, app_conf, experiment_cli_arg
...
-        uncore_efficient_freq = policy['CPU_UNCORE_FREQ_EFFICIENT']
-        uncore_mbm_list = []
-
-        for key,value in policy.items():
-            if key not in ("CPU_FREQ_MAX", "CPU_FREQ_EFFICIENT", "CPU_UNCORE_FREQ_MAX", "CPU_UNCORE_FREQ_EFFICIENT", "CPU_PHI"):
-                uncore_mbm_list.append(value)
+        uncore_config = gen_cpu_activity_constconfig_recommendation.get_config_from_frequency_sweep(df_frequency_sweep,
+                                                                                                    ['intensity_1',
+                                                                                                     'intensity_16'])
+        uncore_efficient_freq = uncore_config['CPU_UNCORE_FREQUENCY_EFFICIENT_HIGH_INTENSITY']['values'][0]
```
avilcheslopez, line 144: Note this variable is set but never used. Is this not needed or are you missing some code somewhere?

src/CPUActivityAgent.cpp
```diff
@@ -136,46 +138,61 @@ namespace geopm
...
+        else {
+            m_freq_uncore_efficient = m_freq_uncore_min;
         }
 
-        // When the policy is all NaNs, this check also verifies that
-        // the system was not left in a bad state with regard to the
-        // UNCORE_FREQUENCY_<MAX/MIN>_CONTROLs.
-        if (in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT] > in_policy[M_POLICY_UNCORE_FREQ_MAX]) {
+        // Grab all (uncore frequency, max memory bandwidth) pairs
+        for (int entry_idx = 0; entry_idx < (int)ALL_NAMES.size(); ++entry_idx) {
```
avilcheslopez: I know we don't follow this strictly in the codebase, but I would recommend using C++-style casts instead of C-style casts.

However, looking at this ``for`` loop more closely, it seems you should be able to use an unsigned int type for ``entry_idx`` and avoid having to do any casts.

src/CPUActivityAgent.cpp
```diff
@@ -136,46 +138,61 @@ namespace geopm
...
+            const std::string KEY_NAME = "CONST_CONFIG::CPU_UNCORE_FREQUENCY_" +
+                                          std::to_string(entry_idx);
+            const std::string VAL_NAME = "CONST_CONFIG::CPU_UNCORE_MAX_MEMORY_BANDWIDTH_" +
+                                          std::to_string(entry_idx);
+            if (ALL_NAMES.find(KEY_NAME) != ALL_NAMES.end() &&
+                ALL_NAMES.find(VAL_NAME) != ALL_NAMES.end()) {
```
avilcheslopez: Previously, you were using the ``count()`` method to search in the ``ALL_NAMES`` set, whereas here you're using the ``find()`` method. I recommend you remain consistent. Using ``find()`` seems more readable, but I like that using the ``count()`` method avoids creating a couple of iterators (and with set the count is going to be either 0 or 1). I have no strong preference, just remain consistent.

src/CPUActivityAgent.cpp
```diff
@@ -136,46 +138,61 @@ namespace geopm
...
+                double uncore_freq = m_platform_io.read_signal(KEY_NAME, GEOPM_DOMAIN_BOARD, 0);
+                double max_mem_bw = m_platform_io.read_signal(VAL_NAME, GEOPM_DOMAIN_BOARD, 0);
+                if (!std::isnan(uncore_freq) && !std::isnan(max_mem_bw) &&
+                    uncore_freq != 0 && max_mem_bw != 0) {
+                    m_qm_max_rate[uncore_freq] = max_mem_bw;
+                }
```
avilcheslopez, line 176: You know, I was thinking there must be a way to know how many pairs of uncore_freq-max_mem_bw are included in the configuration instead of having to sweep the entire domain of indices and seeing which names are valid. Would it be a terrible idea to just include that as one of the configuration parameters/signals in the ConstConfig file? Do any other more elegant ideas come to mind?

src/CPUActivityAgent.cpp
```diff
@@ -299,34 +241,55 @@ namespace geopm
...
+        // Default phi = 0.5 case is full fe to fmax range
+        // Core
+        m_resolved_f_core_max = m_freq_core_max;
+        m_resolved_f_core_efficient = m_freq_core_efficient;
+        // Uncore
+        m_resolved_f_uncore_max = m_freq_uncore_max;
+        m_resolved_f_uncore_efficient = m_freq_uncore_efficient;
+
+        // If phi is not 0.5 we move into the energy or performance biased behavior
+        if (phi > 0.5) {
```
avilcheslopez, line 259: Q: can phi change during the execution? If not, then I'm thinking this code should be executed once (i.e. moved to the init phase).

src/CPUActivityAgent.cpp
Krebrov001: Yeah, like this:
```cpp
size_t ALL_NAMES_size = ALL_NAMES.size();
for (size_t entry_idx = 0; entry_idx < ALL_NAMES_size; ++entry_idx) {
```

integration/test/test_cpu_activity_agent.py
lhlawson: It's used in the experiment setup for the core freq sweep below (~line 156):
experiment_args.min_uncore_frequency = uncore_efficient_freq
experiment_args.max_uncore_frequency = uncore_efficient_freq

src/CPUActivityAgent.cpp
lhlawson: My natural preference is find, which is why inconsistency cropped up, but we use .count() elsewhere in GEOPM so I'll convert to count()

src/CPUActivityAgent.cpp
lhlawson: In our current usage Phi does not change during execution, however we want the capability to change the phi value passed to child agents in the future for potential multi-node algorithms.

Phi is provided via the policy, so I'm not sure how we'd move it to init since we have to call validate_policy.  Previously this logic on checking phi was handled in validate_policy.

src/CPUActivityAgent.cpp
lhlawson: Including a count signal of some sort would be a viable solution, but we have no way of enforcing accuracy (i.e. the count sig says 10 pairs, but there are 11) that aren't reliant upon doing the same sweep.  We could rely on the agent to tell us if there are less pairs than specified by the signal (i.e. missing signal throw from PlatformIO) but not more.  

Ultimately this is called once as part of init so I don't know if the search overhead is significant enough to be a concern.

avilcheslopez: Yeah, that makes sense. I wasn't really concerned with the efficiency of the code, but rather its elegance and readability (because it took me a moment to realize why such a sweep was being done). However, even the idea I suggested would still require a similar sweep.

src/CPUActivityAgent.cpp
avilcheslopez: Right, good point (init wouldn't be the right place for it). So I would recommend leaving it in validate policy until there is a need to accommodate potential multi-node algorithms in the future.

lhlawson: Validate policy is const (can't modify member variables).  Previously we worked around this by modifying the in_policy values for max freq & efficient freq, but those have been pulled out of the policy in this PR (using the GPU IOGroup & ConstConfig).

avilcheslopez: Well, then :)

avilcheslopez (dismissed):

dannosliwcd (dismissed):

test/CPUActivityAgentTest.cpp
```diff
@@ -50,18 +50,11 @@ class CPUActivityAgentTest : public ::testing::Test
...
-            PHI = 4,
-            UNCORE_FREQ_0 = 5,
-            UNCORE_MEM_BW_0 = 6,
-            UNCORE_FREQ_1 = 7,
-            UNCORE_MEM_BW_1 = 8,
+            PHI = 0
         };
 
         void SetUp();
+        void SetUpDefaultExpectCalls();
```
avilcheslopez: Despite GTest's naming convention, we still follow our naming convention in test fixture classes (lowercase with underscores).

test/CPUActivityAgentTest.cpp
```diff
@@ -595,27 +519,27 @@ TEST_F(CPUActivityAgentTest, adjust_platform_signal_out_of_bounds)
...
-                .WillRepeatedly(Return(m_cpu_uncore_freq_max));
-    m_agent->sample_platform(tmp);
+    EXPECT_CALL(*m_platform_io, push_control("CPU_UNCORE_FREQUENCY_MIN_CONTROL", _, _)).Times(M_NUM_PACKAGE);
+    EXPECT_CALL(*m_platform_io, push_control("CPU_UNCORE_FREQUENCY_MAX_CONTROL", _, _)).Times(M_NUM_PACKAGE);
+
+    EXPECT_CALL(*m_platform_io, write_control("MSR::PQR_ASSOC:RMID", _, _, _)).Times(1);
+    EXPECT_CALL(*m_platform_io, write_control("MSR::QM_EVTSEL:RMID", _, _, _)).Times(1);
+    EXPECT_CALL(*m_platform_io, write_control("MSR::QM_EVTSEL:EVENT_ID", _, _, _)).Times(1);
+    EXPECT_CALL(*m_platform_io, read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", _, _)).Times(1);
+    EXPECT_CALL(*m_platform_io, read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", _, _)).Times(1);
```
avilcheslopez: This is almost ``SetUpDefaultExpectCalls``, so I can't help but think there might be an opportunity to factor out some code. If it ends up being more trouble than any benefit that provides, feel free to disregard.

avilcheslopez (dismissed):

src/CPUActivityAgent.cpp
```diff
@@ -136,46 +138,61 @@ namespace geopm
...
+
+        if (m_qm_max_rate.empty()) {
             throw Exception("CPUActivityAgent::" + std::string(__func__) +
-                            "():CPU_UNCORE_FREQ_EFFICIENT (" +
-                            std::to_string(in_policy[M_POLICY_UNCORE_FREQ_EFFICIENT]) +
-                            ") value exceeds CPU_UNCORE_FREQ_MAX (" +
-                            std::to_string(in_policy[M_POLICY_UNCORE_FREQ_MAX]) +
-                            ").", GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                            "(): ConstConfigIO file did not contain" +
+                            " memory bandwidth information.",
```
avilcheslopez:
```suggestion
                            "(): ConstConfigIOGroup configuration file does not contain" +
                            " memory bandwidth information.",
```

test/CPUActivityAgentTest.cpp
```diff
@@ -595,27 +525,16 @@ TEST_F(CPUActivityAgentTest, adjust_platform_signal_out_of_bounds)
...
-    EXPECT_CALL(*m_platform_io, sample(CPU_UNCORE_FREQUENCY_IDX))
-                .WillRepeatedly(Return(m_cpu_uncore_freq_max));
-    m_agent->sample_platform(tmp);
+    std::set<std::string> signal_name_set = {"CPU_FREQUENCY_EFFICIENT_HIGH_INTENSITY",
+                                             "CPU_UNCORE_FREQUENCY_EFFICIENT_HIGH_INTENSITY"};
+
+    ON_CALL(*m_platform_io, signal_names()).WillByDefault(Return(signal_name_set));
+    // leaf agent
+    GEOPM_EXPECT_THROW_MESSAGE(m_agent->init(0, {}, false), GEOPM_ERROR_INVALID,
+                                "ConstConfigIO file did not contain memory bandwidth information");
```
avilcheslopez:
```suggestion
                                "ConstConfigIOGroup configuration file does not contain memory bandwidth information");
```

avilcheslopez (approved):

PR #2670, opened 2022-10-25
Initial config for pre-commit

dannosliwcd:
- Add an initial configuration for pre-commit.com checks to run on this repository
- Related to #2530

Some of these pre-commit tests fail on the current contents of the repo. Some pre-work exists to resolve those failures.

To run a single check on all files in the repo, use the `--all-files` option in `pre-commit run`. E.g., for the geopm license checks (that one succeeds for all files right now):

```
pre-commit run geopm-license --all-files
```

Or to find issues with rst file backtick usage:

```
pre-commit run rst-backticks --all-files
```

PR #2695, opened 2022-11-08
Add a spell-checking linter

dannosliwcd:
- Fix spell-checking errors detected by codespell
- Add codespell to a new linter job in our GitHub Actions automation
- Add an initial pre-commit hook configuration that runs codespell

**Why both an action and a pre-commit hook?**
Actions can be used to gate PRs that are candidates to merge into dev, so every PR contribution goes through those checks. Pre-commit hooks do not run for anyone who chooses not to opt in to installing them. However, pre-commit hooks are are still valuable to contributors because they let the contributor see whether each individual commit introduces new issues, instead of only seeing those issues all at once when a PR is pushed.

Draft PR #2670 has some examples of other light-weight checks that might be useful as pre-commit and action linters.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = googletest-release*,autom4te.cache,./service/docs/build,./service/contrib,libtool*,ltmain*,./integration/apps/*/*,./.git,./ChangeLog,*.patch,./test/ProfileTableTest.cpp,configure,config.*,depcomp,*.spec
+ignore-words-list = atleast,clos,fom,warmup,affinitize,affinitized,whis
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/codespell-project/codespell
+    rev: 'v2.2.0'
+    hooks:
+    -   id: codespell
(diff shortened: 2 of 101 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/codespell-project/codespell
```
Krebrov001, line 4: I think that this is really cool.

Krebrov001 (dismissed):

.codespellrc
```diff
@@ -0,0 +1,3 @@
+[codespell]
+skip = googletest-release*,autom4te.cache,./service/docs/build,./service/contrib,libtool*,ltmain*,./integration/apps/*/*,./.git,./ChangeLog,*.patch,./test/ProfileTableTest.cpp,configure,config.*,depcomp,*.spec
+ignore-words-list = atleast,clos,fom,warmup,affinitize,affinitized,whis
```
cmcantalupo, line 3: I guess atleast and whis are okay?  Are these referring to code?

cmcantalupo (dismissed):
LGTM, only question really is about why "atleast" is valid.

.codespellrc
dannosliwcd: Yeah, we use `atleast` in some test code and `whis` in box-and-whisker plotting code. Both names are provided by external packages so we can't rename them.

cmcantalupo (approved):

Issue #2697, opened 2022-11-09
Add support for controlling default GPU frequency in FrequencyMapAgent


avilcheslopez:
> estimate 1
Parent [feature/story]: #2008 

The implementation for the [...] feature was modified because [...].

Done criteria:

- [ ] Integration tests that cover the feature run and pass
- [ ] Dependencies mocked in unit tests are still valid
- [ ] Documentation changes are not required since this change does
      not modify interfaces or features.
- [ ] [...]
- [ ] [...]

avilcheslopez:
Pending: integration test that leverages the new agent feature.

PR #2699, opened 2022-11-09
Adding GPU support to FrequencyMap Agent

asmaalrawi:
- Relates to #2697

Adds a default GPU frequency setting to the policy for the frequency map agent. Also adds an integration experiment directory for conducting gpu frequency sweeps.

_Note_:

Prior to creating a pull request, please see the documentation at
either of the following links:

https://github.com/geopm/geopm/blob/dev/CONTRIBUTING.rst#change-request

https://geopm.github.io/contrib.html#change-request

about the process to change the GEOPM repository.


integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2015 - 2022, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+'''
+Helper functions for running power sweep experiments.
```
dannosliwcd: power sweep -> gpu frequency sweep

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,125 @@
...
+                        help='bottom gpu frequency setting for the sweep')
+    parser.add_argument('--step-gpu-frequency', dest='step_gpu_frequency',
+                        action='store', type=float, default=None,
+                        help='increment in hertz between gpu frequency settings for the sweep')
+
+
+def setup_gpu_frequency_bounds(mach, min_gpu_freq, max_gpu_freq,
+                                  step_gpu_freq):
+
+    #TODO: Don't hardcode this. What.
```
dannosliwcd: Need to mimic add  `gpu_frequency_*()` functions to machine.py like the `frequency_*()` in there. `GPU_CORE_FREQENCY_MAX_AVAIL` and `GPU_CORE_FREQENCY_MIN_AVAIL` then call those functions here.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,125 @@
...
+    if (max_gpu_freq - min_gpu_freq) % step_gpu_freq != 0:
+        sys.stderr.write('<geopm> Warning: gpu frequency range not evenly divisible by step size.\n')
+    if min_gpu_freq < sys_min or max_gpu_freq > sys_max:
+        raise RuntimeError('GPU frequency bounds are out of range for this system')
+    if min_gpu_freq > max_gpu_freq:
+        raise RuntimeError('GPU frequency min is greater than max')
+
+    num_step = 1 + int((max_gpu_freq - min_gpu_freq) // step_gpu_freq)
+    gpu_freqs = [step_gpu_freq * ss + min_gpu_freq for ss in range(num_step)]
+    gpu_freqs = sorted(gpu_freqs, reverse=True)
```
dannosliwcd: These statements combine to give a decreasing list of integer frequency steps. That can be directly achieved with something like:
`list(range(max_gpu_freq, min_gpu_freq - step_gpu_freq, -step_gpu_freq))`

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,125 @@
...
+
+def launch_configs(app_conf, core_freq_range, gpu_freq_range):
+    agent = 'frequency_map'
+    targets = []
+    for freq in core_freq_range:
+        for gpu_freq in gpu_freq_range:
+            name = '{:.1e}c_{:.1e}u'.format(freq, gpu_freq)
+            options = {'FREQ_DEFAULT': freq,
+                       'FREQ_GPU_DEFAULT': gpu_freq}
+            agent_conf = geopmpy.agent.AgentConf('{}_agent_{}c_{}u.config'.format(agent, freq, gpu_freq), agent, options)
```
dannosliwcd: I think `{}c` in the frequency-sweep script was CPU core frequency and `{}u` was CPU uncore frequency. Maybe in this script `c{}` could be CPU, but if that's the case, then `u{}` should probably be `g{}` for GPU.

src/FrequencyMapAgent.cpp
```diff
@@ -255,6 +269,15 @@ namespace geopm
             }
             m_platform_io.adjust(m_uncore_min_ctl_idx, m_uncore_init_min);
             m_platform_io.adjust(m_uncore_max_ctl_idx, m_uncore_init_max);
+
+            if(m_num_gpu_freq_ctl_domain > 0) {
```
dannosliwcd: You can drop the `if(...)` since it is equivalent to the initial check in the `for(...)`. Or if those aren't equivalent, add a comment to explain.

It might be different if this domain count can be a borrowed/negative value, since the sign gets cast away 🏐. That can be avoided by using ssize_t if negative values are needed.

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +318,25 @@ namespace geopm
             }
             m_last_uncore_freq = m_uncore_freq;
         }
+
+		//TODO: Enable hash gpu freq mapping
+		//adjust fixed gpu freq
+	if (m_last_gpu_freq != m_gpu_freq) {
```
dannosliwcd: I don't think `m_gpu_freq` is defined.

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +318,25 @@ namespace geopm
...
+		if (!std::isnan(m_gpu_freq) && m_num_gpu_freq_ctl_domain > 0 ) {
+			for(size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
+				double val = m_platform_io.read_signal("GPU_FREQUENCY_MAX_CONTROL", 
+						m_gpu_freq_ctl_domain_type, ctl_idx);
+				m_platform_io.adjust(m_gpu_freq, val);
```
dannosliwcd: The first argument to adjust should be a platformIO control index

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +318,25 @@ namespace geopm
             }
             m_last_uncore_freq = m_uncore_freq;
         }
+
+		//TODO: Enable hash gpu freq mapping
+		//adjust fixed gpu freq
+	if (m_last_gpu_freq != m_gpu_freq) {
+		if (!std::isnan(m_gpu_freq) && m_num_gpu_freq_ctl_domain > 0 ) {
+			for(size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
```
dannosliwcd: Add a comment explaining why a read/write-back is needed

src/FrequencyMapAgent.cpp
```diff
@@ -320,9 +362,10 @@ namespace geopm
     std::vector<std::string> FrequencyMapAgent::policy_names(void)
     {
 
-        std::vector<std::string> names{"FREQ_DEFAULT", "FREQ_UNCORE"};
+        std::vector<std::string> names{"FREQ_DEFAULT", "FREQ_UNCORE", "FREQ_GPU_DEFAULT"};
```
dannosliwcd: A user might assume "FREQ_DEFAULT" defines a default frequency for any frequency controls covered by this agent.

Maybe `"FREQ_CPU_CORE_DEFAULT", "FREQ_CPU_UNCORE_DEFAULT", "FREQ_GPU_CORE_DEFAULT"`?

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,135 @@
...
+                        action='store', type=float, default=None,
+                        help='increment in hertz between gpu frequency settings for the sweep')
+
+def setup_gpu_frequency_bounds(mach, min_gpu_freq, max_gpu_freq,
+                                  step_gpu_freq):
+
+    sys_min = mach.gpu_frequency_min()
+    sys_max = mach.gpu_frequency_max()
+
+    #TODO: Is there a GPU signal equivalent for this?
```
dannosliwcd: I don't think so. The way we did it pre-IOGroup was to request a list of allowed frequency limits from the GPU.

We might just need to pick some default number of steps to try, then get the actual applied limit from the reports.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,135 @@
...
+    if min_gpu_freq is None:
+        min_gpu_freq = sys_min
+    if max_gpu_freq is None:
+        max_gpu_freq = sys_max
+    if step_gpu_freq is None:
+        step_gpu_freq = sys_step
+    if step_gpu_freq < sys_step or step_gpu_freq % sys_step != 0:
+        sys.stderr.write('<geopm> Warning: gpu frequency step size may be incompatible with p-states.\n')
+    if (max_gpu_freq - min_gpu_freq) % step_gpu_freq != 0:
+        sys.stderr.write('<geopm> Warning: gpu frequency range not evenly divisible by step size.\n')
```
dannosliwcd: We should test out whether this is actually a problem. Since step size is not known, it means some kind of rounding mechanism needs to be applied anyways. Rounding to an integer step doesn't seem like a special case on top of that.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,135 @@
...
+        raise RuntimeError('GPU frequency min is greater than max')
+
+    gpu_freqs = reversed(arange(min_gpu_freq, max_gpu_freq, step_gpu_freq))
+    return gpu_freqs
+
+def arange(minv, maxv, stepv):
+    rval = [minv]
+    while rval[-1] + stepv <= maxv:
+        rval.append(rval[-1] + stepv)
+    return rval
```
dannosliwcd: Why not just use the built-in `range` or the numpy `arange`? This behaves differently from the typical `arange`, which returns a numpy array instead of a list and is a half-open interval, so a different name should probably be used if we define it here.

dannosliwcd: Also thinking about the TODO question about whether we know step size:

One option for default behavior might be to use np.linspace, e.g., to get 10 steps between min and max when the user doesn't give a step size.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,135 @@
...
+
+def report_signals():
+    return ["CPU_CYCLES_THREAD@package", "CPU_CYCLES_REFERENCE@package",
```
dannosliwcd, line 75: If we can't get step size and if we need to rely on some kind of rounding at the IOGroup or below, then we might benefit from reading back GPU_CORE_FREQUENCY_MAX_CONTROL here.

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +317,21 @@ namespace geopm
...
+	if (m_last_gpu_freq != m_default_gpu_freq) {
+	    if (!std::isnan(m_default_gpu_freq) && m_num_gpu_freq_ctl_domain > 0 ) {
+		for(size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
+		    m_platform_io.adjust(m_gpu_control_idx[ctl_idx], m_default_gpu_freq);
+		    m_do_write_batch = true;
+		}
+	    }
+	    else if (!std::isnan(m_last_gpu_freq) && m_num_gpu_freq_ctl_domain > 0) {
+		m_platform_io.adjust(m_gpu_min_ctl_idx, m_gpu_init_min);
+		m_platform_io.adjust(m_gpu_max_ctl_idx, m_gpu_init_max);
```
dannosliwcd: I don't think either of these idx values have been pushed to PIO.

test/FrequencyMapAgentTest.cpp
```diff
@@ -45,20 +45,23 @@ class FrequencyMapAgentTest : public :: testing :: Test
...
             UNCORE = 1,
-            HASH_0 = 2,
-            FREQ_0 = 3,
-            HASH_1 = 4,
-            FREQ_1 = 5,
+	    GPU_DEFAULT = 2, // M_POLICY_FREQ_GPU_DEFAULT
+            HASH_0 = 3,
+            FREQ_0 = 4,
+            HASH_1 = 5,
+            FREQ_1 = 6,
```
dannosliwcd, line 60: There's a TODO earlier in this diff: `//TODO: In future, add GPU hash frequency mapping support`

I guess you could do something like:
```suggestion
            HASH_0 = 3,
            CPU_FREQ_0 = 4,
            GPU_FREQ_0 = 5,
            HASH_1 = 6,
            CPU_FREQ_1 = 7,
            GPU_FREQ_1 = 8,
```

src/FrequencyMapAgent.hpp
```diff
@@ -64,12 +64,15 @@ namespace geopm
...
+                M_POLICY_FREQ_GPU_DEFAULT,
                 M_POLICY_FIRST_HASH,
                 M_POLICY_FIRST_FREQUENCY,
                 // The remainder of policy values can be additional pairs of
                 // (hash, frequency)
-                M_NUM_POLICY = 64,
+                // TODO: This number was 64. I changed it so the names vector
+                //       would be appropriately sized. Should I fix the logic
+                //       in the code instead?
+                M_NUM_POLICY = 65,
```
dannosliwcd, line 72: Good call. Yeah, something like the following is one way it could be enforced as self-consistent:
```
M_POLICY_FREQ_DEFAULT,
M_POLICY_FREQ_UNCORE,
M_POLICY_FREQ_GPU_DEFAULT,
M_POLICY_AFTER_LAST_HEADER_FIELD,

M_POLICY_FIRST_HASH = M_POLICY_AFTER_LAST_HEADER_FIELD,
M_POLICY_FIRST_FREQUENCY,
M_POLICY_AFTER_LAST_RECURRING_FIELD,

// Note: 31 was originally chosen because it let us use a nice round total policy size of 64, but
// it's pretty arbitrary. Could probably just replace 31 with a named constant so readers know it is the
// number of regions we support in the FrequencyMapAgent's policy.
M_NUM_POLICY = 31 * (M_POLICY_AFTER_LAST_RECURRING_FIELD - M_POLICY_AFTER_LAST_HEADER_FIELD) + M_POLICY_AFTER_LAST_HEADER_FIELD
```

dannosliwcd: If something like that is added, you could also prevent future similar issues if GPU freqs are added to the recurring part. Specifically, there are some places in the `.cpp` file that step through the recurring region of policies by `std::advance(it, 2)`. The `2` could instead use `M_POLICY_AFTER_LAST_RECURRING_FIELD - M_POLICY_AFTER_LAST_HEADER_FIELD` so it would stay consistent if a new type of frequency is added to the recurring part of the policy later on.

src/FrequencyMapAgent.cpp
```diff
@@ -8,6 +8,7 @@
 #include <sstream>
 #include <cmath>
 #include <iomanip>
+// TODO delete this include
```
avilcheslopez: Go for it

src/FrequencyMapAgent.cpp
```diff
@@ -200,9 +206,17 @@ namespace geopm
...
+            m_is_policy_updated = true;
+        }
+
         m_uncore_freq = policy[M_POLICY_FREQ_UNCORE];
-    }
 
+        m_default_gpu_freq = policy[M_POLICY_FREQ_GPU_DEFAULT];
+
+
+        }
```
avilcheslopez: Nit: indentation

src/FrequencyMapAgent.cpp
```diff
@@ -248,17 +262,27 @@ namespace geopm
         m_do_write_batch = false;
 
         if (!m_is_adjust_initialized) {
+            std::cout << "m_is_adjust_initialized = false" << std::endl;
```
avilcheslopez: Debug print statement?

src/FrequencyMapAgent.cpp
```diff
@@ -248,17 +262,27 @@ namespace geopm
...
             // adjust all controls once in case not applied by policy
             for (size_t ctl_idx = 0; ctl_idx < (size_t) m_num_freq_ctl_domain; ++ctl_idx) {
                 double val = m_platform_io.read_signal("CPU_FREQUENCY_MAX_CONTROL", m_freq_ctl_domain_type, ctl_idx);
                 m_platform_io.adjust(m_freq_control_idx[ctl_idx], val);
             }
             m_platform_io.adjust(m_uncore_min_ctl_idx, m_uncore_init_min);
             m_platform_io.adjust(m_uncore_max_ctl_idx, m_uncore_init_max);
+
+            std::cout << "number of gpu ctl domains: " << m_num_gpu_freq_ctl_domain << std::endl;
```
avilcheslopez: Debug print statement?

src/FrequencyMapAgent.cpp
```diff
@@ -248,17 +262,27 @@ namespace geopm
...
+            for(size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
```
avilcheslopez: I would advise against using C-style casts. However, I believe you should be able to get away with making ``ctl_idx`` ``int`` to avoid having to do it. If you do have to perform a cast, I recommend using C++'s cast operators (e.g. ``static_cast<>``).

Update: actually, I'm seeing this same cast is sprinkled all over this file. At this point just leave it consistent with the rest of the code for now. I think it's out of the scope of this PR to address this.

Nit: ``for (`` (space after ``for`` keyword)

src/FrequencyMapAgent.cpp
```diff
@@ -248,17 +262,27 @@ namespace geopm
...
+                    double val = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_CONTROL", 
+                                    m_gpu_freq_ctl_domain_type, ctl_idx);
```
avilcheslopez: Indentation nit: align with parameters in previous line.

src/FrequencyMapAgent.cpp
```diff
@@ -248,17 +262,27 @@ namespace geopm
...
+                    m_platform_io.adjust(m_gpu_control_idx[ctl_idx], val);
+            }
             m_is_adjust_initialized = true;
-        }
+
```
avilcheslopez: Get rid of extra new line

src/FrequencyMapAgent.cpp
```diff
@@ -248,17 +262,27 @@ namespace geopm
...
+        } 
 
         if (is_all_nan(in_policy) && !m_is_real_policy) {
+            std::cout << "m_is_adjust_initialized = true" << std::endl;
```
avilcheslopez: Debug print statement?

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +319,17 @@ namespace geopm
             }
             m_last_uncore_freq = m_uncore_freq;
         }
+
+        // adjust fixed gpu freq
+        if (m_last_gpu_freq != m_default_gpu_freq) {
+            if (!std::isnan(m_default_gpu_freq)) {
+                for(size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
```
avilcheslopez: Nit: space after ``for``

src/FrequencyMapAgent.cpp
```diff
@@ -417,6 +453,12 @@ namespace geopm
                 m_platform_io.write_control("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0,
                                             policy[M_POLICY_FREQ_UNCORE]);
         }
+
+        if (!std::isnan(policy[M_POLICY_FREQ_GPU_DEFAULT])) {
+                m_platform_io.write_control("GPU_CORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0,
+                                            policy[M_POLICY_FREQ_GPU_DEFAULT]);
+        }
+
```
avilcheslopez: Remove extra new line

src/FrequencyMapAgent.cpp
```diff
@@ -427,6 +469,8 @@ namespace geopm
                                             GEOPM_REGION_HASH_UNMARKED);
         m_last_freq = std::vector<double>(m_num_freq_ctl_domain, NAN);
         for (size_t ctl_idx = 0; ctl_idx < (size_t) m_num_freq_ctl_domain; ++ctl_idx) {
+
+
```
avilcheslopez: Remove extra new lines

src/FrequencyMapAgent.cpp
```diff
@@ -441,5 +485,22 @@ namespace geopm
         m_core_freq_max = m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
         m_uncore_init_min = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
         m_uncore_init_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+
+
```
avilcheslopez: Remove extra new line

src/FrequencyMapAgent.cpp
```diff
@@ -441,5 +485,22 @@ namespace geopm
         m_core_freq_max = m_platform_io.read_signal("CPU_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
         m_uncore_init_min = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
         m_uncore_init_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+
+        
+        m_gpu_freq_ctl_domain_type = m_platform_io.control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL");
+        if (m_gpu_freq_ctl_domain_type == GEOPM_DOMAIN_INVALID) {
+            m_num_gpu_freq_ctl_domain = 0;
+        }
```
avilcheslopez: What is the purpose of proceeding in the case of an invalid GEOPM domain for "GPU_CORE_FREQUENCY_MAX_CONTROL"? The reason for asking is because I believe ``control_domain_type`` will throw in the case it's unable to determine the domain type for a control.

src/FrequencyMapAgent.cpp
```diff
@@ -200,9 +206,17 @@ namespace geopm
...
+            m_default_gpu_freq != policy[M_POLICY_FREQ_GPU_DEFAULT]) {
+            m_is_policy_updated = true;
+        }
+
         m_uncore_freq = policy[M_POLICY_FREQ_UNCORE];
-    }
 
+        m_default_gpu_freq = policy[M_POLICY_FREQ_GPU_DEFAULT];
+
+
```
avilcheslopez: Remove extra new line

avilcheslopez (commented):
Forgot to mention: the agent's chapter 7 man page should be updated as well.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
asmaalrawi: Added!

src/FrequencyMapAgent.cpp
asmaalrawi: I assumed that if GPUs are not supported on the system, we would get a GEOPM_DOMAIN_INVALID but still want to operate without throwing an error.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2015 - 2022, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
```
lhlawson, line 6: Make sure to add [co-authors](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) for anything re-used from other pull requests (ex: [gpu_frequency_sweep.py](https://github.com/geopm/geopm/blob/2ddf7e6f7146b2de0bd6f9838a416090a9d4f434/integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py) from #2024)

asmaalrawi: Happy to add Sari; I'm not sure how much similarity there is in the code. I couldn't copy the vendor-specific code as shown in 2024.  Most of the code that is repeated is copied from the other experiment scripts, particularly the frequency sweeps. Should I be adding the original authors in there, too?

lhlawson: I'd attribute Sari as an external collaborator for starting the scripts but wouldn't worry about going back through the full chain of internal team collaborators.  A quick diff should show similarity % if it helps!

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,128 @@
...
+def experiment_steps(minv, maxv, stepv):
+    rval = [minv]
+    while rval[-1] + stepv <= maxv:
+        rval.append(rval[-1] + stepv)
+    return rval
+
+def report_signals():
+    return ["CPU_CYCLES_THREAD@package", "CPU_CYCLES_REFERENCE@package",
+            "TIME@package", "CPU_ENERGY@package", "GPU_CORE_FREQUENCY_STATUS@board",
+            "GPU_CORE_FREQUENCY_MAX_CONTROL@board"]
```
asmaalrawi: This doesn't work as expected--reading GPU_CORE_FREQUENCY_MAX_CONTROL is static, even if it has been written to.

integration/experiment/gpu_frequency_sweep/README.md
```diff
@@ -6,11 +6,27 @@ runs that use a range of fixed frequency settings for GPUs.
...
+  In addition to command line arguments common to all run scripts,
+  GPU frequency sweep runs also accept the following options:
+
+  - `--max-gpu-frequency`: the maximum GPU frequency setting for the sweep.
+                       By default, it uses the maximum frequency for the
+                       GPUs on the node. Within each trial, the experiment 
+                       script will start at the highest frequency and run at 
+                       each limit until it reaches the minimum.
+
+  - `--min-frequency`: the minimum GPU frequency setting for the sweep.
```
dannosliwcd:
```suggestion
  - `--min-gpu-frequency`: the minimum GPU frequency setting for the sweep.
```

integration/experiment/gpu_frequency_sweep/README.md
```diff
@@ -6,11 +6,27 @@ runs that use a range of fixed frequency settings for GPUs.
...
+
+  - `--step-frequency`: the step size in hertz between settings for the sweep.
```
dannosliwcd:
```suggestion
  - `--step-gpu-frequency`: the step size in hertz between settings for the sweep.
```

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
dannosliwcd: Does that also happen for `GPU_CORE_FREQUENCY_MAX_CONTROL@gpu_chip`?

src/FrequencyMapAgent.cpp
avilcheslopez: I see the comment has been removed. Was this in reference to <iostream>? If so, is it still needed?

avilcheslopez: Ah, never mind, I see you now depend on std::cerr (for warning messages).

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +330,17 @@ namespace geopm
             }
             m_last_uncore_freq = m_uncore_freq;
         }
+
+        // adjust fixed gpu freq
+        if(m_num_gpu_freq_ctl_domain > 0){
```
avilcheslopez: Nit: space after ``if``, space before opening brace

src/FrequencyMapAgent.cpp
```diff
@@ -255,8 +284,14 @@ namespace geopm
             }
             m_platform_io.adjust(m_uncore_min_ctl_idx, m_uncore_init_min);
             m_platform_io.adjust(m_uncore_max_ctl_idx, m_uncore_init_max);
+
+            for (size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
+                double val = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_CONTROL", 
+                        m_gpu_freq_ctl_domain_type, ctl_idx);
```
avilcheslopez: Nit: align with parameters in previous line

src/FrequencyMapAgent.cpp
avilcheslopez: That makes sense. To achieve that behavior, you'll have to use a different approach (to avoid the exception thrown by ``PlatformIO::control_domain_type()``). For example, you could catch the exception or you could explicitly check if the signal in question exists ("GPU_CORE_FREQUENCY_MAX_CONTROL").

test/FrequencyMapAgentTest.cpp
```diff
@@ -155,38 +203,63 @@ void FrequencyMapAgentTest::TearDown()
...
+    EXPECT_CALL(*m_platform_io, adjust(UNCORE_MAX_CTL_IDX, m_freq_uncore_max));
+    if (num_gpu > 0) {
+        EXPECT_CALL(*m_platform_io, read_signal("GPU_CORE_FREQUENCY_MAX_CONTROL", _, _))
+            .Times(num_gpu)
+            .WillRepeatedly(Return(m_freq_gpu_max));
+    }
+    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQ_CONTROL_IDX, m_freq_gpu_max))
+        .Times(num_gpu);
+
+    return;
```
avilcheslopez, line 236: Nit: not needed

test/FrequencyMapAgentTest.cpp
```diff
@@ -247,10 +319,96 @@ TEST_F(FrequencyMapAgentTest, adjust_platform_uncore)
...
+
+    //Don't write if GPU frequency changes to NAN
+    policy[GPU_DEFAULT] = NAN;
+    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQ_CONTROL_IDX, policy[GPU_DEFAULT]))
+        .Times(num_gpu);
+
+    m_agent->adjust_platform(policy);
+    EXPECT_FALSE(m_agent->do_write_batch());
+
+
```
avilcheslopez: Nit: remove extra new lines

test/FrequencyMapAgentTest.cpp
```diff
@@ -459,8 +678,54 @@ TEST_F(FrequencyMapAgentTest, validate_policy)
...
+    m_agent->validate_policy(policy);
+    ASSERT_EQ(m_num_policy, policy.size());
+    EXPECT_EQ(m_freq_min, policy[DEFAULT]);
+    EXPECT_EQ(1.2e9, policy[UNCORE]);
+    EXPECT_EQ(100, policy[HASH_0]);
+    EXPECT_EQ(m_freq_max, policy[FREQ_0]);
+    EXPECT_TRUE(std::isnan(policy[HASH_1]));
+    EXPECT_TRUE(std::isnan(policy[FREQ_1]));
+
```
avilcheslopez, line 673: Nit: delete extra new line

test/FrequencyMapAgentTest.cpp
```diff
@@ -135,17 +148,52 @@ void FrequencyMapAgentTest::SetUp()
...
+    EXPECT_CALL(*m_platform_io, push_signal("REGION_HASH", _, _)).Times(M_NUM_CPU);
+    EXPECT_CALL(*m_platform_io, push_control("CPU_FREQUENCY_MAX_CONTROL", _, _)).Times(M_NUM_CPU);
     EXPECT_CALL(*m_platform_io, push_control("CPU_UNCORE_FREQUENCY_MIN_CONTROL", _, _));
     EXPECT_CALL(*m_platform_io, push_control("CPU_UNCORE_FREQUENCY_MAX_CONTROL", _, _));
     EXPECT_CALL(*m_platform_io, read_signal("CPU_FREQUENCY_MIN_AVAIL", _, _));
     EXPECT_CALL(*m_platform_io, read_signal("CPU_FREQUENCY_MAX_AVAIL", _, _));
     EXPECT_CALL(*m_platform_io, read_signal("CPU_UNCORE_FREQUENCY_MIN_CONTROL", _, _));
     EXPECT_CALL(*m_platform_io, read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", _, _));
 
+
```
avilcheslopez, line 164: Nit: remove extra new lines

test/FrequencyMapAgentTest.cpp
```diff
@@ -135,17 +148,52 @@ void FrequencyMapAgentTest::SetUp()
...
+}
+
+//Sets up GPU expecations and performs init. Every test must call this.
+void FrequencyMapAgentTest::setup_gpu(int num_gpu)
+{
+    if (num_gpu > 0){
```
avilcheslopez: Nit: space before brace

test/FrequencyMapAgentTest.cpp
```diff
@@ -135,17 +148,52 @@ void FrequencyMapAgentTest::SetUp()
...
+        ON_CALL(*m_platform_io, control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL"))
+            .WillByDefault(Return(GEOPM_DOMAIN_GPU));
+    }
+    else{
```
avilcheslopez: Nit: space before brace

test/FrequencyMapAgentTest.cpp
```diff
@@ -135,17 +148,52 @@ void FrequencyMapAgentTest::SetUp()
...
+        ON_CALL(*m_platform_io, control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL"))
+            .WillByDefault(Return(GEOPM_DOMAIN_INVALID));
```
avilcheslopez: Careful: this is not actually ``PlatformIO::control_domain_type()``'s behavior when a control name is not found. This will make the test validate a behavior that's not actually there. See earlier comment in ``FrequencyMapAgent.cpp`` regarding how to handle this case.

test/FrequencyMapAgentTest.cpp
```diff
@@ -45,20 +45,25 @@ class FrequencyMapAgentTest : public :: testing :: Test
...
+            FREQ_1 = 6,
         };
 
         void SetUp();
+        void setup_gpu(int num_gpu);
         void TearDown();
-        static const int M_NUM_CPU = 1;
+        void set_expectations_adjust_platform_init(int num_gpu);
+        static const int M_NUM_CPU = 3;   
+        int m_num_gpu = 2;
```
avilcheslopez: I'm seeing this variable is never changed, i.e. the GPU-related expectations are set up with 2 in all tests. I thought the idea was to test everything with ``m_num_gpu = 0`` as well.

integration/experiment/gpu_frequency_sweep/README.md
```diff
@@ -6,11 +6,27 @@ runs that use a range of fixed frequency settings for GPUs.
...
 ## Base Experiment Module
 
-No base experiment module is currently provided.
+#### `gpu_frequency_sweep.py`:
+
+  Contains the helper function for launching a GPU frequency sweep and
+  common command line arguments.
+
+  In addition to command line arguments common to all run scripts,
+  GPU frequency sweep runs also accept the following options:
```
avilcheslopez: Minor suggestion: "the GPU frequency sweep script also accepts the following options:"

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,153 @@
...
+                        help='bottom gpu frequency setting for the sweep')
+    parser.add_argument('--step-gpu-frequency', dest='step_gpu_frequency',
+                        action='store', type=float, default=None,
+                        help='increment in hertz between gpu frequency settings for the sweep')
+
+    parser.set_defaults(agent_list='frequency_map')
+
+
+def setup_gpu_frequency_bounds(mach, min_gpu_freq, max_gpu_freq,
+                                  step_gpu_freq):
```
avilcheslopez: Nit: parameter alignment

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,153 @@
...
+
+    core_freq_range = frequency_sweep.setup_frequency_bounds(mach,
+                                                             args.min_frequency,
+                                                             args.max_frequency,
+                                                             args.step_frequency,
+                                                             args.run_max_turbo)
+    uncore_freq_range = uncore_frequency_sweep.setup_uncore_frequency_bounds(mach,
+                                                      args.min_uncore_frequency,
+                                                      args.max_uncore_frequency,
+                                                      args.step_uncore_frequency)
```
avilcheslopez: Nit: parameter alignment

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,153 @@
...
+
+    gpu_freq_range = setup_gpu_frequency_bounds(mach,
+                                            args.min_gpu_frequency,
+                                            args.max_gpu_frequency,
+                                            args.step_gpu_frequency)
```
avilcheslopez: Nit: parameter alignment

service/docs/source/geopm_agent_frequency_map.7.rst
```diff
@@ -59,6 +60,15 @@ Policy Parameters
       out of range value or specifying NAN will result in
       a runtime error.
 
+  ``FREQ_GPU_DEFAULT``\ :
+      The operating frequency in units of *Hz* set by the
+      agent on GPUs executing the application.
+      The GPUs not associated with the application
+      will run at the default frequency.  This policy
```
avilcheslopez: What do you mean by "default frequency" here? Do you mean this same policy parameter? If so, the documentation above mentions the default GPU frequency applies only to GPUs running the application. Either way, some clarification is needed above and/or here.

service/docs/source/geopm_agent_frequency_map.7.rst
```diff
@@ -126,32 +136,32 @@ Control Loop Rate
 -----------------
 
 The agent gates the control loop to sample the region hash and control
-CPU frequency at 2 millisecond intervals.  If the uncore frequency is
-specified in the policy, then it is set at agent start up time.
+CPU frequency at 2 millisecond intervals.  If the uncore and/or GPU frequency 
+is specified in the policy, then it is set at agent start up time.
```
avilcheslopez: Nit: start-up

service/docs/source/geopm_agent_frequency_map.7.rst
```diff
@@ -126,32 +136,32 @@ Control Loop Rate
...
---------
+Examples 
+-------- 
 
-Besides the default CPU frequency and uncore frequency parameters, the
-policy consists of the map of per-region frequencies given as pairs of
-region hash and CPU frequency values.  The recommended way to generate
-a policy file for this agent is to use the :doc:`geopmagent(1) <geopmagent.1>` command
-line tool.
+Besides the default CPU frequency, default GPU frequency,  and uncore
```
avilcheslopez: Extra space.

```suggestion
Besides the default CPU frequency, default GPU frequency, and uncore 
```

src/FrequencyMapAgent.cpp
```diff
@@ -441,5 +492,21 @@ namespace geopm
...
+        }
+        else {
+            m_num_gpu_freq_ctl_domain = m_platform_topo.num_domain(m_gpu_freq_ctl_domain_type); 
+            m_gpu_min = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", m_gpu_freq_ctl_domain_type, 0);
+            m_gpu_max = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", m_gpu_freq_ctl_domain_type, 0);
+
+            for (size_t ctl_idx = 0; ctl_idx < (size_t) m_num_gpu_freq_ctl_domain; ++ctl_idx) {
+                m_gpu_control_idx.push_back(m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL",
+                                                                       m_gpu_freq_ctl_domain_type,
+                                                                       ctl_idx));
```
lhlawson:
```suggestion
                m_gpu_min_control_idx.push_back(m_platform_io.push_control("GPU_CORE_FREQUENCY_MIN_CONTROL",
                                                                       m_gpu_freq_ctl_domain_type,
                                                                       ctl_idx));
                m_gpu_max_control_idx.push_back(m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL",
                                                                       m_gpu_freq_ctl_domain_type,
                                                                       ctl_idx));
```

Setting only the maximum frequency allows the driver & fw to lower frequency below it in certain conditions, which may lead to noisier data overall (ex: V100s will swing between 135MHz at idle to 1290Mhz - Max while active).  

In addition for any system that imposes a strict min <= max requirement only setting the maximum frequency will cause an agent to fail.  In order to address this both the [NVML](https://github.com/geopm/geopm/blob/dev/service/src/NVMLIOGroup.cpp#L540) & [LevelZero](https://github.com/geopm/geopm/blob/dev/service/src/LevelZeroIOGroup.cpp#L914) IOGroups support control write re-ordering on failure if both the maximum & minimum setting are being written in batch.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,153 @@
...
+    return targets
+
+def launch(app_conf, args, experiment_cli_args):
+    '''
+    Run the application over a range of core, uncore, and gpu frequencies.
+    '''
+
+    print(f"Core freq: {args.min_frequency} - {args.max_frequency} step {args.step_frequency},\n"
+          f"Uncore: {args.min_uncore_frequency} - {args.max_uncore_frequency}, step {args.step_uncore_frequency}.\n"
+          f"GPU: {args.min_gpu_frequency} - {args.max_gpu_frequency}, step {args.step_gpu_frequency}")
```
lhlawson: Is there a reason to add additional messaging here vs the uncore freq sweep & core freq sweep?  Should this be added to all experiments?

asmaalrawi: Left over from debug, I'm removing it.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
asmaalrawi: The aggregate issue is gone in that case, but it is still giving static info, so it's not a useful signal on NVIDIA. Not sure about other architectures.

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
lhlawson: That works for me, but it might be worthwhile to also file a quick feature request for when we update/refactor the experiment infrastructure - A high level human readable experiment config summary should be relatively easy to generate for any set of experiments based upon the data we already have

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
lhlawson: Try with latest commit using @board.

integration/experiment/gpu_frequency_sweep/README.md
```diff
@@ -6,11 +6,27 @@ runs that use a range of fixed frequency settings for GPUs.
...
+#### `gpu_frequency_sweep.py`:
+
+  Contains the helper function for launching a GPU frequency sweep and
+  common command line arguments.
+
+  In addition to command line arguments common to all run scripts,
+  the GPU frequency sweep script also accepts the following options:
+
+  - `--max-gpu-frequency`: the maximum GPU frequency setting for the sweep.
+                       By default, it uses the maximum frequency for the
```
lhlawson:
```suggestion
                           By default, it uses the maximum frequency for the
```
wsp

integration/experiment/gpu_frequency_sweep/README.md
```diff
@@ -6,11 +6,27 @@ runs that use a range of fixed frequency settings for GPUs.
...
+                       GPUs on the node. Within each trial, the experiment 
+                       script will start at the highest frequency and run at 
+                       each limit until it reaches the minimum.
+
+  - `--min-gpu-frequency`: the minimum GPU frequency setting for the sweep.
```
lhlawson, line 23: Add comment on default behavior

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,150 @@
...
+
+def setup_run_args(parser):
+    common_args.setup_run_args(parser)
+    common_args.add_min_frequency(parser)
+    common_args.add_max_frequency(parser)
+    common_args.add_step_frequency(parser)
+    common_args.add_run_max_turbo(parser)
+    common_args.add_agent_list(parser)
+
+    parser.add_argument('--max-uncore-frequency', dest='max_uncore_frequency',
```
lhlawson: Add gpu & uncore freq sweeps to common_args

integration/experiment/gpu_frequency_sweep/README.md
```diff
@@ -6,11 +6,27 @@ runs that use a range of fixed frequency settings for GPUs.
...
+
+  - `--step-gpu-frequency`: the step size in hertz between settings for the sweep.
 
 ## Analysis Scripts to Produce Summary Tables and Visualizations
 
-#### `gen_gpu_activity_policy_recommendation.py`:
+#### gen_gpu_activity_constconfig_recommendation.py
```
lhlawson: `

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,150 @@
...
+                        action='store', type=float, default=None,
+                        help='bottom gpu frequency setting for the sweep')
+    parser.add_argument('--step-gpu-frequency', dest='step_gpu_frequency',
+                        action='store', type=float, default=None,
+                        help='increment in hertz between gpu frequency settings for the sweep')
+
+    parser.set_defaults(agent_list='frequency_map')
+
+
+def setup_gpu_frequency_bounds(mach, min_gpu_freq, max_gpu_freq,
```
lhlawson:
```suggestion
def setup_gpu_frequency_bounds(mach, min_gpu_freq=None, max_gpu_freq=None,
```
Use =None where appropriate to represent optional arg

src/FrequencyMapAgent.hpp
```diff
@@ -62,13 +62,17 @@ namespace geopm
...
                 M_POLICY_FREQ_UNCORE,
+                M_POLICY_FREQ_GPU_DEFAULT,
                 M_POLICY_FIRST_HASH,
                 M_POLICY_FIRST_FREQUENCY,
                 // The remainder of policy values can be additional pairs of
                 // (hash, frequency)
-                M_NUM_POLICY = 64,
+                // TODO: This number was 64. I changed it so the names vector
+                //       would be appropriately sized. Should I fix the logic
+                //       in the code instead?
```
bgeltz:
```suggestion
```

src/FrequencyMapAgent.hpp
```diff
@@ -79,24 +83,32 @@ namespace geopm
...
             int m_freq_ctl_domain_type;
+            int m_gpu_freq_ctl_domain_type;
             int m_num_freq_ctl_domain;
+            int m_num_gpu_freq_ctl_domain;
             double m_core_freq_min;
             double m_core_freq_max;
             double m_uncore_init_min;
             double m_uncore_init_max;
+            double m_gpu_min;
+            double m_gpu_max;
```
bgeltz:
```suggestion
            double m_gpu_freq_min;
            double m_gpu_freq_max;
```

src/FrequencyMapAgent.cpp
```diff
@@ -26,6 +26,7 @@
 #include "geopm_debug.hpp"
 #include "config.h"
 
+
```
bgeltz:
```suggestion
```

src/FrequencyMapAgent.cpp
```diff
@@ -103,19 +110,34 @@ namespace geopm
...
+                            std::to_string(policy[M_POLICY_FREQ_CPU_DEFAULT]) + ".",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
 
+        if (!std::isnan(policy[M_POLICY_FREQ_GPU_DEFAULT])) {
+            if (m_num_gpu_freq_ctl_domain == 0) {
+                std::cerr << "Warning: <geopm> FrequencyMapAgent::validate_policy(): "
+                    << "User set default GPU Frequency "
+                    << policy[M_POLICY_FREQ_GPU_DEFAULT]
+                    << " on a system with no GPUs.\n";
```
bgeltz: Promote this to a full error/exception unless this is a valid use case.

src/FrequencyMapAgent.cpp
```diff
@@ -255,6 +284,13 @@ namespace geopm
             }
             m_platform_io.adjust(m_uncore_min_ctl_idx, m_uncore_init_min);
             m_platform_io.adjust(m_uncore_max_ctl_idx, m_uncore_init_max);
+
+            for (size_t ctl_idx = 0; ctl_idx < (size_t)m_num_gpu_freq_ctl_domain; ++ctl_idx) {
```
bgeltz: Change this to an if() and push the GPU signals initially at the board scope in the init() method.

src/FrequencyMapAgent.cpp
```diff
@@ -441,5 +507,25 @@ namespace geopm
...
+
+            for (size_t ctl_idx = 0; ctl_idx < (size_t) m_num_gpu_freq_ctl_domain; ++ctl_idx) {
+                m_gpu_min_control_idx.push_back(m_platform_io.push_control("GPU_CORE_FREQUENCY_MIN_CONTROL",
+                                                                           m_gpu_freq_ctl_domain_type,
+                                                                           ctl_idx));
+                m_gpu_max_control_idx.push_back(m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL",
+                                                                           m_gpu_freq_ctl_domain_type,
+                                                                           ctl_idx));
+            }
+        } else {
```
bgeltz:
```suggestion
        } 
        else {
```

test/FrequencyMapAgentTest.cpp
```diff
@@ -135,17 +149,76 @@ void FrequencyMapAgentTest::SetUp()
...
+    if (num_gpu > 0) {
+        ON_CALL(*m_platform_io, control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL"))
+            .WillByDefault(Return(GEOPM_DOMAIN_GPU));
+        control_names.insert({
+            "GPU_CORE_FREQUENCY_MAX_CONTROL",
+            "GPU_CORE_FREQUENCY_MIN_CONTROL"});
+    }
+    else {
+        ON_CALL(*m_platform_io, control_domain_type("GPU_CORE_FREQUENCY_MAX_CONTROL"))
+            .WillByDefault(Return(GEOPM_DOMAIN_INVALID));
```
bgeltz:
```
.WillOnce(Throw(std::runtime_error("injected error")));
```
We have code like this in other tests that may be useful to use here.  It should also work for WillRepeatedly, etc.

test/FrequencyMapAgentTest.cpp
```diff
@@ -155,38 +228,67 @@ void FrequencyMapAgentTest::TearDown()
...
+            .Times(0);
         for (int sample = 0; sample < num_samples; ++sample) {
             EXPECT_CALL(*m_platform_io, sample(REGION_HASH_IDX))
-                .WillOnce(Return(m_region_hash[x]));
+                .Times(M_NUM_CPU)
+                .WillRepeatedly(Return(m_region_hash[x]));
             std::vector<double> tmp;
             m_agent->sample_platform(tmp);
             m_agent->adjust_platform(m_default_policy);
+        // initial all-NAN policy is accepted
```
bgeltz: Check spacing.  might be wrong.

test/FrequencyMapAgentTest.cpp
```diff
@@ -247,10 +348,108 @@ TEST_F(FrequencyMapAgentTest, adjust_platform_uncore)
...
+    EXPECT_TRUE(m_agent->do_write_batch());
+
+    //Changing GPU frequency only should result in no changes
+    policy[GPU_DEFAULT] = 1.4e9;
+    m_agent->adjust_platform(policy);
+    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQ_MIN_CONTROL_IDX, policy[GPU_DEFAULT]))
+        .Times(num_gpu);
+    EXPECT_CALL(*m_platform_io, adjust(GPU_FREQ_MAX_CONTROL_IDX, policy[GPU_DEFAULT]))
+        .Times(num_gpu);
+    EXPECT_FALSE(m_agent->do_write_batch());
```
bgeltz: This is going to have to change based on the warning we're promoting to an error.

test/FrequencyMapAgentTest.cpp
```diff
@@ -459,8 +726,54 @@ TEST_F(FrequencyMapAgentTest, validate_policy)
                                "policy maps a NaN region with frequency");
 }
 
+TEST_F(FrequencyMapAgentTest, validate_policy_nogpu)
+{
+    using ::testing::Pointwise;
+    using ::testing::NanSensitiveDoubleEq;
```
bgeltz:
```suggestion
```
These may be able to be removed.

bgeltz (dismissed):

integration/experiment/gpu_frequency_sweep/run_gpu_frequency_sweep_parres_dgemm_oneapi.py
```diff
@@ -0,0 +1,26 @@
...
+'''
+
+import argparse
+
+from experiment.gpu_frequency_sweep import gpu_frequency_sweep
+from experiment import machine
+from apps.parres import parres
+
+if __name__ == '__main__':
+
```
avilcheslopez: Nit: unneeded new line

src/FrequencyMapAgent.cpp
```diff
@@ -103,19 +110,34 @@ namespace geopm
...
             throw Exception("FrequencyMapAgent::" + std::string(__func__) +
-                            "(): default frequency out of range: " +
-                            std::to_string(policy[M_POLICY_FREQ_DEFAULT]) + ".",
+                            "(): default CPU frequency out of range: " +
+                            std::to_string(policy[M_POLICY_FREQ_CPU_DEFAULT]) + ".",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
 
+        if (!std::isnan(policy[M_POLICY_FREQ_GPU_DEFAULT])) {
+            if (! m_do_gpu_ctl) {
```
avilcheslopez:
```suggestion
            if (!m_do_gpu_ctl) {
```

src/FrequencyMapAgent.cpp
```diff
@@ -295,6 +329,16 @@ namespace geopm
             }
             m_last_uncore_freq = m_uncore_freq;
         }
+
+        // adjust fixed gpu freq
+        if (m_do_gpu_ctl) {
+            if (m_last_gpu_freq != m_default_gpu_freq && !std::isnan(m_default_gpu_freq)) {
```
avilcheslopez:
```suggestion
            if (!std::isnan(m_default_gpu_freq) && m_last_gpu_freq != m_default_gpu_freq) {
```

src/FrequencyMapAgent.cpp
```diff
@@ -410,13 +454,31 @@ namespace geopm
...
+        //Ensuring that MIN_CONTROL <= MAX CONTROL at all times to avoid issues
+        if (!std::isnan(policy[M_POLICY_FREQ_GPU_DEFAULT]) && m_do_gpu_ctl) {
+            double gpu_min_control = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MIN_CONTROL", m_gpu_freq_ctl_domain_type, 0);
+            if (policy[M_POLICY_FREQ_GPU_DEFAULT] >= gpu_min_control) {
+                m_platform_io.write_control("GPU_CORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0,
+                                            policy[M_POLICY_FREQ_GPU_DEFAULT]);
+                m_platform_io.write_control("GPU_CORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0,
+                                            policy[M_POLICY_FREQ_GPU_DEFAULT]);
+            }
+            else  {
```
avilcheslopez:
```suggestion
            else {
```

src/FrequencyMapAgent.cpp
```diff
@@ -410,13 +454,31 @@ namespace geopm
...
+                m_platform_io.write_control("GPU_CORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0,
+                                            policy[M_POLICY_FREQ_GPU_DEFAULT]);
+                m_platform_io.write_control("GPU_CORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0,
+                                            policy[M_POLICY_FREQ_GPU_DEFAULT]);
+            }
+        }
```
avilcheslopez: Something is off: I'm seeing that regardless of the branch that's taken, the side effect is going to be the same: to set GPU min and max to the policy default for GPU.

asmaalrawi: The ordering is important on some architectures that enforce that min < max. If the new setting is less than the system's current min, we must set the min first, otherwise the max can get overwritten by a non-geopm mechanism. If the new setting is greater than the current min, it is safe to set the max first.

avilcheslopez (dismissed):

avilcheslopez (dismissed):

test/FrequencyMapAgentTest.cpp
```diff
@@ -245,7 +245,7 @@ TEST_F(FrequencyMapAgentTest, adjust_platform_map)
         set_expectations_adjust_platform_init(m_do_gpu);
         m_agent->adjust_platform(empty_policy);
         // initial all-NAN policy is accepted
```
dannosliwcd, line 247: I would guess the comment should be inverted if we're inverting the assertion, but I don't really know what this comment means by accepting `empty_policy`. My guess is that we are just testing that do_write_batch doesn't throw since the later parts of the test change some stuff then assert that all-nan throws. Can you update (or remove) the comment?

dannosliwcd: Alternative interpretation:

Previously this test shouldn't have cared about the return value of `write_batch`. The test was passing even though the code was wrong because the assertion was wrong. The only thing we cared about with respect to the test's design was that it didn't throw an exception. But now, we are intentionally testing that it returns true because we want the agent to ensure that the default frequency settings are applied during the first adjust cycle. So maybe the comment should specify *that* is why we're asserting true here.

Also, the updated context note could be a test message instead of a comment so if it ever does fail it says why the failure matters. E.g., `EXPECT_TRUE(...) << "The first adjust_platform() call should set default frequency controls"`

integration/experiment/gpu_frequency_sweep/gpu_frequency_sweep.py
```diff
@@ -0,0 +1,144 @@
...
+            for gpu_freq in gpu_freq_range:
+                name = f'{freq:.1e}c_{uncore_freq:.1e}u_{gpu_freq:.1e}g'
+                options = {'FREQ_CPU_DEFAULT': freq,
+                           'FREQ_UNCORE': uncore_freq,
+                           'FREQ_GPU_DEFAULT': gpu_freq}
+                file_name = os.path.join(output_dir, f'{agent}_agent_{name}.config')
+                agent_conf = geopmpy.agent.AgentConf(file_name, agent, options)
+                targets.append(launch_util.LaunchConfig(app_conf=app_conf,
+                                                    agent_conf=agent_conf,
+                                                    name=name))
```
bgeltz:
```suggestion
                targets.append(launch_util.LaunchConfig(app_conf=app_conf,
                                                        agent_conf=agent_conf,
                                                        name=name))
```

integration/experiment/gpu_frequency_sweep/run_gpu_frequency_sweep_parres_dgemm_oneapi.py
```diff
@@ -0,0 +1,25 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    gpu_frequency_sweep.setup_run_args(parser)
+    parres.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = parres.create_dgemm_appconf_oneapi(mach, args)
+    gpu_frequency_sweep.launch(app_conf=app_conf, args=args,
+                        experiment_cli_args=extra_args)
```
bgeltz:
```suggestion
    gpu_frequency_sweep.launch(app_conf=app_conf, args=args,
                               experiment_cli_args=extra_args)
```

service/docs/source/geopm_agent_frequency_map.7.rst
```diff
@@ -126,32 +135,32 @@ Control Loop Rate
...
 The agent gates the control loop to sample the region hash and control
-CPU frequency at 2 millisecond intervals.  If the uncore frequency is
-specified in the policy, then it is set at agent start up time.
+CPU frequency at 2 millisecond intervals.  If the uncore and/or GPU frequency 
+is specified in the policy, then it is set at agent start-up time.
 
-Examples
---------
+Examples 
+--------
```
bgeltz:
```suggestion
Examples
--------
```

src/FrequencyMapAgent.cpp
```diff
@@ -320,7 +365,7 @@ namespace geopm
     std::vector<std::string> FrequencyMapAgent::policy_names(void)
     {
 
-        std::vector<std::string> names{"FREQ_DEFAULT", "FREQ_UNCORE"};
+        std::vector<std::string> names{"FREQ_CPU_DEFAULT", "FREQ_UNCORE", "FREQ_GPU_DEFAULT"};
```
bgeltz: Changing these policy names will break a number of integration tests.  I'll post a list of what is busted on the summary of this review.

src/FrequencyMapAgent.cpp
```diff
@@ -441,5 +504,19 @@ namespace geopm
...
+            m_gpu_init_freq_min = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+            m_gpu_init_freq_max = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+
+            m_gpu_min_control_idx = m_platform_io.push_control("GPU_CORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+            m_gpu_max_control_idx = m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+        }
+        else {
+            m_do_gpu_ctl = false;
+            m_gpu_freq_ctl_domain_type = GEOPM_DOMAIN_INVALID;
+        }
```
bgeltz:
```suggestion
```
Both of these conditions are set in the initializer list of the constructor.  This is redundant.

src/FrequencyMapAgent.cpp
```diff
@@ -441,5 +504,19 @@ namespace geopm
...
         m_uncore_init_max = m_platform_io.read_signal("CPU_UNCORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+
+        const std::set<std::string> &CONTROLS = m_platform_io.control_names();
+        if (CONTROLS.find("GPU_CORE_FREQUENCY_MAX_CONTROL") != CONTROLS.end()) {
+            m_do_gpu_ctl = true;
+            m_gpu_init_freq_min = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+            m_gpu_init_freq_max = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+
+            m_gpu_min_control_idx = m_platform_io.push_control("GPU_CORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0);
+            m_gpu_max_control_idx = m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL", GEOPM_DOMAIN_BOARD, 0);
```
bgeltz, line 513: Was ```m_gpu_freq_ctl_domain_type``` supposed to be set here?  I don't see any setting of that variable besides in the initializer list.  I see it's only used in enforce_policy(), but the rest of this code should be checked to ensure we're passing the ctl domain properly.

How is it that *none* of the unit tests associated with enforce_policy() are failing?  I don't see how they should be passing if you're always passing GEOPM_DOMAIN_INVALID.

bgeltz (changes requested):
The changing policy names will break the following integration tests:
- test_frequency_map.py
- test_environment.py
- test_geopmagent.py
- test_scaling_region.py
- test_timed_scaling_region.py
- test_frequency_hint_usage.py
- test_enforce_policy.py

All of those tests have explicit references to the old policy names.  At a minimum they need to be updated to comprehend the new names.  There may be other issues lurking.  Let me know if you'd like a hand running the integration tests.  All of the files I listed run and pass in the nightlies today.

src/FrequencyMapAgent.cpp
asmaalrawi: We moved to working at BOARD scope, as we don't control per-GPU, so this is residual and ought to have been removed. The enforce_policy tests did not specify domain on the line in question, so I will add that in and check those tests.

src/FrequencyMapAgent.cpp
asmaalrawi: Updated the names, would like some help running them so I can expand them to test GPU frequency settings.

bgeltz (commented):
These changes LGTM.  Thanks for addressing everything.

I think this is ready to go now, put I'm not going to approve it just yet.
I'm going to cherrypick this PR into the nightlies over the weekend to ensure there aren't any issues that were missed with those policy changes.  If the runs come back clean, I'll approve/merge on Monday.

bgeltz:
I've managed to get all the integration tests to pass clean now.  The code changes are good.

What remains is to re-write history on this branch to squash the commits down.  I see at least one merge commit:
```
*   62a14ec66 - Merge branch 'ahalrawi-gpufreqmap' of github.com:asmaalrawi/geopm into ahalrawi-gpufreqmap (4 weeks ago) <Asma H. Al-Rawi>
```

Please re-write history so it is linear (i.e. no merge commits).  I'm fine if you want to squash this down a bit.  139 commits is a lot of commits.

bgeltz (approved):

PR #2711, opened 2022-11-17
In ConstConfigIOGroup, check values against domain size

### LINKED ISSUE #2710: In ConstConfigIOGroup, ensure number of signal values concurs with domain size reported by PlatformTopo
> estimate 1

Currently, ConstConfigIOGroup does not check to ensure the number of signal values concurs with what's actually available for a given domain type. Update the implementation so it performs this check and reports an exception early (when parsing the config file) when such a discrepancy is found.

Done criteria:

- [ ] ~~Integration tests that cover the feature run and pass~~
- [x] Dependencies mocked in unit tests
- [x] Affected unit tests are updated and all tests are passing
- [ ] ~~Documentation changes are not required since this change does
      not modify interfaces or features.~~

avilcheslopez:
- Check the number of values for each signal concurs with domain size reported by PlatformTopo
- Add unit test to validate domain sizes are checked properly

- Fixes #2710

### DIFF
--- a/service/src/ConstConfigIOGroup.cpp
+++ b/service/src/ConstConfigIOGroup.cpp
@@ -39,11 +39,12 @@ namespace geopm
         GEOPM_CONFIG_PATH "/const_config_io.json";
 
     ConstConfigIOGroup::ConstConfigIOGroup()
-        : ConstConfigIOGroup(geopm::get_env(M_CONFIG_PATH_ENV), M_DEFAULT_CONFIG_FILE_PATH)
+        : ConstConfigIOGroup(platform_topo(), geopm::get_env(M_CONFIG_PATH_ENV), M_DEFAULT_CONFIG_FILE_PATH)
     {
     }
 
-    ConstConfigIOGroup::ConstConfigIOGroup(const std::string &user_file_path,
+    ConstConfigIOGroup::ConstConfigIOGroup(const PlatformTopo &topo,
+                                           const std::string &user_file_path,
                                            const std::string &default_file_path)
     {
         std::string config_json;
@@ -68,7 +69,7 @@ namespace geopm
             config_json = geopm::read_file(default_file_path);
         }
 
-        parse_config_json(config_json);
+        parse_config_json(topo, config_json);
     }
 
     std::set<std::string> ConstConfigIOGroup::signal_names(void) const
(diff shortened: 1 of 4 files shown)


service/src/ConstConfigIOGroup.cpp
```diff
@@ -330,18 +331,35 @@ namespace geopm
                     properties["aggregation"].string_value());
 
             auto json_values = properties["values"].array_items();
+            if (json_values.empty()) {
+                throw Exception("ConstConfigIOGroup::parse_config_json(): "
+                                "empty array of values provided for " + name,
+                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+            if (static_cast<int>(json_values.size()) != topo.num_domain(domain_type)) {
+                throw Exception("ConstConfigIOGroup::parse_config_json(): "
```
lhlawson, line 340: How will this interact with the LevelZero IOGroup if the user hasn't set the ZES_SYSMAN_ENABLE env var or when they change the number of GPUs visible using ZE_AFFFINITY_MASK? 

There could be 18 GPUs in the system and if the user sets ZE_AFINITY_MASK=16 GEOPM will only expose 1 GPU, leading to ConstConfig erroring, even though it may have all the information needed (assuming 18 values are populated).

service/test/ConstConfigIOGroupTest.cpp
```diff
@@ -215,14 +251,18 @@ TEST_F(ConstConfigIOGroupTest, input_empty_domain)
...
         "PlatformTopo::domain_name_to_type(): unrecognized domain_name: "
     );
 }
 
 TEST_F(ConstConfigIOGroupTest, input_empty_description)
 {
+    EXPECT_CALL(*m_default_topo, num_domain(_)).Times(0);
+    EXPECT_CALL(*m_default_topo, num_domain(GEOPM_DOMAIN_GPU))
+        .Times(AtMost(1))
+        .RetiresOnSaturation();
```
lhlawson: This appears to be used by a large number of tests, may be something we can refactor, but it isn't necessarily a high priority.

lhlawson (approved):
LGTM

cmcantalupo (approved):

PR #2725, opened 2022-11-22
Use control access list when generating signal access list

### LINKED ISSUE #2707: PlatformIO aggregated writes assume user has read permission for the target control
**Describe the bug**
I tried to use geopmwrite at board scope on a package-scoped control that I only have write permission to use. The write failed due to a missing signal of the same name (the signal exists, but is not reported by GEOPM since I do not have read access to it).

**GEOPM version**
2.0.0

**Expected behavior**
I expect to be allowed to geopmwrite any control that I have been granted permission to write.

**Actual behavior**
If I geopmwrite at a more coarse domain than the control's native domain, then the write operation fails when I do not also have read permission.

**Error messages produced**
```
geopmwrite SST::COREPRIORITY_ENABLE:ENABLE board 0 0
Error: cannot write control: <geopm> Invalid argument: ServiceIOGroup::signal_behavior(): signal_name SST::COREPRIORITY_ENABLE:ENABLE not valid for ServiceIOGroup: at src/ServiceIOGroup.cpp:313
```

Note that although the error message says it comes from `signal_behavior()`, the text of that message is wrong. The line number is from `agg_function()`. That function is called by PlatformIO when writing at a level more coarse than the control's native domain. The output is used to determine whether the write value should be distributed (for sum types) or broadcast (for other types) across its native domain.

**Workaround**
Write the control in its native domain to avoid executing the function that depends on read permissions.

cmcantalupo:
- Users with permission to write a control should always be allowed to read back the value.
- Fixes #2707

### DIFF
--- a/service/README.rst
+++ b/service/README.rst
@@ -127,7 +127,7 @@ Access Management
 -----------------
 
 Access to signals and controls through the GEOPM Service is configured
-by the system administrator.  The administrator controls a default
+by the system administrator.  The administrator maintains a default
 access list that applies to all users of the system.  This list
 may be augmented so that users who are members of particular Unix groups may
 have enhanced access.  The default lists are stored in:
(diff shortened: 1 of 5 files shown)


dannosliwcd (commented):
Yeah, I think this is a viable solution to #2707. With this change, users can be given read-only permission for individual signals and their metadata (same as before), and users can be given read+write permission for controls with read permission on the control metadata.

This change should resolve that specific issue because it means the automatic distribution function on writing a coarser-than-native control will be able to read the control metadata that informs which distribution method to apply.

dannosliwcd (approved):

PR #2726, opened 2022-11-22
Set signal and control access lists independenly

### LINKED ISSUE #2712: geopmaccess unexpectedly prunes access files of unsupported signals or controls
**Describe the bug**
I tried to use ```geopmaccess``` to populate files in ```/etc/geopm-service``` and I expected the generated files to be 1:1 copies of the input if the force flag is used.  Instead, setting signals prunes unsupported controls and/or setting controls prunes unsupported signals.

**GEOPM version**
```2.0.0+dev247gc9ccc4ff```

**Expected behavior**
I have written ```allowed_signals``` and ```allowed_controls``` files that contain entries that are not supported on my test system.  Presently these unsupported signals/controls are the entries provided by the SSTIOGroup.  I expect that if I do the following commands that the files will be written into ```/etc/geopm-service```, and a diff of the original file with the generated file will return no discrepancies:
```
geopmaccess -Fw < allowed_signals
geopmaccess -Fwc < allowed_controls
```

**Actual behavior**
1. Run ```geopmaccess -Fw < allowed_signals``` and inspect ```/etc/geopm-service/0.DEFAULT_ACCESS/allowed_signals``` to verify that any unsupported signals were written to the file.  In my case, I'm simply grepping for "SST".
2. Run ```geopmaccess -Fwc < allowed_controls``` and inspect ```/etc/geopm-service/0.DEFAULT_ACCESS/allowed_controls``` to verify that any unsupported controls were written to the file.
3. Repeat your original grep on the ```allowed_signals``` file.  Observe that the unsupported entries that were present in step 1 have been removed.

This behavior is also seen if you reverse steps 1 and 2.  In this case, the unsupported entries in ```allowed_controls``` will be pruned.  The file that is written first will be pruned when the second file is written.

**Additional context**
The workaround here is to manually copy ```allowed_signals``` or ```allowed_contols``` including any unsupported entries out into ```/etc/geopm-service/0.DEFAULT_ACCESS``` and ensuring that the file and directory permissions are set properly (i.e., only accessible by root).  This way, any unsupported entries are not pruned because ```geopmaccess``` is not invoked.

### PR DESCRIPTION
**Pull request by cmcantalupo**
- Add new DBus interface for signal or control access list write
- Use new control/signal interfaces in geopmaccess
- Fixes #2712

### DIFF
--- a/service/geopmdpy_test/TestAccess.py
+++ b/service/geopmdpy_test/TestAccess.py
@@ -142,14 +142,12 @@ class TestAccess(unittest.TestCase):
         """
         return_value = (self._signals_expect,
                         self._controls_expect)
-        self._geopm_proxy.PlatformSetGroupAccess = mock.Mock()
-        self._geopm_proxy.PlatformGetGroupAccess = mock.Mock(return_value=return_value)
+        self._geopm_proxy.PlatformSetGroupAccessSignals = mock.Mock()
         self._geopm_proxy.PlatformGetAllAccess = mock.Mock(return_value=return_value)
         with mock.patch('sys.stdin.readlines', return_value=self._signals_expect):
             self._access.run(True, False, False, None, False, False, False, False, False)
-            self._geopm_proxy.PlatformGetGroupAccess.assert_called_once_with('')
-            self._geopm_proxy.PlatformSetGroupAccess.assert_called_once_with('',
-                self._signals_expect, self._controls_expect)
+            self._geopm_proxy.PlatformSetGroupAccessSignals.assert_called_once_with('',
+                self._signals_expect)
             self._geopm_proxy.PlatformGetAllAccess.assert_called_once()
 
     def test_write_default_controls(self):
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by avilcheslopez — COMMENTED** [review]
On first pass it looks good to me! I'll spend a little more time getting familiar with the rest of the related code in geopmdpy to give a better review.

**Review by Krebrov001 — DISMISSED** [review]
I don't see anything here that needs to be changed.

File: service/geopmdpy_test/TestAccess.py [hunk 1]
```diff
@@ -142,14 +142,12 @@ def test_write_default_signals(self):
...
-        self._geopm_proxy.PlatformGetGroupAccess = mock.Mock(return_value=return_value)
+        self._geopm_proxy.PlatformSetGroupAccessSignals = mock.Mock()
         self._geopm_proxy.PlatformGetAllAccess = mock.Mock(return_value=return_value)
         with mock.patch('sys.stdin.readlines', return_value=self._signals_expect):
             self._access.run(True, False, False, None, False, False, False, False, False)
-            self._geopm_proxy.PlatformGetGroupAccess.assert_called_once_with('')
-            self._geopm_proxy.PlatformSetGroupAccess.assert_called_once_with('',
-                self._signals_expect, self._controls_expect)
+            self._geopm_proxy.PlatformSetGroupAccessSignals.assert_called_once_with('',
+                self._signals_expect)
```
avilcheslopez, line 150: Overall, I see all the calls to Platform[Get|Set]GroupAccess are being removed in favor of testing the new APIs. However, if we are still going to keep the original/existing APIs for now, I would recommend we still continue to test them.

cmcantalupo: Seems like a good idea :) thanks

avilcheslopez (approved):

Issue #2731, opened 2022-11-23
As a user of ConstConfigIOGroup I would like to be able to specify a default value for a signal (as opposed to having to provide a value per domain index)


avilcheslopez:
**What problem does the story solve? Please describe.**
At the moment, the ConstConfigIOGroup configuration file schema requires specifying an array of values for each signal, with one value per domain index. However, this schema is not very flexible for use cases where the hardware configuration can change more dynamically (i.e. where a domain size can change as a result of a configuration change). In these cases, the configuration file would have to be updated every time the configuration changes.

In order to address this issue, it would be convenient if the schema allowed the flexibility to provide a default value for a signal to be used regardless of the size of the signal's domain (i.e. for all domain indexes).

**Describe the tasks that will be required to complete the story**
- [x] Update the schema
- [x] Update ConstConfigIOGroup to handle the new schema and provide signal values for both cases: a default value or an array of values
- [x] Update any affected unit tests
- [x] Add unit tests to validate the new functionality

PR #2733, opened 2022-11-24
Update ConstConfigIOGroup config file schema

### LINKED ISSUE #2732: Allow specifying a default value for a signal in ConstConfigIOGroup
> estimate 1
Parent [feature/story]: #2731 

Done criteria:
- [x] Schema updated
- [x] ConstConfigIOGroup updated to handle new schema
- [x] Affected unit tests updated and passing
- [x] New unit tests added for new functionality

avilcheslopez:
Update schema to allow specifying a default value for a signal. For each signal, either a default value or an array of values needs to be provided.

- Relates to #2731 
- Fixes #2732

### DIFF
--- a/service/src/ConstConfigIOGroup.cpp
+++ b/service/src/ConstConfigIOGroup.cpp
@@ -335,7 +335,8 @@ namespace geopm
         auto signals = root.object_items();
         for (const auto &signal : signals) {
             check_json_signal(signal);
-            std::string name = M_SIGNAL_PREFIX + signal.first;
+            const std::string &NAME = signal.first;
+            const std::string FULL_NAME = M_SIGNAL_PREFIX + NAME;
 
             auto properties = signal.second.object_items();
             int units = IOGroup::string_to_units(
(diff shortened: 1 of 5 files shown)


service/src/ConstConfigIOGroup.hpp
```diff
@@ -88,6 +88,7 @@ namespace geopm
                 int domain;
                 std::function<double(const std::vector<double> &)> agg_function;
                 std::string description;
+                bool default_value;
```
lhlawson: Nit on naming.  default_value is a boolean indicating if a devalue value exists but might be interpreted as the actual default value.  default_value_provided is used elsewhere and seems somewhat clearer...or maybe default_value_exists or default_value_present?  

Otherwise lgtm.

service/src/ConstConfigIOGroup.cpp
```diff
@@ -203,13 +209,21 @@ namespace geopm
...
         else if (domain_idx < 0 ||
-                 static_cast<std::size_t>(domain_idx) >= it->second->values.size()) {
+                 domain_idx >= m_platform_topo.num_domain(domain_type)) {
             throw Exception("ConstConfigIOGroup::read_signal(): domain_idx " +
                             std::to_string(domain_idx) + " out of range.",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
 
-        return it->second->values[domain_idx];
+        double value = 0.0;
```
lhlawson: Other IOGroups use NAN as a default value as a precaution; if we get to the return without setting the value the NAN indicates something has gone wrong.  Should we follow the same pattern here?

avilcheslopez: Yeah, that makes sense. I'll make that change.

lhlawson (dismissed):

service/src/ConstConfigIOGroup.hpp
```diff
@@ -88,6 +88,7 @@ namespace geopm
                 int domain;
                 std::function<double(const std::vector<double> &)> agg_function;
                 std::string description;
+                bool common_value_provided;
```
cmcantalupo: `common_valud_provided` -> `is_common_value` or `is _common_value_provided`

I personally prefer `is_common_value`, but i'm okay with the longer name.  I try to enforce prefixing bool variable names with short verbs (usually a "to be verb") to aid with reading logic statements in code.  There are a couple of similar stack variables in the parser.

cmcantalupo (commented):
Minor nit on struct name, otherwise looks great.

cmcantalupo (dismissed):

cmcantalupo (approved):

PR #2736, opened 2022-11-29
CPU-CA test updates for CPU-CA constconfig

### LINKED ISSUE #2735: Improve unit testing for CPU Activity Agent
> estimate 1
Parent [feature/story]: #2698

Update CPU-CA agent testing to include explicit tests around core and uncore F efficient settings and calls to the CONST_CONFIG:: signals.  

Done criteria:

- [x] Updates to const_config signal names in tests
- [x] Addition of test covering invalid Fe cases for core & uncore

lhlawson:

- Relates to #2735 issue from github issues
- Fixes #2735 change request from github issues.

Update to CPU-CA to add explicit Fe checking, and update to test to handle Fe checking & add proper CONST_CONFIG::* signal names.

### DIFF
--- a/test/CPUActivityAgentTest.cpp
+++ b/test/CPUActivityAgentTest.cpp
@@ -55,7 +55,7 @@ class CPUActivityAgentTest : public ::testing::Test
 
         void SetUp();
         void setup_default_expect_calls();
-        void setup_default_expect_calls_mem();
+        void setup_default_expect_calls_const_config();
         void TearDown();
         static const int M_NUM_CPU;
         static const int M_NUM_CORE;
@@ -115,6 +115,7 @@ void CPUActivityAgentTest::SetUp()
     ON_CALL(*m_platform_io, agg_function(_))
         .WillByDefault(Return(geopm::Agg::average));
 
+    m_cpu_freq_min = 1000000000.0;
     m_cpu_freq_max = 3700000000.0;
     m_cpu_uncore_freq_min = 1200000000.0;
     m_cpu_uncore_freq_max = 2400000000.0;
@@ -146,7 +147,6 @@ void CPUActivityAgentTest::SetUp()
     ASSERT_LT(m_cpu_uncore_freq_min, m_cpu_uncore_freq_max);
 
     setup_default_expect_calls();
-    setup_default_expect_calls_mem();
 
     m_gov = std::make_shared<MockFrequencyGovernor>();
     ON_CALL(*m_gov, frequency_domain_type())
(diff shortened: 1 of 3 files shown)


test/CPUActivityAgentTest.cpp
```diff
@@ -286,7 +290,7 @@ TEST_F(CPUActivityAgentTest, validate_policy)
 
 TEST_F(CPUActivityAgentTest, control_signal_granularity_check) {
     setup_default_expect_calls();
-    setup_default_expect_calls_mem();
+    setup_default_expect_calls_const_config();
```
avilcheslopez, line 300: Q: why is this method called here if it's already called in ``SetUp()``.

test/CPUActivityAgentTest.cpp
```diff
@@ -538,3 +542,34 @@ TEST_F(CPUActivityAgentTest, no_mem_constconfig)
                                 "ConstConfigIOGroup configuration file does not contain memory bandwidth information");
 
 }
+
+TEST_F(CPUActivityAgentTest, invalid_fe)
+{
+    setup_default_expect_calls();
```
avilcheslopez: Same question/observation as before: this is already called by ``SetUp()``.

test/CPUActivityAgentTest.cpp
```diff
@@ -538,3 +542,34 @@ TEST_F(CPUActivityAgentTest, no_mem_constconfig)
                                 "ConstConfigIOGroup configuration file does not contain memory bandwidth information");
 
 }
+
+TEST_F(CPUActivityAgentTest, invalid_fe)
```
avilcheslopez: Caution: this test contains multiple tests inside, where each is a negative test that's testing a failed agent initialization, and they're all sharing the same agent object. I don't believe we make any guarantees about the state of an agent if ``init()`` fails, so these tests are depending on behavior that's not guaranteed (i.e. that a failed init leaves the agent object in a consistent state ready to attempt subsequent inits). I would recommend breaking this test out into multiple tests. If you want to keep it all in a single test, you could simply instantiate multiple agent objects.

test/CPUActivityAgentTest.cpp
lhlawson: This method covers everything required by m_agent->init().  init is called once as part of setup for the default config used by most tests, but some tests are covering specific agent configs that require calling init() againt to test.

test/CPUActivityAgentTest.cpp
lhlawson: Broken out into 4 tests: invalid_fe_<core/uncore>_<high/low> to indicate which signal (core/uncore) fails & why (higher than range, lower than range).

avilcheslopez (approved):

dannosliwcd (approved):

PR #2739, opened 2022-11-30
Added explicit std::move calls for newer clang compilers

### LINKED ISSUE #2740: Json11 fails to build with newer clang versions
**Describe the bug**
I tried to compile GEOPM with a recent clang-based compiler and it failed to build due to warnings in json11.cpp.

**GEOPM version**
9396f98

**Expected behavior**
GEOPM builds successfully.

**Actual behavior**
GEOPM failed to build due to errors in json11.cpp.

**Error messages produced**
```cpp
	contrib/json11/json11.cpp:154:46: error: unqualified call to 'std::move' [-Werror,-Wunqualified-std-cast-call]
	    explicit Value(T &&value)      : m_value(move(value)) {}
	                                             ^
	                                             std::
```

We either need to ignore that warning or resolve the (few) instances where this occurs in json11.cpp.

**Additional context**
The [clang documentation](https://clang.llvm.org/docs/DiagnosticsReference.html#wunqualified-std-cast-call) doesn't really explain the issue, but more context is available in the change proposal: https://reviews.llvm.org/D119670

lhlawson:

- Relates to #2740 bug report/ from github issues
- Fixes #2740 change request from github issues.

Newer clang versions error on non-explicit std::move calls even when using std::move is provided.  This updates the json11.cpp implementation to explicitly use std::move.

### DIFF
--- a/service/contrib/json11/json11.cpp
+++ b/service/contrib/json11/json11.cpp
@@ -35,7 +35,6 @@ using std::vector;
 using std::map;
 using std::make_shared;
 using std::initializer_list;
-using std::move;
 
 /* Helper for representing null - just a do-nothing struct, plus comparison
  * operators so the helpers in JsonValue work. We can't use nullptr_t because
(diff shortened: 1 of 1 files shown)


dannosliwcd (approved):
LGTM. Covers all of the unqualified `move()` calls. Looks like there are no unqualified `forward()` calls to worry about.

bgeltz (approved):

PR #2628, opened 2022-09-14
Add TensorTwoD class and tests

asmaalrawi:
- Relates to #2617
- Fixes #2678 

Adds TensorTwoD class which builds upon TensorOneD to define 2-dimensional tensor operations and is used by the Feed Forward agent

PR #2759, opened 2022-12-09
Combined LocalNeuralNet code with interface/implementation separation…

asmaalrawi:
- Relates to #2617 
-  Adds LocalNeuralNet class which builds on TensorOneD and TensorTwoD to do neural net json parsing and feed-forward inference, and is used by the Feed Forward agent
- Addresses #2693 
- Replaces #2618 
- Replaces #2628 
- …, tensors moved to inner classes, and JSON11 constructors replaced with STL-based interface.

### DIFF
--- a/src/LocalNeuralNet.cpp
+++ b/src/LocalNeuralNet.cpp
@@ -1,0 +1,271 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "config.h"
+#include "LocalNeuralNet.hpp"
+#include "LocalNeuralNetImp.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <functional>
+#include <numeric>
+
+#include "geopm/Exception.hpp"
+#include "geopm/Helper.hpp"
+
+namespace geopm
+{
+    std::unique_ptr<LocalNeuralNet> LocalNeuralNet::make_unique(std::vector<std::pair<std::vector<std::vector<float> >, std::vector<float> > > input) {
+        return geopm::make_unique<LocalNeuralNetImp>(input);
+    }
+
+    LocalNeuralNetImp::LocalNeuralNetImp(std::vector<std::pair<std::vector<std::vector<float> >, std::vector<float> > > input)
+    {
+        if (input.size() == 0u) {
+            throw geopm::Exception("Empty array is invalid for neural network weights.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        std::size_t nlayers = input.size();
+
+        m_layers.resize(nlayers);
+        for (std::size_t idx = 0; idx < nlayers; idx++) {
+            m_layers[idx] = std::make_pair(LocalNeuralNetImp::TensorTwoD(input[idx].first),
+                                           LocalNeuralNetImp::TensorOneD(input[idx].second));
+            if (m_layers[idx].first.get_rows() != m_layers[idx].second.get_dim()) {
+                throw geopm::Exception("Incompatible dimensions for weights and biases.",
+                                       GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+            if (idx > 0 && m_layers[idx].first.get_cols() != m_layers[idx-1].first.get_rows()) {
+                throw geopm::Exception("Incompatible dimensions for consecutive layers.",
+                                       GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+    }
+
+    std::vector<float>
+    LocalNeuralNetImp::model(std::vector<float> inp_vector)
+    {
+        LocalNeuralNetImp::TensorOneD inp(inp_vector);
+
+        if (inp.get_dim() != m_layers[0].first.get_cols()) {
+            throw geopm::Exception("Input vector dimension is incompatible with network.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        LocalNeuralNetImp::TensorOneD tmp = inp;
+        std::size_t nlayers = m_layers.size();
+        for (std::size_t idx = 0; idx < nlayers; idx++) {
+            // Tensor operations
+            tmp = m_layers[idx].first * tmp + m_layers[idx].second;
+
+            // Apply a sigmoid on all but the last layer
+            if (idx != m_layers.size() - 1) {
+                tmp = tmp.sigmoid();
+            }
+        }
+        return tmp.to_vector();
+    }
+
+    LocalNeuralNetImp::TensorOneD::TensorOneD(std::size_t dim)
+    {
+        set_dim(dim);
+    }
+
+    LocalNeuralNetImp::TensorOneD::TensorOneD(const TensorOneD &other)
+        : m_vec(other.m_vec)
+    {
+    }
+
+    LocalNeuralNetImp::TensorOneD::TensorOneD(TensorOneD &&other)
+        : m_vec(std::move(other.m_vec))
+    {
+    }
+
+    void LocalNeuralNetImp::TensorOneD::set_dim(std::size_t dim)
+    {
+        m_vec.resize(dim);
+    }
+
+    std::size_t LocalNeuralNetImp::TensorOneD::get_dim() const
+    {
+        return m_vec.size();
+    }
+
+    LocalNeuralNetImp::TensorOneD::TensorOneD(std::vector<float> input)
+    {
+        set_dim(input.size());
+        std::size_t vec_size = m_vec.size();
+
+        for (std::size_t idx = 0; idx < vec_size; ++idx) {
+            m_vec[idx] = input[idx];
+        }
+    }
+
+    LocalNeuralNetImp::TensorOneD LocalNeuralNetImp::TensorOneD::operator+(const LocalNeuralNetImp::TensorOneD& other)
+    {
+        if (get_dim() != other.get_dim()) {
+            throw geopm::Exception("Adding vectors of mismatched dimensions.",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        LocalNeuralNetImp::TensorOneD rval(m_vec.size());
+        std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.m_vec.begin(), std::plus<float>());
+
+        return rval;
+    }
+
+    LocalNeuralNetImp::TensorOneD LocalNeuralNetImp::TensorOneD::operator-(const LocalNeuralNetImp::TensorOneD& other)
+    {
+        if (get_dim() != other.get_dim()) {
+            throw geopm::Exception("Subtracting vectors of mismatched dimensions.",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        LocalNeuralNetImp::TensorOneD rval(m_vec.size());
+        std::transform(m_vec.begin(), m_vec.end(), other.m_vec.begin(), rval.m_vec.begin(), std::minus<float>());
+
+        return rval;
+    }
+
+
+    float LocalNeuralNetImp::TensorOneD::operator*(const LocalNeuralNetImp::TensorOneD& other)
+    {
+        if (get_dim() != other.get_dim()) {
+            throw geopm::Exception("Inner product of vectors of mismatched dimensions.",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        return std::inner_product(m_vec.begin(), m_vec.end(), other.m_vec.begin(), 0);
+    }
+
+    LocalNeuralNetImp::TensorOneD& LocalNeuralNetImp::TensorOneD::operator=(const LocalNeuralNetImp::TensorOneD &other)
+    {
+        m_vec = other.m_vec;
+        return *this;
+    }
+
+    LocalNeuralNetImp::TensorOneD& LocalNeuralNetImp::TensorOneD::operator=(LocalNeuralNetImp::TensorOneD &&other)
+    {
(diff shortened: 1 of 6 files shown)


src/LocalNeuralNet.hpp
```diff
@@ -0,0 +1,47 @@
...
+            /// @brief Returns a unique_ptr to a concrete object
+            ///        constructed using the underlying implementation
+            ///        from an array of doubles of weights and biases.
+            /// 
+            /// @param [in] vector of pairs of weight matrices (as
+            ///             double vectors of floats) and bias vectors
+            ///             as vectors of floats.
+            /// 
+            /// @throws geopm::Exception if dimensions are incompatible.
+            static std::unique_ptr<LocalNeuralNet> make_unique(std::vector<std::pair<std::vector<std::vector<float> >, std::vector<float> > > input);
```
avilcheslopez: Suggestion: given the complexity of the parameter's type, for readability, use typedefs to simplify its declaration and at the same time describe what it is. For example:

`typedef std::pair<...> Layer // or NNLayer, NeuralNetLayer... (I'm sure you can think of a nicer name)`

or

```
typedef std::vector<std::vector<float> > Weights
typedef std::vector<float> Biases
typedef std::pair<Weights, Biases> Layer
```

This way the method declaration would naturally be more readable:

`static std::unique_ptr<LocalNeuralNet> make_unique(std::vector<Layer> input)`

src/LocalNeuralNet.hpp
```diff
@@ -0,0 +1,47 @@
...
+{
+    ///  @brief Class to manage data and operations of feed forward neural nets
+    ///         required for neural net inference.
+
+    class LocalNeuralNet
+    {
+        public:
+            /// @brief Returns a unique_ptr to a concrete object
+            ///        constructed using the underlying implementation
+            ///        from an array of doubles of weights and biases.
```
avilcheslopez: Minor discrepancy: this comment talks about doubles, while the actual code uses floats.

src/LocalNeuralNet.hpp
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2015 - 2022, Intel Corporation
```
avilcheslopez: 2015 - 2023

Same applies to the other source files

src/LocalNeuralNet.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+            /// @brief Perform inference using the instance weights and biases.
+            /// 
+            /// @param [in] std::vector<float> of input signals.
+            ///
+            /// @return Returns a std::vector<float> of output values.
+            ///
+            /// @throws geopm::Exception if input dimension is incompatible
+            /// with network.
+            virtual std::vector<float> model(std::vector<float> inp) = 0;
```
avilcheslopez: Suggestion: use a verb to make the methods functionality clearer. For example: create_model, generate_model, perform_inference, do_inference, ... (I'm sure you can think of a nicer name)

src/LocalNeuralNetImp.hpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        public:
+            LocalNeuralNetImp(std::vector<std::pair<std::vector<std::vector<float> >, std::vector<float> > > input);
+            std::vector<float> model(std::vector<float> inp);
+
+        private:
+            /// @brief Class to store and perform operations on 1D Tensors,
+            ///        aka vectors, suitable for use in feed-forward neural
+            ///        networks.
+            class TensorOneD
```
avilcheslopez, line 23: Q: is there a reason why you opted to make the tensor classes internal?

If they are really particular to LocalNeuralNetImp, then it makes sense. Otherwise, my concern is that this causes unnecessary back-and-forth conversion between the STL types and the tensor classes every time you want to interact with a LocalNeuralNetImp object.

src/LocalNeuralNetImp.hpp
```diff
@@ -0,0 +1,176 @@
...
+                    /// @brief Oerload = operator with an in-place deep copy
+                    /// 
+                    /// @param [in] TensorTwoD& Reference to the 2D Tensor to copy
+                    TensorTwoD& operator=(const TensorTwoD&);
+
+                private:
+                    std::vector<TensorOneD> m_mat;
+            };
+
+            std::vector<std::pair<TensorTwoD, LocalNeuralNetImp::TensorOneD> > m_layers;
```
avilcheslopez:
```suggestion
            std::vector<std::pair<TensorTwoD, TensorOneD> > m_layers;
```

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include <cmath>
+#include <functional>
+#include <numeric>
+
+#include "geopm/Exception.hpp"
+#include "geopm/Helper.hpp"
+
+namespace geopm
+{
+    std::unique_ptr<LocalNeuralNet> LocalNeuralNet::make_unique(std::vector<std::pair<std::vector<std::vector<float> >, std::vector<float> > > input) {
```
avilcheslopez: Nit: new line before opening brace

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+            }
+            if (idx > 0 && m_layers[idx].first.get_cols() != m_layers[idx-1].first.get_rows()) {
+                throw geopm::Exception("Incompatible dimensions for consecutive layers.",
+                                       GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+            }
+        }
+    }
+
+    std::vector<float>
+    LocalNeuralNetImp::model(std::vector<float> inp_vector)
```
avilcheslopez: Nit:

```suggestion
    std::vector<float> LocalNeuralNetImp::model(std::vector<float> inp_vector)
```

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+    {
+        if (input.size() == 0u) {
+            throw geopm::Exception("Empty array is invalid for neural network weights.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        std::size_t nlayers = input.size();
+
+        m_layers.resize(nlayers);
+        for (std::size_t idx = 0; idx < nlayers; idx++) {
```
avilcheslopez: Nit:

```suggestion
        for (std::size_t idx = 0; idx < nlayers; ++idx) {
```
Same suggestion applies in other places.

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+    float LocalNeuralNetImp::TensorOneD::operator[] (std::size_t idx) const
+    {
+        return m_vec.at(idx);
+    }
+
+    LocalNeuralNetImp::TensorOneD LocalNeuralNetImp::TensorOneD::sigmoid() const
+    {
+        std::size_t vec_size = m_vec.size();
+        LocalNeuralNetImp::TensorOneD rval(vec_size);
+        for(std::size_t idx = 0; idx < vec_size; idx++) {
```
avilcheslopez: Nit:

```suggestion
        for (std::size_t idx = 0; idx < vec_size; idx++) {
```

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+            rval[idx] = 1/(1 + expf(-(m_vec.at(idx))));
```
avilcheslopez: Nit: you've ensured `0 < vec_size < m_vec.size()`, so you could simply use the indexing operator.

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    LocalNeuralNetImp::TensorOneD::TensorOneD(std::vector<float> input)
+    {
+        set_dim(input.size());
+        std::size_t vec_size = m_vec.size();
+
+        for (std::size_t idx = 0; idx < vec_size; ++idx) {
+            m_vec[idx] = input[idx];
+        }
```
avilcheslopez: You could leverage `std::vector`'s assignment operator

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,271 @@
...
+    {
+        m_vec.resize(dim);
+    }
+
+    std::size_t LocalNeuralNetImp::TensorOneD::get_dim() const
+    {
+        return m_vec.size();
+    }
+
+    LocalNeuralNetImp::TensorOneD::TensorOneD(std::vector<float> input)
```
avilcheslopez: Put this constructor next to the other constructors

src/LocalNeuralNet.hpp
alawibaba: That's a great call. I think I misremembered what this was called in other libraries and went with this. It looks like it's typically called "forward" and overload function invocation to use this method. I think it might make sense for us to do that, too -- principle of least surprise. What do you think?

src/LocalNeuralNetImp.hpp
alawibaba: This is what @cmcantalupo and I (@alawibaba) came up with -- totally happy to have a discussion about this.

alawibaba:
Thanks for the review @avilcheslopez! I've written responses to a couple -- I'll address the others and send an update.

src/LocalNeuralNet.hpp
alawibaba: I meant pairs!

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,265 @@
...
+    std::vector<float> LocalNeuralNetImp::forward(std::vector<float> inp_vector)
+    {
+        LocalNeuralNetImp::TensorOneD inp(inp_vector);
+
+        if (inp.get_dim() != m_layers[0].first.get_cols()) {
+            throw geopm::Exception("Input vector dimension is incompatible with network.\n",
+                                   GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        LocalNeuralNetImp::TensorOneD tmp = inp;
```
avilcheslopez, line 58: Very minor note: you only need one of these TensoreOneD objects. Is `inp` being copied into another object mainly for formality to clearly separate between the input parameter and the return value? If so, perhaps this isn't necessary since due to the conversion you have to do from std::vector to TensorOneD, you're already doing this. So maybe you can just rename `inp` to `tmp` or a more descriptive name. Anyway, just a suggestion to avoid unnecessarily creating another copy of the object, not to stop the PR.

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,265 @@
...
+    {
+        if (&other == this) {
+            return *this;
+        }
+
+        m_vec = std::move(other.m_vec);
+        return *this;
+    }
+
+    float& LocalNeuralNetImp::TensorOneD::operator[] (std::size_t idx)
```
avilcheslopez, line 154: Nit:

```suggestion
    float& LocalNeuralNetImp::TensorOneD::operator[](std::size_t idx)
```

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,265 @@
...
+    {
+        return m_vec.at(idx);
+    }
+
+    float LocalNeuralNetImp::TensorOneD::operator[] (std::size_t idx) const
```
avilcheslopez, line 159: Nit:

```suggestion
    float LocalNeuralNetImp::TensorOneD::operator[](std::size_t idx) const
```

src/LocalNeuralNet.hpp
```diff
@@ -0,0 +1,64 @@
...
+            /// @brief Perform inference using the instance weights and biases.
+            ///        (Short-hand for the forward method.)
+            /// 
+            /// @param [in] std::vector<float> of input signals.
+            ///
+            /// @return Returns a std::vector<float> of output values.
+            ///
+            /// @throws geopm::Exception if input dimension is incompatible
+            /// with network.
+            std::vector<float> operator()(std::vector<float> inp) {
```
avilcheslopez, line 56: Cool!

Nit:

```suggestion
            std::vector<float> operator()(std::vector<float> inp)
            {
```

src/LocalNeuralNet.cpp
```diff
@@ -0,0 +1,265 @@
...
+    {
+        m_vec.resize(dim);
+    }
+
+    std::size_t LocalNeuralNetImp::TensorOneD::get_dim() const
+    {
+        return m_vec.size();
+    }
+
+    LocalNeuralNetImp::TensorOneD LocalNeuralNetImp::TensorOneD::operator+(const LocalNeuralNetImp::TensorOneD& other)
```
avilcheslopez, line 102: Nit: we're putting the `&` next to the identifier (rather than the type).

Multiple instances of this throughout.

test/LocalNeuralNetTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "gmock/gmock.h"
+#include "geopm/Exception.hpp"
+
+#include "LocalNeuralNetImp.hpp"
+
+using geopm::LocalNeuralNetImp;
+
+class LocalNeuralNetTest : public ::testing::Test
+{
+};
```
avilcheslopez, line 16: If a test fixture is not really needed, please get rid of this and use the regular `TEST` macro

test/LocalNeuralNetTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    std::vector<float> inp = {1, 2};
+    std::vector<float> out = net(inp);
+    EXPECT_EQ(1u, out.size());
+    EXPECT_NEAR(1/(1 + expf(-43)), out[0], 1e-6);
+}
+
+TEST_F(LocalNeuralNetTest, test_bad_layers)
+{
+    EXPECT_THROW(LocalNeuralNetImp (
```
avilcheslopez, line 41: You could use `GEOPM_EXPECT_THROW` for deeper checking or is there a reason why you preferred `EXPECT_THROW`?

PR #2772, opened 2023-01-01
[WIP] Draft to enable GEOPM Service as a k8 service using gRPC and pidfd_open()

cmcantalupo:
- Different approach than #2771
  + Here we have dropped all DBus and systemd requirements
  + Instead process management is enabled through the new Linux pidfd_open() syscall:
     https://man7.org/linux/man-pages/man2/pidfd_open.2.html
  + Requests to the service are routed through a gRPC socket in /run/geopm-service instead of DBus
  + The pidfd file descriptors are passed through a different socket to initiate a session
- There are several TODO items before this PR is ready including:
    - TODO: finish gRPC version of C++ GEOPMProxy interface
    - TODO: Test new code
    - TODO: write cffi wrappers for geopm::pidfd_* C++ interafaces
- The purpose of this early PR is to show POC and general approach


service/protoc-gen.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+#  Copyright (c) 2015 - 2022, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+
+if ! which protoc >& /dev/null || \
+       ! which grpc_cpp_plugin >& /dev/null || \
+       ! which grpc_python_plugin >& /dev/null; then
+    echo "Error: Install the grpc and grpc development packages packages" 1>&2
```
dannosliwcd, line 10:
```suggestion
    echo "Error: Install the grpc and grpc development packages" 1>&2
```

service/geopm_service.proto
```diff
@@ -0,0 +1,139 @@
+//
```
dannosliwcd, line 1: Do we need to somehow inform automake that this file is a dependency of autogen.sh, or does it know that since this is in the dist files list?

service/configure.ac
```diff
@@ -327,6 +339,8 @@ AX_CHECK_COMPILE_FLAG([-fp-model="precise"], [NO_FASTMATH="-fp-model=precise"],
 
 [EXTRA_CFLAGS="$EXTRA_CFLAGS -Wall -Werror $NO_FASTMATH"]
 [EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS -Wall -Werror $NO_FASTMATH"]
+[LIBS="$LIBS $GRPC_LIBS"]
+[INCLUDE="$INCLUCE $GRPC_INCLUDE"]
```
dannosliwcd, line 343:
```suggestion
[INCLUDE="$INCLUDE $GRPC_INCLUDE"]
```

Is INCLUDE an automake thing or a make thing, or a geopm builder thing? I can't find it in the user manuals. I assume one layer uses that to add `-I<paths>` to relevant compiler flag variables.

service/geopmdpy/access.py
```diff
@@ -28,15 +29,18 @@ def __init__(self, geopm_proxy):
...
-        try:
-            geopm_proxy.PlatformGetGroupAccess
-        except DBusError as ee:
-            if 'io.github.geopm was not provided' in str(ee):
-                err_msg = """The geopm systemd service is not enabled.
+        if geopm_proxy is None:
+            self._access_lists = system_files.AccessLists()
+        else:
+            try:
+                geopm_proxy.PlatformGetGroupAccess
```
dannosliwcd, line 36: This looks surprising to me since it's not making a function call. But it was already working before. Is PlatformGetGroupAccess an attribute that makes a dbus call in its accessor's implementation?

service/geopmdpy/access.py
```diff
@@ -373,6 +383,8 @@ def main():
     parser = ArgumentParser(description=main.__doc__)
     parser.add_argument('-c', '--controls', dest='controls', action='store_true', default=False,
                         help='Command applies to controls not signals')
+    parser.add_argument('-x', '--direct', dest='direct', action='store_true', default=False,
```
dannosliwcd, line 386: Stopping review here on a context switch. Will resume later.

service/geopmdpy/grpc_service.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def TopoGetCache(self, request, context):
+        result = geopm_service_pb2.TopoCache()
+        result.cache = self._topo_service.get_cache()
+        return result
+
+    def OpenSession(self, request, context):
+        result = geopm_service_pb2.SessionKey()
+        result.name = str(self._serial_id)
+        self._platform_service.open_session("", self._serial_id)
```
dannosliwcd, line 106: Is this `geopmd.service.open_session()`? If so, we'll need to update the docs there to allow this usage of the `client_pid` argument.

service/geopmdpy/grpc_service.py
```diff
@@ -0,0 +1,131 @@
...
+        self._timeout = 0.01
+
+    def _watch_client(self, client_pidfd):
+        return 0
+
+    def _write_mode_pid(self, client_pidfd):
+        return None
+
+    def run(self):
+        server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
```
dannosliwcd, line 124: Does gRPC do any automatic synchronization, or do our handlers need to be thread safe? Should we lock around `open_session .. increment session id`?

service/src/GRPCServiceProxy.cpp
```diff
@@ -0,0 +1,200 @@
...
+        GEOPMPackage::SessionKey response;
+        grpc::Status status = m_client->OpenSession(&context,
+                                                    request,
+                                                    &response);
+        if (!status.ok()) {
+            throw Exception("GRPCSericeProxy::platform_open_session(): " +
+                            status.error_message(),
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        m_session_key = response.name();
```
dannosliwcd, line 143: Can this be reached if a session already exists, or does the service return an error in that case? If this is reachable, is it okay to lose the previous session key?

service/src/Helper.cpp
```diff
@@ -268,4 +270,148 @@ namespace geopm
...
+            if (my_pidfd == -1) {
+                throw Exception("pidfd_open(): ",
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+            int err = setns(pidfd, CLONE_NEWPID);
+            if (err == -1) {
+                throw Exception("setns(): ",
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+            std::string info_stream_path = "/proc/self/dfinfo/" + std::to_string(pidfd);
```
dannosliwcd, line 294:
```suggestion
            std::string info_stream_path = "/proc/self/fdinfo/" + std::to_string(pidfd);
```

service/src/Helper.cpp
```diff
@@ -268,4 +270,148 @@ namespace geopm
...
+            std::ifstream info_stream(info_stream_path);
+            for (std::string line; std::getline(info_stream, line); ) {
+                std::string key = "Pid:";
+                if (string_begins_with(line, key)) {
+                    pid = std::stoi(line.substr(key.size()));
+                    break;
+                }
+            }
+            if (pid == -1) {
+                throw Exception("pidfd_to_sidpid(): Failed to parse dfinfo file",
```
dannosliwcd, line 304:
```suggestion
                throw Exception("pidfd_to_sidpid(): Failed to parse fdinfo file",
```

service/src/Helper.cpp
```diff
@@ -268,4 +270,148 @@ namespace geopm
...
+                                errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+            }
+            err = setns(my_pidfd, CLONE_NEWPID);
```
dannosliwcd, line 307: This should probably happen before the preceding `if(..){ throw }`. Or put `getmypidfd/setNS ... resetNS/closemypidfd` inside a scoped wrapper.

service/src/Helper.cpp
```diff
@@ -268,4 +270,148 @@ namespace geopm
         }
         return stat_struct.st_gid;
     };
+
+    int pidfd_open(pid_t pid, int flags)
+    {
+        return syscall(SYS_pidfd_open, pid, flags);
+    }
+
+    pid_t pidfd_to_pid(int pidfd)
```
dannosliwcd, line 279: When do we plan on translating pidfd to its self-viewed pid?

If we plan to translate pidfds to pids for use in our existing control mechanisms, then I think we'd need to include the namespaces themselves as part of the identifier. Or maybe just use the pidfd? Otherwise, we need to get the process's PID from a common ancestor namespace instead of getting it from the target process's namespace (to avoid collisions between same PID numbers in different namespaces).

cmcantalupo: The goal is to be able to call getsid() in the namespace of the pid that is requesting write control.  

You can change PID namespaces with the pidfd using the setns(2) call.  This is the intended use of the pidfd_ns paramters.

dannosliwcd: An example of my concern is if two separate applications attach to the GEOPM service from separate containers. If app1's SID is 1 in its PID namespace and app2's SID is also 1 in its PID namespace, then GEOPM would wrongly think they are in the same session (if we only use self-viewed PIDs as identifiers), even if they are actually different processes.

service/geopm_service.proto
cmcantalupo: This would be cleaner if we added it as a make file target: we could identify the dependency between the output of protoc and the geopm_service.proto file.  This could enable the automatic regeneration of the output files if the proto file changed.  

Here, I'm just assuming that autogen.sh is always run first, and the user "knows" to rerun protoc-gen.sh if the geopm_service.proto file changes.  The output source files are included in the distribution tarball, so this step is not required when building from a distribution tarball (potentially without running autogen.sh).

service/geopmdpy/access.py
cmcantalupo: This is the best way I could make trigger an exception without *calling* one of our methods.  We don't have a "Hello" method or the like.

service/geopmdpy/grpc_service.py
cmcantalupo: This is more of a work-around.  In the DBus use case we get the client pid from the DBus service. We need to figure out a way to reliably get the user credentials through gRPC, and then associate those credentials with a pidfd.  

Instead we are doing anonymous access for read only.  I haven't implemented any write features, and the access list used is the default access list that applies to all users.

service/geopmdpy/grpc_service.py
cmcantalupo: gRPC is explicitly multithreaded.  The increment of session id is also a work around.  I would prefer that we were recording the pidfd as the session id.

service/src/GRPCServiceProxy.cpp
cmcantalupo: The service increments a reference count.  The key that is returned should be the same in this case (not true for the incrementing work around).

service/src/Helper.cpp
cmcantalupo: We will store the pidfd for the session leader PID.  This pidfd will always be used with setns() prior to calling getsid() on the user provided pid.  We wouldn't use the pidfd namespace of the user provided pid (the problematic case you are referring to).  The session leader PID namespace would always be the reference namespace when determining another pid's session leader.

PR #2825, opened 2023-01-26
Fix issues with building on Ubuntu 22

### LINKED ISSUE #2829: Several issues were discovered when updating to Ubuntu 22 in github ci
- Read off end of array issue with ApplicationSampler test
- Improper initialization of array values in implementation accounting for hint time
  + Could use uninitialized values for hint time start depending on compiler interpretation
  + Could lead to arbitrary output for hint time or leaked data.
- Some issues around implicit casting in constructors.

cmcantalupo:
- Fixes #2829


configure.ac
```diff
@@ -345,7 +345,7 @@ AC_SUBST([MPI_FCLIBS])
 AC_SUBST([MPI_FLIBS])
 
 [EXTRA_CFLAGS="$EXTRA_CFLAGS $OPENMP_CXXFLAGS -std=c99 -msse4.2"]
-[EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS $OPENMP_CXXFLAGS -std=c++11 -msse4.2"]
+[EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS $OPENMP_CXXFLAGS -std=c++17 -msse4.2"]
```
dannosliwcd, line 348: ✨
I guess we can start doing some things differently after this. First things that come to mind are things we've pulled into the geopm namespace, like std::make_unique and std::hardware_destructive_interference_size.

service/configure.ac
```diff
@@ -218,7 +218,7 @@ AC_DEFINE_UNQUOTED([GEOPM_CONFIG_PATH],
           [The prefix to the path where GEOPM config files are stored])
 
 [EXTRA_CFLAGS="$EXTRA_CFLAGS -std=c99 -msse4.2"]
-[EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS -std=c++11 -msse4.2"]
```
cmcantalupo, line 221: docs?

dannosliwcd (dismissed):

test/ApplicationSamplerTest.cpp
```diff
@@ -529,24 +529,24 @@ TEST_F(ApplicationSamplerTest, sampler_cpu)
     // inactive.
 
     // The topo is queried to discover the total number of cores
-    EXPECT_CALL(*m_mock_topo, num_domain(GEOPM_DOMAIN_CORE))
+    EXPECT_CALL(*m_mock_topo, num_domain(int(GEOPM_DOMAIN_CORE)))
```
cmcantalupo: check again on these casts

configure.ac
cmcantalupo: Yeah, we should open an issue to deprecate their use.

test/ApplicationSamplerTest.cpp
cmcantalupo: was an artifact of the array access is issue, not actually required.

configure.ac
cmcantalupo: https://github.com/geopm/geopm/issues/2827

bgeltz (approved):

Issue #2728, opened 2022-11-23
Design solution to disable profiling


avilcheslopez:
> estimate 1

Done criteria:
- [x] Design discussed with the team
- [x] Scope of work understood

PR #2836, opened 2023-02-06
Enable non-MPI applications and get rid of controller / application handshake

cmcantalupo:
Relates to #2728
Relates to #2875

This draft PR differs from the issue-2728 branch, in that this preserves application feedback when MPI is disabled.


service/geopmdpy/system_files.py
```diff
@@ -723,6 +731,51 @@ def remove_batch_server(self, client_pid):
...
+        shmem.create_prof('record-log', size, client_pid, uid, gid)
+        self._update_session_file(client_pid)
+
+    def stop_profile(self, client_pid, region_names):
+        self.check_client_active(client_pid, 'stop_profile')
+        try:
+            profile_name = self._sessions[client_pid].pop('profile_name')
+        except KeyError:
+            raise RuntimeError(f'Client PID {client_pid} requested to stop profiling, but it had not been started.')
+        # TODO: store region names in file in /run/geopm-service to enable clean restart
```
avilcheslopez, line 764: It's fine. Delete :)

src/ApplicationSampler.cpp
```diff
@@ -284,51 +266,37 @@ namespace geopm
...
-        for (int cpu_idx = 0; cpu_idx != m_num_cpu; ++cpu_idx) {
-            result[cpu_idx] = m_status->get_process(cpu_idx);
-        }
-        return result;
-#endif
+        return m_per_cpu_process;
     }
 
-    void ApplicationSamplerImp::connect(const std::string &shm_key)
+    void ApplicationSamplerImp::connect(void)
```
avilcheslopez: Refactor (break this down into more functions, perhaps a helper function)

src/ApplicationSampler.cpp
```diff
@@ -337,12 +305,32 @@ namespace geopm
...
+                int err = geopm_sched_proc_cpuset_pid(pid, m_num_cpu, cpuset.get());
+                if (err) {
+                    throw Exception("Failed to get cpuset for pid: " + std::to_string(pid),
+                                    err, __FILE__, __LINE__);
+                }
+                CPU_OR_S(set_size, all_cpuset.get(), all_cpuset.get(), cpuset.get());
+                for (int cpu_idx = 0; cpu_idx < m_num_cpu; ++cpu_idx) {
+                    if (CPU_ISSET_S(cpu_idx, set_size, cpuset.get())) {
+                        process.cpus.push_back(cpu_idx);
+                        // TODO: Last PID wins if the affinity masks overlap
```
avilcheslopez: Emit a warning that warns user about overlap

src/ApplicationIO.cpp
```diff
@@ -40,67 +56,83 @@ namespace geopm
...
+                // PID is no longer valid
+                errno = 0;
+                pid_it = m_profile_pids.erase(pid_it);
+            }
+            else {
+                // We don't need to prune all values if we find one
+                // valid PID
+                break;
+            }
         }
```
bgeltz: Need a unit test to prove that a process that is erased from the middle of the vector does not lose the first elements.  E.g. a vector of 1,2,3 where 2 is erased will have pid_id point to 3.  What happened to 1?

src/Reporter.cpp
```diff
@@ -53,7 +53,8 @@ namespace geopm
                       nullptr,
                       environment().report_signals(),
                       environment().policy(),
-                      environment().do_endpoint())
+                      environment().do_endpoint(),
+                      environment().timeout() != -1)
```
bgeltz, line 57: Reconsider if this is the best way for a user to toggle profiling on and off.  environment().do_profile() for example.

src/DefaultProfile.cpp
```diff
@@ -10,40 +12,24 @@
...
     {
-        g_pmpi_prof_enabled = m_is_enabled;
-        g_pmpi_tprof_enabled = m_is_enabled;
-    }
-
-    DefaultProfile::~DefaultProfile()
-    {
-        g_pmpi_prof_enabled = 0;
-        g_pmpi_tprof_enabled = 0;
+        g_prof_enabled = 1;
```
bgeltz, line 30: Why are you enabling profiling in the destructor?  The seems to be the opposite of the previous behavior.

bgeltz (changes requested):

src/DefaultProfile.cpp
```diff
@@ -10,40 +12,24 @@
...
         : ProfileImp()
     {
-        g_pmpi_prof_enabled = m_is_enabled;
-        g_pmpi_tprof_enabled = m_is_enabled;
-    }
-
-    DefaultProfile::~DefaultProfile()
-    {
-        g_pmpi_prof_enabled = 0;
-        g_pmpi_tprof_enabled = 0;
```
bgeltz, line 39: Why do we no longer need to set the enable vars to 0 in the destructor?

bgeltz:
Are there any guards against oversubscribing the cores between 2+ apps?

service/geopmdpy/system_files.py
```diff
@@ -723,6 +731,51 @@ def remove_batch_server(self, client_pid):
...
+            os.unlink(shmem.path_prof('status', client_pid, uid, gid))
+        self._update_session_file(client_pid)
+
+    def get_profile_pids(self, profile_name):
+        return self._profiles.get(profile_name)
+
+    def get_profile_region_names(self, profile_name):
+        result = None
+        if profile_name in self._region_names:
+            result = self._region_names.pop(profile_name)
```
avilcheslopez, line 792: Just curious: if (based on the method name) the intention is to retrieve the region names for a given profile, why pop the profile entry from `_region_names` altogether?

src/ApplicationSampler.cpp
```diff
@@ -98,7 +95,9 @@ namespace geopm
                                 std::map<int, m_process_s> {},
                                 environment().do_record_filter(),
                                 environment().record_filter(),
-                                {})
+                                {},
+                                environment().timeout() != -1,
```
bgeltz, line 101: Another spot to revist if we want the timeout env var to control do_profile.

src/ApplicationStatus.hpp
```diff
@@ -61,23 +61,7 @@ namespace geopm
...
-            ///        application side, and one ApplicationRecordLog
-            ///        per process on each side of the shared memory.
-            /// @param [in] cpu_idx Set of Linux logical CPUs for the process
-            virtual void set_process(const std::set<int> &cpu_idx, int process) = 0;
-            /// @brief Get the process ID for the process the CPU is
-            ///        currently assigned to.
-            /// @param [in] cpu_idx Index of the Linux logical CPU
-            /// @return ID of the process running on the given CPU.
-            virtual int get_process(int cpu_idx) const = 0;
+            virtual void set_valid_cpu(const std::set<int> &cpu_idx, bool is_valid) = 0;
```
bgeltz: Needs docu

src/Controller.cpp
```diff
@@ -156,16 +156,18 @@ namespace geopm
                      environment().agent(),
                      Agent::num_policy(environment().agent()),
                      Agent::num_sample(environment().agent()),
-                     std::unique_ptr<TreeComm>(new TreeCommImp(ppn1_comm,
+                     geopm::make_unique<TreeCommImp>(
```
bgeltz: Do we still need to use the geopm flavor of make_unique here, or can this be moved over to the GCC implementation?  Maybe we should save it and do it all at once?

src/Controller.cpp
```diff
@@ -323,22 +322,20 @@ namespace geopm
...
-        m_application_io->controller_ready();
-        geopm_time_s curr_time;
-        geopm_time(&curr_time);
-        m_application_sampler.update(curr_time);
         m_platform_io.read_batch();
         m_reporter->update();
         m_tracer->update(m_trace_sample);
         m_profile_tracer->update(m_application_sampler.get_records());
-
+        m_reporter->update();
```
bgeltz: Do we need to call update() here and on line 338?

src/ApplicationStatus.hpp
avilcheslopez: Please add code documentation for this method

src/CSV.cpp
```diff
@@ -29,9 +30,11 @@ namespace geopm
         , m_buffer_limit(buffer_size)
         , m_is_active(false)
     {
+#ifdef ENABLE_MPI
```
cmcantalupo: ENABLE_MPI comes automatically with the MPI m4 macros that we are using (which are not GEOPM sourced)
https://github.com/geopm/geopm/blob/dev/m4/lx_find_mpi.m4

src/Controller.cpp
cmcantalupo: no i think this is a rebase error.  The better place is below, just because it matches the sequence of calls in the step() function.

src/Controller.cpp
cmcantalupo: can't hurt to use the std version in new code.

service/geopmdpy/system_files.py
cmcantalupo: We need some mechanism for clean up of the resources associated with the application run, and since we currently have only one consumer of this information (the controller), I put the clean up in the function the controller calls once at the end.  This is a nasty side effect of this method, and we should have a better story for cleaning up.

src/Profile.cpp
```diff
@@ -56,75 +55,33 @@ namespace geopm
...
+                           bool do_profile)
         : m_is_enabled(false)
         , m_prof_name(prof_name)
-        , m_key_base(key_base)
         , m_report(report)
-        , m_timeout(timeout)
-        , m_comm(comm)
         , m_curr_region_id(0)
         , m_current_hash(GEOPM_REGION_HASH_UNMARKED)
         , m_ctl_shmem(nullptr)
```
avilcheslopez: It seems like this property is no longer needed

src/Profile.cpp
```diff
@@ -134,212 +91,74 @@ namespace geopm
...
-                tmp_msg[PATH_MAX-1] = '\0';
-                std::cerr << tmp_msg << std::endl;
-            }
-            m_is_enabled = false;
+            char tmp_msg[PATH_MAX];
+            geopm_error_message(err, tmp_msg, sizeof(tmp_msg));
+            tmp_msg[PATH_MAX-1] = '\0';
+            std::cerr << tmp_msg << std::endl;
         }
+        m_is_enabled = true;
```
avilcheslopez: Did you intend to make `m_is_enabled = true` even in the case where the prior init steps throw an exception?

src/Reporter.cpp
```diff
@@ -110,6 +103,9 @@ namespace geopm
         if (m_do_profile && m_proc_region_agg == nullptr) {
             // ProcessRegionAggregator should not be constructed until
             // application connection is established.
+            init_sync_fields();
+            init_environment_signals();
```
cmcantalupo, line 111: Does this impact the report if profiling is disabled?  The update function is a noop in this case, so it seems like the generate function should also be a noop.

integration/test/test_multi_app.py
```diff
@@ -0,0 +1,76 @@
...
+    TEST_NAME = 'test_multi_app'
+    TIME_LIMIT = 30
+    NUM_NODE = 1
+    EXPECTED_REGIONS = {'model-init', 'stream', 'dgemm'}
+
+    @classmethod
+    def setUpClass(cls):
+        sys.stdout.write('(' + os.path.basename(__file__).split('.')[0] +
+                         '.' + cls.__name__ + ') ...')
+        subprocess.run(['/bin/bash', 'test_multi_app.sh'],
```
cmcantalupo: A simple solution to the PATH problem here is to assume that test_multi_app.sh script is in the same directory as the test_multi_app.py script.  The path to the script itself

```
os.path.join(os.path.dirname(os.path.abspath(__file__)), 'test_multi_app.sh') 
```

or something similar

scripts/geopmpy/launcher.py
```diff
@@ -396,6 +394,7 @@ def __init__(self, argv, num_rank=None, num_node=None, cpu_per_rank=None, timeou
         # Calculate derived values
         if self.rank_per_node is None and self.num_rank and self.num_node:
             self.rank_per_node = int_ceil_div(self.num_rank, self.num_node)
+        self.environ_ext['GEOPM_NUM_PROC'] = str(self.rank_per_node)
```
avilcheslopez, line 398: Does it matter that `self.rank_per_node` could be adjusted a bit further down in the code (line 439)?

scripts/geopmpy/launcher.py
```diff
@@ -1133,7 +1148,7 @@ def preload_option(self):
         result = []
         if self.config and self.config.get_preload():
             value = ':'.join((ll for ll in
-                              ('libgeopm.so', os.getenv('LD_PRELOAD'))
+                              ('libgeopm.so.1.0.0', os.getenv('LD_PRELOAD'))
```
avilcheslopez: I'm worried that I'm seeing this change multiple times. Is there a reason why each implementation of `Launcher` implements `preload_option()`? If it is necessary, I would recommend at least putting this string in a constant.

service/geopmdpy/service.py
```diff
@@ -860,6 +938,22 @@ def PlatformReadSignal(self, signal_name, domain, domain_idx, **call_info):
...
+    def PlatformStopProfile(self, region_names, **call_info):
+        return self._platform.stop_profile(self._get_pid(**call_info), region_names)
+
+    # TODO: This method should check credentials
+    def PlatformGetProfilePids(self, profile_name):
+        return self._platform.get_profile_pids(profile_name)
+
+    # TODO: This method should check credentials
+    def PlatformGetProfileRegionNames(self, profile_name):
+        return self._platform.get_profile_region_names(profile_name)
```
avilcheslopez: Are these TODOs for this PR or for a later PR?

cmcantalupo: Yes, lets open an issue.

src/GEOPMBenchConfig.cpp
```diff
@@ -0,0 +1,32 @@
...
+namespace geopm
+{
+    const GEOPMBenchConfig &geopmbench_config(void)
+    {
+        static GEOPMBenchConfigImp instance;
+        return instance;
+    }
+
+    GEOPMBenchConfigImp::GEOPMBenchConfigImp()
+        : GEOPMBenchConfigImp(getenv("GEOPMBENCH_NO_MPI") == nullptr)
```
bgeltz, line 19: This should be included in the geopmbench.1 man page.

bgeltz (dismissed):

src/GEOPMBenchConfig.cpp
cmcantalupo: Just pushed a patch.

integration/test/test_hint_time.cpp
```diff
@@ -22,6 +23,7 @@ int main(int argc, char **argv)
     int comm_rank;
     MPI_Init(&argc, &argv);
     MPI_Comm_rank(MPI_COMM_WORLD, &comm_rank);
+    sleep(5);
```
bgeltz, line 26: Is this still required?  If yes, we need to add some documentation somewhere that makes it clear that if you try to profile regions that occur too quickly after app startup, data will be lost.

integration/test/test_hint_time.py
```diff
@@ -153,9 +157,9 @@ def test_hint_time(self):
             expect = 1.0
             actual = raw_totals['time-hint-memory (s)']
             util.assertNear(self, expect, actual, msg=msg)
-            msg = "Application totals should have four seconds of total time"
-            expect = 4.0
-            actual = raw_totals['sync-runtime (s)']
+            msg = "Application totals should have nine seconds of total time"
+            expect = 9.0 + overhead_time
```
bgeltz, line 159: There should be a comment here explaining why the total time is different than what is setup in the test_hint_time.cpp (4 seconds presently).

bgeltz (changes requested):

integration/test/test_hint_time.py
```diff
@@ -99,9 +100,11 @@ def setUpClass(cls):
             launcher = geopm_test_launcher.TestLauncher(app_conf,
                                                         agent_conf,
                                                         cls._report_path,
+                                                        cls._trace_path,
                                                         time_limit=time_limit)
             launcher.set_num_node(cls._num_node)
             launcher.set_num_rank(num_rank)
+            launcher.set_pmpi_ctl('application')
```
bgeltz, line 105: What prompted this change?

integration/test/test_hint_time.py
```diff
@@ -76,6 +76,7 @@ def setUpClass(cls):
                          '.' + cls.__name__ + ') ...')
         cls._test_name = 'hint_time'
         cls._report_path = 'test_{}.report'.format(cls._test_name)
+        cls._trace_path = f'test_{cls._test_name}.trace'
```
bgeltz:
```suggestion
```
Traces are not used in this test.  Was this just for debug?

integration/test/test_hint_time.py
```diff
@@ -99,9 +100,11 @@ def setUpClass(cls):
             launcher = geopm_test_launcher.TestLauncher(app_conf,
                                                         agent_conf,
                                                         cls._report_path,
+                                                        cls._trace_path,
```
bgeltz:
```suggestion
```

integration/test/test_hint_time.py
```diff
@@ -167,7 +171,7 @@ def test_hint_time(self):
             expect = 1.0
             actual = raw_epoch['time-hint-memory (s)']
             util.assertNear(self, expect, actual, msg=msg)
-            msg = "Epoch should have three seconds of total time".format(region_name)
+            msg = "Epoch should have two seconds of total time".format(region_name)
```
bgeltz, line 172: Same thing here.  Somewhere there should be a comment explaining what to expect and why based on the code in the cpp.

src/geopmbench_main.cpp
```diff
@@ -152,6 +157,7 @@ int main(int argc, char **argv)
             err = geopm_prof_exit(init_rid);
         }
         if (!err) {
+            sleep(5);
```
bgeltz, line 160: There should be a comment here, and probably also in the geopmbench.1 man page about why this is necessary.

integration/test/test_monitor.py
```diff
@@ -96,21 +96,24 @@ def test_runtime(self):
...
-            total_runtime = (spin_data['runtime (s)'] +
+            total_runtime = (app_total['GEOPM overhead (s)'] +
+                             spin_data['runtime (s)'] +
                              sleep_data['runtime (s)'] +
                              unmarked_data['runtime (s)'])
             util.assertNear(self, total_runtime, app_total['runtime (s)'])
 
     def test_runtime_epoch(self):
         '''Test that region and epoch total runtimes match.'''
+        initial_sleep_time = 5.0
```
bgeltz, line 107: Include a comment about where this magic number came from.  Also, perhaps this should be moved to a class member variable if it's used multiple times.

integration/test/test_multi_app.sh
```diff
@@ -0,0 +1,34 @@
...
+EOF
+
+source ${GEOPM_SOURCE}/integration/config/run_env.sh
+
+TEST_NAME=test_multi_app
+export GEOPM_REPORT=${TEST_NAME}_report.yaml
+export GEOPM_TRACE=${TEST_NAME}_trace.csv
+export GEOPM_TRACE_PROFILE=${TEST_NAME}_trace_profile.csv
+export GEOPM_PROFILE=${TEST_NAME}
+#export GEOPM_REPORT_SIGNALS=CPU_TIMESTAMP_COUNTER@package
```
bgeltz:
```suggestion
```
Debug cruft?

integration/test/test_multi_app.sh
```diff
@@ -0,0 +1,34 @@
...
+export GEOPM_REPORT_SIGNALS=TIME@package
+export GEOPM_NUM_PROC=2
+
+geopmctl &
+sleep 2
```
bgeltz: geombench has a sleep 5 in it.  Is this still needed?  If yes, please add a comment explaining why.  I'm guessing this is just a problem for stress-ng?

integration/test/test_omp_outer_loop.cpp
```diff
@@ -21,6 +22,7 @@ int main(int argc, char **argv)
     int comm_rank = -1;
     int comm_size = -1;
     err = MPI_Init(&argc, &argv);
+    sleep(5);
```
bgeltz, line 25: Comment on why.  Please do this in all the places that have a newly added sleep 5.

integration/test/test_time.py
```diff
@@ -0,0 +1,123 @@
...
+    def tearDown(self):
+        if sys.exc_info() != (None, None, None):
+            TestIntegration_hint_time._keep_files = True
+
+    def test_hint_time(self):
+        host_names = self._report.host_names()
+        self.assertEqual(len(host_names), self._num_node)
+        for host in host_names:
+            found_nw = False
+            found_nw_mem = False
```
bgeltz:
```suggestion
```
Unused

integration/test/test_time.py
```diff
@@ -0,0 +1,123 @@
...
+            self.assertEqual(0, len(self._report.region_names(host)))
+            raw_totals = self._report.raw_totals(host_name=host)
+            msg = "Application totals should have 10 seconds of total time"
+            expect = 10.0
+            actual = raw_totals['runtime (s)']
+            util.assertNear(self, expect, actual, msg=msg)
```
bgeltz, line 110: Why doesn't this test need to account for the extra magic 5 seconds that the others do?

bgeltz (dismissed):

integration/test/test_hint_time.cpp
cmcantalupo: The data isn't "lost" we do not collect hardware samples until the controller is ready, but the app runs immediately.  For this reason, we cannot associate hardware metrics with the application data that *is* collected and processed (time stamped application data is not lost).  So, for tests that require hardware data to be collected in regions that occur early in the app, the sleep is required.

integration/test/test_hint_time.py
cmcantalupo: That is the extra 5 seconds of wait time at the beginning of geopmbench of course.

integration/test/test_hint_time.py
cmcantalupo: Both changes may have been made to debug (application launch mode and creation of the trace data).

integration/test-no-mpi/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Notes
+
+This is a work in progress.
+
+TODO: Add some tests that use the REPORT_SIGNALS and TRACE_SIGNALS features.
+
+TODO: Add some assertions to your "tests".
+
+TODO: Add command line arguments to ``geopmctl`` that align with the
```
bgeltz: Have tracking issues been created for these TODOs?

service/geopmdpy/system_files.py
```diff
@@ -728,6 +735,61 @@ def remove_batch_server(self, client_pid):
...
+            size = 64 * os.cpu_count()
+            shmem.create_prof('status', size, client_pid, uid, gid)
+        if profile_name in self._profiles:
+            self._profiles[profile_name].add(client_pid)
+        else:
+            self._profiles[profile_name] = {client_pid}
+        self._sessions[client_pid]['profile_name'] = profile_name
+        size = 57384
+        shmem.create_prof('record-log', size, client_pid, uid, gid)
+        self._update_session_file(client_pid)
```
avilcheslopez: There is a scenario that could cause problems with this code. Given how the client is going to call into these APIs, it should never happen, but it's still worth protecting against in case of a bug in the client or a malicious actor calling into the DBus APIs.

The scenario happens as a result of:
- A client can call `PlatformStartProfile` multiple times
- The code tracks the profile a client belongs to in two data structures: a map of profile names to PIDs (as a profile can have multiple PIDs), and the session data map (where a PID's profile name is stored)

The scenario is: a client calls `PlatformStartProfile` twice with two different profile names. The way the code currently is, it will add the client's PID to two different profiles in the profiles map, and would only track the last profile name with which it was called in the session data map. So when this client stops profiling, its profile will be removed from the session data map, but it will be removed only from one of the profiles in the profiles map.

I think the `shmem.create_prof()` call might fail in this scenario the second time its called and might prevent this scenario from happening, but I think it would be better to guard against it more explicitly.

service/geopmdpy/system_files.py
avilcheslopez: So maybe it's just a matter of changing the method's name, for example, something like `pop_profile_region_names()`

service/geopmdpy/shmem.py
```diff
@@ -0,0 +1,32 @@
...
+        raise RuntimeError('geopm_shmem_create() failed: {}'.format(error.message(err)))
+
+def path_prof(shm_key, pid, uid, gid):
+    global _dl
+    name_max = 1024
+    shm_key_cstr = gffi.gffi.new("char[]", shm_key.encode())
+    result_cstr = gffi.gffi.new("char[]", name_max)
+    err = _dl.geopm_shmem_path_prof(shm_key_cstr, pid, uid, name_max, result_cstr)
+    if err < 0:
+        raise RuntimeError('geopm_shmem_destroy() failed: {}'.format(error.message(err)))
```
avilcheslopez: Did you mean 'geopm_shmem_path_prof() failed'?

service/docs/source/runtime.rst
```diff
@@ -1,23 +1,76 @@
 
-Guide for HPC Runtime Users
-===========================
-The GEOPM HPC Runtime enables interactions between GEOPM's :doc:`application
+Guide for Runtime Users
+=======================
+
+The GEOPM Runtime is software designed to enhance energy efficiency of
+applications though active hardware configuration.  The architecture is
+designed to provide an secure infrastructure to support a wide range
```
bgeltz:
```suggestion
designed to provide a secure infrastructure to support a wide range
```

service/docs/source/runtime.rst
```diff
@@ -1,23 +1,76 @@
...
+of tuning algorithms while solving three related challenges.
```
bgeltz:
```suggestion
of tuning algorithms while solving three related challenges:
```

service/docs/source/runtime.rst
```diff
@@ -66,19 +119,18 @@ individual MPI application and enable the management of system power
...
-Agent algorithm to control the compute application.  The
-:doc:`geopmlaunch(1) <geopmlaunch.1>` command is part of the geopmpy python
-package that is included in the GEOPM installation.  See the :doc:`GEOPM
-overview man page <geopm.7>` for further documentation and links.
-
-The GEOPM HPC Runtime provides some built-in algorithms, each as an
+The GEOPM Runtime package provides the libgeopm shared object library.
+There are several command line tools included in GEOPM which have
+dedicated manual pages.  The :doc:`geopmlaunch(1) <geopmlaunch.1>`
+command line tool is used to launch an MPI application while enabling
```
bgeltz, line 125:
```suggestion
command line tool is used to launch an application while enabling
```

service/docs/source/runtime.rst
```diff
@@ -108,14 +160,14 @@ Build Requirements
...
+to build the GEOPM Runtime some additional requirements must be
+met.  If the user is not interested in building the GEOPM Runtime,
 these extra build requirements may be ignored.  The user may also opt
-out of the specific GEOPM HPC Runtime features enabled by any of these
+out of the specific GEOPM Runtime features enabled by any of these
 requirements by providing the appropriate disable flag to the base
 build configure command line.
 
-The GEOPM HPC runtime requires support for MPI, the Message Passing
+The GEOPM Runtime requires support for MPI, the Message Passing
```
bgeltz, line 170: Where's the right spot to document the nuances about controller launch modes and MPI requirements?  The way this is, it's saying we require MPI in all cases.

service/geopmdpy/gffi.py
```diff
@@ -34,25 +34,25 @@ def get_dl_geopmd():
...
+    if type(_dl_geopm) is OSError:
+        raise _dl_geopm
+    return _dl_geopm
 
-# Enforce load order of libgeopmpolicy.so and libgeopmd.so
+# Enforce load order of libgeopm.so and libgeopmd.so
 try:
-    _dl_geopmpolicy = gffi.dlopen('libgeopmpolicy.so.1',
+    _dl_geopm = gffi.dlopen('libgeopm.so.1',
                                   gffi.RTLD_GLOBAL|gffi.RTLD_LAZY)
```
bgeltz:
```suggestion
    _dl_geopm = gffi.dlopen('libgeopm.so.1',
                            gffi.RTLD_GLOBAL|gffi.RTLD_LAZY)
```

service/src/IOGroup.cpp
```diff
@@ -88,41 +88,42 @@ namespace geopm
...
+#endif
+#ifdef GEOPM_ENABLE_DCGM
+            register_plugin(DCGMIOGroup::plugin_name(),
+                            DCGMIOGroup::make_plugin);
+#endif
+#ifdef GEOPM_ENABLE_NVML
+            register_plugin(NVMLIOGroup::plugin_name(),
+                            NVMLIOGroup::make_plugin);
 #endif
+        }
```
avilcheslopez: Minor suggestion to change the structure of the `if` statement to avoid possibly generating an empty true branch:

```
    if (getuid() == 0) {
        // register plugins that require root privilege
    }
#ifdef GEOPM_ENABLE_SYSTEMD
    else {
        // register ServiceIOGroup
    }
#endif
```

src/All2allModelRegion.cpp
```diff
@@ -102,40 +110,52 @@ namespace geopm
...
-                ModelRegion::loop_exit();
+                    ModelRegion::loop_exit();
+                }
+                ModelRegion::region_exit();
+            }
+            else {
+                ModelRegion::region_enter();
+                for (uint64_t i = 0; i < m_num_progress_updates; ++i) {
+                    ModelRegion::loop_enter(i);
+                    std::this_thread::sleep_for(std::chrono::microseconds(100));
```
bgeltz, line 154: geopmbench.1.rst should probably be updated to say that all2all becomes sleep when MPI is disabled.

We should document all the changed behaviors for the other regions as applicable.

src/ApplicationIO.cpp
```diff
@@ -2,105 +2,139 @@
...
-            m_application_sampler.get_sampler()->initialize();
-            m_is_connected = true;
+        std::vector<int> result(m_profile_pids.begin(), m_profile_pids.end());
+        if (m_is_connected) {
+            return result;
+        }
+        double timeout = m_timeout;
+        if (timeout < 0.0) {
+            timeout = 60.0;
         }
```
bgeltz, line 63: Is this just a paranoia check?

src/ApplicationRecordLog.cpp
```diff
@@ -169,21 +158,90 @@ namespace geopm
...
+            if (CPU_ISSET(cpu_idx, cpu_set.get())) {
+                affinity(time, cpu_idx);
+            }
+        }
+    }
+
+    void ApplicationRecordLogImp::affinity(const geopm_time_s &time, int cpu_idx)
+    {
+        std::unique_ptr<SharedMemoryScopedLock> lock = m_shmem->get_scoped_lock();
+        m_layout_s &layout = *((m_layout_s *)(m_shmem->pointer()));
```
bgeltz, line 189: I think we could consider a refactor here to use dynamic_cast instead of explicit C style casting.

tutorial/admin/README.md
```diff
@@ -31,22 +31,22 @@ message will be displayed recommending corrective action or referring the user
 back to the main GEOPM README for guidance.
 
 ## 1. GEOPM+SLURM static policy plugin
-GEOPM provides two libraries: `libgeopmpolicy` and `libgeopm`. `libgeopmpolicy`
+GEOPM provides two libraries: `libgeopm` and `libgeopm`. `libgeopm`
```
bgeltz, line 34:
```suggestion
GEOPM provides two libraries: `libgeopm` and `libgeopmd`. `libgeopmd`
```

tutorial/admin/README.md
```diff
@@ -31,22 +31,22 @@ message will be displayed recommending corrective action or referring the user
...
 contains tools for interacting with hardware signals and controls, such as
 `geopmread`, and the supporting library functions. `libgeopm` contains all of
 these functions and also provides tools for launching and interacting with MPI
 applications.
 
-The `libgeopmpolicy` library must be available in order to set up a system with
+The `libgeopm` library must be available in order to set up a system with
 the GEOPM+SLURM static policy plugin. The SLURM plugin requires
-`libgeopmpolicy` alone; it does not use `libgeopm` or MPI. If the GEOPM runtime
+`libgeopm` alone; it does not use `libgeopm` or MPI. If the GEOPM runtime
```
bgeltz, line 42:
```suggestion
`libgeopmd` alone; it does not use `libgeopm` or MPI. If the GEOPM runtime
```

tutorial/admin/README.md
```diff
@@ -31,22 +31,22 @@ message will be displayed recommending corrective action or referring the user
...
 will be installed (including `libgeopm` and MPI launch tools), it should be
 installed in a different location, preferably using the module system. The
 packages are available through OpenHPC
 (<https://openhpc.community/downloads/>). See the [GEOPM runtime
 capabilities](#2-geopm-runtime-capabilities) section in this guide.
 
-A.  To build `libgeopmpolicy` and a compatible `libgeopm_slurm.so`:
+A.  To build `libgeopm` and a compatible `libgeopm_slurm.so`:
```
bgeltz, line 49:
```suggestion
A.  To build `libgeopmd` and a compatible `libgeopm_slurm.so`:
```

bgeltz: We should revisit all of this docu.

integration/test-no-mpi/Makefile
```diff
@@ -0,0 +1,12 @@
+#  Copyright (c) 2015 - 2023, Intel Corporation
```
bgeltz: Delete this entire directory.  It's redundant with the original integration tests when you disable MPI in geopmbench.

integration/test/test_hint_time.cpp
bgeltz, line 26: Comment please on why this sleep is here.

integration/test/test_hint_time.py
bgeltz, line 105:
```suggestion
```

integration/test/test_multi_app.sh
bgeltz: Remove the sleep and re-test.

src/Environment.cpp
```diff
@@ -431,6 +437,11 @@ namespace geopm
         return std::stoi(lookup("GEOPM_TIMEOUT"));
```
bgeltz:
```suggestion
```

src/ApplicationIO.cpp
bgeltz:
```suggestion
```

bgeltz: The Environment.cpp should also enforce that the timeout is positive.

src/ApplicationIO.hpp
```diff
@@ -27,10 +30,10 @@ namespace geopm
             virtual ~ApplicationIO() = default;
             /// @brief Connect to the application via
             ///        shared memory.
```
avilcheslopez, line 32: Minor note: maybe add a note about what the function now returns.

src/ApplicationIO.hpp
```diff
@@ -39,34 +42,34 @@ namespace geopm
...
         private:
+            std::set<int> get_profile_pids(void);
             static constexpr size_t M_SHMEM_REGION_SIZE = 2*1024*1024;
 
             bool m_is_connected;
-            ApplicationSampler &m_application_sampler;
+            std::shared_ptr<ServiceProxy> m_service_proxy;
+            const std::string m_profile_name;
+            const std::string m_report_name;
+            const int m_timeout;
```
avilcheslopez, line 70: Nit: should these be uppercase?

cmcantalupo: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es9-avoid-all_caps-names

I have never been a fan of constant members being all caps, although I am not in the majority wrt contributors to the project.  Consistency is important, but there are a number of arguments against all caps members.

avilcheslopez (approved):

bgeltz (approved):

src/ApplicationRecordLog.hpp
```diff
@@ -183,22 +166,28 @@ namespace geopm
             static size_t max_region(void);
         protected:
             ApplicationRecordLog() = default;
-            static constexpr size_t M_LAYOUT_SIZE = 49192;
+            static constexpr size_t M_LAYOUT_SIZE = 57384;
```
avilcheslopez, line 169: It's out of scope for this review, but still wanted to ask: why does the interface contain implementation details? Is it due to the static methods just above? If so, do those need to be static?

src/ApplicationSampler.hpp
```diff
@@ -119,37 +103,19 @@ namespace geopm
...
             /// @brief Connect with an application using a key
             ///
             /// Called by the Controller to set up all channels of
             /// communication with the application using the provided
-            /// string as a key.  This string corresponds to the
-            /// default shared memory key specified by the system,
-            /// used by the application that is connected.
-            /// The default shared memory key consists of the prefix
-            /// "/geopm-shm-" concatenated with the euid.
+            /// string as a key.
```
avilcheslopez, line 110: This code documentation corresponded to the `connect()` method that was deleted. The new connect method and other new methods need are missing their code documentation.

src/ApplicationSampler.cpp
```diff
@@ -220,6 +201,83 @@ namespace geopm
                                          proc_it.short_regions.begin(),
                                          proc_it.short_regions.end());
         }
+        update_start();
+        m_status->update_cache();
+        double time_delta;
+        if (m_is_first_update) {
+            for (int cpu_idx = 0; cpu_idx != m_num_cpu; ++cpu_idx) {
```
avilcheslopez, line 208:
```suggestion
            for (int cpu_idx = 0; cpu_idx < m_num_cpu; ++cpu_idx) {
```

src/ApplicationSampler.cpp
```diff
@@ -220,6 +201,83 @@ namespace geopm
...
+                m_hint_last[cpu_idx] = cpu_hint(cpu_idx);
+            }
+        }
+        else {
+            time_delta = geopm_time_diff(&m_update_time, &curr_time);
+            for (int cpu_idx = 0; cpu_idx != m_num_cpu; ++cpu_idx) {
```
avilcheslopez, line 214:
```suggestion
            for (int cpu_idx = 0; cpu_idx < m_num_cpu; ++cpu_idx) {
```

src/ApplicationSampler.cpp
```diff
@@ -236,27 +294,28 @@ namespace geopm
...
+        return result;
     }
 
     uint64_t ApplicationSamplerImp::cpu_hint(int cpu_idx) const
     {
+        uint64_t result = GEOPM_REGION_HINT_INACTIVE;
         if (!m_status) {
-            return GEOPM_REGION_HINT_INACTIVE;
+            result = GEOPM_REGION_HINT_INACTIVE;
+        }
```
avilcheslopez, line 314: Could simply delete this `if` statement, if the intention is to return `GEOPM_REGION_HINT_INACTIVE` two cases:
- `!m_status`
- `m_status && !m_is_cpu_active`

src/ApplicationSampler.cpp
```diff
@@ -282,67 +341,61 @@ namespace geopm
...
+        int num_active_cpu = 0;
+        std::fill(m_is_cpu_active.begin(), m_is_cpu_active.end(), false);
+        for (const auto &client_it : m_client_cpu_map) {
+            const std::set<int> &cpu_set = client_it.second;
+            for (int cpu_idx : cpu_set) {
+                m_is_cpu_active[cpu_idx] = true;
+                ++num_active_cpu;
             }
         }
+        for (int cpu_idx = 0; cpu_idx != m_num_cpu; ++cpu_idx) {
```
avilcheslopez, line 392:
```suggestion
        for (int cpu_idx = 0; cpu_idx < m_num_cpu; ++cpu_idx) {
```

src/ApplicationSampler.cpp
```diff
@@ -353,6 +406,58 @@ namespace geopm
...
+    {
+        std::set<int> result;
+        auto client_it = m_client_cpu_map.find(client_pid);
+        if (client_it != m_client_cpu_map.end()) {
+            result = client_it->second;
+        }
+        return result;
+    }
+
+    std::vector<int> ApplicationSamplerImp::client_pids(void) const
```
avilcheslopez, line 431: Does this need to return a vector? Could it return a set, instead?

Issue #2839, opened 2023-02-08
As a user of GPU hardware I would like an agent to target energy savings during periods of low utilization


lhlawson:
**What problem does the story solve? Please describe.**
I'm always frustrated when additional energy savings are lost due to not heavily throttling frequency during long idle periods.  

**Describe the tasks that will be required to complete the story**
Update the GPU-CA to incorporate a 'long idle' detection and take actions to respond by further lowering frequency.  

**Describe alternatives you've considered**
Not targeting idle power reductions, leaving some energy savings on the table.

**Additional context**
This is a rework of the agent proposed in #1907 to fit into the context of the GPU-CA.

PR #2841, opened 2023-02-08
Added GPU-CA On time tracking & idle time reaction

### LINKED ISSUE #2840: Update GPU-CA to provide long idle handling
> estimate 1
Parent [feature/story]: #2839 

The implementation for the GPU Activity agent feature was modified because it did not target periods of long idle.

Done criteria:

- [ ] Documentation changes are required

lhlawson:

- Relates to #2839 story from github issues
- Fixes #2840 change request from github issues.

Provides a report entry for tracking the GPU energy usage when the GPU is above a utilization threshold.
Provides an algorithm update to target long idle periods.

### DIFF
--- a/service/docs/source/geopm_agent_gpu_activity.7.rst
+++ b/service/docs/source/geopm_agent_gpu_activity.7.rst
@@ -51,6 +51,11 @@ extremely short phase behavior a 1 ms polling rate can be used.
 As the 1 ms polling rate is not officially recommended by the DCGM API the 100 ms
 setting should be used by default.
 
+If a ``phi`` value of 0.5 or greater is used and a long idle period, defined as
+10 samples with a ``GPU_UTILIZATION`` of 0, occurs the agent will request the
+minimum frequency for GPU as specified by the ``GPU_CORE_FREQUENCY_MIN_AVAIL``
+signal.
+
 Agent Name
 ----------
 
--- a/service/docs/source/geopm_agent_gpu_activity.7.rst
+++ b/service/docs/source/geopm_agent_gpu_activity.7.rst
@@ -41,6 +41,10 @@ agent towards higher frequencies by increasing the ``Fe`` value.
 In the extreme case (``phi`` of 0) ``Fe`` will be raised to ``Fmax``.  A ``phi`` value greater than
 0.5 biases the agent towards lower frequencies by reducing the ``Fmax`` value.
 In the extreme case (``phi`` of 1.0) ``Fmax`` will be lowered to ``Fe``.
+If a ``phi`` value of 0.5 or greater is used and a long idle period, defined as
+10 samples with a ``GPU_UTILIZATION`` of 0, occurs the agent will request the
+minimum frequency for GPU as specified by the ``GPU_CORE_FREQUENCY_MIN_AVAIL``
+signal.
 
 For NVIDIA based systems the agent should be used with DCGM settings of
 ``DCGM::FIELD_UPDATE_RATE`` = 100 ms, ``DCGM::MAX_STORAGE_TIME`` = 1 s, and ``DCGM::MAX_SAMPLES``
@@ -50,11 +54,6 @@ microsecond range. If the agent is intended to be used with workloads that exhib
 extremely short phase behavior a 1 ms polling rate can be used.
 As the 1 ms polling rate is not officially recommended by the DCGM API the 100 ms
 setting should be used by default.
-
-If a ``phi`` value of 0.5 or greater is used and a long idle period, defined as
-10 samples with a ``GPU_UTILIZATION`` of 0, occurs the agent will request the
-minimum frequency for GPU as specified by the ``GPU_CORE_FREQUENCY_MIN_AVAIL``
-signal.
 
 Agent Name
 ----------
--- a/src/GPUActivityAgent.cpp
+++ b/src/GPUActivityAgent.cpp
@@ -295,8 +295,9 @@ namespace geopm
                 // Frequency selection is based upon the gpu compute activity.
                 // For active regions this means that we scale with the amount of work
                 // being done (such as SM_ACTIVE for NVIDIA GPUs).
-                //
-                // The compute activity is scaled by the GPU Utilization, to help
+                f_request = m_resolved_f_gpu_efficient + m_f_range * gpu_core_activity;
+
+                // If possible, the compute activity is scaled by the GPU Utilization, to help
                 // address the issues that come from workloads that have short phases that are
                 // frequency sensitive.  If a workload has a compute activity of 0.5, and
                 // is resident on the GPU for 50% of cycles (0.5) it is treated as having
@@ -314,13 +315,27 @@ namespace geopm
                 // achieved through tracking the GPU Utilization signal and setting frequency
                 // to a separate idle value (f_idle) during regions where GPU Utilization is
                 // zero (or below some bar).
-                if (!std::isnan(gpu_utilization) &&
-                    gpu_utilization > 0) {
+                if (!std::isnan(gpu_utilization)) {
                     gpu_utilization = std::min(gpu_utilization, 1.0);
-                    f_request = m_resolved_f_gpu_efficient + m_f_range * (gpu_core_activity / gpu_utilization);
+                    if(gpu_utilization > 0) {
+                        f_request = m_resolved_f_gpu_efficient + m_f_range * (gpu_core_activity / gpu_utilization);
+                        m_gpu_idle_timer.at(domain_idx) = M_IDLE_SAMPLE_COUNT;
+                    }
+                    else if (phi >= 0.5) {
+                        if (m_gpu_idle_timer.at(domain_idx) > 0) {
+                            m_gpu_idle_timer.at(domain_idx) -= 1;
+                        }
+                        else {
+                            // If no activity has been observed for M_IDLE_SAMPLE_COUNT samples,
+                            // we assume it is safe to reduce the frequency to a minimum value.
+                            f_request = m_freq_gpu_min;
+                            m_gpu_idle_samples.at(domain_idx) += 1;
+                        }
+                    }
                 }
                 else {
-                    f_request = m_resolved_f_gpu_efficient + m_f_range * gpu_core_activity;
+                    // If activity is observed or is NAN, reset the sample count tracking
+                    m_gpu_idle_timer.at(domain_idx) = M_IDLE_SAMPLE_COUNT;
                 }
 
                 // We're using the activity of the first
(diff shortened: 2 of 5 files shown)


src/GPUActivityAgent.cpp
```diff
@@ -140,6 +146,8 @@ namespace geopm
         m_freq_gpu_min = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
         m_freq_gpu_max = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
 
+        //m_platform_io.write_control("GPU_CORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0, m_freq_gpu_min);
+
```
dannosliwcd:
```suggestion
```

src/GPUActivityAgent.cpp
```diff
@@ -314,6 +322,23 @@ namespace geopm
...
 
+            if (phi >= 0.5) {
+                if (!std::isnan(gpu_utilization) &&
+                    gpu_utilization == 0) {
+                    if (m_gpu_idle_timer.at(domain_idx) > 0) {
+                        m_gpu_idle_timer.at(domain_idx) = m_gpu_idle_timer.at(domain_idx) - 1;
+                    }
+                }
+                else {
+                    m_gpu_idle_timer.at(domain_idx) = 10;
```
dannosliwcd: The width of the idle filter should probably be calculated or at least #defined close to the agent's wait time in case wait time becomes variable in the future

dannosliwcd: That comment assumes this is supposed to approximate a time-based filter. If it is actually meant to be 10 samples regardless of time, then also add a comment explaining why.

lhlawson: Updated to a define of 10 samples regardless of time requested, with a comment explaining why.

src/GPUActivityAgent.cpp
```diff
@@ -138,6 +150,8 @@ namespace geopm
             m_gpu_freq_max_control.push_back(m_control{m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL",
                                                        m_agent_domain,
                                                        domain_idx), NAN});
+            m_gpu_idle_timer.push_back(10);
```
dannosliwcd:
```suggestion
            m_gpu_idle_timer.push_back(IDLE_SAMPLE_COUNT);
```

src/GPUActivityAgent.cpp
```diff
@@ -23,6 +23,14 @@
 #include "Waiter.hpp"
 #include "Environment.hpp"
 
+// IDLE SAMPLE COUNT of 10 is based upon a study of the idle behavior of CORAL-2
+// workloads of interest assuming the default 20ms sample rate (200ms idle).
+// We could use 200ms as the default for the agent, but this does not provide a
+// mechanism for user control of the idle period.  Using a count provides partial
+// user control in that the idleness period is defined by the requested agent
+// control loop time.
```
dannosliwcd: This makes it sound like the goal is actually time-based rather than sample-based. Should `IDLE_SAMPLE_COUNT` be computed from `environment().period(M_WAIT_SEC)` and a `200ms` target instead?

lhlawson: The ideal solution would be to provide a control for both the control loop time (via waiter) and the idle time until the agent takes action.  We don't currently have the agent built with this in mind, so having the idle time be a function of the user input we do have seemed preferable as there could be a workload that has the behavior
... N milliseconds active, 200 ms idle, N milliseconds active ...
We can either:
1. Make it sample based, giving the user some control
2. Make it time based, and risk a pathological case described above
3. Add the idle time/sample count to the policy

Of these I like either 1 or 2

lhlawson:
- [ ] Prior to merge the GPU-CA should be re-run with all review fixes on the target hardware with one workload of interest.

service/docs/source/geopm_agent_gpu_activity.7.rst
```diff
@@ -51,6 +51,11 @@ extremely short phase behavior a 1 ms polling rate can be used.
 As the 1 ms polling rate is not officially recommended by the DCGM API the 100 ms
 setting should be used by default.
 
+If a ``phi`` value of 0.5 or greater is used and a long idle period, defined as
+10 samples with a ``GPU_UTILIZATION`` of 0, occurs the agent will request the
+minimum frequency for GPU as specified by the ``GPU_CORE_FREQUENCY_MIN_AVAIL``
+signal.
+
```
bgeltz: I feel like this paragraph belongs up with the other discussion of the phi policy parameter (one paragraph up).

src/GPUActivityAgent.cpp
```diff
@@ -23,6 +23,14 @@
 #include "Waiter.hpp"
 #include "Environment.hpp"
 
+// IDLE SAMPLE COUNT of 10 is based upon a study of the idle behavior of CORAL-2
```
```suggestion
// IDLE_SAMPLE_COUNT of 10 is based upon a study of the idle behavior of CORAL-2
```

src/GPUActivityAgent.cpp
```diff
@@ -328,6 +342,26 @@ namespace geopm
             f_request = std::min(f_request, m_resolved_f_gpu_max);
             f_request = std::max(f_request, m_resolved_f_gpu_efficient);
 
+            if (phi >= 0.5) {
+                if (!std::isnan(gpu_utilization) &&
+                    gpu_utilization == 0) {
+                    if (m_gpu_idle_timer.at(domain_idx) > 0) {
+                        m_gpu_idle_timer.at(domain_idx) = m_gpu_idle_timer.at(domain_idx) - 1;
```
```suggestion
                        m_gpu_idle_timer.at(domain_idx) -= 1;
```

src/GPUActivityAgent.cpp
```diff
@@ -328,6 +342,26 @@ namespace geopm
...
+                    }
+                }
+                else {
+                    // If no activity has been observed for a number of samples
+                    // IDLE_SAMPLE_COUNT we assume it is safe to reduce the frequency
+                    // to a minimum value.
+                    m_gpu_idle_timer.at(domain_idx) = IDLE_SAMPLE_COUNT;
+                }
+
+                if (m_gpu_idle_timer.at(domain_idx) <= 0) {
```
```suggestion
                // If no activity has been observed for IDLE_SAMPLE_COUNT samples,
                // we assume it is safe to reduce the frequency to a minimum value.
                if (m_gpu_idle_timer.at(domain_idx) <= 0) {
```

src/GPUActivityAgent.cpp
```diff
@@ -328,6 +342,26 @@ namespace geopm
...
+                if (!std::isnan(gpu_utilization) &&
+                    gpu_utilization == 0) {
+                    if (m_gpu_idle_timer.at(domain_idx) > 0) {
+                        m_gpu_idle_timer.at(domain_idx) = m_gpu_idle_timer.at(domain_idx) - 1;
+                    }
+                }
+                else {
+                    // If no activity has been observed for a number of samples
+                    // IDLE_SAMPLE_COUNT we assume it is safe to reduce the frequency
+                    // to a minimum value.
```
```suggestion
                    // If activity is observed or is NAN, reset the IDLE_SAMPLE_COUNT tracking.
```

src/GPUActivityAgent.cpp
```diff
@@ -328,6 +342,26 @@ namespace geopm
...
+                    m_gpu_idle_timer.at(domain_idx) = IDLE_SAMPLE_COUNT;
+                }
+
+                if (m_gpu_idle_timer.at(domain_idx) <= 0) {
+                    f_request = m_freq_gpu_min;
```
bgeltz: I think all of the code in this method that has anything to do with calculating f_request could be refactored into it's own helper method.  I started thinking about this because I'm seeing repeated code already:
```
                if (!std::isnan(gpu_utilization) &&
                    gpu_utilization == 0) {                 
```

I think this new code can be combined into the logic that starts on line 318 (new code), but further I think all of that can have an extract method refactor done on it to clean up ```adjust_platform()``` considerably, i.e. if the idle tracking time has expired and we're about to set the freq to min, we don't need to do any of the other frequency calculation above.

src/GPUActivityAgent.cpp
bgeltz: Why a ```#define``` instead of a const like M_CPU_ACTIVITY_CUTOFF and set in initializer list?

If we expect this is something the user would modify, it should be a policy parameter.

src/GPUActivityAgent.cpp
```diff
@@ -338,14 +372,27 @@ namespace geopm
...
+
+                    // GPU on time tracking
+                    if (m_time.value > m_prev_time) {
+                        m_gpu_on_time.at(domain_idx) += m_time.value - m_prev_time;
+                    }
+
+                    // TODO: handle roll-over more gracefully than dropping a sample
+                    if (m_gpu_energy.at(domain_idx).value > m_prev_gpu_energy.at(domain_idx)) {
+                        m_gpu_on_energy.at(domain_idx) += m_gpu_energy.at(domain_idx).value - m_prev_gpu_energy.at(domain_idx);
+                    }
```
bgeltz, line 386: I'm pretty surprised that this is necessary here and not handled by the IOGroup.
AFAICT, the GPU_ENERGY signal and the CPU_ENERGY signal both have this same problem.  The agents should not have to track this, and the IOGroups should be updated to deal with this.

The code in this agent could then be simplified, i.e. you don't need m_prev_gpu_energy at all if this was handled by the IOGroup.

bgeltz (changes requested):
Integration test(s)?

test/GPUActivityAgentTest.cpp
```diff
@@ -288,7 +288,197 @@ TEST_F(GPUActivityAgentTest, adjust_platform_signal_out_of_bounds_low)
...
+                                                                          {"Agent Idle Time (estimate in seconds) Required to Request Minimum Frequency", "0"},
+                                                                          {"GPU Chip 0 Idle Agent Actions", "1"}};
+    std::vector<std::pair<std::string, std::string> > report_header = m_agent->report_host();
+
+    EXPECT_EQ(expected_header.size(), report_header.size());
+    for (long unsigned int i = 0; i < expected_header.size(); ++i) {
+        EXPECT_EQ(expected_header.at(i).first, report_header.at(i).first);
+        if (expected_header.at(i).first != "Agent Domain") {
+            EXPECT_EQ(std::stod(expected_header.at(i).second), std::stod(report_header.at(i).second));
+        };
```
bgeltz:
```suggestion
        }
```

test/GPUActivityAgentTest.cpp
```diff
@@ -288,7 +288,197 @@ TEST_F(GPUActivityAgentTest, adjust_platform_signal_out_of_bounds_low)
...
+                                                                          {"Agent Idle Time (estimate in seconds) Required to Request Minimum Frequency", "0"},
+                                                                          {"GPU Chip 0 Idle Agent Actions", "0"}};
+    std::vector<std::pair<std::string, std::string> > report_header = m_agent->report_host();
+
+    EXPECT_EQ(expected_header.size(), report_header.size());
+    for (long unsigned int i = 0; i < expected_header.size(); ++i) {
+        EXPECT_EQ(expected_header.at(i).first, report_header.at(i).first);
+        if (expected_header.at(i).first != "Agent Domain") {
+            EXPECT_EQ(std::stod(expected_header.at(i).second), std::stod(report_header.at(i).second));
+        };
```
bgeltz:
```suggestion
        }
```

bgeltz: Also have a look at this, it may simplify this check:
https://stackoverflow.com/a/12340578

test/GPUActivityAgentTest.cpp
```diff
@@ -288,7 +288,197 @@ TEST_F(GPUActivityAgentTest, adjust_platform_signal_out_of_bounds_low)
...
```
bgeltz:
```suggestion
        }
```

bgeltz (changes requested):

src/GPUActivityAgent.cpp
lhlawson: I agree this likely exists for CPU_ENERGY as well, but I have confirmed that the NVML and L0 IOGroups report the monotonic energy provided by the API (translated to SI units).  
We could add a "GPU_ENERGY_DELTA" signal (final naming TBD) using the geopm DifferenceSignal class that provides the difference in the last sample and current sample and handles the rollover.

PR #2845, opened 2023-02-11
Use service API to get topo cache

### LINKED ISSUE #2832: The `TopoGetCache()` GEOPM service is not used by the `PlatformTopo` class which reads a file in `/tmp` instead
> estimate 1
Relates to #2800 

Done criteria:

- [ ] ~Integration tests that cover the feature run and pass~
- [x] Documentation changes are not required since this change does
- [x] Add issue about tracking class local exceptions through inheritance rather than through error message strings which may change.
- [x] Create a "legacy-dev" branch prior to merging this PR since it will be the first time we require the GEOPM service to get basic functionality 
- [x] Fix up missing doxygen and unit tests as specified in comments.

cmcantalupo:
- Resolves #2832

### DIFF
--- a/service/src/PlatformTopo.cpp
+++ b/service/src/PlatformTopo.cpp
@@ -30,6 +30,7 @@
 #include "geopm/Exception.hpp"
 #include "geopm/Helper.hpp"
 #include "GPUTopo.hpp"
+#include "ServiceProxy.hpp"
 
 
 int geopm_read_cpuid(void)
@@ -110,13 +111,29 @@ namespace geopm
     }
 
     PlatformTopoImp::PlatformTopoImp()
-        : PlatformTopoImp("")
+        : PlatformTopoImp("", try_service_proxy())
     {
 
     }
 
-    PlatformTopoImp::PlatformTopoImp(const std::string &test_cache_file_name)
+    std::unique_ptr<ServiceProxy> PlatformTopoImp::try_service_proxy(void)
+    {
+        std::unique_ptr<ServiceProxy> result;
+        try {
+            result = ServiceProxy::make_unique();
+        }
+        catch (const Exception &ex) {
+            if (std::string(ex.what()).find("Failed to open system bus") == std::string::npos) {
+                throw;
+            }
+        }
+        return result;
+    }
+
+    PlatformTopoImp::PlatformTopoImp(const std::string &test_cache_file_name,
+                                     std::shared_ptr<ServiceProxy> service_proxy)
         : M_TEST_CACHE_FILE_NAME(test_cache_file_name)
+        , m_service_proxy(service_proxy)
     {
         std::map<std::string, std::string> lscpu_map;
         lscpu(lscpu_map);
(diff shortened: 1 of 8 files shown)


service/src/PlatformTopo.cpp
```diff
@@ -110,13 +111,29 @@ namespace geopm
...
 
-    PlatformTopoImp::PlatformTopoImp(const std::string &test_cache_file_name)
+    std::unique_ptr<ServiceProxy> PlatformTopoImp::try_service_proxy(void)
+    {
+        std::unique_ptr<ServiceProxy> result;
+        try {
+            result = ServiceProxy::make_unique();
+        }
+        catch (const Exception &ex) {
+            if (std::string(ex.what()).find("Failed to open system bus") == std::string::npos) {
```
dannosliwcd, line 126: Is there any way to make this and the other string comparison resilient to future message changes? E.g., either extract those strings to a constant used here and wherever it is generated, or add some kind of test case that ensures the throwing scenario matches this catching scenario.

avilcheslopez: To add more ideas, I like the approach of having different exceptions for different types of errors (i.e. the exception itself you're catching is an indicator of what went wrong). I realize this is not the approach currently being used in the codebase, however, I do see geopm::Exception has an error code. Could this be leveraged to achieve a similar effect? In general, I would suggest leveraging any metadata that the exception can carry to provide information about the type of error other than the error message.

Minor separate note: in the case where the code proceeds, I would recommend logging a warning.

service/src/PlatformTopo.cpp
```diff
@@ -693,21 +710,33 @@ namespace geopm
...
             }
-            else {
-                create_cache(M_CACHE_FILE_NAME);
-                result = geopm::read_file(M_CACHE_FILE_NAME);
+            catch (const Exception &ex) {
+                std::string err_msg = ex.what();
+                std::string key = "Failed to call sd-bus function";
+                if (err_msg.find(key) == std::string::npos) {
+                    throw;
+                }
```
avilcheslopez, line 728: Similar comments as before apply here as well

service/src/ServiceProxy.hpp
```diff
@@ -143,6 +143,7 @@ namespace geopm
                                                 int domain,
                                                 int domain_idx,
                                                 double setting) = 0;
+            virtual std::string topo_get_cache(void) = 0;
```
avilcheslopez, line 150: Minor note: missing code documentation. The method might be self-explanatory, but figured it was worth mentioning considering all other methods have code documentation, and in case there were any helpful notes worth mentioning.

service/src/ServiceProxy.cpp
```diff
@@ -157,6 +157,12 @@ namespace geopm
                                  setting);
     }
 
+    std::string ServiceProxyImp::topo_get_cache(void)
```
avilcheslopez, line 160: I see a unit test was added for the changes to PlatformTopo, however, a test should be added for this new method in ServiceProxy as well.

service/src/PlatformTopo.cpp
bgeltz: Let's take this to an architecture meeting and discuss.

cmcantalupo:
- Add issue about tracking class local exceptions through inheritance rather than through error message strings which may change.
- Create a "legacy-dev" branch prior to merging this PR since it will be the first time we require the GEOPM service to get basic functionality 
- Fix up missing doxygen and unit tests as specified in comments.

service/src/PlatformTopo.cpp
cmcantalupo: https://github.com/geopm/geopm/issues/2965

service/src/PlatformTopo.cpp
cmcantalupo: https://github.com/geopm/geopm/issues/2965

bgeltz (dismissed):

avilcheslopez (approved):

PR #2862, opened 2023-02-18
Add PBS batch script generator to support experiment infrastructure

### LINKED ISSUE #2817: Add PALS generator script for experiments in the integration infrastructure
> estimate 1

Done criteria:
- [ ] genpals.sh script
- [ ] Update existing documentation
- [ ] Verify against previously delivered frequency sweeps scripts

bgeltz:
- Resolves #2817.
- Change geopmlaunch - PALSLauncher to use the list option with cpu-bind.

Example usage:
```
$ GEOPM_SYSTEM_DEFAULT_QUEUE=R1161 ./gen_pbs.sh 1 minife monitor
```

### DIFF
--- a/integration/experiment/gen_pbs.sh
+++ b/integration/experiment/gen_pbs.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#
+#  Copyright (c) 2015 - 2023, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+# GEOPM PBS run script generator
+
+# Check command line args
+if [ $# -eq 3 ]; then
+    NUM_NODES=$1
+    APP=$2
+    EXP_DIR=$3
+    EXP_TYPE=$3
+elif [ $# -eq 4 ]; then
+    NUM_NODES=$1
+    APP=$2
+    EXP_DIR=$3
+    EXP_TYPE=$4
+else
+    echo "Usage:"
+    echo "       $0 NUM_NODES APP EXPERIMENT_DIR [EXPERIMENT_TYPE]"
+    echo
+    echo "  EXPERIMENT_TYPE: Optional, required when the directory name does not match"
+    echo "                   the Python experiment file."
+    echo
+    echo "Examples:"
+    echo "  $ ./gen_pbs.sh 1 nekbone monitor"
+    echo "  $ ./gen_pbs.sh 1 nekbone energy_efficiency power_balancer_energy"
+    echo "  $ GEOPM_SYSTEM_DEFAULT_QUEUE=R1259 ./gen_pbs.sh 1 arithmetic_intensity monitor"
+    echo
+    exit 1
+fi
+
+if [ -f $HOME/.geopmrc ]; then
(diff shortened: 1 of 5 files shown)


avilcheslopez (commented):
I know it's still a draft, and I'm guessing you've already thought about this, but just in case: seeing how there's a lot of common code between this and the `gen_slurm.sh` script, there's opportunity to factor out common code somehow. It seems like the main difference between the two is the file generation portion. This reminds me of design patterns where you have a piece of higher-level code managing the overall algorithm/steps, but you can inject/incorporate some of the details in a flexible way. For example, here you could use something like "function pointers". You could have a common `.sh` file (which the `gen_*.sh` files would include) that provides a function with the overall script generation steps, and the `gen_*.sh` files, which would define a function that knows how to generate the actual script (i.e. it has the knowledge of the particular syntax for the resource manager they're supporting). Then, this function can be passed as parameter to the main function. Anyway, just an idea.

The other note that comes to mind is that we'll have to update the documentation that depends on/mentions `gen_slurm.sh`. This perhaps should be done in a separate PR, but just wanted to mention it so we don't lose track of it.

bgeltz:
> I know it's still a draft, and I'm guessing you've already thought about this, but just in case: seeing how there's a lot of common code between this and the `gen_slurm.sh` script, there's opportunity to factor out common code somehow. It seems like the main difference between the two is the file generation portion. This reminds me of design patterns where you have a piece of higher-level code managing the overall algorithm/steps, but you can inject/incorporate some of the details in a flexible way. For example, here you could use something like "function pointers". You could have a common `.sh` file (which the `gen_*.sh` files would include) that provides a function with the overall script generation steps, and the `gen_*.sh` files, which would define a function that knows how to generate the actual script (i.e. it has the knowledge of the particular syntax for the resource manager they're supporting). Then, this function can be passed as parameter to the main function. Anyway, just an idea.
> 
> The other note that comes to mind is that we'll have to update the documentation that depends on/mentions `gen_slurm.sh`. This perhaps should be done in a separate PR, but just wanted to mention it so we don't lose track of it.

I'm going to hold off on refactoring this to use a common.sh for common functions for now.  I don't want to further complicate this code just yet, and I'm signing up for dealing with the immediate maintenance issues of having 2 scripts.  I like the idea though, and something I want to consider for the future.

Docu updated.

cmcantalupo (approved):
LGTM i'm not sure i understand why the uniq filtering is required.

integration/experiment/gen_pbs.sh
```diff
@@ -0,0 +1,80 @@
...
+cd \${GEOPM_WORKDIR}
+
+${GEOPM_SOURCE}/integration/experiment/${EXP_DIR}/run_${EXP_TYPE}_${APP}.py \\
+    --node-count=${NUM_NODES} \\
+    --output-dir=\${OUTPUT_DIR} \\
+    # end
+
+EOF
+
+uniq ${SCRIPT_NAME} .${SCRIPT_NAME}
```
cmcantalupo, line 78: Why is this needed?

PR #2864, opened 2023-02-22
Fix bug in reporting

### LINKED ISSUE #2863: The controller does not update the reporter every time the ApplicationRecordLog::dump() call is made resulting in lost data
**Describe the bug**
I tried to run with the controller and application starting asynchronously and I expected the report to be correct instead data was missing.

**GEOPM version**
2.0.1

**Expected behavior**
The problem discovered in the ``cloud`` branch exists in the ``dev`` branch as well, but is seen less frequently because the application and controller are started more synchronously.

**Actual behavior**
Time spent by the application was missing from the report, as well as counts for regions.  

**Error messages produced**
Reports could contain negative time for regions, an missing region counts.

**Additional context**
May be related to #2239 but, unclear.

cmcantalupo:
- Always update the reporter object each time the m_application_sampler is updated.
- Failure to do this can result is lost events that are not included in the reporter accounting.
- Fixes #2863

### DIFF
--- a/src/Controller.cpp
+++ b/src/Controller.cpp
@@ -337,6 +337,7 @@ namespace geopm
         m_platform_io.read_batch();
         m_tracer->update(m_trace_sample);
         m_profile_tracer->update(m_application_sampler.get_records());
+        m_reporter->update();
 
         while (!m_application_io->do_shutdown()) {
             step();
@@ -346,6 +347,7 @@ namespace geopm
         m_platform_io.read_batch();
         m_tracer->update(m_trace_sample);
         m_profile_tracer->update(m_application_sampler.get_records());
+        m_reporter->update();
         generate();
         m_platform_io.restore_control();
     }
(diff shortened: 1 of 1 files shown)


src/Controller.cpp
```diff
@@ -337,6 +337,7 @@ namespace geopm
         m_platform_io.read_batch();
         m_tracer->update(m_trace_sample);
         m_profile_tracer->update(m_application_sampler.get_records());
+        m_reporter->update();
```
dannosliwcd: One other difference between these header/footer blocks and the `step()`'s `walk_up()` block:

```
        m_tracer->update(m_trace_sample);
        m_profile_tracer->update(m_application_sampler.get_records());
```
are called before `m_reporter->update()` here, but after the reporter inside `walk_up()`. Does that difference matter?

cmcantalupo: It shouldn't matter, but i can make them consistent.

cmcantalupo: done!

cmcantalupo: Done now...

dannosliwcd (approved):

PR #2867, opened 2023-02-25
Add a Waiter class to support Agent::wait()

### LINKED ISSUE #1065: Add control loop period as parameter to all agents that wait() on time

cmcantalupo:
- Update all agents to use Waiter
- Add GEOPM_PERIOD environment variable that can be used to override agent default value.
- Update unit tests with mock waiter
- Added documentation for GEOPM_PERIOD environment variable and --geopm-period geopmlaunch CLI
- Uses sleep not spin which reduces CPU overhead for controller and reduces overall energy consumption
- Resolves #1065

### DIFF
--- a/service/docs/source/geopmlaunch.1.rst
+++ b/service/docs/source/geopmlaunch.1.rst
@@ -555,6 +555,16 @@ GEOPM Options
                       Disable OMPT detection of OpenMP regions.
                       See the :ref:`INTEGRATION WITH OMPT section of geopm(7)<geopm.7:Integration With OMPT>`
                       for more information about OpenMP region detection.
+--geopm-period  .. _geopm-period option:
+
+               Override the control loop period specified by the Agent.  All
+	       agents have a default control loop period, and this command
+	       line option allows users to set a different value in units of
+	       seconds.  With longer control loop periods, the overhead for
+	       using GEOPM will be reduced, but more interpolation will be
+	       required when aligning the sparsely sampled hardware signals
+	       with the application feedback.  Additionally agent reaction
+	       time is reduced with longer control loop period.
 
 Examples
 --------
--- a/service/docs/source/geopmlaunch.1.rst
+++ b/service/docs/source/geopmlaunch.1.rst
@@ -563,7 +563,7 @@ GEOPM Options
 	       seconds.  With longer control loop periods, the overhead for
 	       using GEOPM will be reduced, but more interpolation will be
 	       required when aligning the sparsely sampled hardware signals
-	       with the application feedback.  Additionally agent reaction
+	       with the application feedback.  Additionally, agent reaction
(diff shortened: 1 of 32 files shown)


lhlawson (commented):
Agent changes LGTM

src/PowerGovernorAgent.hpp
```diff
@@ -8,9 +8,10 @@
 
 #include <vector>
 #include <functional>
+#include <memory>
+#include <memory>
```
dannosliwcd:
```suggestion
#include <memory>
```
duplicated include

src/Environment.cpp
```diff
@@ -228,6 +229,16 @@ namespace geopm
         return lookup("GEOPM_AGENT");
     }
 
+    double EnvironmentImp::period(double default_period) const
+    {
+        double result = default_period;
+        std::string period_str = lookup("GEOPM_PERIOD");
```
cmcantalupo, line 236: Do exception management for stod() throwing on invalid input.  Throw and give a useful message.

src/Waiter.cpp
```diff
@@ -0,0 +1,72 @@
...
+        m_period = period;
+        reset();
+    }
+
+    void SleepWaiter::wait(void)
+    {
+        int err = 0;
+        do {
+            err = clock_nanosleep(CLOCK_REALTIME, TIMER_ABSTIME,
+                                  &(m_time_target.t), nullptr);
```
bgeltz, line 58: AFAICT following MonitorAgent::wait() and the calls to geopm_time_since() and geopm_time(), it's all relying on CLOCK_MONOTONIC_RAW.

cmcantalupo: To use clock_nanosleep with TIMER_ABSTIME we need to use CLOCK_REALTIME.  I don't see any reason that the sample interval could not be driven by realtime even if we use monotonic time elsewhere.

service/docs/source/geopmlaunch.1.rst
```diff
@@ -555,6 +555,16 @@ GEOPM Options
...
+--geopm-period  .. _geopm-period option:
+
+               Override the control loop period specified by the Agent.  All
+	       agents have a default control loop period, and this command
+	       line option allows users to set a different value in units of
+	       seconds.  With longer control loop periods, the overhead for
+	       using GEOPM will be reduced, but more interpolation will be
+	       required when aligning the sparsely sampled hardware signals
+	       with the application feedback.  Additionally agent reaction
+	       time is reduced with longer control loop period.
```
cmcantalupo:
```suggestion
	       with the application feedback.  Additionally, agent reaction
	       time is reduced with longer control loop period.
```

bgeltz (approved):

PR #2885, opened 2023-03-02
Fix sampling issues with the GEOPM endpoint

### LINKED ISSUE #2830: geopm_endpoint_read_sample() does not update the caller's sample array
**Describe the bug**
The geopm endpoint's C wrapper does not copy read samples to the caller's buffer.

In `geopm_endpoint_read_sample(struct geopm_endpoint_c *endpoint, size_t agent_num_sample, double *sample_array, double *sample_age_sec))`:
```
    try {
        std::vector<double> sample(agent_num_sample);
        *sample_age_sec = end->read_sample(sample);     <=== We read to the local vector and do not copy to sample_array
    }
```

**GEOPM version**
af58d71

**Expected behavior**
After calling `geopm_endpoint_read_sample()`, the given `sample_array` contains the most recent agent sample.

**Actual behavior**
`sample_array` is unmodified after calling `geopm_endpoint_read_sample()`.

### LINKED ISSUE #2873: Endpoint does not initialize sample data
**Describe the bug**
I tried to get a sample from the GEOPM endpoint after calling `wait_for_agent_attach()`. I expected to get a sample (even if empty/nan) since the agent is attached, but the endpoint threw an exception instead.

**GEOPM version**
5bae57c

**Expected behavior**
I expect to get a sample (even if empty/nan) after calling `wait_for_agent_attach()`.

**Actual behavior**
The endpoint throws an exception instead.

**Error messages produced**
```
Error: <geopm> Invalid argument: EndpointImpUser::read_sample(): Data read from shmem does not match number of samples.: at src/Endpoint.cpp:139
```

**Additional Context**
The EndpointUser constructor initializes several of the fields held in its shmem. It should also initialize the sample-related fields. E.g.,
```
        data->count = m_num_sample;
        std::fill(data->values, data->values + m_num_sample, NAN);
        geopm_time(&data->timestamp);
```

dannosliwcd:
- Fixes #2830
- geopm_endpoint_read_sample() reads samples to a local vector. The
  caller of that function provides a destination buffer, but this function does
  not populate that destination buffer. This change copies the local
  vector's contents to the caller's destination buffer.
- Fixes #2873
- The GEOPM endpoint was not initializing sample data (values, length,
  and age). Attempts to read a sample from the endpoint between agent
  initialization and an agent's first sample would result in an
  exception from the EndpointImpUser. This change initializes sample
  data, setting NAN for unknown values.

### DIFF
--- a/service/docs/source/geopm_endpoint.3.rst
+++ b/service/docs/source/geopm_endpoint.3.rst
@@ -196,10 +196,49 @@ All functions described on this man page return an error code.  See
 :doc:`geopm_error(3) <geopm_error.3>` for a full description of the error numbers and how
 to convert them to strings.
 
+Example
+-------
+The endpoint interface needs to be opened and attached to an agent. The
+following pseudocode illustrates the order of function calls that would enable
+an endpoint user to interact with an agent after a job has already started
+executing, instead of writing a static policy to a file and waiting for job
+completion to receive feedback.
+
+Error-checking is omitted from this example for brevity. See
+:doc:`geopm_error(3) <geopm_error.3>` for interpretation of the return values
+from these functions.
+
+.. code-block:: c
+
+    char agent_name[128];
+    char profile_name[128];
+    char node_name[128];
+    int num_node;
+    struct geopm_endpoint_c* endpoint;
+    geopm_endpoint_create("my_endpoint", &endpoint);
+    geopm_endpoint_open(endpoint);
+    geopm_endpoint_wait_for_agent_attach(endpoint, 1.0);
+    geopm_endpoint_agent(endpoint, sizeof agent_name, agent_name);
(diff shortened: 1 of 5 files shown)


lhlawson (dismissed):
LGTM Overall.  Any documentation updates required to describe the new/intended behavior?

dannosliwcd:
> Any documentation updates required to describe the new/intended behavior?

This change makes the following part of the documentation true:
> provides the sample telemetry from the endpoint's agent in sample_array 

Maybe the docs could do with an example user of the endpoint. I can either add some descriptions to the docs here, or we can work toward doing that in the form of example scripts in the time-varying power cap experiments.

lhlawson (approved):

cmcantalupo (approved):

PR #2897, opened 2023-03-15
Addition of LevelZero metric support to the LevelZero device pool and IO Group

lhlawson:
- Relates to #2148 feature quest from github issues
- Provides Intel hardware support analogous to the NVIDIA support added in #1917 and #1914

The addition of ZET metric polling capability to the LevelZero IO Group, Device Pool, and API shim.   

General Metric Streamer flowchart implemented follows the LevelZero approach shown [here](https://spec.oneapi.io/level-zero/latest/_images/tools_metric_streamer.png)

Details of change:
- Updated LevelZero.cpp to set ZET environment variables and follow the general time based metric query approach outlined in the [LevelZero Tools API](https://spec.oneapi.io/level-zero/latest/tools/PROG.html#metrics). 
- Added querying for the 'ComputeBasics' group of Intel Hardware.  An additional metric group may be queried but is not supported at this time.  

Additional Tasks:
- [x] Define Default LevelZero Tools report/event rate based on workload performance
- [x] Add Test for signal and control of LevelZero Tools.
- [ ] Later PR: Consider adding LEVELZERO::METRICS:XVE_*:AGGREGATION.  Currently all signals are averaged over the number of reports received, but this could be modified to be AVG, LAST, MIN, MAX, etc
- [x] Test dynamic tracing with geopm controller (read_signal and batch read have been tested, but not with the controller) 
- [x] Add multi-tile support
- [x] Re-implement metric_destroy/teardown mechanism
- [x] Remove geopmread functionality until an appropriate tear-down handling with read_batch and read_signal can occur.
- [ ] Later PR: Add 'check and fail' if the metric group is changed out  from under the user

### DIFF
--- a/libgeopmd/src/LevelZero.cpp
+++ b/libgeopmd/src/LevelZero.cpp
@@ -220,7 +220,7 @@ namespace geopm
             engine_domain_cache(gpu_idx);
             temperature_domain_cache(gpu_idx);
             ras_domain_cache(gpu_idx);
-            metric_group_cache(gpu_idx);
+            metric_group_init(gpu_idx);
         }
     }
 
@@ -551,11 +551,11 @@ namespace geopm
     //      - metric group handle
     //      - sampling period
     //      - data storage for the ComputeBasics metric group
-    void LevelZeroImp::metric_group_cache(unsigned int device_idx) {
+    void LevelZeroImp::metric_group_init(unsigned int device_idx) {
         for (int subdevice_idx = 0;
              subdevice_idx < m_devices.at(device_idx).m_num_subdevice;
              ++subdevice_idx) {
-            // Setup tracking of the caching and initilization steps
+            // Setup tracking of the caching and initialization steps
             m_devices.at(device_idx).subdevice.metric_domain_cached.push_back(false);
             m_devices.at(device_idx).subdevice.metrics_initialized.push_back(false);
 
@@ -603,70 +603,72 @@ namespace geopm
 
             for (unsigned int metric_group_idx = 0; metric_group_idx < num_metric_group;
                  metric_group_idx++) {
-                 zet_metric_group_properties_t metric_group_properties;
-                 ze_result = zetMetricGroupGetProperties(metric_group_handle.at(metric_group_idx),
-                                             &metric_group_properties);
-                 check_ze_result(ze_result,GEOPM_ERROR_RUNTIME,
-                                 "LevelZero::" + std::string(__func__) +
-                                 ": LevelZero Metric Group property acquisition failed",
-                                 __LINE__);
+                zet_metric_group_properties_t metric_group_properties;
+                ze_result = zetMetricGroupGetProperties(metric_group_handle.at(metric_group_idx),
+                                            &metric_group_properties);
+                check_ze_result(ze_result,GEOPM_ERROR_RUNTIME,
+                                "LevelZero::" + std::string(__func__) +
+                                ": LevelZero Metric Group property acquisition failed",
+                                __LINE__);
 
-                 std::string metric_group_name (metric_group_properties.name);
+                std::string metric_group_name (metric_group_properties.name);
 
-                 //Confirm metric groups of interest exist
-                 if (metric_group_properties.samplingType == ZET_METRIC_GROUP_SAMPLING_TYPE_FLAG_TIME_BASED
-                     && metric_group_name == "ComputeBasic") {
+                // Confirm metric groups of interest exist
+                // Eventually the metric group of interest may be configurable,
+                // to start we're using compute basics
+                if (metric_group_properties.samplingType == ZET_METRIC_GROUP_SAMPLING_TYPE_FLAG_TIME_BASED
+                    && metric_group_name == "ComputeBasic") {
 
-                    //cache compute basic metric group
-                    m_devices.at(device_idx).subdevice.metric_group_handle.push_back(metric_group_handle.at(metric_group_idx));
+                   //cache compute basic metric group
+                   m_devices.at(device_idx).subdevice.metric_group_handle.push_back(metric_group_handle.at(metric_group_idx));
 
-                    // could likely use metric_group_properties.metricCount instead
-                    uint32_t num_metric = 0;
-                    ze_result = zetMetricGet(metric_group_handle.at(metric_group_idx), &num_metric, nullptr );
-                    check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
-                                    "LevelZero::" + std::string(__func__) +
-                                    ": LevelZero Metric Count query failed",
-                                    __LINE__);
+                   // could likely use metric_group_properties.metricCount instead
+                   uint32_t num_metric = 0;
+                   ze_result = zetMetricGet(metric_group_handle.at(metric_group_idx), &num_metric, nullptr );
+                   check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                   "LevelZero::" + std::string(__func__) +
+                                   ": LevelZero Metric Count query failed",
+                                   __LINE__);
 
-                    //Cache compute basic number of metrics
-                    m_devices.at(device_idx).subdevice.num_metric.push_back(num_metric);
+                   //Cache compute basic number of metrics
+                   m_devices.at(device_idx).subdevice.num_metric.push_back(num_metric);
 
-                    //Build metric map
-                    for (unsigned int metric_idx = 0; metric_idx < num_metric; metric_idx++)
-                    {
+                   //Build metric map
+                   for (unsigned int metric_idx = 0; metric_idx < num_metric; metric_idx++)
+                   {
 
-                        std::vector<zet_metric_handle_t> metric_handle(num_metric);
-                        ze_result = zetMetricGet(metric_group_handle.at(metric_group_idx),
-                                                 &num_metric, metric_handle.data());
+                       std::vector<zet_metric_handle_t> metric_handle(num_metric);
+                       ze_result = zetMetricGet(metric_group_handle.at(metric_group_idx),
+                                                &num_metric, metric_handle.data());
 
-                        check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
-                                        "LevelZero::" + std::string(__func__) +
-                                        ": LevelZero Metric handle acquisition failed",
-                                        __LINE__);
-                        zet_metric_properties_t metric_properties;
-                        ze_result = zetMetricGetProperties(metric_handle.at(metric_idx), &metric_properties);
+                       check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                       "LevelZero::" + std::string(__func__) +
+                                       ": LevelZero Metric handle acquisition failed",
+                                       __LINE__);
+                       zet_metric_properties_t metric_properties;
+                       ze_result = zetMetricGetProperties(metric_handle.at(metric_idx), &metric_properties);
 
-                        check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
-                                        "LevelZero::" + std::string(__func__) +
-                                        ": LevelZero Metric Property acquisition failed",
-                                        __LINE__);
+                       check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                       "LevelZero::" + std::string(__func__) +
+                                       ": LevelZero Metric Property acquisition failed",
+                                       __LINE__);
 
-                        std::string metric_name (metric_properties.name);
+                       std::string metric_name (metric_properties.name);
 
-                        //m_devices is a std::vector of structs, indexed by GPU
-                        // subdevice is a struct inside of it.  One per GPU
-                        //  m_metric_data is a vector, indexed by CHIP
-                        //    it contains a map of metric name (string) -> report data for N reports (vector<double>)
-                        if (m_devices.at(device_idx).subdevice.m_metric_data.at(subdevice_idx).count(metric_name) != 0) {
-                            throw Exception("LevelZero::" + std::string(__func__) +
-                                            ": Metric group has two metrics with the same name",
-                                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-                        }
-                        m_devices.at(device_idx).subdevice.m_metric_data.at(subdevice_idx)[metric_name] = {};
-                        m_devices.at(device_idx).subdevice.m_metric_data.at(subdevice_idx)["NUM_REPORTS"] = {};
-                    }
-                    // Break out of loop once we've found the group of interest
-                    break;
+                       //m_devices is a std::vector of structs, indexed by GPU
+                       // subdevice is a struct inside of it.  One per GPU
+                       //  m_metric_data is a vector, indexed by CHIP
+                       //    it contains a map of metric name (string) -> report data for N reports (vector<double>)
+                       if (m_devices.at(device_idx).subdevice.m_metric_data.at(subdevice_idx).count(metric_name) != 0) {
+                           throw Exception("LevelZero::" + std::string(__func__) +
+                                           ": Metric group has two metrics with the same name",
+                                           GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                       }
+                       m_devices.at(device_idx).subdevice.m_metric_data.at(subdevice_idx)[metric_name] = {};
+                       m_devices.at(device_idx).subdevice.m_metric_data.at(subdevice_idx)["NUM_REPORTS"] = {};
+                   }
+                   // Break out of loop once we've found the group of interest
+                   break;
                 }
             }
             m_devices.at(device_idx).subdevice.metric_domain_cached.at(subdevice_idx) = true;
@@ -675,6 +677,16 @@ namespace geopm
 
     void LevelZeroImp::metric_destroy(unsigned int l0_device_idx, unsigned int l0_domain_idx)
     {
+        GEOPM_DEBUG_ASSERT(m_devices.at(l0_device_idx).subdevice.metric_domain_cached.at(l0_domain_idx) == true,
+                           "metric caching for GPU " + std::to_string(l0_device_idx) +
+                           ", CHIP " + std::to_string(l0_domain_idx) +
+                           " not completed prior to metric_calc call.");
+
+        GEOPM_DEBUG_ASSERT(m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx) == true,
+                           "metric initialization for GPU " + std::to_string(l0_device_idx) +
+                           ", CHIP " + std::to_string(l0_domain_idx) +
+                           " not completed prior to metric_calc call.");
+
         ze_result_t ze_result;
 
         if (m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx)) {
@@ -714,7 +726,7 @@ namespace geopm
     }
 
 
-    void LevelZeroImp::metric_init(unsigned int l0_device_idx, unsigned int l0_domain_idx)
+    void LevelZeroImp::metric_execute(unsigned int l0_device_idx, unsigned int l0_domain_idx)
     {
         ze_result_t ze_result;
         ze_context_handle_t context = m_devices.at(l0_device_idx).subdevice.context.at(l0_domain_idx);
@@ -878,7 +890,7 @@ namespace geopm
     {
         if (m_devices.at(l0_device_idx).subdevice.metric_domain_cached.at(l0_domain_idx)) {
             if (!m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx)) {
-                metric_init(l0_device_idx, l0_domain_idx);
+                metric_execute(l0_device_idx, l0_domain_idx);
                 m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx) = true;
             }
 
--- a/libgeopmd/src/LevelZero.cpp
+++ b/libgeopmd/src/LevelZero.cpp
@@ -11,8 +11,6 @@
 #include <cstdlib>
 #include <utility>
 
-#include <chrono>
-
 #include "geopm/Exception.hpp"
 #include "geopm/Agg.hpp"
 #include "geopm/Helper.hpp"
@@ -652,7 +650,6 @@ namespace geopm
         }
     }
 
-    //TODO: need metric destory
     void LevelZeroImp::metric_destroy(unsigned int l0_device_idx, unsigned int l0_domain_idx)
     {
 ////        //std::cout << "ZET_DESTROY - gpu " << std::to_string(l0_device_idx) << std::endl;
@@ -760,9 +757,6 @@ namespace geopm
         //////////////////////
         // Convert Raw Data //
         //////////////////////
-        std::chrono::time_point<std::chrono::system_clock> start, end;
-        std::chrono::duration<double> elapsed_seconds;
-
         size_t data_size = 0;
         uint32_t report_count_req = 100;//UINT32_MAX; //100;
         ze_result = zetMetricStreamerReadData(metric_streamer, report_count_req, &data_size, nullptr );
@@ -784,23 +778,14 @@ namespace geopm
         uint32_t num_metric_values = 0;
         uint32_t data_count = 0;
         zet_metric_group_calculation_type_t calculation_type = ZET_METRIC_GROUP_CALCULATION_TYPE_METRIC_VALUES;
-        //ze_result = zetMetricGroupCalculateMultipleMetricValuesExp(m_devices.at(l0_device_idx).metric_group_handle, calculation_type, data_size, data.data(), &data_count, &num_metric_values, nullptr, nullptr);
         ze_result = zetMetricGroupCalculateMultipleMetricValuesExp(m_devices.at(l0_device_idx).subdevice.metric_group_handle.at(l0_domain_idx), calculation_type, data_size, data.data(), &data_count, &num_metric_values, nullptr, nullptr);
-
-        end = std::chrono::system_clock::now();
-        elapsed_seconds = end - start;
-//        std::cout << "\t\tgpu " << std::to_string(l0_device_idx) << " pre-calculate time: " << elapsed_seconds.count() << "s" << std::endl;
         check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
                         "LevelZero::" + std::string(__func__) +
                         ": LevelZero Metric group calculate metric values to find num metrics failed",
                         __LINE__);
-//        std::cout << "\t\tNum metric values from calculate metric values: " << std::to_string(num_metric_values) << std::endl;
-//        std::cout << "\t\tNum data count from calculate metric values: " << std::to_string(data_count) << std::endl;
 
-        start = std::chrono::system_clock::now();
         std::vector<uint32_t> metric_count(data_count);
         std::vector<zet_typed_value_t> metric_values(num_metric_values);
-        //ze_result = zetMetricGroupCalculateMultipleMetricValuesExp(m_devices.at(l0_device_idx).metric_group_handle, calculation_type, data_size, data.data(), &data_count, &num_metric_values, metric_count.data(), metric_values.data());
         ze_result = zetMetricGroupCalculateMultipleMetricValuesExp(m_devices.at(l0_device_idx).subdevice.metric_group_handle.at(l0_domain_idx), calculation_type, data_size, data.data(), &data_count, &num_metric_values, metric_count.data(), metric_values.data());
 //        check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
 //                        "LevelZero::" + std::string(__func__) +
@@ -809,30 +794,12 @@ namespace geopm
 //        //std::cout << "\tmetric group calculate metric values: " << std::to_string(num_metric_values) << std::endl;
 //        ze_result = ZE_RESULT_ERROR_UNKNOWN;
 
-        end = std::chrono::system_clock::now();
-        elapsed_seconds = end - start;
-//        std::cout << "\t\tgpu " << std::to_string(l0_device_idx) << " calculate time: " << elapsed_seconds.count() << "s" << std::endl;
-
-        //uint32_t num_metric = m_devices.at(l0_device_idx).num_metric;
         uint32_t num_metric = m_devices.at(l0_device_idx).subdevice.num_metric.at(l0_domain_idx);
-        //unsigned int num_reports = num_metric_values/num_metric;
-        //std::cout << "\tgpu " << std::to_string(l0_device_idx) << " num reports: " << std::to_string(num_reports) <<
-                  //std::endl;
-
         if (ze_result == ZE_RESULT_SUCCESS) {
-            start = std::chrono::system_clock::now();
-
-
             unsigned int num_reports = num_metric_values / num_metric;
 
             for (unsigned int metric_idx = 0; metric_idx < num_metric; metric_idx++)
             {
-                std::chrono::time_point<std::chrono::system_clock> start2, end2;
-                std::chrono::duration<double> elapsed_seconds;
-                start2 = std::chrono::system_clock::now();
-////
-
-
                 //TODO: It is possible that simply parsing all the metrics is
                 //      faster than the additional API calls to check the metric
                 //      name.  This should be studied
@@ -853,9 +820,6 @@ namespace geopm
                                 __LINE__);
 
                 std::string metric_name (metric_properties.name);
-                end2 = std::chrono::system_clock::now();
-                elapsed_seconds = end2 - start2;
-                //std::cout << "\t\tgpu " << std::to_string(l0_device_idx) << " name find time: " << elapsed_seconds.count() << "s" << std::endl;
 
                 //TODO: check timing impact of only processing metrics supported by IOGroup
                 if(metric_name.compare("XVE_ACTIVE") == 0 ||
@@ -900,40 +864,21 @@ namespace geopm
             m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).at("XVE_ACTIVE") = {};
             m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).at("XVE_STALL") = {};
         }
-        end = std::chrono::system_clock::now();
-        elapsed_seconds = end - start;
-//        std::cout << "\t\tgpu " << std::to_string(l0_device_idx) << " process time: " << elapsed_seconds.count() << "s" << std::endl;
     }
 
     void LevelZeroImp::metric_read(unsigned int l0_device_idx, unsigned int l0_domain_idx)
     {
         if (m_devices.at(l0_device_idx).subdevice.metric_domain_cached.at(l0_domain_idx)) {
             if (!m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx)) {
-                std::chrono::time_point<std::chrono::system_clock> start, end;
-                std::chrono::duration<double> elapsed_seconds;
-                start = std::chrono::system_clock::now();
-
                 metric_init(l0_device_idx, l0_domain_idx);
                 m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx) = true;
-
-                end = std::chrono::system_clock::now();
-                elapsed_seconds = end - start;
-//                std::cout << "\tgpu " << std::to_string(l0_device_idx) << " metric_init time: " << elapsed_seconds.count() << "s" << std::endl;
             }
 
             ze_result_t ze_host_result = zeEventHostSynchronize(m_devices.at(l0_device_idx).subdevice.event.at(l0_domain_idx), 0);
 
             if (ze_host_result != ZE_RESULT_NOT_READY) {
-                std::chrono::time_point<std::chrono::system_clock> start, end;
-                std::chrono::duration<double> elapsed_seconds;
-                start = std::chrono::system_clock::now();
-
                 metric_calc(l0_device_idx, l0_domain_idx,
                             m_devices.at(l0_device_idx).subdevice.metric_streamer.at(l0_domain_idx));
-
-                end = std::chrono::system_clock::now();
-                elapsed_seconds = end - start;
-//                std::cout << "\tgpu " << std::to_string(l0_device_idx) << " metric_calc time: " << elapsed_seconds.count() << "s" << std::endl;
             }
             else {
                 metric_read(l0_device_idx, l0_domain_idx); //TODO: possible infinite loop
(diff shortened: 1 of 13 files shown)


service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                                                    std::string metric_name) const
+    {
+        std::vector<double> result = {};
+        if (!m_devices.at(l0_device_idx).subdevice.metric_domain_cached.at(l0_domain_idx)) {
+            throw Exception("LevelZero::" + std::string(__func__) +
+                            ": Metric groups not cached" ,
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+
+        if(m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).count(metric_name) == 0) {
```
avilcheslopez: nit:
```suggestion
        if (m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).count(metric_name) == 0) {
```

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+            }
+
+            ze_result_t ze_host_result = zeEventHostSynchronize(m_devices.at(l0_device_idx).subdevice.event.at(l0_domain_idx), 0);
+
+            if (ze_host_result != ZE_RESULT_NOT_READY) {
+                metric_calc(l0_device_idx, l0_domain_idx,
+                            m_devices.at(l0_device_idx).subdevice.metric_streamer.at(l0_domain_idx));
+            }
+            else {
+                metric_read(l0_device_idx, l0_domain_idx); //TODO: possible infinite loop
```
avilcheslopez: :o 
Please address this when moving out of draft

service/src/LevelZero.hpp
```diff
@@ -249,8 +249,23 @@ namespace geopm
...
+                                                      unsigned int l0_domain_idx,
+                                                      std::string metric_name) const = 0;
+            virtual uint32_t metric_update_rate(unsigned int l0_device_idx) const = 0;
+
+            virtual void metric_read(unsigned int l0_device_idx,
+                                     unsigned int l0_domain_idx) = 0;
+            virtual void metric_init(unsigned int l0_device_idx,
+                                     unsigned int l0_domain_idx) = 0;
+            virtual void metric_destroy(unsigned int l0_device_idx,
+                                        unsigned int l0_domain_idx) = 0;
```
avilcheslopez: Are these methods really needed? I see the init is being handled automatically internally, which means we can do without temporal couplings. I do see the code is still in draft form as the `metric_destroy()` implementation is commented out. However, if it is needed, I'm thinking it could be managed similarly how the init is being done.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
             }
         }
     }
+    void LevelZeroImp::metric_group_cache(unsigned int device_idx) {
+        for (int subdevice_idx = 0;
+         subdevice_idx < m_devices.at(device_idx).m_num_subdevice;
+         ++subdevice_idx) {
```
avilcheslopez: nit: please fix indentation

service/src/LevelZeroImp.hpp
```diff
@@ -114,9 +128,25 @@ namespace geopm
...
+                std::vector<uint32_t> num_metric;
+                std::vector<uint32_t> num_reports;
+                std::vector<bool> metric_domain_cached;
+                std::vector<ze_event_pool_handle_t> event_pool;
+                std::vector<ze_event_handle_t> event; //TODO: rename metric_notification_event?
+                std::vector<zet_metric_streamer_handle_t> metric_streamer;
+                std::vector<zet_metric_group_handle_t> metric_group_handle; //compute basic only
+
+                // required for L0 metric result tracking
+                mutable std::vector<std::map<std::string, std::vector<double>>> m_metric_data;
```
avilcheslopez: Following the convention I'm seeing in the rest of our codebase, we're not prefixing properties in these inner structs with "m_".

service/src/LevelZeroImp.hpp
```diff
@@ -114,9 +128,25 @@ namespace geopm
...
+                mutable std::vector<bool> metrics_initialized;
             };
 
             struct m_device_info_s {
+                ze_driver_handle_t driver;
                 zes_device_handle_t device_handle;
                 ze_device_properties_t property;
                 uint32_t m_num_subdevice;
```
avilcheslopez: Same comment as above regarding "m_" prefix. I know it's not part of this review, but it would be something very simple to make the code consistent.

service/src/LevelZeroImp.hpp
```diff
@@ -114,9 +128,25 @@ namespace geopm
...
 
+                //ZE Context used for ZET data collection
+                std::vector<ze_context_handle_t> context;
+
+                // required for L0 metric querying
+                std::vector<uint32_t> num_metric;
+                std::vector<uint32_t> num_reports;
+                std::vector<bool> metric_domain_cached;
+                std::vector<ze_event_pool_handle_t> event_pool;
+                std::vector<ze_event_handle_t> event; //TODO: rename metric_notification_event?
```
avilcheslopez: Sure

service/src/LevelZero.cpp
```diff
@@ -19,11 +19,13 @@
 static void __attribute__((constructor)) geopm_levelzero_init(void)
 {
     setenv("ZES_ENABLE_SYSMAN", "1", 1);
+//    setenv("ZET_ENABLE_METRICS", "1", 1);
```
dannosliwcd:
```suggestion
```

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
             }
         }
     }
+    void LevelZeroImp::metric_group_cache(unsigned int device_idx) {
+        for (int subdevice_idx = 0;
+         subdevice_idx < m_devices.at(device_idx).m_num_subdevice;
+         ++subdevice_idx) {
+            //assume false
```
dannosliwcd: Seems to just be truth rather than assumption at this point, since it isn't cached or initialized yet

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                                              &num_metric_group, nullptr);
+
+                check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                "LevelZero::" + std::string(__func__) +
+                                ": LevelZero Metric Group enumeration failed.",
+                                 __LINE__);
+
+                std::vector<zet_metric_group_handle_t> metric_group_handle(num_metric_group);
+
+                //TODO: save the relevant per GPU pointers
```
dannosliwcd: Which pointers?

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                ze_result = zetMetricGroupGet(m_devices.at(device_idx).subdevice_handle.at(subdevice_idx),
+                                              &num_metric_group, metric_group_handle.data());
+
+                check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                "LevelZero::" + std::string(__func__) +
+                                ": LevelZero Metric Group handle acquisition failed",
+                                __LINE__);
+
+                //sampling period in nanoseconds
+                m_devices.at(device_idx).metric_sampling_period = 2000000;
```
dannosliwcd: Optionally rename to make sure the units are known everywhere this is used. E.g., `sampling_period_ns`

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+    void LevelZeroImp::metric_group_cache(unsigned int device_idx) {
+        for (int subdevice_idx = 0;
+         subdevice_idx < m_devices.at(device_idx).m_num_subdevice;
+         ++subdevice_idx) {
+            //assume false
+            m_devices.at(device_idx).subdevice.metric_domain_cached.push_back(false);
+            m_devices.at(device_idx).subdevice.metrics_initialized.push_back(false);
+
+            char *zet_enable_metrics = getenv("ZET_ENABLE_METRICS");
+            if (zet_enable_metrics == NULL || strcmp(zet_enable_metrics, "1") != 0) {
```
avilcheslopez: Avoid use of `strcmp()`. Alternatives:
- Convert `char*` to `std::string` and use `std::string`'s `==` operator.
- Use `strncmp()`.

I know you're only comparing one character (and you first check `zet_enable_metrics` is not null), so there is no risk of `strcmp()` going out of bounds, so this comment is more for formality and to prevent future changes to inadvertently introduce an issue.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+#ifdef GEOPM_DEBUG
+                std::cerr << "Warning: <geopm>: ZET_ENABLE_METRICS not set to 1.  Skipping metric caching" <<
+                             " for device " << std::to_string(device_idx) << " subdevice " <<
+                             std::to_string(subdevice_idx) << std::endl;
+#endif
```
avilcheslopez: If `GEOPM_DEBUG` is not defined, this will result in an empty block for the true branch of this `if`. Rework this to avoid that situation. You could change the condition such that you invert the branches and have this warning be printed in the `else` branch.

service/src/LevelZero.cpp
avilcheslopez:
```suggestion
            if (zet_enable_metrics == nullptr || strcmp(zet_enable_metrics, "1") != 0) {
```

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                     ze_result = zetMetricGroupGetProperties(metric_group_handle.at(metric_group_idx),
+                                                 &metric_group_properties);
+                     check_ze_result(ze_result,GEOPM_ERROR_RUNTIME,
+                                     "LevelZero::" + std::string(__func__) +
+                                     ": LevelZero Metric Group property acquisition failed",
+                                     __LINE__);
+
+                     //Confirm metric groups of interest exist
+                     if (metric_group_properties.samplingType == ZET_METRIC_GROUP_SAMPLING_TYPE_FLAG_TIME_BASED
+                         && strcmp(metric_group_properties.name, "ComputeBasic") == 0) {
```
avilcheslopez: Same comment as before regarding `strcmp()`.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
             }
         }
     }
+    void LevelZeroImp::metric_group_cache(unsigned int device_idx) {
```
avilcheslopez: I would recommend you rename this method to something that describes what the method does. After reading the code, it seems like this method is initializing the metric group cache. Is that correct?

I just noticed you're following the same naming pattern as the other methods that perform similar cache initialization. At this point, it might be too much trouble to rename all the methods for consistency. So please keep this in mind for the impending refactor of this code and in general for method naming.

service/src/LevelZeroImp.hpp
```diff
@@ -155,6 +188,10 @@ namespace geopm
 
             std::vector<ze_driver_handle_t> m_levelzero_driver;
             std::vector<m_device_info_s> m_devices;
+
+            void metric_group_cache(unsigned int l0_device_idx);
```
dannosliwcd: Need a comment explaining what this is caching and/or what data or functions are invalid until this is cached. My read-through understanding is:
* Caches to several parts of m_devices. I guess this function needs to be called after initialization and before attempting to request any metric data.
* Caches the names of metrics in the ComputeBasic L0 metric group. 

Maybe rename this to indicate that it is hard-coded to computebasic. Or parameterize the computebasic part.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                            }
+                        }
+                    }
+                }
+                m_devices.at(device_idx).subdevice.metric_domain_cached.at(subdevice_idx) = true;
+            }
+        }
+    }
+
+    void LevelZeroImp::metric_destroy(unsigned int l0_device_idx, unsigned int l0_domain_idx)
```
dannosliwcd: Unused (unless that's just temporary)

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+//        }
+    }
+
+
+    void LevelZeroImp::metric_init(unsigned int l0_device_idx, unsigned int l0_domain_idx)
+    {
+        ze_result_t ze_result;
+        ze_context_handle_t context = m_devices.at(l0_device_idx).subdevice.context.at(l0_domain_idx);
+        ze_result = zetContextActivateMetricGroups(context, m_devices.at(l0_device_idx).subdevice_handle.at(l0_domain_idx),
+                                                   1, &m_devices.at(l0_device_idx).subdevice.metric_group_handle.at(l0_domain_idx));
```
dannosliwcd: Maybe comment 1 since this is initialized for computebasic only

service/src/LevelZero.cpp
avilcheslopez: On a separate note about this method, I feel it's a bit lengthy and requires drilling down into nested blocks, so it's not straightforward to quickly grok what it's doing, so if you can, please break it down a bit.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+
+        ze_event_pool_handle_t event_pool_handle = nullptr;
+        ze_event_pool_desc_t event_pool_desc = {ZE_STRUCTURE_TYPE_EVENT_POOL_DESC, nullptr, 0, 1};
+
+        ze_result = zeEventPoolCreate(context, &event_pool_desc, 1,
```
dannosliwcd: Knowledge check for myself. Docs say 1 here is the number of devices. Is this 1 because we're initializing one GPU at a time?

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                        "LevelZero::" + std::string(__func__) +
+                        ": LevelZero Event Create failed",
+                        __LINE__);
+
+        m_devices.at(l0_device_idx).subdevice.event.push_back(event);
+
+        zet_metric_streamer_desc_t metric_streamer_desc = {
+            ZET_STRUCTURE_TYPE_METRIC_STREAMER_DESC,
+            nullptr,
+            32768,
```
dannosliwcd: Why this number of reports per notification?

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+
+        std::vector<uint32_t> metric_count(data_count);
+        std::vector<zet_typed_value_t> metric_values(num_metric_values);
+        ze_result = zetMetricGroupCalculateMultipleMetricValuesExp(m_devices.at(l0_device_idx).subdevice.metric_group_handle.at(l0_domain_idx), calculation_type, data_size, data.data(), &data_count, &num_metric_values, metric_count.data(), metric_values.data());
+//        check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+//                        "LevelZero::" + std::string(__func__) +
+//                        ": LevelZero Metric group calculate metric values to calculate data failed",
+//                        __LINE__);
+//        //std::cout << "\tmetric group calculate metric values: " << std::to_string(num_metric_values) << std::endl;
+//        ze_result = ZE_RESULT_ERROR_UNKNOWN;
```
dannosliwcd:
```suggestion
```

Looks like you removed this since you added a different check a few lines later

service/src/LevelZero.cpp
avilcheslopez: How did you come up with this number? I do see it's the max + 1 of a 2-byte int type, but why choose that number? Is that the max value for that descriptor's property? If you do actually need the max int value, use `<limits>`.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                        "LevelZero::" + std::string(__func__) +
+                        ": LevelZero Metric Streamer Open failed",
+                        __LINE__);
+
+        m_devices.at(l0_device_idx).subdevice.metric_streamer.push_back(metric_streamer);
+    }
+
+    void LevelZeroImp::metric_calc(unsigned int l0_device_idx, unsigned int l0_domain_idx,
+                                   zet_metric_streamer_handle_t metric_streamer) const
+    {
```
avilcheslopez: Does this method require `metric_init()` and `metric_group_cache()` to already be called? If so, I would recommend adding an assertion (`GEOPM_DEBUG_ASSERT`) to ensure this.

service/src/LevelZero.cpp
dannosliwcd: From reading further down I guess it also means we're sampling one device at a time, since this is the event that read attempts block on while waiting for metric data

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                ze_result = zetMetricGetProperties(metric_handle.at(metric_idx), &metric_properties);
+                check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                "LevelZero::" + std::string(__func__) +
+                                ": LevelZero Metric property acquisition failed",
+                                __LINE__);
+
+                std::string metric_name (metric_properties.name);
+
+                //TODO: check timing impact of only processing metrics supported by IOGroup
+                if(metric_name.compare("XVE_ACTIVE") == 0 ||
```
avilcheslopez: nit:
```suggestion
                if (metric_name.compare("XVE_ACTIVE") == 0 ||
```

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                   metric_name.compare("XVE_STALL") == 0) {
+                    for (unsigned int report_idx = 0; report_idx < num_reports; report_idx++)
```
avilcheslopez: Fix indentation

service/src/LevelZero.cpp
dannosliwcd: Should this have a timeout mechanism? E.g., change it to a loop around line 858 and trigger an error if too many attempts or too much time has passed without the event

service/src/LevelZero.cpp
avilcheslopez: Looking at the documentation, it seems `std::string::compare()` (without specifying the number of characters to compare) has the same vulnerability/risk as `strcmp()`. Please use `operator==()`.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                    {
+                        //This is the actual gathering of the data
+                        zet_typed_value_t data = metric_values.at(report_idx*num_metric + metric_idx);
+                        double data_double = NAN;
+                        switch( data.type )
```
avilcheslopez: nit:
```suggestion
                        switch (data.type)
```

service/src/LevelZero.cpp
dannosliwcd: Might want to add a non-zero timeout to the last arg in zeEventHostSynchronize if going that route

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                            data_double = data.value.fp64;
+                            break;
+                        case ZET_VALUE_TYPE_BOOL8:
+                            data_double = data.value.ui32;
+                            break;
+                        default:
+                            break;
+                        };
+                        //Clear cached values
+                        if(report_idx == 0) {
```
avilcheslopez: nit:
```suggestion
                        if (report_idx == 0) {
```

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                            break;
+                        case ZET_VALUE_TYPE_FLOAT64:
+                            data_double = data.value.fp64;
+                            break;
+                        case ZET_VALUE_TYPE_BOOL8:
+                            data_double = data.value.ui32;
+                            break;
+                        default:
+                            break;
+                        };
```
avilcheslopez: Does the API provide a helper function to do this? If not, consider adding a helper function to do this. If it's not needed anywhere else it would at least reduce this function's size and make it more readable.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                m_devices.at(l0_device_idx).subdevice.metrics_initialized.at(l0_domain_idx) = true;
+            }
+
+            ze_result_t ze_host_result = zeEventHostSynchronize(m_devices.at(l0_device_idx).subdevice.event.at(l0_domain_idx), 0);
+
+            if (ze_host_result != ZE_RESULT_NOT_READY) {
+                metric_calc(l0_device_idx, l0_domain_idx,
+                            m_devices.at(l0_device_idx).subdevice.metric_streamer.at(l0_domain_idx));
+            }
+            else {
```
dannosliwcd: Should probably be something like `if == NOT_READY then retry else if SUCCESS then calc() else error`

service/src/LevelZeroImp.hpp
dannosliwcd: Or docs for the other functions might cover that (e.g., they say they'll throw if not yet cached)

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+
+                std::string metric_name (metric_properties.name);
+
+                //TODO: check timing impact of only processing metrics supported by IOGroup
+                if(metric_name.compare("XVE_ACTIVE") == 0 ||
+                   metric_name.compare("XVE_STALL") == 0) {
+                    for (unsigned int report_idx = 0; report_idx < num_reports; report_idx++)
+                    {
+                        //This is the actual gathering of the data
+                        zet_typed_value_t data = metric_values.at(report_idx*num_metric + metric_idx);
```
avilcheslopez: Preferably, it would be better if a data structure was used that more naturally represented the data being stored. If it's not convenient for whatever reason, please add a comment explaining this formula and how `metric_values` is structured such that this formula correctly indexes into the data item of interest.

I'm a bit nervous that `report_idx` depends on `num_reports` which is `num_metric_values / num_metric`. Does this formula take into account what happens when `num_metric_values` is not divisible by `num_metric`? If needed, please add checks and/or assertions to ensure any preconditions needed for this formula to work.

service/src/LevelZero.cpp
```diff
@@ -523,6 +526,366 @@ namespace geopm
...
+                            m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).at(metric_name) = {};
+                        }
+                        m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).at(metric_name).push_back(data_double);
+                    }
+                }
+            }
+        }
+        else {
+            m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).at("XVE_ACTIVE") = {};
+            m_devices.at(l0_device_idx).subdevice.m_metric_data.at(l0_domain_idx).at("XVE_STALL") = {};
```
avilcheslopez: Consider putting these strings (the metric names) in symbolic constants, as they're used in multiple places (including the IOGroup).

service/src/LevelZeroImp.hpp
avilcheslopez: Q: why does this need the mutable specifier?

service/src/LevelZero.cpp
lhlawson: The recursive call is removed in the most recent version.  In cases where the event is not ready the old data will be provided to the user.
We may want to add a 'collected' timestamp so the user can identify if this happens.

service/src/LevelZero.cpp
lhlawson: Updated to 4 reports (1ms per report target)

service/src/LevelZero.cpp
lhlawson: We're targeting 1 event in each event pool, and 1 sub-device per event.

service/src/LevelZeroImp.hpp
lhlawson: Comment added

service/src/LevelZero.cpp
lhlawson: TODO removed, this was a hold over from when the metric group handle, context, etc were not being stored in the tracking struct.

service/src/LevelZero.cpp
lhlawson: Re-enabled ZET by default.

service/src/LevelZero.cpp
lhlawson: Retry logic removed in most recent version.

service/src/LevelZero.hpp
lhlawson: These have been removed int he most recent commit, however metric_destroy may be re-added at some point if we determine we want to be able to stop using ZET dynamically as discussed in the arch meeting

service/src/LevelZero.cpp
lhlawson: Removed in the most recent commit, ZET_ENABLE_METRICS is set by default as part of the GEOPM run

service/src/LevelZero.cpp
lhlawson: How does this look:

- metric_group_cache -> metric_group_init 
- metric_init -> metric_execute

My main concern would be confusion of our metric group initialization vs the ZET concept of initializing the metric group.

service/src/LevelZero.cpp
lhlawson: Updated to 4.

service/src/LevelZero.cpp
lhlawson: The API does not, I've moved it to metric_data_convert(...)

service/src/LevelZeroImp.hpp
lhlawson: It doesn't anymore!

service/src/LevelZero.cpp
```diff
@@ -13,17 +13,20 @@
...
 #include "geopm/Agg.hpp"
 #include "geopm/Helper.hpp"
+#include "geopm_debug.hpp"
 
 #include "LevelZeroImp.hpp"
 
 static void __attribute__((constructor)) geopm_levelzero_init(void)
 {
     setenv("ZES_ENABLE_SYSMAN", "1", 1);
+    setenv("ZET_ENABLE_METRICS", "1", 1);
```
bgeltz: Should this be set conditionally somehow?

If it's too late to set this when push_signal() is called with metrics based signals/controls, then I think this has to be done at compile time with a #define.

service/src/LevelZero.hpp
```diff
@@ -249,8 +249,17 @@ namespace geopm
...
+                                                      unsigned int l0_domain_idx,
+                                                      std::string metric_name) const = 0;
+            virtual uint32_t metric_update_rate(unsigned int l0_device_idx) const = 0;
+
+            virtual void metric_read(unsigned int l0_device_idx,
+                                     unsigned int l0_domain_idx) = 0;
     };
 
-    const LevelZero &levelzero();
+    LevelZero &levelzero();
```
bgeltz: Why was this required?

service/src/LevelZero.cpp
```diff
@@ -178,7 +181,18 @@ namespace geopm
             perf_domain_cache(gpu_idx);
             engine_domain_cache(gpu_idx);
             temperature_domain_cache(gpu_idx);
-       }
+            metric_group_init(gpu_idx);
+        }
+    }
+
+    LevelZeroImp::~LevelZeroImp() {
+        for (unsigned int gpu_idx = 0; gpu_idx < m_num_gpu; gpu_idx++) {
```
```suggestion
        for (unsigned int gpu_idx = 0; gpu_idx < m_num_gpu; ++gpu_idx) {
```

Please be consistent with the variable incrementing.

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,425 @@ namespace geopm
...
+            ze_result = zetMetricGroupGet(m_devices.at(device_idx).subdevice_handle.at(subdevice_idx),
+                                          &num_metric_group, metric_group_handle.data());
+
+            check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                            "LevelZero::" + std::string(__func__) +
+                            ": LevelZero Metric Group handle acquisition failed",
+                            __LINE__);
+
+            // set metric_notifcation_event sampling period in nanoseconds
+            m_devices.at(device_idx).metric_sampling_period_ns =   1000000;
```
```suggestion
            m_devices.at(device_idx).metric_sampling_period_ns = 1000000; // 1 ms
```

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,425 @@ namespace geopm
...
+                   check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                                   "LevelZero::" + std::string(__func__) +
+                                   ": LevelZero Metric Count query failed",
+                                   __LINE__);
+
+                   //Cache compute basic number of metrics
+                   m_devices.at(device_idx).subdevice.num_metric.push_back(num_metric);
+
+                   //Build metric map
+                   for (unsigned int metric_idx = 0; metric_idx < num_metric; metric_idx++)
```
```suggestion
                   for (unsigned int metric_idx = 0; metric_idx < num_metric; ++metric_idx)
```

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,425 @@ namespace geopm
...
+                       if (m_devices.at(device_idx).subdevice.metric_data.at(subdevice_idx).count(metric_name) != 0) {
+                           throw Exception("LevelZero::" + std::string(__func__) +
+                                           ": Metric group has two metrics with the same name",
+                                           GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                       }
+                       m_devices.at(device_idx).subdevice.metric_data.at(subdevice_idx)[metric_name] = {};
+                       m_devices.at(device_idx).subdevice.metric_data.at(subdevice_idx)["NUM_REPORTS"] = {};
+                   }
+                   // Break out of loop once we've found the group of interest
+                   break;
```
```suggestion
```
This ```break``` is inside the if() block that starts on line 612.  I don't believe this was intentional.

bgeltz (changes requested):

service/src/LevelZero.hpp
lhlawson: This change was required for the lazy init approach being used to update some level zero metric containers.  After more recent code changes it might not be required.

service/src/LevelZero.cpp
lhlawson: We're setting this at all times, but having the actual usage of ZET (and context running on the GPU) be conditional on the user doing a push signal of a ZET metric and the first read of those metrics.

service/src/LevelZero.cpp
lhlawson: The intent is to break out of the for loop surrounding the if once we fine the ComputeBasics group that allows for ZET_METRIC_GROUP_SAMPLING_TYPE_FLAG_TIME_BASED sampling

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,425 @@ namespace geopm
...
+                       //    it contains a map of metric name (string) -> report data for N reports (vector<double>)
+                       if (m_devices.at(device_idx).subdevice.metric_data.at(subdevice_idx).count(metric_name) != 0) {
+                           throw Exception("LevelZero::" + std::string(__func__) +
+                                           ": Metric group has two metrics with the same name",
+                                           GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+                       }
+                       m_devices.at(device_idx).subdevice.metric_data.at(subdevice_idx)[metric_name] = {};
+                       m_devices.at(device_idx).subdevice.metric_data.at(subdevice_idx)["NUM_REPORTS"] = {};
+                   }
+                   // Break out of loop once we've found the group of interest
```
lhlawson:
```suggestion
                   // Break out of the metric group for loop once we've found the group of interest (ComputeBasic, time based sampling).  
```

service/src/LevelZero.cpp
```diff
@@ -702,167 +737,203 @@ namespace geopm
...
-        m_devices.at(l0_device_idx).subdevice.event.push_back(event);
+        // TODO: nothing guarantees CHIP 0 was called first
+        m_devices.at(l0_device_idx).subdevice.metric_notifcation_event.push_back(metric_notifcation_event);
 
         zet_metric_streamer_desc_t metric_streamer_desc = {
             ZET_STRUCTURE_TYPE_METRIC_STREAMER_DESC,
             nullptr,
-            32768,
-            m_devices.at(l0_device_idx).metric_sampling_period};
+            4, // number of reports to notify on.
```
dannosliwcd: Same question as before: why 4 reports at 1 ms per sample? Is that something to do with our typical 5ms agent loops?

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,426 @@ namespace geopm
...
+                        ": LevelZero Event Create failed",
+                        __LINE__);
+
+        // TODO: nothing guarantees CHIP 0 was called first
+        m_devices.at(l0_device_idx).subdevice.metric_notifcation_event.push_back(metric_notifcation_event);
+
+        zet_metric_streamer_desc_t metric_streamer_desc = {
+            ZET_STRUCTURE_TYPE_METRIC_STREAMER_DESC,
+            nullptr,
+            4, // number of reports to notify on.
```
lhlawson:
```suggestion
            4, // number of reports to notify on.  Targeting 4 reports, 1 per millisecond as that will, generally, 
                // fall within the 5ms control loop and is a good tradeoff in terms of overhead vs visibility within the sampling period.
```

dannosliwcd (dismissed):

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+                        ": LevelZero Metric Streamer Open failed",
+                        __LINE__);
+
+        // TODO: nothing guarantees CHIP 0 was called first
+        m_devices.at(l0_device_idx).subdevice.metric_streamer.push_back(metric_streamer);
+
+        // Allocate memory for future reads
+        size_t data_size = 0;
+        ze_result = zetMetricStreamerReadData(m_devices.at(l0_device_idx).subdevice.metric_streamer.at(l0_domain_idx),
+                                              UINT32_MAX, &data_size, nullptr); //TODO: this value should match the report_count_req in metric_read
```
avilcheslopez: nit: alignment

service/src/LevelZero.cpp
```diff
@@ -178,7 +181,18 @@ namespace geopm
...
-       }
+            metric_group_init(gpu_idx);
+        }
+    }
+
+    LevelZeroImp::~LevelZeroImp() {
+        for (unsigned int gpu_idx = 0; gpu_idx < m_num_gpu; ++gpu_idx) {
+            for (int subdevice_idx = 0;
+             subdevice_idx < m_devices.at(gpu_idx).num_subdevice;
+             ++subdevice_idx) {
```
avilcheslopez: nit: fix indentation

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+            zet_metric_properties_t metric_properties;
+            ze_result = zetMetricGetProperties(metric_handle.at(metric_idx), &metric_properties);
+            check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                            "LevelZero::" + std::string(__func__) +
+                            ": LevelZero Metric property acquisition failed",
+                            __LINE__);
+
+            std::string metric_name (metric_properties.name);
+
+            //TODO: check timing impact of only processing metrics supported by IOGroup
```
avilcheslopez: TODO for this PR?

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+
+            // process metrics
+            zet_metric_properties_t metric_properties;
+            ze_result = zetMetricGetProperties(metric_handle.at(metric_idx), &metric_properties);
+            check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                            "LevelZero::" + std::string(__func__) +
+                            ": LevelZero Metric property acquisition failed",
+                            __LINE__);
+
+            std::string metric_name (metric_properties.name);
```
avilcheslopez: nit:
```suggestion
            std::string metric_name(metric_properties.name);
```

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+
+            //TODO: check timing impact of only processing metrics supported by IOGroup
+            if (metric_name == "XVE_ACTIVE" ||
+                metric_name == "XVE_STALL") {
+                for (unsigned int report_idx = 0; report_idx < num_reports; report_idx++) {
+                    // Each report contains num_metric entries of data.  We need to access
+                    // this metric (metric_idx) from each report ( report_idx * num_metric)
```
avilcheslopez: nit:
```suggestion
                    // this metric (metric_idx) from each report (report_idx * num_metric)
```

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+                         ": ZET metric_calc call retunred a number of metric values "
+                         "that is not evenly divisible by the number of metrics.  This "
+                         "may indicate a ZET report erroor, skipping data processing." << std::endl;
+#endif
+            num_metric = 0;
+        }
+
+        for (unsigned int metric_idx = 0; metric_idx < num_metric; metric_idx++)
+        {
+            //TODO: It is possible that simply parsing all the metrics is
```
avilcheslopez: TODO for this PR?

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+                                      ZE_EVENT_SCOPE_FLAG_HOST, ZE_EVENT_SCOPE_FLAG_HOST};
+
+        ze_event_handle_t metric_notifcation_event = nullptr;
+        ze_result = zeEventCreate(event_pool_handle, &event_desc, &metric_notifcation_event);
+        check_ze_result(ze_result, GEOPM_ERROR_RUNTIME,
+                        "LevelZero::" + std::string(__func__) +
+                        ": LevelZero Event Create failed",
+                        __LINE__);
+
+        // TODO: nothing guarantees CHIP 0 was called first
```
avilcheslopez: :o

service/src/LevelZero.cpp
```diff
@@ -524,6 +538,427 @@ namespace geopm
...
+        m_devices.at(l0_device_idx).subdevice.metric_notifcation_event.push_back(metric_notifcation_event);
+
+        zet_metric_streamer_desc_t metric_streamer_desc = {
+            ZET_STRUCTURE_TYPE_METRIC_STREAMER_DESC,
+            nullptr,
+            4, // number of reports to notify on.  Targeting 4 reports, 1 per millisecond as that will, generally,
```
avilcheslopez: This would be better to place in a symbolic constant, to make it easier to update when needed (as I know you've had to test different values), and make the code easier to read.

avilcheslopez (commented):
I'm still seeing multiple "TODO" comments, which makes it seem like this is still in draft form. Which of those TODOs need to be addressed as part of this PR?

bgeltz:
Closing in favor of #4052

PR #2902, opened 2023-03-27
Create parent directories in system_files.secure_make_dirs()

### LINKED ISSUE #2901: geopmdpy.system_files.secure_make_dirs() does not create parent directories
**Describe the bug**
Using `system_files`'s `secure_make_dirs()` function in `geopmdpy`, I tried to create a path of directories where some of the parent directories in the path didn't exist. As the code documentation indicates, I expected the request to succeed and for all parent directories in the path to be created. Instead, I the request resulted in an error saying "no such file or directory".

**GEOPM version**
\>=2.0.1

**Expected behavior**
According to code documentation, parent directories in the specified path that don't exist should be created as well.

**Actual behavior**
Parent directories are not being created.

**Error messages produced**
When one or more parent directories in the path don't exist, an error saying "no such file or directory" is returned.

avilcheslopez:
Recursively create directories when needed so parent directories in a path that don't exist get created as well.

- Fixes #2901

### DIFF
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -117,7 +117,12 @@ def secure_make_dirs(path, perm_mode=0o700):
             os.mkdir(path, mode=perm_mode)
     # If the path doesn't exist
     else:
-        os.makedirs(path, mode=perm_mode)
+        umask = ~perm_mode & 0o777
+        old_mask = os.umask(umask)
+        try:
+            os.makedirs(path, mode=perm_mode)
+        finally:
+            os.umask(old_mask)
 
 
 def secure_make_file(path, contents):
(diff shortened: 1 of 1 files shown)


service/geopmdpy/system_files.py
```diff
@@ -117,7 +117,7 @@ def secure_make_dirs(path, perm_mode=0o700):
             os.mkdir(path, mode=perm_mode)
     # If the path doesn't exist
     else:
-        os.mkdir(path, mode=perm_mode)
+        os.makedirs(path, mode=perm_mode)
```
bgeltz: In the future, it would be nice to have a test that tries to create a path with an intermediate directory to prove that this all works.

bgeltz (dismissed):

service/geopmdpy/system_files.py
cmcantalupo: There is an issue here. The os.makedirs() call only applies the perm_mode to the leaf directory.  It would be good indeed to have a unit test that shows that the permissions are set correctly on intermediate directories as well.  The recommended solution in the documentation is to set the umask first:

https://docs.python.org/3/library/os.html#os.makedirs

cmcantalupo (approved):

PR #2905, opened 2023-03-29
Add SECURITY.md doc for vulnerability reporting

bgeltz:


cmcantalupo (changes requested):
Should also be added to EXTRA_DIST in geopm/Makefile.am

cmcantalupo (approved):

PR #2907, opened 2023-03-29
Add unit test for secure_make_dirs() to handle nested dirs

### LINKED ISSUE #2906: Improve unit testing for geopmdpy.system_files.secure_make_dirs()
> estimate 1
Relates to: #2901 

The recent changes to address #2901 have updated `secure_make_dirs()` in the following way:
- When the requested path needs to be created, it will create any intermediate directories that don't exist
- In this case, it will set a umask to limit permissions for the intermediate directories that get created

Unit tests are needed to verify this new behavior.

Done criteria:

- [ ] [...]
- [ ] [...]

avilcheslopez:
Ensure secure_make_dirs() can create a requested path when parent directories don't exist.

- Fixes #2906

### DIFF
--- a/service/geopmdpy_test/TestSecureFiles.py
+++ b/service/geopmdpy_test/TestSecureFiles.py
@@ -23,12 +23,15 @@ class TestSecureFiles(unittest.TestCase):
         """
         self._test_name = 'TestSecureFiles'
         self._TEMP_DIR = tempfile.TemporaryDirectory(self._test_name)
+        self._old_umask = os.umask(0)
+        os.umask(self._old_umask)
 
     def tearDown(self):
         """Clean up temporary directory
 
         """
         self._TEMP_DIR.cleanup()
+        os.umask(self._old_umask)
 
     def check_dir_perms(self, path, perm_mode=0o700):
         """Assert that the path points to a file with mode 0o700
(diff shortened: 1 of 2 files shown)


service/geopmdpy_test/TestSecureFiles.py
```diff
@@ -246,6 +249,26 @@ def test_creation_bad_group_owner(self):
...
+        Test secure_make_dirs() properly sets a umask to limit
+        permissions for intermediate directories that need to
+        be created as part of the path.
+        """
+        base_path = f"{self._TEMP_DIR.name}/level1"
+        sess_path = f"{base_path}/level2"
+        os.umask(0o000)
+        perm_mode = 0o711
+        secure_make_dirs(sess_path, perm_mode)
+        self.check_dir_perms(sess_path, perm_mode)
```
bgeltz, line 268:
```suggestion
        self.check_dir_perms(sess_path, perm_mode)
        self.check_dir_perms(base_path, perm_mode)
```
Maybe this is redundant with what you have below?

avilcheslopez: I don't believe so. The check below ensures the base directory doesn't get any of the denied permissions set by the umask `secure_make_dirs()` is supposed to set.

Perhaps I should just mock out `os.umask()` and verify `secure_make_dirs()` calls it with the right parameter.

cmcantalupo (approved):
LGTM

PR #2926, opened 2023-04-13
Add NPB application configs to the integration infrastructure

### LINKED ISSUE #2890: As a user of the integration experiment infrastructure I would like to launch apps that do not utilize the same number of cores across all nodes
Some applications have restrictions on the total number of MPI ranks over which they can decompose their computations. For example, several of the NPB benchmarks require power-of-two or square-number rank counts. Depending on the node count and the experiment platform's count of cores per node, some nodes may contain fewer ranks than other nodes.

Right now, the appconf class only supports defining a count of ranks per node, which the experiment launcher multiplies against node count to determine the total rank count. This results in an incorrect total rank count. Although the application is able to execute in that case, GEOPM does not count epochs on nodes that do not check in to an epoch with every assigned rank.

I would like for appconfs to be able to declare the total rank count needed for an application. Many apps are fine with the current behavior, so it would also be nice if it is possible for appconfs to omit such a declaration (defaulting to the current nodes*ranks_per_node behavior) so that we don't need to modify existing configs.

dannosliwcd:
- Related to #2669: Add appconfs for BT, FT, IS, CG, LU, EP, SP, and MG benchmarks
 - Resolves #2890: Let an appconf define its total rank count


cmcantalupo (commented):
LGTM, just need the build fixed

integration/experiment/launch_util.py
```diff
@@ -70,7 +70,7 @@ def launch_run(agent_conf, app_conf, run_id, output_dir, extra_cli_args,
     bash_path = apps.make_bash(app_conf, run_id, log_path)
     argv.extend([bash_path])
 
-    num_ranks = app_conf.get_rank_per_node() * num_nodes
+    num_ranks = app_conf.get_total_ranks(num_nodes)
```
bgeltz, line 73: This change will break all the apps that do not implement get_total_ranks.  Unclear just now why this was required to change.

bgeltz (changes requested):

integration/apps/nascg/build.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Run helper functions
+clean_source ${DIRNAME}
+get_archive ${ARCHIVE} ${URL}
+unpack_archive ${ARCHIVE}
+setup_source_git ${DIRNAME}
+
+# Build application
+cd ${DIRNAME}/NPB3.4-MPI
+make suite
```
bgeltz, line 25: I'm seeing this error repeatedly when I try to run build.sh on mcfly:
```
mpif90 -c -I/home/bgeltz/build/geopm/include -O3 -fopenmp -I/home/bgeltz/build/geopm/lib/ifx/modules/geopm-x86_64 cg.f90
f951: Fatal Error: Reading module ‘geopm_prof’ at line 1 column 2: Unexpected EOF
compilation terminated.
make[1]: *** [Makefile:17: cg.o] Error 1
```

bgeltz: All these new apps compile fine on borealis.

bgeltz: This is an mpif90 problem on mcfly.  If you want to build there, you need to modify the NPB build to use mpiifort instead of mpif90.  I tested and confirmed that solution. 
 Related thread:
https://community.intel.com/t5/Intel-MPI-Library/Intel-MPI-Fortran-2008-Error/m-p/1360122

Since the preferred solution is to stay with mpif90 here, if we ever need to build on mcfly we'd have to create a build variant for it that applies an extra patch to switch out the compiler.

bgeltz (approved):

cmcantalupo (approved):

PR #2960, opened 2023-05-26
Add PBS hooks for automatic power capping

### LINKED ISSUE #3028: As a cluster administrator I would like to ensure that concurrently executing jobs stay under a total power limit
When I want active jobs to stay under a power limit, I would like if the job scheduler could decide when individual jobs should start for me.

I want jobs to start only when the overlapping executing jobs would not cause my specified limit to be exceeded. I also want the executing jobs to run under a power limit that enforces any assumptions the scheduler made about maximum power demand.

Alternatively, I would like for executing jobs to receive power-capping updates from the cluster. However, this would require regular communication between a cluster power manager and each executing job.

### LINKED ISSUE #3029: As a job submitter I would like my power-capped jobs to experience similar performance across compute nodes
Different compute nodes may react differently to the same power cap.

When I submit a power-capped job, I would like for the nodes assigned to my job to receive non-uniform power caps in a way that reaches closer to uniform performance across the job's nodes, while respecting the total power limit assigned to my job.

dannosliwcd:
- Resolves #3028
- Use desired max slowdown with performance models to set job power caps
- Resolves #3029
- Use job power caps and node performance models to counteract node
  performance variation

### DIFF
--- a/service/integration/open_pbs/README.rst
+++ b/service/integration/open_pbs/README.rst
@@ -1,84 +1,303 @@
-geopm_open_pbs_power_limit(7) -- OpenPBS hook for providing job-level power limits
-==================================================================================
+geopm_open_pbs_power_limit(7) -- OpenPBS hooks for providing job-level power limits
+===================================================================================
 
 Overview
 --------
 
 The GEOPM OpenPBS power limit hook extends OpenPBS's functionality to allow
-the user to request a job-level power limit at the time of job submission. Via
-a command-line option, the user can specify a per-node power limit to be
+the user to request a job-level power limit through a ``geopm-job-power-limit``
+resource at the time of job submission and to enforce cluster-level power
+limits through that mechanism. Via a command-line option, the user can specify
+a per-node power limit through a ``geopm-node-power-limit`` resource to be
 applied to every node that is part of the job. The power limit is applied just
-before the job runs, as part of the job's prologue, and is restored upon job
-completion, as part of the job's epilogue.
+before the job runs as part of the job's prologue, and it is restored upon job
+completion as part of the job's epilogue. The scheduler can be configured
+with constraints over those power limits across a cluster in order to enforce a
+cluster power limit. The GEOPM OpenPBS hook enables automatic configuration of
+job power limits to increase utilization of the available cluster power.
+
+.. image:: https://geopm.github.io/images/geopm_power_resources.svg
+     :width: 600
+     :alt: Block diagram of the relationship between geopm node and job power resources
 
 Save-restore Mechanism
 ----------------------
-
-The hook will keep track of the original power limit setting and restore it at
+The hook keeps track of the original power limit setting and restores it at
 the end of the job (for each node that's part of the job). To achieve this,
 the hook needs to be configured to run during the job prologue and epilogue
 events (see the Installation section for more details).
 
-In the case where a job is interrupted unexpectedly prior to the hook being
-able to restore the power limit setting, the hook will rectify the setting the
-next time it is triggered on the affected nodes. For this, the hook saves the
-original setting to a file prior to applying any changes. This is leveraged by
-the hook during the epilogue for the restore operation and is also used to
-correct the power limit setting if it was not restored as part of the previous
-job. More specifically, if a new power limit is not requested in the next job
-involving the affected nodes (and the save file is present), the hook will
-take the opportunity to restore the power limit setting during the prologue so
+The hook saves the original power limit to a file prior to applying any new
+power limits. The job's epilogue hook leverages the save file for the restore
+operation and in the prologue if it was not restored as part of the previous
+job. The save file is removed when the power limits are restored (regardless of
+whether the prologue or epilogue handles the restore operation).
+
+If a job is interrupted unexpectedly prior to restoring the power limit
+setting, the hook will rectify the power setting the next time the hook
+executes on the affected nodes. More specifically, if a new power limit is not
+requested in the next job involving the affected nodes (and the save file is
+present), the hook will restore the power limit setting during the prologue so
 the new job is not affected by a prior power limit setting.
+
+Automatic Power Limits
+----------------------
+The goal of the automatic power limiter is to provide additional scheduling
+flexibility when the cluster is power constrained. If the cluster power cap is
+sufficiently low, then jobs may not begin executing even when sufficient
+non-power resources are available on idle nodes. At a high level,
+this automatic limiter attempts to set the lowest acceptable power cap to
+each job, based on modeled power-performance relationships and a desired
+slowdown limit.
+
+Automatic power limits are selected through a handler for ``queuejob`` PBS
+events. The ``queuejob`` event handler initially sets the
+``geopm-job-power-limit`` of each submitted job to be the lowest tolerable
+power limit for that job, according to the configured maximum job slowdown and
+the job's selected power-performance model. If the user explicitly requests a
+``geopm-job-power-limit`` or a ``geopm-node-power-limit`` on job submission,
+then automatic power limit selection is not applied.
+
+If both the ``geopm-job-power-limit`` and the ``geopm-node-power-limit`` are
+requested by a user, then the ``queuejob`` hook sets a
+``geopm-job-power-limit`` limit that is equivalent to the greater of the two,
+to conservatively ensure that the user receives at least the amount of
+requested power under either interpretation. For example, in a 2-node job with
+a 200 W node power limit and a 200 W job power limit, the job is assigned a 200
+W node power limit and a 400 W job power limit.
 
 Usage
 -----
+This section describes user interfaces for end users of the cluster to specify
+power limits for their jobs and for system administrators to configure cluster
+power limits.
 
-To request a power limit for a job, the user can provide the
-``geopm-node-power-limit`` resource when queueing a job via ``qsub``. The
-general syntax is as follows:
+Setting Node Power Limits For a Job
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To request a uniform power limit across all nodes in a job, a user can
+provide the ``geopm-node-power-limit`` resource when queueing a job via
+``qsub``. The general syntax is as follows:
 
 ::
 
    qsub [-l geopm-node-power-limit=<power limit (watts)>]
       [<other qsub options>] [- | <script> | -- <executable> [<arguments>]]
 
+Enforcing a Cluster Power Limit Through Job Power Limits
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To request a power limit for the cluster, an administrator can set the total
+watts of node power available across reservations and running jobs associated
+with the PBS server. The syntax is as follows:
+
+::
+
+    qmgr -c 'set server resources_available.geopm-job-power-limit=<power limit (watts)>'
+
+This tells PBS not to start executing a job if the job's assigned power limit
+plus the power limits of other overlapping jobs and reservations in the
+schedule would exceed the assigned cluster power limit.
+
+The ``resources_available`` limit *does not* account for any resources that are
+not tracked in PBS or that are used outside of the context of a running job.
+For example, idle power consumption is not automatically accounted by this
+mechanism.
+
+If no power limit is specified for a job or for its nodes, then the job will
+default to requesting the maximum allowed power. To set the maximum allowed
+power per node:
+
+::
+
+    qmgr -c 'set server resources_available.geopm-max-node-power-limit=<power per node (W)>'
+
+If a cluster power cap is set, then a queued job may be delayed due to waiting
+for a period with sufficient power in the PBS server's available resources. The
+chance of delay may be reduced by requesting a lower power cap for the job. A
+user can limit the maximum tolerable slowdown by specifying the slowdown factor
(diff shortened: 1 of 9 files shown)


service/integration/open_pbs/README.rst
```diff
@@ -30,10 +38,46 @@ involving the affected nodes (and the save file is present), the hook will
...
+according to that job's selected power-performance model. Once the PBS
+scheduler has determined that there are sufficient resources (power and
+otherwise) to execute the job, the ``runjob`` event handler is triggered. That
+event handler opportunistically increases the power limit for that job, up to
+the configured maximum power per node (times the number of nodes in the job),
+bound by the amount of power that is still unallocated in the cluster.
+
+Both the ``queuejob`` and ``runjob`` event handlers reconcile differences
+between the automatically configured ``geopm-job-power-limit`` and the
+user-selectable ``geopm-node-power-limit``. In the even that both limits are
```
dannosliwcd:
```suggestion
user-selectable ``geopm-node-power-limit``. In the event that both limits are
```

service/integration/open_pbs/README.rst
```diff
@@ -42,43 +86,109 @@ general syntax is as follows:
...
 
    qmgr -c "create hook geopm_power_limit"
    qmgr -c "import hook geopm_power_limit application/x-python default geopm_power_limit.py"
-   qmgr -c "set hook geopm_power_limit event='execjob_prologue,execjob_epilogue'"
+   qmgr -c "set hook geopm_power_limit event='execjob_prologue,execjob_epilogue,queuejob,runjob'"
 
-Note how the hook needs to be configured to run in two events: the job
-prologue and the job epilogue.
+Note how the hook needs to be configured to run in three events: job
+prologue, job epilogue, and runjob.
```
dannosliwcd:
```suggestion
Note how the hook needs to be configured to run in four events: job
prologue, job epilogue, queuejob, and runjob.
```

service/integration/open_pbs/README.rst
```diff
@@ -1,20 +1,28 @@
...
+===================================================================================
 
 Overview
 --------
 
 The GEOPM OpenPBS power limit hook extends OpenPBS's functionality to allow
-the user to request a job-level power limit at the time of job submission. Via
-a command-line option, the user can specify a per-node power limit to be
+the user to request a job-level power limit through a ``geopm-job-power-limit``
+resource at the time of job submission, and to enforce cluster-level power
```
asmaalrawi:
```suggestion
resource at the time of job submission and to enforce cluster-level power
```

service/integration/open_pbs/README.rst
```diff
@@ -1,20 +1,28 @@
...
+limits through that mechanism. Via a command-line option, the user can specify
+a per-node power limit through a ``geopm-node-power-limit`` resource to be
 applied to every node that is part of the job. The power limit is applied just
 before the job runs, as part of the job's prologue, and is restored upon job
```
asmaalrawi:
```suggestion
before the job runs as part of the job's prologue, and it is restored upon job
```

service/integration/open_pbs/README.rst
```diff
@@ -1,20 +1,28 @@
...
-completion, as part of the job's epilogue.
+completion, as part of the job's epilogue. The scheduler can be configured
```
asmaalrawi:
```suggestion
completion as part of the job's epilogue. The scheduler can be configured
```

service/integration/open_pbs/README.rst
```diff
@@ -1,20 +1,28 @@
...
+     :alt: Block diagram of the relationship between geopm node and job power resources
 
 Save-restore Mechanism
 ----------------------
-
-The hook will keep track of the original power limit setting and restore it at
+The hook keeps track of the original power limit setting and restores it at
 the end of the job (for each node that's part of the job). To achieve this,
 the hook needs to be configured to run during the job prologue and epilogue
 events (see the Installation section for more details).
```
asmaalrawi, line 28: Below:

`If the job is unexpectedly interrupted before the hook is able to restore the power limit setting, the power limit will be rectified the next time the hook is triggered on affected nodes. To accomplish this, the hook saves the original setting to a file before applying any changes. Typically, the hook will access the stored setting for the restore operation during the epilogue. During the prologue of a new job, if a new power limit is not requested and a save file is present, the hook will restore the saved power limit so that the new job is not affected by prior power limit settings.`

Question: Is the save file deleted following a successful restore? Might be worth stating explicitly if so

service/integration/open_pbs/README.rst
```diff
@@ -30,10 +38,46 @@ involving the affected nodes (and the save file is present), the hook will
...
 
+Automatic Power Limits
+----------------------
+The goal of the automatic power limiter is to provide additional scheduling
+flexibility when the cluster is power constrained. If the cluster power cap is
+sufficiently low, then jobs may not begin executing even when sufficient
+non-power resources are available on idle nodes. The high-level approach of
+this automatic limiter is to initially set the lowest acceptable power cap to
+each job (based on modeled power-performance relationships and a desired
+slowdown limit) until the scheduler offers to allocate resources to a job. When
```
asmaalrawi:
```suggestion
non-power resources are available on idle nodes. At a high level,
this automatic limiter attempts to set the lowest acceptable power cap to
each job (based on modeled power-performance relationships and a desired
slowdown limit) until the scheduler offers to allocate resources to a job. When
```

service/integration/open_pbs/README.rst
```diff
@@ -30,10 +38,46 @@ involving the affected nodes (and the save file is present), the hook will
...
+the scheduler offers to start the job, the automatic limiter opportunistically
+increases power to that job if there is additional unallocated power available
+in the cluster.
+
+Automatic power limits are selected through handlers for ``queuejob`` and
+``runjob`` PBS events. The ``queuejob`` event handler initially sets the
+``geopm-job-power-limit`` of each submitted job to be the lowest tolerable
+power limit for that job, according to the configured maximum job slowdown and
+according to that job's selected power-performance model. Once the PBS
```
asmaalrawi:
```suggestion
the job's selected power-performance model. Once the PBS
```

service/integration/open_pbs/README.rst
```diff
@@ -42,43 +86,109 @@ general syntax is as follows:
...
+configuration file (located at ``$PBS_HOME/sched_priv/sched_config``). That is,
+the line beginning with ``resources:`` should include an entry for
+``geopm-job-power-limit``. The other resources mentioned in this guide do not
+need to be added to that configuration file.
 
-The hook is available in the ``service/integration/open_pbs`` directory
-(``geopm_power_limit.py``) and it can be installed with the following
-commands:
+The hook to enforce node power caps is available in the GEOPM
+``service/integration/open_pbs`` directory (``geopm_power_limit.py``) and it
```
asmaalrawi:
```suggestion
``service/integration/open_pbs`` directory (``geopm_power_limit.py``) and
```

asmaalrawi (commented):
Minor wording nits/suggestions.

service/integration/open_pbs/README.rst
dannosliwcd: It is. I'll add a note.

service/integration/open_pbs/README.rst
```diff
@@ -28,57 +36,250 @@ correct the power limit setting if it was not restored as part of the previous
...
+in the cluster.
+
+Automatic power limits are selected through handlers for ``queuejob`` and
+``runjob`` PBS events. The ``queuejob`` event handler initially sets the
+``geopm-job-power-limit`` of each submitted job to be the lowest tolerable
+power limit for that job, according to the configured maximum job slowdown and
+the job's selected power-performance model. Once the PBS
+scheduler has determined that there are sufficient resources (power and
+otherwise) to execute the job, the ``runjob`` event handler is triggered. That
+event handler opportunistically increases the power limit for that job, up to
```
cmcantalupo: Would the runjob event handler ever reduce the power limit back down to the original lower bound to make space for a new job?

service/integration/open_pbs/README.rst
```diff
@@ -28,57 +36,250 @@ correct the power limit setting if it was not restored as part of the previous
...
+the configured maximum power per node (times the number of nodes in the job),
+bound by the amount of power that is still unallocated in the cluster.
+
+Both the ``queuejob`` and ``runjob`` event handlers reconcile differences
+between the automatically configured ``geopm-job-power-limit`` and the
+user-selectable ``geopm-node-power-limit``. In the event that both limits are
+specified, the greater of the two is applied. For example, in a 2-node job with
```
cmcantalupo: pehaps provide a justification for using the greater of the two rather than the lesser of the two?

service/integration/open_pbs/README.rst
```diff
@@ -28,57 +36,250 @@ correct the power limit setting if it was not restored as part of the previous
...
+across nodes assigned to the job.
+
+Each host or profile model is defined in a *model* attribute that contains
+members for the model coefficients ``x0``, ``A``, ``B``, and ``C``. The root
+level of the configuration file also has a ``max_power`` value that maps to
+$P_{max}$. The following example configuration file defines coefficients for
+job classes named ``jobtype_a`` and ``jobtype_b``, and defines per-host models
+under a profile named ``node-characterization``:
+
+.. code-block::
```
cmcantalupo, line 161: Perhaps we can reference a schema here?

cmcantalupo (commented):
Looks great.  Made a few comments in the documentation portion.  Additionally, we typically post images here:

https://github.com/geopm/geopm.github.io/tree/master/images

and then just reference the URL in the documentation.  This avoids us putting binary files in the man Git repo.

service/integration/open_pbs/README.rst
```diff
@@ -28,57 +36,250 @@ correct the power limit setting if it was not restored as part of the previous
...
+The hook to enforce node power caps is available in the GEOPM
+``service/integration/open_pbs`` directory (``geopm_power_limit.py``) and
+can be installed with the following commands:
 
 ::
 
    qmgr -c "create hook geopm_power_limit"
    qmgr -c "import hook geopm_power_limit application/x-python default geopm_power_limit.py"
-   qmgr -c "set hook geopm_power_limit event='execjob_prologue,execjob_epilogue'"
+   qmgr -c "set hook geopm_power_limit event='execjob_prologue,execjob_epilogue,queuejob'"
```
bgeltz, line 279: runjob seems to be missing

service/integration/open_pbs/generate_coefficients_from_reports.py
```diff
@@ -0,0 +1,224 @@
...
+# 
+#     if args.verbose:
+#         y_train_pred = model.predict(X)
+#         r2score = sklearn.metrics.r2_score(y, y_train_pred)
+#         print(f'{host} training R2 Score: {r2score}', file=sys.stderr)
+# 
+#     A = model.named_steps["linear_regression"].coef_[1 if len(model.named_steps["linear_regression"].coef_) > 1 else 0]
+#     B = model.named_steps["linear_regression"].coef_[0]
+#     C = model.named_steps["linear_regression"].intercept_
+#     return (A, B, C)
```
bgeltz: Remove unused code?

service/integration/open_pbs/geopm_install_pbs_power_limit.sh
```diff
@@ -67,9 +108,15 @@ remove() {
 
 if [ $# -eq 0 ]; then
     install
+    # TODO: Print recommendation to restart the PBS server/scheduler? Or do it automatically?
```
bgeltz: I'd vote to print the reccomendation.  The admin may want to wait till an opportune time to restart PBS.

service/integration/open_pbs/geopm_install_pbs_power_limit.sh
```diff
@@ -67,9 +108,15 @@ remove() {
...
+    # TODO: Echo a note mentioning that cluster power limit is not yet set?
```
bgeltz: I'm in favor of echoing a note that things are setup but not yet configured.

bgeltz (commented):
Looks pretty good.  I don't have any other specific feedback just yet on usage expectations.

service/integration/open_pbs/README.rst
dannosliwcd: This is an escaped part of documentation from my latest updates, so I'll remove that note. The rest of this comment is just additional context.

I originally implemented the capability described here but I removed it after realizing that it does not play nice with backfilling. As far as I can tell, it doesn't cause any constraint violations, but it would potentially cause a candidate fill job to lose its backfill spot in the schedule, which breaks a key motivation to get users to opt in to do this.

Example scenario: There is a power-and-time hole in the schedule, which can be assigned to a particular fill job that was assigned a low power cap on queue entry. Now that it is identified as a viable fill job, we execute the runjob event handler, which increases the power assigned to this job to use whatever slack power **currently** exists. But even if there is currently sufficient power, that may not be true for all times between now and the job's time limit. The scheduler will re-evaluate viability of this job after changing the job's resource requirements. If the increased power cap causes an expected delay for the top job, then this job will no longer be used as a fill job.

I think we could avoid that problem if it was possible to ask the scheduler whether the resource change would make us lose our spot, but I haven't figured out a way to do that. So this implementation just keeps the initial cap set by queuejob.

service/integration/open_pbs/README.rst
dannosliwcd: It's erroneously mentioned in other parts of the docs. I'll remove those references from the upcoming revision.

cmcantalupo (approved):

PR #2997, opened 2023-06-30
Fix issue with reporter

cmcantalupo:
- Do not use epoch count if it is NAN
- Do not add epoch or unmarked regions if they are all zeros


sidjana (approved):
LGTM

PR #2998, opened 2023-07-01
Fix test_name for test_application_totals_pinning

cmcantalupo:
- It overlaps with test_monitor, and output files may be overwritten.


sidjana (approved):
LGTM

Issue #2509, opened 2022-08-03
8-byte memory leak in NVMLGPUTopo 


dannosliwcd:
**Describe the bug**
I tried to run service unit tests under ASAN and several NVMLGPUTopo tests fail due to a leak in the constructor.

**GEOPM version**
ae97f97

**Expected behavior**
All unit tests pass under ASAN.

**Actual behavior**
NVMLGPUTopo tests fail under ASAN due to an 8-byte leak in the NVMLGPUTopo constructor.

Specifically, the following allocation is never freed:
https://github.com/geopm/geopm/blob/ae97f97b913f5433f47626645de73129b7f9cd07/service/src/NVMLGPUTopo.cpp#L35

The `CPU_FREE()` macro can be used to release the resources allocated by `CPU_ALLOC()`

**Note**: The `NVMLGPUTopo` class is instantiated as a singleton via `gpu_topo()`, so the 8-byte leak is expected to happen up to 1 time.

PR #3001, opened 2023-07-12
Enable clang sanitizers for C++ code and unit tests

### LINKED ISSUE #2978: Create Github-CI workflows that apply Clang sanitizers to the unit test execution
> estimate 2

Done criteria:
- [ ] Update github workflow to compile using two sanitization build work flows
- [ ] Figure out how to run the unit tests with the sanitizers engaged

avilcheslopez:
- Relates to #2978 
- Fixes #2978 #2509

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: codespell-project/actions-codespell@v1.0
   build_and_test:
-    name: "build_and_test: ${{ matrix.config.name }} ${{ matrix.debug-flag }}"
+    name: "build: ${{ matrix.config.name }} ${{ matrix.debug-flag }} service:${{ matrix.service-asan-flag }} base:${{ matrix.base-asan-flag  }}"
     runs-on: ubuntu-22.04
     strategy:
       matrix:
@@ -45,6 +45,18 @@ jobs:
           cc: "gcc-9", cxx: "g++-9"
         }
         debug-flag: ["enable-debug", "disable-debug"]
+        service-asan-flag: ["disable-asan"]
+        base-asan-flag: ["disable-asan"]
+        include:
+        - config: {name: clang-12, cc: "clang-12", cxx: "clang++-12"}
+          debug-flag: "enable-debug"
+          service-asan-flag: "enable-asan"
+          base-asan-flag: "disable-asan"
+        - config: {name: clang-12, cc: "clang-12", cxx: "clang++-12"}
+          debug-flag: "enable-debug"
+          service-asan-flag: "disable-asan"
+          base-asan-flag: "enable-asan"
+
     env:
       CC: ${{ matrix.config.cc }}
       CXX: ${{ matrix.config.cxx }}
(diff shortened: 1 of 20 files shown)


.github/workflows/build.yml
```diff
@@ -45,6 +45,16 @@ jobs:
           cc: "gcc-9", cxx: "g++-9"
         }
         debug-flag: ["enable-debug", "disable-debug"]
+        service-sanitizers-flag: ["disable-sanitizers", "enable-sanitizers"]
+        base-sanitizers-flag: ["disable-sanitizers", "enable-sanitizers"]
```
cmcantalupo: maybe "base" -> "runtime" ?

service/configure.ac
```diff
@@ -47,6 +47,20 @@ fi
 [enable_coverage="0"]
 )
 
+AC_ARG_ENABLE([sanitizers],
+  [AS_HELP_STRING([--enable-sanitizers], [Build and test with sanitizers enabled])],
```
cmcantalupo: I thought that there were two mutually exclusive sets of sanitizer options?  If there is more than one sanitizer configuration setting, we may need to be more specific, e.g. --enable-address-sanitizer / --enable-memory-sanitizer

service/configure.ac
```diff
@@ -148,6 +162,11 @@ if test "x$enable_coverage" = "x1" ; then
 fi
 AC_SUBST([enable_coverage])
 
+if test "x$enable_sanitizers" = "x1" ; then
+  EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS -fsanitize=address -mllvm -asan-use-private-alias=1"
+  #EXTRA_LDFLAGS="$EXTRA_LDFLAGS -fsanitize=address -mllvm -asan-use-private-alias=1"
```
cmcantalupo: we can remove this comment since it doesn't seem to be necessary with C++ code.

cmcantalupo (commented):
Some small suggestions.  Looks good though!

cmcantalupo:
```

Invalid workflow file: .github/workflows/build.yml#L53
The workflow is not valid. .github/workflows/build.yml (Line: 53, Col: 13): Matrix exclude key 'config.name' does not match any key within the matrix .github/workflows/build.yml (Line: 55, Col: 13): Matrix exclude key 'config.name' does not match any key within the matrix
```

.github/workflows/build.yml
avilcheslopez: Discussed with Chris: we'll use the term "base" for now for consistency and make the change everywhere in a separate PR.

configure.ac
```diff
@@ -188,6 +165,52 @@ fi
...
+  EXTRA_CXXFLAGS="$EXTRA_CXXFLAGS -fsanitize=address -fno-omit-frame-pointer -mllvm -asan-use-private-alias=1"
+fi
+
+if test "x$enable_overhead" = "x1" ; then
+  AC_DEFINE([GEOPM_OVERHEAD], [ ], [Enables code for measuring overhead])
+fi
+AC_SUBST([enable_overhead])
+
+AC_SUBST([enable_asan])
+AM_CONDITIONAL([ENABLE_SANITIZERS], [test "x$enable_asan" = "x1"])
```
cmcantalupo, line 211: maybe `ENABLE_SANITIZERS` -> `ENABLE_ASAN`

cmcantalupo (approved):
Thanks for doing all this clean up.  The Controller test in particular was a mess. Looks great.

PR #3048, opened 2023-08-07
Fuzz test for GEOPM hash function

### LINKED ISSUE #3003: Figure out interfaces that we will fuzz test first, enable first fuzz testing harness
> estimate 2

Done criteria:
- Fuzz testing harness for first API developed and in place
- Fuzz test should be incorporated into build via a configure option (to be built manually/explicitly, not as part of CI)
- Look into where we could deploy fuzz tests, understand requirements and scope of work to deploy
- Run fuzz test, confirm it works

avilcheslopez:
- Fixes #3003


service/Makefile.am
```diff
@@ -122,6 +127,12 @@ geopmwrite_SOURCES = src/geopmwrite_main.cpp
 geopmwrite_CXXFLAGS = $(AM_CXXFLAGS) -std=c++17
 geopmwrite_LDADD = libgeopmd.la
 
+if ENABLE_FUZZTESTS
```
cmcantalupo: Maybe better to put this change in ``service/integration/Makefile.mk`` and move ``service/src/fuzztest_geopmhash_main.cpp`` to ``service/integration/test/fuzztest_geopmhash_main.cpp``?

avilcheslopez: Will do

cmcantalupo (dismissed):

service/integration/test/fuzztest_geopmhash_main.cpp
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2015 - 2023, Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <cstdint>
+
```
bgeltz:
```suggestion
```

bgeltz (commented):
Tiny nit.  LGTM otherwise

bgeltz:
What's the expectation here if someone is building with the GNU toolchain but configures with ``--enable-fuzz``?
Also, what about configuring with ``--disable-debug``?

I think this is the first time we're adding support for CLANG specific features, and we may want to take care to isolate them properly.

bgeltz (approved):

PR #3051, opened 2023-08-08
Change REGION_PROGRESS to Agg::sum

### LINKED ISSUE #2916: REGION_PROGRESS does not get updated on single-threaded applications if OMP_NUM_THREADS is not set
**Describe the bug**
I tried to look at REGION_PROGRESS in a geopm trace, but it was always either NAN or zero even though the application was annotated with geopm_tprof_post(). It looks like this happens when I use geopmlaunch without explicitly setting OMP_NUM_THREADS, even though this application does not use openmp threading.

**GEOPM version**
A branch based on dev at af58d710.

**Expected behavior**
I expect REGION_PROGRESS in the GEOPM trace file to increase as a region makes progress.

**Actual behavior**
REGION_PROGRESS is always NAN or zero in the trace.

This branch has an annotated version of the MPI implementation of NPB IS, with GEOPM tprof annotations around its initialization region and its two main sorting regions. NPB IS uses a power of 2 processes. On my test node, this means I can assign 32 processes out of 44 cores. geopmlaunch chooses to set OMP_NUM_THREADS=2 if I don't set it to 1 explicitly in this configuration.

**Error messages produced**
_None_

**Workaround**
Set OMP_NUM_THREADS=1 explicitly for non-OMP applications.

bgeltz:
- Aggregated REGION_PROGRESS will now have a max = the number of ranks in the run.
- Add integration test for no_omp use case.
- Resolves #2916.

### DIFF
--- a/service/docs/source/geopm_pio_profile.7.rst
+++ b/service/docs/source/geopm_pio_profile.7.rst
@@ -53,11 +53,12 @@ Signals
     * **Unit**: geopm_region_hint_e
 
 ``PROFILE::REGION_PROGRESS``
-    Minimum per-rank reported progress through the current region.  When
-    utilizing OpenMP, the per-CPU returned value will be on the interval [0, 1
-    / *num_thread*] where *num_thread* is the number of requested OpenMP threads
-    per rank.  When running single-threaded (i.e. not leveraging OpenMP), the
-    per-CPU returned value will be on the interval [0, 1].
+    Sum of the fractional progress reported by all CPUs over the aggregated
+    domain.  When utilizing OpenMP, the per-CPU returned value will be on the
+    interval [0, 1 / *num_thread*] where *num_thread* is the number of
+    requested OpenMP threads per rank.  When running single-threaded (i.e. not
+    leveraging OpenMP), the per-CPU returned value will either be zero (no
+    active threads), or on the interval [0, 1].
 
     When aggregating this signal to the board domain, the aggregated value will
     always be on the interval [0, *num_rank*] where *num_rank* is equal to the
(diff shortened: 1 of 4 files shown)


dannosliwcd (dismissed):

service/docs/source/geopm_pio_profile.7.rst
```diff
@@ -53,11 +53,23 @@ Signals
     * **Unit**: geopm_region_hint_e
 
 ``PROFILE::REGION_PROGRESS``
-    Minimum per-rank reported progress through the current region.  The
-    returned value will be on the interval [0, 1].  0 indicates no progress
-    while 1 indicates the region is complete.
-
-    * **Aggregation**: min
+    Minimum per-rank reported progress through the current region.  When
```
cmcantalupo:
```suggestion
    Sum of the fractional progress reported by all threads running within the aggregated domain.  When
```

service/docs/source/geopm_pio_profile.7.rst
```diff
@@ -53,11 +53,23 @@ Signals
...
+    utilizing OpenMP, the per-CPU returned value will be on the interval [0, 1
+    / *num_thread*] where *num_thread* is the number of requested OpenMP threads
+    per rank.  When running single-threaded (i.e. not leveraging OpenMP), the
+    per-CPU returned value will be on the interval [0, 1].
```
cmcantalupo:
```suggestion
    per-CPU returned value will either zero (no active threads), or on the interval [0, 1].
```

service/docs/source/geopm_pio_profile.7.rst
```diff
@@ -53,11 +53,24 @@ Signals
...
+    active threads), or on the interval [0, 1].
+
+    When aggregating this signal to the board domain, the aggregated value will
+    always be on the interval [0, *num_rank*] where *num_rank* is equal to the
+    number of ranks (processes) in the run.  0 indicates no progress while 1 (if
+    examining the per-CPU signal) or *num_rank* (if examining the board
+    aggregated signal) indicates the region is complete.
+
+    This signal is not valid outside of a region.  Specifically, this signal
+    should be ignored any time the board aggregated ``REGION_HASH`` signal
```
cmcantalupo:
```suggestion
    should be ignored any time the ``REGION_HASH`` signal aggregated to the same domain
```

cmcantalupo (approved):

PR #3057, opened 2023-08-09
Refactor LevelZero.cpp

### LINKED ISSUE #2757: ze_* variables are not initialized in all cases in LevelZero.cpp
**Describe the bug**
I tried to make changes to LevelZero.cpp and I expected ze_result to be initialized in all cases. 

**GEOPM version**
2.0.0+dev232gb0510fa1

**Expected behavior**
All ze_* variables should be initialized to a safe value.

**Actual behavior**
Not all ze_* variables are initialized.

**Error messages produced**
By @bgeltz inspection.

bgeltz:
- Resolves #2757.

### DIFF
--- a/service/src/LevelZero.cpp
+++ b/service/src/LevelZero.cpp
@@ -34,69 +34,59 @@ namespace geopm
         , m_num_gpu_subdevice(0)
     {
         if (getenv("ZE_AFFINITY_MASK") != nullptr) {
-            throw Exception("LevelZero: Cannot be used directly when ZE_AFFINITY_MASK environment variable is set, must use service to access LevelZero in this case",
+            throw Exception("LevelZero: Cannot be used directly when ZE_AFFINITY_MASK environment "
+                            "variable is set, must use service to access LevelZero in this case.",
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
 
-        ze_result_t ze_result;
         //Initialize
-        ze_result = zeInit(ZE_INIT_FLAG_GPU_ONLY);
-        check_ze_result(ze_result, GEOPM_ERROR_RUNTIME, "LevelZero::"
-                        + std::string(__func__) +
+        check_ze_result(zeInit(ZE_INIT_FLAG_GPU_ONLY),
+                        GEOPM_ERROR_RUNTIME, "LevelZero::" + std::string(__func__) +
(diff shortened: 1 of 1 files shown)


service/src/LevelZero.cpp
```diff
@@ -38,7 +38,7 @@ namespace geopm
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
 
-        ze_result_t ze_result;
+        ze_result_t ze_result = ZE_RESULT_ERROR_UNKNOWN;
         //Initialize
         ze_result = zeInit(ZE_INIT_FLAG_GPU_ONLY);
```
cmcantalupo:
```suggestion
        //Initialize
        ze_result_t ze_result = zeInit(ZE_INIT_FLAG_GPU_ONLY);
```

cmcantalupo (changes requested):
I think a better solution is to scope the ze_result variable so that it is declared on the line where the level zero function is called, rather than at the top of each function.

service/src/LevelZero.cpp
cmcantalupo: This is preferable, It could also be reasonable to not declare a local variable at all, since the only use in all cases is to pass it to check_ze_result():

```
        check_ze_result(
            zeInit(ZE_INIT_FLAG_GPU_ONLY),
            GEOPM_ERROR_RUNTIME, "LevelZero::" + std::string(__func__) +
            ": LevelZero Driver failed to initialize.", __LINE__);
```

But I'm okay with either solution.

service/src/LevelZero.cpp
```diff
@@ -226,26 +209,23 @@ namespace geopm
                 }
                 else {
                     if (property.type == ZES_FREQ_DOMAIN_GPU) {
-                        m_devices.at(device_idx).subdevice.freq_domain.at(
-                                  geopm::LevelZero::M_DOMAIN_COMPUTE).push_back(handle);
+                        m_devices.at(device_idx).
+                            subdevice.freq_domain.at(geopm::LevelZero::M_DOMAIN_COMPUTE).push_back(handle);
```
cmcantalupo, line 213:
```suggestion
                        m_devices.at(device_idx)
                            .subdevice.freq_domain.at(geopm::LevelZero::M_DOMAIN_COMPUTE).push_back(handle);
```

cmcantalupo (approved):
LGTM

PR #3063, opened 2023-08-14
Fix static analyis issues

cmcantalupo:
- Use std::move() where appropriate
- Delete unimplemented copy constructor and copy assignment operator for IOUringImp

### DIFF
--- a/service/src/IOUringImp.cpp
+++ b/service/src/IOUringImp.cpp
@@ -48,8 +48,8 @@ namespace geopm
         // to the location the user requested we write operation return values.
         // Ensure that point remains valid by keeping the shared_ptr in our
         // result destinations vector until we finish cleaning up the batch.
-        m_result_destinations.push_back(destination);
         io_uring_sqe_set_data(sqe, destination.get());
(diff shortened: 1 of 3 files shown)


service/src/IOUringImp.cpp
```diff
@@ -87,7 +87,7 @@ namespace geopm
         auto sqe = get_sqe_or_throw();
         // Available since Linux 5.10
         io_uring_prep_read(sqe, fd, buf, nbytes, offset);
-        set_sqe_return_destination(sqe, ret);
+        set_sqe_return_destination(sqe, std::move(ret));
```
dannosliwcd, line 90: Should we also use std::move inside this called function, at https://github.com/geopm/geopm/pull/3063/files#diff-8c130e3e1ad539642628f8fd31118ca7ae5ae4851680f4dd633146df5d29f5baR51 ?

The function called here adds the raw pointer as context in the io_uring request, and pushes the shared_ptr to a vector inside this class to ensure that the pointer remains valid until we handle the completion event (where we write to the context-provided raw pointer).

dannosliwcd (approved):

PR #3072, opened 2023-08-19
Check if GEOPM_PLUGIN_PATH is secure

### LINKED ISSUE #2229: The IOGroup plugin path may be used by the service, but there is no check that the plugin path is secure
**Describe the bug**
If the GEOPM_PLUGIN_PATH environment variable is set in the geopm.service config file (this is not set in the default geopm.service unit), then it will be used by the service.  In this use case we must check that the directory is not writable by unprivileged users, otherwise the service may load IOGroup plugins which are not secure.  

**GEOPM version**
d538866

**Expected behavior**
Add a check when the plugin path is used such that if the calling process is privileged then the directory permissions are check prior to use.  

**Actual behavior**
No check is present, but the admin still must add an insecure directory to the geopm.service unit configuration file to create this problem. 

**Error messages produced**

**Additional context**

bgeltz:
- Fixes #2229.

### DIFF
--- a/service/docs/source/GEOPM_CXX_MAN_PluginFactory.3.rst
+++ b/service/docs/source/GEOPM_CXX_MAN_PluginFactory.3.rst
@@ -206,6 +206,15 @@ detailed example of plugin load order can be found in
        export GEOPM_PLUGIN_PATH=$GEOPM_HOME/tutorial/iogroup
        export GEOPM_PLUGIN_PATH=$HOME/plugin/iogroup:$GEOPM_PLUGIN_PATH
 
+.. warning::
+
+    It is imperative that any dynamically loaded library that is a dependency
+    of a plugin is available in the environment of the systemd initiated geopmd
+    service.  If dependent libararies are not available on the root user's
+    LD_LIBRARY_PATH, the plugin will fail to load.
+    For this reason, it is recommended that any plugins are compiled with
+    system installed gcc/g++ compilers.
+
 Plugin Load Constructor Function
 --------------------------------
 
--- a/service/docs/source/GEOPM_CXX_MAN_PluginFactory.3.rst
+++ b/service/docs/source/GEOPM_CXX_MAN_PluginFactory.3.rst
@@ -179,10 +179,11 @@ Plugin Search Path And Load Order
 
 The ``GEOPM_PLUGIN_PATH`` is a colon-separated list of directories
 that contain plugin shared object files to be loaded by the GEOPM
-runtime.  See :doc:`geopm(7) <geopm.7>` for details about ``GEOPM_PLUGIN_PATH``.
-Note that an Exception will be thrown by the ``register_plugin()``
-method if an attempt is made to register a plugin with the same name
-as a previously registered plugin.
+runtime.  Any symlink that resides in a directory specified by this
+variable will be ignored.  See :doc:`geopm(7) <geopm.7>` for details
+about ``GEOPM_PLUGIN_PATH``.  Note that an Exception will be thrown by
+the ``register_plugin()`` method if an attempt is made to register a
+plugin with the same name as a previously registered plugin.
 
 In the case of ``IOGroup`` plugins, the most recently loaded plugin to
 register a signal or control name provides the implementation at
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -892,13 +892,17 @@ class ActiveSessions(object):
         else:
            os.rename(sess_path, renamed_path)
 
-def _get_names(exec_name):
+def _get_names(exec_name, quiet=True):
     try:
         stdout_encoding = sys.stdout.encoding
+        stderr_encoding = sys.stderr.encoding
     except AttributeError:
         stdout_encoding = None
+        stderr_encoding = None
     if stdout_encoding is None:
         stdout_encoding = locale.getpreferredencoding()
+    if stderr_encoding is None:
+        stderr_encoding = locale.getpreferredencoding()
 
     try:
         stderr_encoding = sys.stderr.encoding
@@ -910,6 +914,8 @@ def _get_names(exec_name):
                                    stdout=subprocess.PIPE,
                                    stderr=subprocess.PIPE)
     stdout_bytes, stderr_bytes = signals_pid.communicate()
+    if not quiet:
+        sys.stderr.write(stderr_bytes.decode(encoding=stderr_encoding))
     return stdout_bytes.decode(encoding=stdout_encoding).split()
 
 
@@ -921,7 +927,7 @@ class AccessLists(object):
         self._CONFIG_PATH = config_path
         secure_make_dirs(self._CONFIG_PATH)
         self._DEFAULT_ACCESS = '0.DEFAULT_ACCESS'
-        self._signal_names = _get_names('/usr/bin/geopmread')
+        self._signal_names = _get_names('/usr/bin/geopmread', quiet=False)
         self._control_names = _get_names('/usr/bin/geopmwrite')
 
     def _validate_group(self, group):
--- a/service/src/MSRIOGroup.cpp
+++ b/service/src/MSRIOGroup.cpp
@@ -39,6 +39,7 @@
 #include "geopm/Helper.hpp"
 #include "geopm_debug.hpp"
 #include "SaveControl.hpp"
+#include "SecurePath.hpp"
 
 using json11::Json;
 
@@ -157,8 +158,14 @@ namespace geopm
         }
         auto custom_files = msr_data_files();
         for (const auto &filename : custom_files) {
-            std::string data = read_file(filename);
-            parse_json_msrs(data);
+            try {
+                SecurePath sp (filename);
+                std::string data = read_file(sp.secure_path());
+                parse_json_msrs(data);
+            }
+            catch (const geopm::Exception &ex) {
+                std::cerr << "Warning: " << ex.what() << std::endl;
+            }
         }
 
         // HWP enable is checked via an MSR, so we cannot do this as part of
@@ -1256,12 +1263,17 @@ namespace geopm
         }
         std::vector<std::unique_ptr<MSR> > msr_arr_custom;
         for (const auto &dir : plugin_paths) {
-            auto files = list_directory_files(dir);
-            for (const auto &file : files) {
-                std::string filename = dir + "/" + file;
-                if (string_begins_with(file, "msr_") && string_ends_with(file, ".json")) {
-                    data_files.insert(filename);
+            try {
+                auto files = list_directory_files(dir);
+                for (const auto &file : files) {
+                    std::string filename = dir + "/" + file;
+                    if (string_begins_with(file, "msr_") && string_ends_with(file, ".json")) {
+                        data_files.insert(filename);
+                    }
                 }
+            }
+            catch (const geopm::Exception &ex) {
+                std::cerr << ex.what() << std::endl;
             }
         }
         return data_files;
--- a/service/src/SecurePath.cpp
+++ b/service/src/SecurePath.cpp
@@ -7,15 +7,15 @@
 
 #include "SecurePath.hpp"
 
-#include <string>
-#include <iostream>
-#include <sstream>
-
 #include <fcntl.h>
 #include <unistd.h>
 #include <string.h>
 #include <sys/stat.h>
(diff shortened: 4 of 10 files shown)


service/geopmdpy/service.py
```diff
@@ -57,6 +59,20 @@ def __init__(self):
...
 
+    def _check_plugins(self):
+        # Similar logic to geopm_plugin.cpp -> plugin_load()
+        so_suffix = '.so.' + version.__abi_version__.replace(':', '.')
+        plugin_path = os.getenv('GEOPM_PLUGIN_PATH')
+        if plugin_path is not None:
+            user_paths = plugin_path.split(':')
+            for path in user_paths:
+                if system_files.is_secure_path(path):
+                    for file_path in glob.glob(os.path.join(path, '*')):
```
cmcantalupo: at least currently we also have .json files in the plugin directory.  We don't want to error when these .json files don't end in .so.1.0.0.

cmcantalupo:
```suggestion
                    for file_path in glob.glob(os.path.join(path, 'lib*.so*')):
```

service/geopmdpy/service.py
```diff
@@ -14,7 +14,6 @@
 import shutil
 import psutil
 import uuid
-import gi
```
bgeltz: Was this important for any reason?  Seems unnecessary.

tutorial/iogroup/README.md
```diff
@@ -197,6 +197,15 @@ filename must begin with "libgeopmiogroup_" and end in
 An alternative is to install the plugin by copying the .so file into
 the GEOPM install directory, <GEOPM_INSTALL_DIR>/lib/geopm.
 
+> [!WARNING]
```
bgeltz, line 199: GitHub markdown now supports admonition:
https://stackoverflow.com/a/72327818

Rendering:
https://github.com/bgeltz/geopm/tree/issue-2229/tutorial/iogroup

service/geopmdpy/system_files.py
```diff
@@ -270,6 +271,42 @@ def is_secure_file(path, fid):
...
+
+    """
+    plugin_path = os.getenv('GEOPM_PLUGIN_PATH')
+    if plugin_path is not None:
+        user_paths = plugin_path.split(':')
+        for path in user_paths:
+            secure_make_dirs(path)
+            scanned_paths = set()
+            # Custom IOGroup/Agent plugins: geopm_plugin.cpp -> plugin_load()
+            for file_path in glob.glob(os.path.join(path, 'lib*.so*')):
```
cmcantalupo: we have a more strict filter for loading plugins libgeopmpi*.so if i remember.

service/geopmdpy/system_files.py
```diff
@@ -270,6 +271,42 @@ def is_secure_file(path, fid):
...
+                if not file_path.endswith('INVALID'):
+                    with open(file_path, 'r') as ff:
+                        scanned_paths.add(file_path)
+                        if is_secure_file(file_path, ff):
+                            so_suffix = '.so.' + version.__abi_version__.replace(':', '.')
```
cmcantalupo: I guess these file checks are better than not having them, but there is a TOCTOU issue with this solution.

service/geopmdpy/system_files.py
```diff
@@ -270,6 +271,42 @@ def is_secure_file(path, fid):
...
+    files that do not meet the permissions requirements will be renamed
+    such that they are ignored by the runtime.  Files that pass the check
+    will be loaded by the runtime.
+
+    """
+    plugin_path = os.getenv('GEOPM_PLUGIN_PATH')
+    if plugin_path is not None:
+        user_paths = plugin_path.split(':')
+        for path in user_paths:
+            secure_make_dirs(path)
```
cmcantalupo: potentially moves files that are tracked by an RPM or deb file to /usr/lib64/geopm-INVALID, not sure if this is desired, as they will not be removed with an uninstall.

service/src/Helper.cpp
```diff
@@ -269,6 +271,46 @@ namespace geopm
...
+        int err = 0;
+        struct stat stat_struct;
+        err = fstat(fd, &stat_struct);
+        if (err) {
+            throw Exception("fd_to_mode(): ", errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        return stat_struct.st_mode;
+    }
+
+    bool is_fd_secure(const int fd, const std::string &plugin) {
```
cmcantalupo: maybe a more specific name is required for Helper.hpp, e.g. `is_plugin_secure`.

service/src/MSRIOGroup.cpp
```diff
@@ -1260,7 +1261,20 @@ namespace geopm
...
-                    data_files.insert(filename);
+                    bool do_insert = true;
+                    int fd = open(filename.c_str(), O_RDONLY);
+                    if (fd < 0) {
+                        std::cerr << "Warning: <geopm> Failed to open() MSR plugin: "
+                                  << strerror(errno) << std::endl;
+                        do_insert = false;
+                    }
+                    else {
+                        // TODO: Insert fd into tracking vector to close on destruction
```
bgeltz: This TODO is actually more difficult than I want it to be.  This function is static and is used in geopmadmin to generate the msr-safe allowlist.  Not sure if it's even worth it as the fd's will be close when the process ends.

service/geopmdpy/system_files.py
```diff
@@ -910,6 +914,8 @@ def _get_names(exec_name):
                                    stdout=subprocess.PIPE,
                                    stderr=subprocess.PIPE)
     stdout_bytes, stderr_bytes = signals_pid.communicate()
```
cmcantalupo, line 916: check return code for subprocess.

service/src/Helper.cpp
```diff
@@ -269,6 +271,46 @@ namespace geopm
...
+        bool is_secure = true;
+        if (getuid() == 0) {
+            int uid = geopm::fd_to_uid(fd);
+            if (uid != 0) {
+                is_secure = false;
+                std::cerr << "Warning: <geopm> Plugin not owned by root and will be ignored: "
+                          << plugin << std::endl;
+            }
+            int mode = geopm::fd_to_mode(fd);
+            if (mode != (S_IFREG | S_IRWXU)) { // 0o700
```
cmcantalupo: Default plugins should have permissions 755 like other files in /usr/lib64.  So the validation here must be that the write bits are not set. For example 755 is okay, 700 is okay, but 777 is not okay

service/src/MSRIOGroup.cpp
```diff
@@ -1256,13 +1257,25 @@ namespace geopm
         }
         std::vector<std::unique_ptr<MSR> > msr_arr_custom;
         for (const auto &dir : plugin_paths) {
-            auto files = list_directory_files(dir);
-            for (const auto &file : files) {
-                std::string filename = dir + "/" + file;
-                if (string_begins_with(file, "msr_") && string_ends_with(file, ".json")) {
-                    data_files.insert(filename);
+            try {
```
cmcantalupo, line 1266:
```
class SecurePath {
    SecurePath(const std::string &orig_path)
    : SecurePath(orig_path, 0o022, getuid() == 0)
    {}
    SecurePath(const std::string &orig_path, int umask, bool is_same_user);
    ~SecurePath()
    std::string secure_path();
    int fid();
}
```

cmcantalupo: SecurePath sp("foobar");
// replace all uses of "foobar" with sp.secure_path()

service/src/SecurePath.cpp
```diff
@@ -0,0 +1,85 @@
...
+    SecurePath::SecurePath(const std::string &orig_path)
+        : SecurePath(orig_path,
+                     (S_IWGRP | S_IWOTH), // 0o022
+                     (getuid() == 0))
+    {
+
+    }
+
+    SecurePath::SecurePath(const std::string &orig_path, mode_t umask, bool enforce)
+        : m_fd(open(orig_path.c_str(), O_RDONLY))
```
bgeltz: Could add O_NOFOLLOW here if we want to disallow symlinks.

open() follows symlinks by default, and fstat cannot tell the difference between a link and a regular file when doing things this way.

service/src/geopm_plugin.cpp
```diff
@@ -77,32 +77,41 @@ namespace geopm
...
-            }
-            else {
-                dl_handle = dlopen(plugin.c_str(), RTLD_LAZY|RTLD_GLOBAL);
+            try {
+                SecurePath sp (plugin.c_str());
+                void *dl_handle = dlopen(sp.secure_path().c_str(), RTLD_NOLOAD);
                 if (dl_handle != nullptr) {
                     DLRegistry::dl_registry().add(dl_handle);
                 }
-#ifdef GEOPM_DEBUG
```
bgeltz, line 99: Note I'm removing the ifdef.  I think it's important to print into the journal when plugins fail to load unexpectedly.

service/geopmdpy/system_files.py
bgeltz: Open an issue @bgeltz

service/docs/source/GEOPM_CXX_MAN_PluginFactory.3.rst
```diff
@@ -206,6 +206,15 @@ detailed example of plugin load order can be found in
...
 
+.. warning::
+
+    It is imperative that any dynamically loaded library that is a dependency
+    of a plugin is available in the environment of the systemd initiated geopmd
+    service.  If dependent libararies are not available on the root user's
+    LD_LIBRARY_PATH, the plugin will fail to load.
+    For this reason, it is recommended that any plugins are compiled with
+    system installed gcc/g++ compilers.
+
```
bgeltz, line 218: Add blurb about not allowing symlinks.

tutorial/iogroup/README.md
```diff
@@ -197,6 +197,15 @@ filename must begin with "libgeopmiogroup_" and end in
...
+> When installing plugins system-wide for use through the GEOPM service,
+> it is imperative that any dynamically loaded library that is a dependency
+> of a plugin is available in the environment of the systemd initiated geopmd
+> service.  If dependent libararies are not available on the root user's
+> LD_LIBRARY_PATH, the plugin will fail to load.
+> For this reason, it is recommended that any plugins are compiled with
+> the system installed gcc/g++ compilers.
+
```
bgeltz, line 207: Add blurb about not allowing symlinks.

service/test/SecurePathTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+TEST_F(SecurePathTest, umask)
+{
+    mode_t test_perms = (S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH); // 0o644
+    chmod(m_file_name.c_str(), test_perms);
+
+    // Assert write permissions for the group/world are not set
+    SecurePath* sp = new SecurePath(m_file_name, (S_IWGRP | S_IWOTH), true);
+    EXPECT_EQ("/proc/self/fd/3", sp->secure_path());
+    delete sp;
```
bgeltz: Try with brackets instead of new/delete.

cmcantalupo:
```suggestion
    {
        SecurePath sp(m_file_name, (S_IWGRP | S_IWOTH), true);
        EXPECT_EQ("/proc/self/fd/3", sp.secure_path());
    }
```

service/src/SecurePath.cpp
```diff
@@ -0,0 +1,85 @@
...
+            throw Exception("SecurePath::" + std::string(__func__) +
+                            "(): Failed to open file: ",
+                            errno ? errno : GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+
+        if (enforce) {
+            struct stat stat_struct;
+            int err = fstat(m_fd, &stat_struct);
+            if (err) {
+                (void)close(m_fd);
```
cmcantalupo:
```suggestion
                (void)!close(m_fd);
```
We need to put the ! in there with all these ignored return codes now

cmcantalupo (commented):
Noticed the missing `!` from the ignored return values on this pass.  Also just wondering why the scoping trick doesn't enable us to avoid the new/delete pattern?

cmcantalupo (approved):

PR #3078, opened 2023-08-23
Avoid accidental profiling on import of geopmdpy.gffi

bgeltz:
- Since profiling is now opt-out, ensure that undesired processes (e.g. python) are not accidently included for profiling.
- Add loop to print the PIDs that will be profiled on debug builds.

### DIFF
--- a/src/ApplicationIO.cpp
+++ b/src/ApplicationIO.cpp
@@ -9,6 +9,7 @@
 #include <utility>
 #include <unistd.h>
 #include <ctime>
+#include <iostream>
 
 #include "geopm_time.h"
 #include "geopm/Exception.hpp"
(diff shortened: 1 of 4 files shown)


src/ApplicationIO.cpp
```diff
@@ -83,6 +84,13 @@ namespace geopm
             clock_nanosleep(CLOCK_REALTIME, 0, &delay, NULL);
             geopm_time(&time_curr);
         } while (!m_is_connected && geopm_time_diff(&time_zero, &time_curr) < timeout);
+#ifdef GEOPM_DEBUG
+        std::cerr << "Warning: <geopm> Controller will profile PIDs: ";
```
cmcantalupo:
```suggestion
        std::cerr << "Debug: <geopm> Controller will profile PIDs: ";
```

cmcantalupo (approved):

Issue #2995, opened 2023-06-30
Python agents are not working after non-MPI support branch merged


cmcantalupo:
**Describe the bug**
I tried to create a report with the python agent tutorial and I expected correct data instead all fields were zero.

When using the python agents, it is recommended that the [legacy-dev](https://github.com/geopm/geopm/tree/legacy-dev) branch be used until this issue is resolved in the dev branch.

**GEOPM version**
28e1581c5b2df12bc2ab7404a2d03f386ff7ff46

**Expected behavior**
No regressions in test_tutorial_python_agents.py assertions

**Actual behavior**
Test is failing and report has no timing or sampled data.

**Error messages produced**

```
======================================================================
FAIL: test_frequency_limit_report (integration.test.test_tutorial_python_agents.TestIntegration_tutorial_python_agents)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/test/geopm/integration/test/test_tutorial_python_agents.py", line 91, in test_frequency_limit_report
    places=0)
AssertionError: 5 != 0 within 0 places

======================================================================
FAIL: test_monitor_report (integration.test.test_tutorial_python_agents.TestIntegration_tutorial_python_agents)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/test/geopm/integration/test/test_tutorial_python_agents.py", line 64, in test_monitor_report
    places=0)
AssertionError: 5 != 0 within 0 places

======================================================================
FAIL: test_package_energy_report (integration.test.test_tutorial_python_agents.TestIntegration_tutorial_python_agents)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/test/geopm/integration/test/test_tutorial_python_agents.py", line 74, in test_package_energy_report
    places=0)
AssertionError: 5 != 0 within 0 places

```

**Additional information**
When bug is fixed, please also revert skip decorator added here:
b2628e82bd6a835c7076d4f98fa6dc45bbb41efc

cmcantalupo:
Python agent feature is removed for 3.0 release.  May take a different form in future release now that non-MPI applications are supported by the C++ runtime.

PR #3082, opened 2023-08-24
Remove python runtime

cmcantalupo:
- Relates to #2995 
- See #3074 for work to fix issues with python agent.
- This PR removes the python based runtime


service/docs/source/geopmpy.7.rst
```diff
@@ -116,6 +116,13 @@ geopmpy.reporter
    :undoc-members:
    :show-inheritance:
 
+geopmpy.runtime
```
cmcantalupo: remove

src/Reporter.cpp
```diff
@@ -594,7 +594,7 @@ int geopm_reporter_init(void)
     int err = geopm_time_string(NAME_MAX, start_time);
     if (!err) {
         try {
-            geopm::basic_reporter(start_time);
+            geopm::basic_reporter(start_time).init();
```
bgeltz: AFAICT, init() was never called in the past (by the C++ runtime).  What are all the implications here?  I'm wondering what we should have seen not working properly before this change.

integration/test/test_reporter.py
bgeltz, line 1: So we're removing the runtime, but the reporter and pio modules are still working OK?
I'm wondering why we started skipping this test in the first place.

src/Reporter.cpp
cmcantalupo: This change should be reverted.  Thanks!

integration/test/test_reporter.py
cmcantalupo: I have also removed the reporter.py module and supporting C and tests.

src/Reporter.hpp
```diff
@@ -196,6 +196,7 @@ namespace geopm
             double m_total_time;
             double m_overhead_time;
             double m_sample_delay;
+            std::string m_profile_name;
```
bgeltz:
```suggestion
            const std::string m_profile_name;
```
I think this can be const.  I only see it being set in the initializer list.

bgeltz (changes requested):

bgeltz (approved):

PR #3105, opened 2023-09-12
Quiesce warnings about direct writes to the MSR module

### LINKED ISSUE #2461: MSR writes with the stock MSR driver no longer work by default with kernels >= 5.9
**Describe the bug**
I tried to write to a MSR with the stock MSR driver and I expected it to succeed instead an error was raised.

**GEOPM version**
```41d68f17```

**Expected behavior**
MSR writes through GEOPM work as expected.

**Actual behavior**
MSR writes through GEOPM fail and produce the following message in the journal:
```
Jul 14 14:42:48 compute_node_1 kernel: msr: Write to unrecognized MSR 0x38f by geopmd (pid: 42577). Please report to x86@kernel.org.
```

**Additional context**
Starting with Linux kernel 5.9, writes to MSRs are disallowed by default:
https://www.phoronix.com/scan.php?page=news_item&px=Linux-Filter-Tightening-MSRs
https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=x86/misc&id=a7e1f67ed29f0c339e2aa7483d13b085127566ab

Viable workarounds are to set ```msr.allow_writes=on``` on the kernel command line at boot, or if the system is already booted use the following sysfs entry:
```
echo "on" > /sys/module/msr/parameters/allow_writes
```

*NOTE:* This does not apply when using the msr-safe kernel driver.

### PR DESCRIPTION
**Pull request by dannosliwcd**
- OSes with Linux version 5.9 or greater emit warnings about unexpected MSR writes unless an option is set to allow writes in the MSR driver
- This change writes the setting to allow writes in the MSR driver, restoring the initial setting when the geopm service exits.
- Resolves #2461

In addition to the unit tests in this change, the following manual validation was performed on an Ubuntu 22.04.3 LTS system (Kernel 5.15):
1. Install geopm, start the service
2. `sudo cat /var/run/geopm-service/msr-saved-allow-writes` prints `default`
3. `sudo cat /sys/module/msr/parameters/allow_writes` prints `on`
4. `geopmwrite CPU_FREQUENCY_MAX_CONTROL core 0 3700000000`
5. `sudo dmesg -H` shows no warnings about unexpected MSR writes at the time of the test
6. `sudo systemctl stop geopm`
7. `sudo cat /var/run/geopm-service/msr-saved-allow-writes` does not exist
8. `sudo cat /sys/module/msr/parameters/allow_writes` prints `default`
9. `sudo pkill -kill geopmd`
10. `sudo journalctl -u geopm` shows a warning that a pre-existing backup file is being reused. `msr-saved-allow-writes` is still `default` and `allow_writes` is `on`.

### DIFF
--- a/service/geopmdpy/restorable_file_writer.py
+++ b/service/geopmdpy/restorable_file_writer.py
@@ -1,0 +1,87 @@
+#  Copyright (c) 2015 - 2023, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+import os
+from typing import Union, Callable
+from geopmdpy import system_files
+
+
+class RestorableFileWriter:
+    def __init__(self,
+                 write_path: Union[str, os.PathLike],
+                 backup_path: Union[str, os.PathLike],
+                 warning_handler: Union[Callable[[str], None], None] = None):
+        """Create a context manager that restores the contents of backup_path
+        to write_path and deletes backup_path on context exit.
+        Note that only one RestorableFileWriter should use a given backup_path
+        at a time since the backup file is removed as soon as any instance
+        restores the backup.
+        """
+        self._write_path: Union[str, os.PathLike] = write_path
+        self._backup_path: Union[str, os.PathLike] = backup_path
+        # If no warning handler is requested, ignore warnings
+        self._warning_handler: Callable[[str], None] = (
+            warning_handler if warning_handler is not None else lambda warning: None)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # This function can potentially be called while handling another
+        # exception already. Log this function's exceptions so we don't clobber
+        # some other exception's logged stacktrace.
+        try:
+            self.restore_and_cleanup()
+        except Exception as e:
+            self._warning_handler(f'Encountered exception in restore to {self._write_path}: {repr(e)}')
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: service/geopmdpy/restorable_file_writer.py [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+
+        with open(self._write_path, 'w') as f:
+            f.write(new_file_contents)
+
+    def restore_and_cleanup(self):
+        """Restore the contents of the backup file to the target file, deleting
+        the backup file after successful restore.
+        """
+        try:
+            with open(self._backup_path, 'r') as f:
```
cmcantalupo: can we replace these open()/read/write() calls with the secure methods?
https://github.com/geopm/geopm/blob/dev/service/geopmdpy/system_files.py#L130-L194

Maybe overkill for such short buffers as in the current use case, but could be helpful if the file was bigger.

cmcantalupo (changes requested):
Suggestion to use the secure file methods.  Other wise looks good!

service/geopmdpy/restorable_file_writer.py
```diff
@@ -0,0 +1,78 @@
...
+                backup_contents = f.read()
```
cmcantalupo:
```suggestion
            backup_contents = system_files.secure_read_file(self._backup_path)
```

cmcantalupo (approved):

PR #3107, opened 2023-09-13
Remove system_files.py from gitignore

dannosliwcd:
- After change 1e9bc15e system_files.py is not a generated file so it should no longer be ignored


cmcantalupo (approved):
Thanks!

PR #3113, opened 2023-09-15
Implement ServiceIOGroup save/restore

### LINKED ISSUE #2892: Controller does not restore controls when they are accessed through the service: the ServiceIOGroup does not implement save/restore
**Describe the bug**
I tried to run the static_agent.py tutorial Python agent and I expected controls to be restored to their initial value at the end of the run instead any requested control values persist after the run is complete.

**GEOPM version**
2c8cb6160

**Expected behavior**
```
$ geopmread CPU_POWER_LIMIT_CONTROL board 0
280

$ ./static_agent.py --geopm-report stress-report.report --geopm-initialize-control CPU_POWER_LIMIT_CONTROL=150 -- stress-ng --cpu 44 --taskset 0-43 --timeout 10

$ geopmread CPU_POWER_LIMIT_CONTROL board 0
280
```

**Actual behavior**
```
$ geopmread CPU_POWER_LIMIT_CONTROL board 0
280

$ ./static_agent.py --geopm-report stress-report.report --geopm-initialize-control CPU_POWER_LIMIT_CONTROL=150 -- stress-ng --cpu 44 --taskset 0-43 --timeout 10

$ geopmread CPU_POWER_LIMIT_CONTROL board 0
150
```

**Additional context**
My user does not have direct access to msr-safe, and is leveraging the service for all reading/writing.

The real problem here is that the ServiceIOGroup has no implementation for save_control() or restore_control().  This is what's called on the user's side when they call into pio.save_control() from the runtime.py.

It would be nice if we had a way to trigger the service to do a restore operation without having to close the currently active session.

https://github.com/geopm/geopm/blob/2c8cb6160c21f83ceb8e9eecf3644aac9aa93838/tutorial/python_agents/static_agent.py#L36-L38
The lines above only apply when you have direct access to the IOGroup that provides the control.  If you're relying on the service to provide signals/controls, the service only restores control values when the session ends (e.g. by job termination or logging out of an allocation).

avilcheslopez:
Fix bug in controller where controls accessed through the service are not restored. Add DBus API to restore controls (called by ServiceIOGroup) to allow controller to trigger controls to be restored.

- Fixes #2892

### DIFF
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -676,8 +676,7 @@ class PlatformService(object):
         self._pio.write_control(control_name, domain, domain_idx, setting)
 
     def restore_control(self, client_pid):
-        """Restore the state of all controls recorded at the start of a
-        session.
+        """Restore all controls recorded at the start of a session.
 
         For the session associated to the given process, restore the state
         of all controls, which is recorded at the beginning of the session.
--- a/service/geopmdpy/service.py
+++ b/service/geopmdpy/service.py
@@ -675,6 +675,23 @@ class PlatformService(object):
         self._write_mode(client_pid)
         self._pio.write_control(control_name, domain, domain_idx, setting)
 
+    def restore_control(self, client_pid):
+        """Restore the state of all controls recorded at the start of a
+        session.
+
+        For the session associated to the given process, restore the state
+        of all controls, which is recorded at the beginning of the session.
+
+        Raises
+            RuntimeError: If client_pid does not have an open session or
+                          if a different client currently has an open
+                          session in write mode
+        """
+        self._active_sessions.check_client_active(client_pid, 'PlatformRestoreControl')
+        self._write_mode(client_pid)
+        save_dir = os.path.join(self._RUN_PATH, self._SAVE_DIR)
+        self._pio.restore_control_dir(save_dir)
+
     def start_profile(self, user, client_pid, profile_name):
         """Begin profiling a user PID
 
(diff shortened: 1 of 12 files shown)


service/io.github.geopm.xml
```diff
@@ -632,6 +632,24 @@ range.
         </doc:description>
       </doc:doc>
     </method>
+    <method name="PlatformRestoreControl">
+     <doc:doc>
+        <doc:description>
+          <doc:summary>Restore the state of all controls recorded at the start of a
+          </doc:summary>
+          <doc:para>session.
```
cmcantalupo: This looks weird.

service/geopmdpy/service.py
```diff
@@ -675,6 +675,23 @@ def write_control(self, client_pid, control_name, domain, domain_idx, setting):
         self._write_mode(client_pid)
         self._pio.write_control(control_name, domain, domain_idx, setting)
 
+    def restore_control(self, client_pid):
+        """Restore the state of all controls recorded at the start of a
+        session.
```
cmcantalupo:
```suggestion
        """Restore all controls recorded at the start of a session.
```

Just to keep the summary on one line.

cmcantalupo (changes requested):
Looks good!  Just a nit about the line wrap issue in the xml.

service/io.github.geopm.xml
```diff
@@ -632,6 +632,24 @@ range.
...
+
+For the session associated to the given process, restore the state
+of all controls, which is recorded at the beginning of the session.
```
cmcantalupo:
```suggestion
          <doc:summary>Restore all controls recorded at the start of a session.
          </doc:summary>
          <doc:para>For the session associated to the given process, restore the state
of all controls, which is recorded at the beginning of the session.
```

cmcantalupo (approved):

PR #3115, opened 2023-09-15
Updates to geopmlaunch's man page to reflect changed default

alawibaba:
Also noted that non-MPI apps can only be used with the application control mode and removed the note about aprun.


bgeltz (changes requested):
These changes LGTM.

The other spot where I know this is discussed is in the overview.rst here:
https://geopm.github.io/overview.html#straight-ruler-use-the-runtime-to-measure-application-performance

I think that might be it?  But you might want to check the other docu we have and make sure I didn't miss yet another spot.

alawibaba:
Fixed. I also grepped through the rest of the docs in service/docs/source; afaict these are the only two spots where the launch methods are listed.

bgeltz (approved):

PR #3116, opened 2023-09-15
Cleaned up derived Launcher class constructors

alawibaba:
This code deleted by this PR is identical to code in the parent constructor (lines 410-417); in fact, the error that is raised on line 416 means that the second if statement in the IMPIExecLauncher constructor can never be True. I believe the IMPIExecLauncher constructor can be removed entirely, but I'd like someone who's more familiar with this code to confirm it.

### DIFF
--- a/scripts/geopmpy/launcher.py
+++ b/scripts/geopmpy/launcher.py
@@ -304,6 +304,9 @@ class Launcher(object):
     Abstract base class for MPI job launch application abstraction.
     Defines common methods used by all Launcher objects.
     """
+
+    is_slurm_enabled = False
+
     def __init__(self, argv, num_rank=None, num_node=None, cpu_per_rank=None, timeout=None,
                  time_limit=None, job_name=None, node_list=None, exclude_list=None, host_file=None,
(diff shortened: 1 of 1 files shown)


scripts/geopmpy/launcher.py
```diff
@@ -304,6 +304,9 @@ class Launcher(object):
     Abstract base class for MPI job launch application abstraction.
     Defines common methods used by all Launcher objects.
     """
+
+    is_slurm_enabled = False
```
bgeltz:
```suggestion
    _is_slurm_enabled = False
```
To denote this as a static member of the class.  There are very few examples of this in the codebase anymore, but this is how ```_is_once``` is defined for the IMPIExecLauncher.

bgeltz (changes requested):

scripts/geopmpy/launcher.py
alawibaba: Done. FWIW as I understand it, this is to indicate that a member is protected, which is completely suitable.

bgeltz (approved):

PR #3124, opened 2023-09-19
Add environment variable GEOPM_CTL_LOCAL

### LINKED ISSUE #3023: As a user of the GEOPM runtime I would like to run the geopmctl(1) command and opt out of using MPI
I'm always frustrated when I have compiled the GEOPM Runtime with MPI enabled, but I do not want to run the GEOPM Controller using MPI.  This may be because I either do not wish for the GEOPM runtime to utilize the MPI fabric, nor do I wish to launch the GEOPM Controller using an MPI launcher.  This would enable the exact same use case as when the GEOPM Runtime is configured with `--disable-mpi` but allowing a single binary to serve both use cases.

Add an environment variable like `GEOPM_CTL_LOCAL` which would be used to skip calls to MPI by the `Controller`, and utilize the `NullComm` rather  than the `MPIComm` class.

Currently this feature is only available as a compile time option, which restricts usablity.

cmcantalupo:
- When set disable controller use of MPI
- Fixes #3023

### DIFF
--- a/service/docs/source/geopmlaunch.1.rst
+++ b/service/docs/source/geopmlaunch.1.rst
@@ -559,7 +559,9 @@ GEOPM Options
                    running on different compute nodes.  This will
                    result in one report file per host, and each report
                    file will have the ``hostname`` appended to the file
-                   path requested by the user.
+                   path requested by the user.  Note this option has no
+                   effect if ``--geopm-ctl`` or ``--geopm-ctl-disable``
+                   are provided.
 --geopm-ompt-disable  .. _geopm-ompt-disable option:
 
                       Disable OMPT detection of OpenMP regions.
--- a/service/docs/source/geopmlaunch.1.rst
+++ b/service/docs/source/geopmlaunch.1.rst
@@ -553,6 +553,13 @@ GEOPM Options
                      launch the GEOPM runtime in process mode.  When this
                      option is specified, the GEOPM runtime will not be
                      launched.
+--geopm-ctl-local  .. _geopm-ctl-local option:
+
+                   Disable all communication between controllers
+                   running on different compute nodes.  This will
+                   result in one report file per host, and each report
+                   file will have the ``hostname`` appended to the file
+                   path requested by the user.
 --geopm-ompt-disable  .. _geopm-ompt-disable option:
 
                       Disable OMPT detection of OpenMP regions.
--- a/src/Controller.cpp
+++ b/src/Controller.cpp
@@ -12,6 +12,9 @@
 
 #include <algorithm>
 #include <memory>
+#ifdef GEOPM_ENABLE_MPI
+#include <mpi.h>
+#endif
 
 #include "ApplicationIO.hpp"
 #include "Environment.hpp"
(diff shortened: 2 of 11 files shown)


src/Controller.cpp
```diff
@@ -63,12 +66,32 @@ extern "C"
...
+                    char error_str[NAME_MAX + 1] = {};
+                    PMPI_Error_string(err, error_str, &str_size);
+                    fprintf(stderr, "Error: %s\n", error_str);
+                    return err;
+                }
+            }
+#endif
+            {
+                std::unique_ptr<geopm::Controller> ctl = std::make_unique<geopm::Controller>();
+                err = geopm_ctl_run((struct geopm_ctl_c *)ctl.get());
```
cmcantalupo: maybe should just be calling  ``ctl->run()`` no need for the C interface anymore.

src/Controller.cpp
```diff
@@ -63,12 +66,32 @@ extern "C"
...
         int err = 0;
+        bool do_ctl_local = geopm::environment().do_ctl_local();
         try {
-            geopm::Controller ctl;
-            err = geopm_ctl_run((struct geopm_ctl_c *)&ctl);
+#ifdef GEOPM_ENABLE_MPI
+            if (!do_ctl_local) {
+                err = PMPI_Init(&argc, &argv);
+                if (err) {
+                    int str_size = NAME_MAX;
```
dannosliwcd: Should this be `MPI_MAX_ERROR_STRING` instead?

src/Environment.cpp
```diff
@@ -214,7 +215,11 @@ namespace geopm
 
     std::string EnvironmentImp::comm(void) const
     {
-        return lookup("GEOPM_COMM");
+        std::string ret = "NullComm";
+        if (!is_set("GEOPM_CTL_LOCAL")) {
```
avilcheslopez: Suggestion: leverage `do_ctl_local()`

service/docs/source/geopmlaunch.1.rst
```diff
@@ -553,6 +553,13 @@ GEOPM Options
                      launch the GEOPM runtime in process mode.  When this
                      option is specified, the GEOPM runtime will not be
                      launched.
+--geopm-ctl-local  .. _geopm-ctl-local option:
+
+                   Disable all communication between controllers
+                   running on different compute nodes.  This will
+                   result in one report file per host, and each report
+                   file will have the ``hostname`` appended to the file
+                   path requested by the user.
```
cmcantalupo: Add comment that this option has no effect when --geopm-ctl is set to process or pthread.

src/Controller.cpp
```diff
@@ -63,12 +66,33 @@ extern "C"
...
+                    PMPI_Error_string(err, error_str, &str_size);
+                    fprintf(stderr, "Error: %s\n", error_str);
+                    return err;
+                }
+            }
+#endif
+            {
+                std::unique_ptr<geopm::Controller> ctl = std::make_unique<geopm::Controller>();
+                ctl->run();
+            }
```
bgeltz, line 90: Was the extra {} here necessary for something?

bgeltz (approved):

src/Controller.cpp
cmcantalupo: that forces the controler destructor to be called, which needs to happen before MPI_FInalize()

PR #3125, opened 2023-09-19
Consolidate geopm configuration directories

### LINKED ISSUE #2914: As a system administrator I would like GEOPM's config files to reside in the same directory
I'm always frustrated when I'm deploying GEOPM service updates to my compute image and I have to manually place configuration files in multiple directories.

If manual configuration files are required, they are all contained in the **_same_** directory.  This makes maintenance/audits of the compute image much easier.

**Additional context**
Plugins (MSRs or otherwise) go in: ```/usr/lib64/geopm```
Access lists go in: ```/etc/geopm-service```
Environment default/overrides go in: ```/etc/geopm```
Const config default file path: ```/etc/geopm```

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html

Agent plugins are SOs, currently loaded from ```/usr/lib64/geopm```.
MSR additions are JSON files, currently loaded from ```/usr/lib64/geopm```.

Tasks
- [x] #3118
- [x] #3119

dannosliwcd:
- Resolves #2914
- Use `/etc/geopm` instead of `/etc/geopm-service`
- Use `/run/geopm` instead of `/run/geopm-service`
- Read MSRIOGroup's json files from a new environment variable separate from the plugin path, with a default search location in `/etc/geopm`
- Continue reading from old file locations for `/etc` and MSR files, but emit warnings that those locations may not be read in future releases.

### DIFF
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -6,11 +6,11 @@
 
 Provides secure interfaces for manipulating the files in
 
-    /run/geopm-service
+    /run/geopm
 
 that enable the service to be restarted and
 
-   /etc/geopm-service
+   /etc/geopm
 
 where the access lists are located.  These interfaces provide
 guarantees about the security of these system files, and an
@@ -30,13 +30,16 @@ import grp
 import pwd
 import fcntl
 import subprocess # nosec
+from functools import lru_cache
 
 from . import pio
 from . import schemas
 from . import shmem
 
-GEOPM_SERVICE_RUN_PATH = '/run/geopm-service'
-GEOPM_SERVICE_CONFIG_PATH = '/etc/geopm-service'
+GEOPM_SERVICE_RUN_PATH = '/run/geopm'
+GEOPM_SERVICE_CONFIG_PATH = '/etc/geopm'
+# Deprecated since 3.0. May remove in a future release.
+LEGACY_GEOPM_SERVICE_CONFIG_PATH = '/etc/geopm-service'
 
 GEOPM_SERVICE_RUN_PATH_PERM = 0o711
 """Default permissions for the GEOPM service run path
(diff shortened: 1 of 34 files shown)


service/integration/open_pbs/geopm_power_limit.py
```diff
@@ -23,7 +23,7 @@
 from geopmdpy import system_files
 
 
-_SAVED_CONTROLS_PATH = "/run/geopm-pbs-hooks/SAVE_FILES"
+_SAVED_CONTROLS_PATH = "/run/geopm/pbs-hooks/SAVE_FILES"
```
cmcantalupo, line 26:
```
diff --git a/service/integration/open_pbs/geopm_install_pbs_power_limit.sh b/service/integration/open_pbs/geopm_install_pbs_power_limit.sh
index 37041ca5d..ab6f65503 100755
--- a/service/integration/open_pbs/geopm_install_pbs_power_limit.sh
+++ b/service/integration/open_pbs/geopm_install_pbs_power_limit.sh
@@ -11,7 +11,7 @@ JOB_TYPE_RESOURCE="geopm-job-type"
 DEFAULT_SLOWDOWN_RESOURCE="geopm-default-slowdown"
 NODE_CAP_HOOK="geopm_power_limit"
 REMOVE_OPT="--remove"
-SAVED_CONTROLS_BASE_DIR="/run/geopm-pbs-hooks"
+SAVED_CONTROLS_BASE_DIR="/run/geopm/pbs-hooks"
 
 SCHED_CONFIG_MODIFIED_MESSAGE="\
 Note: ${PBS_HOME}/sched_priv/sched_config has been modified, but modifications
```

service/src/geopm/MSRIOGroup.hpp
```diff
@@ -112,9 +112,18 @@ namespace geopm
...
+                EMIT_CONFIG_DEPRECATION_WARNING,
+            };
             /// @brief Returns the filenames for user-defined MSRs if
             ///        found in the plugin path.
-            static std::set<std::string> msr_data_files(void);
+            /// @param [in] emit_config_deprecation_warning Whether to emit a warning
+            ///        if MSR config files are found while searching the plugin
+            ///        path instead of while searching config file locations.
+            static std::set<std::string> msr_data_files(
+                    MsrConfigWarningPreference_e warning_preference = SILENCE_CONFIG_DEPRECATION_WARNING);
```
cmcantalupo, line 126: An enum in a function definition 😲 I guess since its private it makes sense.  We may want to rethink our use of enum variables always being cast to integers at a future date.

cmcantalupo (approved):
Nice work.  Looks good to me.  One question and one comment, but nothing blocking.

service/geopmdpy/system_files.py
```diff
@@ -49,6 +52,32 @@
 """
 
 
+@lru_cache(None)
```
cmcantalupo, line 55: Looks like this decorator is what enables the single write of the warning message?

dannosliwcd: That's right. Since `get_config_path` takes no args, the decorator ensures this function's body is only executed once.

service/src/geopm/MSRIOGroup.hpp
dannosliwcd: I'm not super attached to this approach. My main motivation was to try going for calls easier to interpret than `msr_data_files(true)`.

PR #3127, opened 2023-09-19
Download corpus dirs only when enabling fuzz tests

### LINKED ISSUE #3126: Don't download fuzz test corpus dirs when not enabling fuzz tests
> estimate 1

avilcheslopez:
- Fixes #3126


avilcheslopez:
Not an issue. We want corpus dirs included in dist tarball regardless.

PR #3141, opened 2023-09-22
Add --direct to use geopmaccess without the service

### LINKED ISSUE #2777: As a user of `geopmaccess` I would like an option `--direct` that would read and write directly to `/etc/geopm-service`
I'm always frustrated that a functioning and active GEOPM Systemd Service is required in order to read and write the allowed list files.

Provide a command line `--direct` / `-x` that will use the system_files.AccessLists object directly, rather than going through the DBus interface.  

Otherwise administrators will just write these files with a bash script, but there are permissions issues that may result from this solution.

**Additional context**
Partial implementation is here #2771

dannosliwcd:
- Resolves #2777
- When --direct is used, directly interact with GEOPM access lists instead of going through a service proxy.

### DIFF
--- a/service/geopmdpy/access.py
+++ b/service/geopmdpy/access.py
@@ -13,6 +13,34 @@ import subprocess # nosec
 from argparse import ArgumentParser
 from dasbus.connection import SystemMessageBus
 from dasbus.error import DBusError
+from geopmdpy import system_files
+
+
+class DirectAccessProxy:
+    """A proxy for access-related GEOPM service interactions that attempts
+    in-process operations instead of issuing D-Bus calls.
+    """
+
+    def __init__(self):
+        self._access_lists = system_files.AccessLists(system_files.get_config_path())
+
+    def PlatformGetGroupAccess(self, group):
+        return self._access_lists.get_group_access(group)
+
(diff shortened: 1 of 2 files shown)


service/geopmdpy/access.py
```diff
@@ -392,10 +422,19 @@ def main():
...
+    if args.direct and not (args.group or args.default):
+        # This option is intended for early admin access list management before
+        # the GEOPM service is running. The option does not have a meaningful
+        # use case when run for the root user.
+        parser.error('Must specify either --group or --default with --direct.')
+
     try:
-        acc = Access(SystemMessageBus().get_proxy('io.github.geopm',
-                                                  '/io/github/geopm'))
+        geopm_proxy = (None if args.direct
```
dannosliwcd: As an alternative to `None`, we could add a `DirectAccessProxy` that implements the proxy interface but makes direct calls. Doing so would mean we can avoid changes to the `TestAccess.py` file

dannosliwcd: That would also remove all of the other `if none` checks throughout this file

cmcantalupo: that seems like a cleaner approach.  I think we can wait until after monday for this change.  Thanks!

cmcantalupo (approved):
LGTM!

PR #3145, opened 2023-09-25
Use NamedTemporaryFile to create temporary files securely

### LINKED ISSUE #3144: Use NamedTemporaryFiles instead of attempting to create a secure named temporary file manually
The implementation for the secure file write feature was modified because of duplicate effort with the Python standard library.

Done criteria:

- [x] Manual testing - the behavior was tested manually and aligns with the expected behavior in the docstring
- [x] Documentation changes - updated docstring to reflect change

tjs-intel:
- Fixes #3144 change request from github issues.

The temporary file creation for secure writes is duplicate effort with the Python standard library. Using the standard library is preferable because it is studied carefully by security researchers.

### DIFF
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -30,6 +30,7 @@ import grp
 import pwd
 import fcntl
 import subprocess # nosec
+import tempfile
 from functools import lru_cache
 
 from . import pio
@@ -161,12 +162,14 @@ def secure_make_file(path, contents):
 
     The file is created by first writing the output to a temporary
     file with restricted permissions (mode 0o600) within the same
-    directory, and then renaming the file after it is closed.  This
-    temporary file with have the path `<path>-<uuid>-tmp`.
+    directory, and then renaming the file after it is closed.
 
     The rename operation is atomic, and will overwrite any existing
     file located in the specified path.  The permissions on the output
     file are mode 0o600 with uid and gid matching the process values.
+
+    The temporary file has the prefix `tmp.`. This file may persist if
+    the program exits abruptly.
 
     Args:
         path (str): The path where the file is created
@@ -174,12 +177,13 @@ def secure_make_file(path, contents):
         contents (str): The contents of the created file
 
     """
-    temp_path = f'{path}-{uuid.uuid4()}-tmp'
+    directory = os.path.dirname(os.path.abspath(path))
+    prefix = f'{os.path.basename(path)}-tmp-'
     old_mask = os.umask(0o077)
     try:
-        with open(os.open(temp_path, os.O_CREAT | os.O_WRONLY, 0o600), 'w') as file:
+        with tempfile.NamedTemporaryFile(mode="w+", prefix=prefix, dir=directory, delete=False) as file:
             file.write(contents)
-        os.rename(temp_path, path)
+        os.rename(file.name, path)
     finally:
         os.umask(old_mask)
 


service/geopmdpy/system_files.py
```diff
@@ -161,25 +162,26 @@ def secure_make_file(path, contents):
...
 
         contents (str): The contents of the created file
 
     """
-    temp_path = f'{path}-{uuid.uuid4()}-tmp'
-    old_mask = os.umask(0o077)
+    old_mask = os.umask(0o477)
     try:
-        with open(os.open(temp_path, os.O_CREAT | os.O_WRONLY, 0o600), 'w') as file:
+        with tempfile.NamedTemporaryFile(mode="w+", prefix="tmp.", dir=os.path.dirname(path), delete=False) as file:
```
cmcantalupo, line 182: I think the reason for using os.open() directly is to enable the use of the flags here.  Does NamedTemporaryFile use your umask, or does it always create with 0o600 permissions?

tjs-intel:
The permission 0o600 is hardcoded in the NamedTemporaryFile implementation https://hg.python.org/cpython/file/63bde882e311/Lib/tempfile.py#l235

Is this an issue? I suppose this means the umask no longer has any effect.

cmcantalupo:
Yeah, I commented, and then looked up the docs:
https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

Looks like it should be effectively the same as adding the flags to open(), so seems like a good change.  You will have to update the unit tests however.  There are several broken in CI.

tjs-intel:
Looking into broken tests now

cmcantalupo:
```
FAIL: test_context_manager_restores_backup (__main__.TestRestorableFileWriter)
  File "/home/runner/work/geopm/geopm/service/geopmdpy_test/TestRestorableFileWriter.py", line 134, in test_context_manager_restores_backup
    self.assertCountEqual([], warnings)
AssertionError: Element counts were not equal:
First has 0, Second has 1:  "Encountered exception in secure read of /tmp/tmpqk4_98g_TestRestorableFileWriter/__main__.TestRestorableFileWriter.test_context_manager_restores_backup.backup: [Errno 13] Permission denied: '/tmp/tmpqk4_98g_TestRestorableFileWriter/__main__.TestRestorableFileWriter.test_context_manager_restores_backup.backup'"
First has 0, Second has 1:  'Unable to securely read /tmp/tmpqk4_98g_TestRestorableFileWriter/__main__.TestRestorableFileWriter.test_context_manager_restores_backup.backup'

Ran 1 test in 0.001s

FAILED (failures=1)
test_context_manager_warns_on_cleanup_failure (__main__.TestRestorableFileWriter) ... FAIL

FAIL: test_context_manager_warns_on_cleanup_failure (__main__.TestRestorableFileWriter)
  File "/home/runner/work/geopm/geopm/service/geopmdpy_test/TestRestorableFileWriter.py", line 156, in test_context_manager_warns_on_cleanup_failure
    self.assertEqual(1, len(warnings))
AssertionError: 1 != 2

Ran 1 test in 0.001s

FAILED (failures=1)
test_modify_does_not_overwrite_existing_backup (__main__.TestRestorableFileWriter) ... ERROR

ERROR: test_modify_does_not_overwrite_existing_backup (__main__.TestRestorableFileWriter)
  File "/home/runner/work/geopm/geopm/service/geopmdpy_test/TestRestorableFileWriter.py", line 74, in test_modify_does_not_overwrite_existing_backup
    with open(self._backup_path) as f:
PermissionError: [Errno 13] Permission denied: '/tmp/tmpz74w5h2mTestRestorableFileWriter/__main__.TestRestorableFileWriter.test_modify_does_not_overwrite_existing_backup.backup'

```
Seems to be trying to create a file in /tmp.  We would like these temporary files created in /run/geopm/ so that they cannot be listed.

tjs-intel:
I have a hypothesis of the issue, `os.path.dirname("test")` yields `''`, which `NamedTemporaryFile` might interpret as "use the default". I should canonicalize the path before passing it to `os.path.dirname`.

tjs-intel:
It looks like reverting the umask was required for the CI tests. Ready to merge I think

cmcantalupo:
Great, the permissions on the output file are enforced by mkstemp:

https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp
```
The file is readable and writable only by the creating user ID. If the platform uses permission bits to indicate whether a file is executable, the file is executable by no one. 
```

so we should avoid the umask() call (it changes global state so we need exception management when it is used).

cmcantalupo:
I've squashed history and added a small change for traceability:
https://github.com/geopm/geopm/pull/3145/commits/6418de88c6e9eb2dbffbbd84b345f814215156ce

cmcantalupo (approved):
LGTM.  Thanks!

bgeltz (approved):
LGTM; I'm going to rebase on the tip of dev to try and clean up CI.

I would like to understand the nuances of the umask exploration you were both on about.

service/geopmdpy/system_files.py
```diff
@@ -161,25 +162,28 @@ def secure_make_file(path, contents):
...
     Args:
         path (str): The path where the file is created
 
         contents (str): The contents of the created file
 
     """
-    temp_path = f'{path}-{uuid.uuid4()}-tmp'
+    directory = os.path.dirname(os.path.abspath(path))
+    prefix = f'{os.path.basename(path)}-tmp-'
     old_mask = os.umask(0o077)
```
cmcantalupo, line 182: This existing call to umask had no impact on the old code, it also has no impact on the new code.  In both cases the open() call is made with mode set to 0o600, so masking off the "other" and "group" modes has no impact.  If we wanted to remove the umask() call we could, but it also does no harm.

bgeltz (approved):

Issue #3129, opened 2023-09-20
Product release validation on multiple systems


cmcantalupo:
> estimate 2

- [x] Run integration tests on Aurora-type system
- [x] Validate geopm-runtime packages
- [x] Go through any examples in web documenation
- [x] Run hand testing on --geopm-ctl-local flag when MPI is and is not enabled at configure time

PR #3156, opened 2023-09-27
Fix integration tests on HPCM stack

bgeltz:
- Relates to #3129.


lhlawson (commented):
LEVELZERO signal exclusions and helper text updates LGTM.

bgeltz (commented):
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
https://gcc.gnu.org/onlinedocs/gcc/Standards.html

integration/test/test_gpu_activity_agent.py
```diff
@@ -26,8 +26,8 @@
 
 @util.skip_unless_do_launch()
 @util.skip_unless_gpu()
-@util.skip_unless_workload_exists("apps/parres/Kernels/Cxx11/dgemm-mpi-cublas")
-@util.skip_unless_workload_exists("apps/parres/Kernels/Cxx11/nstream-mpi-cuda")
+@util.skip_unless_workload_exists("apps/parres/Kernels/Cxx11/dgemm-multigpu-onemkl")
```
lhlawson:
```suggestion
@util.skip_unless_workload_exists("apps/parres/Kernels/Cxx11/dgemm-onemkl")
```

scripts/geopmpy/io.py
```diff
@@ -877,7 +877,10 @@ def load_reports(self, reports, dir_name, dir_cache, verbose, do_cache):
                     self._unmarked_reports_df = pandas.read_hdf(self._report_h5_name, 'unmarked_report')
                 except:
                     self._unmarked_reports_df = self._reports_df.loc[self._reports_df['region'] == 'unmarked-region']
-                self._epoch_reports_df = pandas.read_hdf(self._report_h5_name, 'epoch_report')
+                try:
+                    self._epoch_reports_df = pandas.read_hdf(self._report_h5_name, 'epoch_report')
+                except:
+                    pass
```
cmcantalupo, line 883: Can we be more specific about what exception is being handled here, and maybe print a warning?

scripts/geopmpy/io.py
```diff
@@ -1026,13 +1029,16 @@ def _try_float(val):
...
+                try:
+                    epoch_data = rr.raw_epoch(host)
+                    for key, val in epoch_data.items():
+                        epoch_data[key] = _try_float(val)
+                    for cc in epoch_data.keys():
+                        _add_column('epoch', cc)
+                    epoch_row.update(epoch_data)
+                    epoch_df_list.append(pandas.DataFrame(epoch_row, index=[0]))
+                except:
+                    pass
```
cmcantalupo, line 1042: Similar issue as on line 882

scripts/geopmpy/io.py
```diff
@@ -1057,9 +1063,12 @@ def _try_float(val):
...
         self._unmarked_reports_df = unmarked_df
-        epoch_df = pandas.concat(epoch_df_list, ignore_index=True)
-        epoch_df = epoch_df.reindex(columns=self._columns_order['epoch'])
-        self._epoch_reports_df = epoch_df
+        try:
+            epoch_df = pandas.concat(epoch_df_list, ignore_index=True)
+            epoch_df = epoch_df.reindex(columns=self._columns_order['epoch'])
+            self._epoch_reports_df = epoch_df
+        except:
+            pass
```
cmcantalupo, line 1071: Again...

scripts/geopmpy/launcher.py
```diff
@@ -918,25 +920,15 @@ def get_idle_nodes(self):
         Returns a list of the names of compute nodes that are currently
         available to run jobs.
         """
-        if self._is_slurm_enabled:
```
cmcantalupo: i'm not sure i understand this change.  I think the original idea is that you may have Slurm as a *resource manager* but use, say, mpiexec from Intel MPI to do *application launch*.  In this way it does make sense to have logic about Slurm in the base class.

cmcantalupo (commented):
Some questions about parts of the code where all exceptions are quietly suppressed.  Also not  sure about how the Launcher base class should abstract the resource manager as opposed to the MPI application launcher.

scripts/geopmpy/io.py
bgeltz: Issue opened to deal with this everywhere: https://github.com/geopm/geopm/issues/3193

cmcantalupo (dismissed):
LGTM

cmcantalupo (approved):

PR #3180, opened 2023-10-04
Add exception handling for all calls to _pid_info()

### LINKED ISSUE #3179: Issues with PID validation in certain cases
**Describe the bug**
I tried to run the example to profile sleep(1) and I expected a run without issue instead multiple failures occurred. 

**GEOPM version**
https://github.com/geopm/geopm/commit/b2add9c57e24faa15486e0532c91c5ccf90c9136

**Expected behavior**
Report and trace generated without error.

**Actual behavior**
TBD

**Error messages produced**
TBD

cmcantalupo:
- A process may terminate while a callback is being executed.
- In these cases warn to syslog
- Do not check if a PID is active when destroying resources (batch server, and profiling)
- Fixes #3179

### DIFF
--- a/service/geopmdpy/system_files.py
+++ b/service/geopmdpy/system_files.py
@@ -431,27 +431,35 @@ class ActiveSessions(object):
         is_registered = client_pid in self._sessions
         is_valid = False
         if is_registered:
-            uid, gid, create_time = self._pid_info(client_pid)
-            session_uid = self._sessions[client_pid]['client_uid']
-            session_gid = self._sessions[client_pid]['client_gid']
-            session_time = self._sessions[client_pid]['create_time']
-            is_valid = session_uid == uid and session_gid == gid and session_time == create_time
-            if not is_valid:
+            try:
+                uid, gid, create_time = self._pid_info(client_pid)
+                session_uid = self._sessions[client_pid]['client_uid']
+                session_gid = self._sessions[client_pid]['client_gid']
+                session_time = self._sessions[client_pid]['create_time']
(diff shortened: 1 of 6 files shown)


service/geopmdpy/system_files.py
```diff
@@ -813,7 +822,11 @@ def start_profile(self, client_pid, profile_name):
         self.check_client_active(client_pid, 'start_profile')
         if 'profile_name' in self._sessions[client_pid]:
             raise RuntimeError(f'Client pid {client_pid} has requested profiling twice')
-        uid, gid, _ = self._pid_info(client_pid)
+        try:
+            uid, gid, _ = self._pid_info(client_pid)
+        except psutil.NoSuchProcess:
+            sys.stderr.write('Warning: Profile cannot be created for PID {client_pid}, it is no longer active')
```
bgeltz, line 828: The only way I can see to hit this is if the client_pid dies between the check_client_active() call on 813 and here.
If the client_pid is dead here, whose responsibility is it to call remove_client() to get it out of self._sessions?

src/ApplicationSampler.cpp
```diff
@@ -507,7 +507,7 @@ namespace geopm
             std::cerr << "." << std::endl;
         }
 
-        if (!found_inactive_cpu) {
+        if (!found_inactive_cpu && ! is_slow_monitor) {
```
bgeltz:
```suggestion
        if (!found_inactive_cpu && !is_slow_monitor) {
```

bgeltz (approved):

PR #3182, opened 2023-10-04
Address function signature mismatches in Controller and CSV

### LINKED ISSUE #3181: Address function signature mismatches
> estimate 1

Static analysis tools revealed a mismatch in function signatures in Controller.cpp and CSV.cpp. Ensure function signatures match and update affected code/tests.

avilcheslopez:
- Fixes #3181


src/Controller.cpp
```diff
@@ -214,8 +214,8 @@ namespace geopm
     Controller::Controller(std::shared_ptr<Comm> comm,
                            PlatformIO &plat_io,
                            const std::string &agent_name,
-                           int num_send_down,
                            int num_send_up,
+                           int num_send_down,
```
dannosliwcd: These should map to the number of agent sample outputs (going up) and agent policy inputs (going down) but the only changes to callers of this function are in test code. Check to make sure there aren't other users of the constructor that need to be updated.

dannosliwcd (dismissed):

src/Controller.cpp
avilcheslopez: Found one spot in the code that needed to be updated. Should be good to go now.

bgeltz: I'd prefer we don't make any changes to the Controller this close to the tag.

bgeltz (changes requested):
Instead of modifying the implementation, please change the documentation (and header if necessary) to match what we currently have implemented.  It is very risky to change the Controller implementation this close to a release.

src/Controller.cpp
avilcheslopez: Reverted the changes to the Controller interface, updated implementation and tests accordingly.

bgeltz (commented):
LGTM; I'm going to run it through the test-service integration tests to be sure, then merge if OK.

bgeltz (approved):

PR #3185, opened 2023-10-09
Clean up StandaloneFuzzTargetMain.c

cmcantalupo:


service/fuzz_test/StandaloneFuzzTargetMain.cpp
bgeltz, line 1: Line 1 should be updated if you're going to rename this file?  Maybe that whole comment block should be updated to reflect that this is a rewrite of that upstream source.

bgeltz (dismissed):

avilcheslopez (approved):

PR #3243, opened 2023-11-16
Add script to combine config files

### LINKED ISSUE #3232: As a user of the const config files I would like to be able to combine files derived from different hosts
In order to support building up a single file with characterization of all compute nodes from a set files with the  characterization for each node included a simple python script should be added.

The script will be added to the integration infrastructure around the system characterization tools.

- Use a directory with a single file per host and load only the file pertaining to the host where the process is executing
- This has two down sides
  +  administrators have to manage thousands of files
  + global parameters would be repeated in every file.

**Additional context**

cmcantalupo:
- Simple python script to create a host specific version of input file
- Can be used to append to existing file
- Fixes #3232


integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,49 @@
...
+#
+
+import sys
+import json
+from argparse import ArgumentParser
+
+
+def run(host_name, append_file):
+    if host_name is None:
+        raise RuntimeError('The command line option -n/--host-name is required')
```
avilcheslopez: Make this a required argument

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,49 @@
...
+#  Copyright (c) 2015 - 2023, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+import sys
+import json
+from argparse import ArgumentParser
+
+
+def run(host_name, append_file):
```
avilcheslopez: I'm thinking it would be great if this feature could be leveraged by the integration tests or other scripts (in addition to being available as a command-line tool). This implementation doesn't prevent that, but thinking of that idea, perhaps a more descriptive function name would be helpful.

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,49 @@
...
+    if host_name is None:
+        raise RuntimeError('The command line option -n/--host-name is required')
+    host_data = json.load(sys.stdin)
```
avilcheslopez: Taking the config from `stdin` would be a useful use case. However, thinking of how this might be used, we should also support taking it from a file. Let's think of the different use cases that might arise, but the one I'm thinking of is: all the individual config files will be generated separately and dumped somewhere, and then run through this script to combine them. I'm also thinking it would be good to support appending one or multiple files at a time.

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
avilcheslopez, line 1: I like the idea of creating a `system_characterization` directory. We could move the existing scripts related to system characterization here. We have the new `[cpu|gpu]_ca_characterization.py` scripts and config generation scripts (currently embedded in experiment-specific directories).

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,49 @@
...
+    if host_name is None:
+        raise RuntimeError('The command line option -n/--host-name is required')
+    host_data = json.load(sys.stdin)
+    if append_file:
+        with open(append_file) as fid:
+            append_data = json.load(fid)
+    else:
+        append_data = dict()
+    if type(host_data) is not dict:
+        raise TypeError('Host file is not json object')
```
avilcheslopez: Nit: json -> JSON?

integration/experiment/system_characterization/host_config.py
cmcantalupo: Agreed!

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,63 @@
...
+import os
+from argparse import ArgumentParser
+
+
+def combine(directory, append_file):
+    if append_file:
+        with open(append_file) as fid:
+            output = json.load(fid)
+        if type(output) is not dict:
+            raise TypeError('Append file is not json object')
```
cmcantalupo:
```suggestion
            raise TypeError('Append file is not JSON object')
```

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,93 @@
...
+    else:
+        output = dict()
+    host_files = glob.glob(f'{directory}/*')
+    for host_file in host_files:
+        host_name = os.path.basename(host_file).split('.json')[0]
+        try:
+            with open(host_file) as fid:
+                host_data = json.load(fid)
+        except json.decoder.JSONDecodeError:
+            sys.stderr.write(f'Warning: File {host_file} is not a JSON file, it will be ignored\n')
```
avilcheslopez: Should we treat this (and other warnings below) as an error?

I'm thinking of a scenario where the user is running this within a script and some files get skipped (unintentionally) without the user noticing the warning message.

integration/experiment/system_characterization/test_host_config.py
```diff
@@ -0,0 +1,112 @@
...
+        for path in self.all_paths:
+            os.unlink(path)
+        os.rmdir(self.test_dir)
+
+    def test_combine(self):
+        """Check that the combined output is as expected
+        """
+        result = host_config.combine(self.test_dir, self.global_config_path)
+        self.assertEqual(self.expected, result)
+
```
avilcheslopez, line 149: Some additional tests to consider:
- Script accepts an empty append file
- Append file is created when not provided
- Script raises error when append file is not in JSON format
- Script raises error when append file doesn't contain a top-level JSON object
- Script raises [warning|error] when host file is not in JSON format
- Script raises [warning|error] when host file doesn't contain a top-level JSON object

integration/experiment/system_characterization/host_config.py
```diff
@@ -0,0 +1,93 @@
...
+
+    """
+    err = 0
+    prog = sys.argv[0]
+    parser = ArgumentParser(description=main.__doc__)
+    parser.add_argument('directory', type=str,
+                        help='Directory containing a configuration file per host')
+    parser.add_argument('-a', '--append', dest='file', type=str,
+                        help='File with data that is not host specific to be appended without modification')
+    opts = parser.parse_args(sys.argv[1:])
```
avilcheslopez, line 76: A feature idea just came to mind: how about we add an option/flag so the user can optionally request validating the JSON content against the ConstConfigIO config file schema? I say "JSON content" generally because there are a couple of approaches we could consider.

avilcheslopez (commented):
Looking good!

integration/experiment/system_characterization/host_config.py
cmcantalupo: I made these errors, but now we force the files to have the .json extension.  This will still allow for someone to populate the directory with a README or other documentation.

integration/experiment/system_characterization/host_config.py
cmcantalupo: I have added schema validation which makes the code simpler (we don't have to check that a json file is an object for instance).

integration/experiment/system_characterization/test_host_config.py
cmcantalupo: I've added tests for when a JSON file is not properly formatted.

avilcheslopez (approved):

PR #3247, opened 2023-11-16
Add soft dependency on msr-safe to service

### LINKED ISSUE #3246: geopmread shows no MSR signals even though msr-safe is loaded
**Describe the bug**
I tried to use geopmread to display all the signals and I expected to see MSR signals instead none were present.

**GEOPM version**
3.0

**Expected behavior**
When I count the number of available signals with "MSR" in the title, it is non-zero:
```
[bgeltz@mcfly11 ~]$ geopmread | grep MSR | wc -l
456
```

**Actual behavior**
```
[bgeltz@mcfly12 ~]$ geopmread | grep MSR | wc -l
0
```

**Error messages produced**
The journal shows:
```
Nov 16 10:50:19 mcfly12 systemd[1]: Starting Global Extensible Open Power Manager Service...
Nov 16 10:50:19 mcfly12 geopmd[3634]: Warning <geopm-service> Using an old GEOPM configuration directory. The directory at "/etc/geopm-service" is deprecated and may not be used in future GEOPM releases. Migrate configuration data to the directory at "/etc/geopm"
Nov 16 10:50:20 mcfly12 systemd[1]: Started Global Extensible Open Power Manager Service.
Nov 16 14:05:36 mcfly12 geopmd[3634]: Warning: <geopm> Unable to read MSR configuration from /etc/geopm. Reason: <geopm> Runtime error: Helper::list_directory_files(): failed to open directory '/etc/geopm': No such file or directory: at src/Helper.cpp:146
```

**Additional context**
This was on mcfly after installing the 3.0 RPMs in the compute image, and no other configuration changes.  The access lists still reside in ```/etc/geopm-service```.

This behavior is only evident at boot-up.  If you restart the geopm service, it goes away.

There appears to be a race between the msr-safe.service and the geopm.service.  If the msr-safe.service is not started before geopm.service, then the MSRIOGroup will fail to load because no msr-safe access list has been loaded.

bgeltz:
- Resolves #3246.
- https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html

### DIFF
--- a/service/geopm.service
+++ b/service/geopm.service
@@ -6,6 +6,8 @@
 Description=Global Extensible Open Power Manager Service
 StartLimitIntervalSec=0
 StartLimitBurst=3
+Wants=msr-safe.service
+After=msr-safe.service
 
 [Service]
 Environment=PYTHONUNBUFFERED=true


service/geopm.service
```diff
@@ -6,6 +6,8 @@
 Description=Global Extensible Open Power Manager Service
 StartLimitIntervalSec=0
 StartLimitBurst=3
+Wants=msr-safe.service
```
bgeltz, line 9: I could see an argument for using "Requires=" here instead of "Wants=", but that would make the msr-safe.service a hard dependency.
The way this is now, if the msr-safe.service is not present, it will *not* block the geopm.service from starting.

cmcantalupo (approved):

PR #3251, opened 2023-11-21
Update geopmsession with features from cloud branch

dannosliwcd:
- Enable PID monitoring in geopmsession 
- Optionally print CSV headers in geopmsession

### DIFF
--- a/service/geopmdpy_test/TestSession.py
+++ b/service/geopmdpy_test/TestSession.py
@@ -158,11 +158,31 @@ class TestSession(unittest.TestCase):
                         return_value=rrq_return_value) as srrq, \
              mock.patch('geopmdpy.session.Session.check_read_args') as scra, \
              mock.patch('geopmdpy.session.Session.run_read') as srr:
-            self._session.run(runtime, period, None, request_stream, out_stream)
+            self._session.run(runtime, period, None, False, request_stream, out_stream)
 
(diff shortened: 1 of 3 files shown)


service/geopmdpy_test/TestSession.py
```diff
@@ -101,11 +158,31 @@ def test_run(self):
...
+             mock.patch('geopmdpy.session.Session.check_read_args') as scra, \
+             mock.patch('geopmdpy.topo.domain_name', side_effect=lambda x: x), \
+             mock.patch('geopmdpy.session.Session.run_read') as srr:
+            self._session.run(runtime, period, None, True, request_stream, out_stream)
 
             srrq.assert_called_once_with(request_stream)
             scra.assert_called_once_with(runtime, period)
-            srr.assert_called_once_with(rrq_return_value, runtime, period, out_stream)
+            srr.assert_called_once_with(rrq_return_value, runtime, period, None, out_stream)
+        self.assertEqual('"signal1-board-0","signal2-package-1"\n', out_stream.getvalue())
```
cmcantalupo: FWIW we omit the "-board-0" extension in the existing Tracer implemenation:

https://github.com/geopm/geopm/blob/dev/src/Tracer.cpp#L128-L131

I'm not sure which i prefer...

cmcantalupo (approved):

PR #3286, opened 2023-12-11
Make two different MSRIOGroup factory functions

### LINKED ISSUE #3273: Create distinct IOGroup objects for msr and msr-safe device drivers and make the msr driver least preference for aliases

cmcantalupo:
- Register two distinct IOGroups, one for the msr driver and another for the msr-safe driver.
- Put stock msr driver as IOGroup of least preference.
- Make MSRIO construction depend on driver selection
- Fallback logic is disabled.
- Fixes #3273

### DIFF
--- a/service/configure.ac
+++ b/service/configure.ac
@@ -134,6 +134,27 @@ else
   ENABLE_LIBCAP=False
 fi
 
+AC_ARG_ENABLE([rawmsr],
+  [AS_HELP_STRING([--disable-rawmsr], [Disable direct use of standard msr(4) device driver])],
+[if test "x$enable_rawmsr" = "xno" ; then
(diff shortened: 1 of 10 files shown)


dannosliwcd (dismissed):
Looks good. Just need to change the default build options to match the current behavior. Either add a warning that this default may change in the future, or open an issue targeted to the next release.

Warning may be tricky. I guess we'd want to warn once per geopmd process. Not sure if/when we'd want to log in client geopmd.so users.

cmcantalupo:
Posted the deprecation warning issue here:

https://github.com/geopm/geopm/issues/3288

bgeltz (approved):

PR #3287, opened 2023-12-12
Remove crc32 dependency

### LINKED ISSUE #3290: As a user of non-x86 platforms I would prefer not to have a dependency on SSE and the crc32 instruction
> estimate 1

Replace the crc32 instruction wrapper with std::hash() equivalent and change build requirements to avoid sse2 requirements.

cmcantalupo:
This pull request enables users to build on non-x86 architectures.

- Fixes #3290


bgeltz (commented):
LGTM; I'll set it up in the nightlies

bgeltz:
This is causing a failure in the nightlies:
```
FAIL: test_zero_progress (integration.test.test_progress.TestIntegration_progress)
Test that all regions that do not contain the geopm_tprof_post()
  File "/home/test/geopm/integration/test/test_progress.py", line 280, in test_zero_progress
    self.assertEqual(1, len(uu), msg=err_msg)
AssertionError: 1 != 2 : Expected all zero, group 0x725e8066 had [0.       0.999931]
```

and from another run:
```
FAIL: test_zero_progress (integration.test.test_progress.TestIntegration_progress)
Test that all regions that do not contain the geopm_tprof_post()
  File "/home/test-service/geopm/integration/test/test_progress.py", line 280, in test_zero_progress
    self.assertEqual(1, len(uu), msg=err_msg)
AssertionError: 1 != 16 : Expected all zero, group 0x725e8066 had [0.       0.963326 0.965854 0.968358 0.970919 0.973485 0.976033 0.978525
 0.981079 0.983649 0.986196 0.988715 0.991266 0.993853 0.996412 0.99891 ]
```

cmcantalupo:
Integration test has been fixed.  There was an assumption about how to get the constant hashes defined in C by hashing the enum.  This is no longer valid.

https://github.com/geopm/geopm/pull/3287/files#diff-020f6843b9866fc3ea8fbf80f91d51755a326a0169495e9fb165621c3304771eL269-R269

cmcantalupo:
Since the requirements of std::hash are not strict enough (not required to be the same for two different processes), we cannot use this approach.

PR #3297, opened 2024-01-02
Make use of cpuid.h optional at configure time

cmcantalupo:
merge after #3296 because this PR removes the ExclusiveArch flag from the spec files.

### DIFF
--- a/service/configure.ac
+++ b/service/configure.ac
@@ -482,6 +482,10 @@ if test "x$enable_io_uring" = "x1" ; then
         [#define _DEFAULT_SOURCE
          #define _POSIX_C_SOURCE 200809L
          #include <liburing.h>])
+    AC_CHECK_LIB([uring], [io_uring_free_probe], [liburing_has_free=1], [liburing_has_free=0])
+    if test "x$liburing_has_free" = "x1"; then
+        AC_DEFINE([GEOPM_IO_URING_HAS_FREE], [ ], [Use io_uring_free_probe instead of free])
+    fi
 fi
 
 AC_CHECK_HEADERS([fcntl.h limits.h stddef.h stdint.h stdlib.h string.h sys/ioctl.h unistd.h])
--- a/service/configure.ac
+++ b/service/configure.ac
@@ -482,6 +482,11 @@ if test "x$enable_io_uring" = "x1" ; then
         [#define _DEFAULT_SOURCE
          #define _POSIX_C_SOURCE 200809L
          #include <liburing.h>])
+
+    # Newer versions of liburing provide an function to release probe
+    # resources, but with older versions the user must call free(3) to
+    # release the resources.  This check sets a preprocessor define
+    # to identify which is required.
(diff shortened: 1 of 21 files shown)


service/test/CpuinfoIOGroupTest.cpp
```diff
@@ -32,6 +21,7 @@ extern "C"
 #include "geopm/PluginFactory.hpp"
 #include "geopm/PlatformTopo.hpp"
 #include "geopm_test.hpp"
+#include "MockCpuid.hpp"
```
dannosliwcd: I think this newly-added header is unused. Did you plan to add the new cpuid class to this the CpuinfoIOGroup interface, or is the header left over from an earlier design?

cmcantalupo: Left overs! Thanks

service/configure.ac
```diff
@@ -454,6 +482,10 @@ if test "x$enable_io_uring" = "x1" ; then
         [#define _DEFAULT_SOURCE
          #define _POSIX_C_SOURCE 200809L
          #include <liburing.h>])
+    AC_CHECK_LIB([uring], [io_uring_free_probe], [liburing_has_free=1], [liburing_has_free=0])
+    if test "x$liburing_has_free" = "x1"; then
+        AC_DEFINE([GEOPM_IO_URING_HAS_FREE], [ ], [Use io_uring_free_probe instead of free])
```
dannosliwcd, line 492: Maybe add a comment about this since there's no issue to track the problem we encountered. I guess key bits are that some versions of liburing expect us to use free and other versions give us the free wrapper. When we use the wrong one, the program aborts.

service/test/CpuinfoIOGroupTest.cpp
```diff
@@ -10,17 +10,6 @@
...
-    static int mock_cpuid(unsigned int leaf, unsigned int *eax, unsigned int *ebx,
-                          unsigned int *ecx, unsigned int *edx)
-    {
-        *eax = g_cpuid_sticker;
-        return 0;
-    }
-#define __get_cpuid(p0, p1, p2, p3, p4) mock_cpuid(p0, p1, p2, p3, p4)
-}
 #define GEOPM_TEST
 #include "src/CpuinfoIOGroup.cpp"
```
dannosliwcd: I suspect we don't need this "include the implementation" pattern here any more. This is pre-existing, so I don't think it must be addressed now (maybe record in a tech debt issue). It might still be needed if any of the test cases rely on parts of the interface hidden from the header.

I think the reason we needed it before was so we could map the cpuid functions to our mock implementations at preprocessing time. Now that we go through an object wrapper, we can use our usual execution-time dependency injection instead. It might still be needed if any of the test cases rely on parts of the interface hidden from the header.

This is pretty low priority. I think the main related tech-debt issue is that these kinds of implementation dependencies would make it more difficult for us to separate public and non-public components of our interface.

dannosliwcd (approved):

PR #3333, opened 2024-01-20
Resolve issues around profiling unexpected processes

### LINKED ISSUE #3332: Arbitrary process is emitting ECHILD when LD_PRELOAD and GEOPM_PROGRAM_FILTER are used
**Describe the bug**
I tried to use GEOPM's runtime with LD_PRELOAD and GEOPM_PROGRAM_FILTER and I expected no errors in the log for processes that should *not* be profiled.  Instead processes that should not be profiled are emitting an sd-bus error.

**GEOPM version**
c47d5d315

**Expected behavior**
Reproducer script (repro.sh):
```
#!/bin/bash

echo "Outer PID = $$"

# Having a pipeline was required to reproduce error
echo $(cat /proc/self/status | grep ^Pid)
```

Expected output:
```
[bgeltz@node0 ~]$ LD_PRELOAD=libgeopm.so.2 GEOPM_PROGRAM_FILTER=testing ./repro.sh
Outer PID = 204547
Pid: 204549
```

**Actual behavior**
```
[bgeltz@node0 ~]$ LD_PRELOAD=libgeopm.so.2 GEOPM_PROGRAM_FILTER=testing ./repro.sh
Outer PID = 204547
  what():  <geopm> Runtime error: SDBus: Failed to call sd-bus function sd_bus_call(), error:-10 name: System.Error.ECHILD: No child processes: at src/SDBus.cpp:32
Pid: 204549
```

**Additional context**
It appears as though the ```GEOPM_PROGRAM_FILTER``` env var is not being propagated in some instances where ```LD_PRELOAD``` is.  And since ```LD_PRELOAD``` is being propagated (to the bash interpreter that is running ```repro.sh```, and to the subshell in the script) the runtime seems to be trying to profile processes that should be ignored.

### LINKED ISSUE #3334: No output files generated from tutorials when runtime build with MPI disabled
**Describe the bug**
I tried to exercise the tutorials and I expected them to produce reports/traces with the Runtime instead the scripts run to completion, but no GEOPM Runtime output is created.

**GEOPM version**
c47d5d315

**Expected behavior**
```
$ ./tutorial_0.sh
MPI_COMM_WORLD size: 8
Sleeping for five seconds

$ ls *report* *trace*
tutorial_0_report_x1001c1s0b0n0  tutorial_0_trace-x1001c1s0b0n0  tutorial_0_trace-x1001c1s2b0n0
```

**Actual behavior**
```
$ ./tutorial_0.sh
MPI_COMM_WORLD size: 10
Sleeping for five seconds

$ ls *report* *trace*
ls: cannot access '*report*': No such file or directory
ls: cannot access '*trace*': No such file or directory
```

**Additional context**
This is seen when testing the tutorials against the OBS RPMs for the Runtime or configuring with ```--disable-mpi``` manually.  The OBS RPMs are setup to disable MPI by default.

When executing the Runtime this way, the Controller is never initialized.  This is taken care of with ```MPI_Init()``` when MPI is enabled.  Because the Controller is never initialized, no profiling occurs.

Also note that the "Actual Behavior" case has 2 more ranks reported in ```MPI_COMM_WORLD``` (10 instead of 8).  These 2 extra ranks (1 per node) are the ranks that the Controller should have been executing on, but since it was never initialized, are instead exposed to the app.

### LINKED ISSUE #3335: Multi-node runs only produce a single trace file when MPI is disabled in the runtime
**Describe the bug**
I tried to build the Runtime without MPI support and I expected a trace file to be generated for each host the app ran on, instead only a single trace file was generated.  The second host's data is lost.

**GEOPM version**
https://github.com/geopm/geopm/commit/c47d5d315c81d549ac542a716538e5ecdabe4302

**Expected behavior**
Build the Runtime with ```--disable-mpi``` at configure time.
```
$ ./tutorial_0.sh
MPI_COMM_WORLD size: 8
Sleeping for five seconds

$ ls *trace*
tutorial_0_trace-x1001c1s0b0n0  tutorial_0_trace-x1001c1s2b0n0
```

**Actual behavior**
```
$ ./tutorial_0.sh
MPI_COMM_WORLD size: 8
Sleeping for five seconds

$ ls *trace*
tutorial_0_trace
```

**Additional context**
In the above case, I modified ```tutorial_0.sh``` to run when the Controller is configured with ```--disable-mpi```.  ```--geopm-ctl=application``` is required, along with the ```GEOPM_PROGRAM_FILTER``` opt-in changes on this branch: https://github.com/bgeltz/geopm/tree/issue-3332

bgeltz:
- Resolves #3332.
- Resolves #3334.
- Resolves #3335.

### DIFF
--- a/service/docs/source/overview.rst
+++ b/service/docs/source/overview.rst
@@ -791,7 +791,7 @@ Profiling Non-MPI Applications
 The ``geopmlaunch(1)`` command may not be best suited for your needs
 if you are running a non-MPI application, or if you are running an MPI
 application but the launch command is embedded in scripts that are
-difficult to modify.  Alternately to using ``geopmlaunch(1)``, the
+difficult to modify.  Instead of using ``geopmlaunch(1)``, the
 user may use the ``geopmctl(1)`` application in conjunction with
 environment variables that control the GEOPM Runtime behavior.
 
--- a/service/docs/source/overview.rst
+++ b/service/docs/source/overview.rst
@@ -788,6 +788,13 @@ option to ``geopmlaunch``.  For more information about ``geompmlaunch`` see:
 Profiling Non-MPI Applications
 """"""""""""""""""""""""""""""
 
+The ``geopmlaunch(1)`` command may not be best suited for your needs
+if you are running a non-MPI application, or if you are running an MPI
+application but the launch command is embedded in scripts that are
+difficult to modify.  Alternately to using ``geopmlaunch(1)``, the
+user may use the ``geopmctl(1)`` application in conjunction with
+environment variables that control the GEOPM Runtime behavior.
+
 In this simple example we run the ``sleep(1)`` command for 10 seconds
 and monitor the system during its execution.  Rather than using the
 ``geopmlaunch`` tool as in the above example, we will run the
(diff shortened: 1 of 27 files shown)


service/docs/source/overview.rst
```diff
@@ -819,13 +828,11 @@ showcases two optional features of the GEOPM Runtime:
...
-   invovation names that will request profiling, other programs will
-   not be affected by ``LD_PRELOAD`` of ``libgeopm.so``.  For this
-   reason a user will typically set these two environment variables
-   together.  This is especially important when profiling programs
-   within a bash script.
+3. **Disable Network Use** The ``GEOPM_CTL_LOCAL`` environment
+   variable may be set which disables all intra-node communication
+   between the controllers on each node, thereby generating a unique
+   report file per host node over which the application processes are
+   launched.
```
bgeltz, line 835: Also note that when MPI is disabled in the runtime, GEOPM_CTL_LOCAL is used by default.  This is a change in this PR.

service/docs/source/overview.rst
```diff
@@ -788,6 +788,13 @@ option to ``geopmlaunch``.  For more information about ``geompmlaunch`` see:
 Profiling Non-MPI Applications
 """"""""""""""""""""""""""""""
 
+The ``geopmlaunch(1)`` command may not be best suited for your needs
+if you are running a non-MPI application, or if you are running an MPI
+application but the launch command is embedded in scripts that are
+difficult to modify.  Alternately to using ``geopmlaunch(1)``, the
```
bgeltz:
```suggestion
difficult to modify.  Instead of using ``geopmlaunch(1)``, the
```
Maybe reads clearer?

cmcantalupo (approved):

PR #3343, opened 2024-02-01
Add a performance test for the batch interface of PlatformIO

cmcantalupo:
- Measure performance of the batch interface and print metrics to files
- The script does not enforce a performance target


service/integration/test/test_batch_perf.py
```diff
@@ -0,0 +1,27 @@
...
+    time_samples = dict()
+    for cc in cases:
+        time_samples[cc] = []
+        with open(f'{cc}.csv') as fid:
+            fid.readline()
+            for line in fid.readlines():
+                (count, time) = line.split(',')
+                time_samples[cc].append(float(time))
+
+    height = [sum(vv)/len(vv) for vv in time_samples.values()]
```
dannosliwcd: Some of the cases spend more time on the first sample than on the rest of the samples (e.g., due to obtaining a session). Can we add error bars and/or separate bars for first&rest iterations so that information isn't lost in the mean delay?

service/src/IOGroup.cpp
```diff
@@ -5,7 +5,7 @@
 #include "config.h"
 
 #include <unistd.h>
-
+#include <cstdlib>
```
bgeltz, line 8:
```suggestion
```

service/src/IOGroup.cpp
```diff
@@ -99,7 +99,8 @@ namespace geopm
 #ifdef GEOPM_ENABLE_RAWMSR
             // Only use /dev/cpu/*/msr if configured with
             // --enable-rawmsr and msr-safe driver is not available.
-            bool use_msr_safe = (access("/dev/cpu/msr_batch", R_OK | W_OK) == 0);
+            bool use_msr_safe = (access("/dev/cpu/msr_batch", R_OK | W_OK) == 0 &&
+                                 std::getenv("GEOPM_DISABLE_MSR_SAFE") == nullptr);
```
bgeltz, line 103: From Helper.[ch]pp:
```suggestion
                                 geopm::get_env("GEOPM_DISABLE_MSR_SAFE").empty);
```
Aside from the LevelZero.cpp checking ```ZE_AFFINITY_MASK``` directly with ```getenv``` (which imo is another issue we should correct), there are  *no* direct usages of ```getenv``` anywhere in the service code.

service/src/IOUring.cpp
```diff
@@ -34,7 +36,8 @@ namespace geopm
     std::unique_ptr<IOUring> IOUring::make_unique(unsigned entries)
     {
 #ifdef GEOPM_HAS_IO_URING
-        if (IOUringImp::is_supported()) {
+        if (IOUringImp::is_supported() &&
+            std::getenv("GEOPM_DISABLE_IO_URING") == nullptr) {
```
bgeltz, line 40: Same suggestion as in IOGroup.cpp here.

bgeltz (changes requested):

service/integration/test/plot_batch_perf.py
```diff
@@ -0,0 +1,52 @@
...
+    aspect=0.5,
+    hue='Which Iteration',
+    ci='sd',
+    y='DURATION',
+#    log_scale=True
+)
+g.set(ylim=(args.min_y, None), xlabel=None)
+g.set_axis_labels(y_var='Latency')
+for ax in g.axes.flat:
+    ax.set_yscale('log')
```
bgeltz, line 45: Confused why this is set to 'log' while ```log_scale``` is commented out above.

bgeltz (changes requested):

service/src/IOGroup.cpp
cmcantalupo: We really want to know if the environment variable is set or not, we don't care about its contents.  In this way, the geopm::get_env() is ambiguous.  Since we are not using the C string in this case, I think the direct use of std::getenv() is valid.  If a user simply called `export GEOPM_DISABLE_MSR_SAFE` the original code would disable msr-safe use, but the proposed code would not.

service/src/IOUring.cpp
cmcantalupo: Same argument applies that this is preferred to geopm::get_env().

service/integration/test/plot_batch_perf.py
cmcantalupo: The commented out use of log_scale above didn't work properly (the bars became just a line where the top of the bar should have been).  Changing the scale afterward as above does preserve the bars.

service/src/IOGroup.cpp
bgeltz: SGTM, then this PR needs to include some documentation about the expected usage of ```GEOPM_DISABLE_MSR_SAFE```.  IIRC we don't have other env vars that are exported without a value set, but I may be mistaken.

bgeltz (changes requested):

bgeltz (approved):

Issue #1874, opened 2021-09-17
Spack and geopm


maiterth:
I got a ping on geopm usage with spack on a cray machine.

Installation issue: geopm #26003 
https://github.com/spack/spack/issues/26003

Spack integration with geopm was a task on one of the projects with LLNL, at the time I took on the task. 
After having the initial version of geopm ( up to geopm1.1) on there working, I have not looked into this.

Since this is not the primary (or even recommended mode of installation according to geopm's readme) I'm not sure if this is a priority. But it may be an interesting platform, therefore figuring out the issue might be interesting to engage with.

Thanks,
--Matthias

cmcantalupo:
Thanks for bringing this to our attention @maiterth!  I have left a  note on the spack issue.  
https://github.com/spack/spack/issues/26003#issuecomment-921839007

cmcantalupo:
#1875

bgeltz:
I cannot repro the issue described in https://github.com/spack/spack/issues/26003#issuecomment-921839007 just yet, but I did identify a few things around building the Fortran code that I just put fixes up for.

It's all in PR #1875 right now.

cmcantalupo:
Done criteria:
- [ ] Working recipes for all packages: covering all installed content from geopm service and base directory
- [ ] Recipes support building from release tag and also from an un-tagged version of the dev branch
- [ ] Pull request has been submitted to upstream Spack repo

PR #3349, opened 2024-02-07
Make geopmctl_main a cpp file

bgeltz:
- Resolves linking issues when building inside spack.
- Relates to #1874.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -484,7 +484,7 @@ ompt_source_files = src/OMPT.cpp \
                     src/ompt_callback.cpp \
                     # end
 
-geopmctl_SOURCES = src/geopmctl_main.c
+geopmctl_SOURCES = src/geopmctl_main.cpp
 
 if ENABLE_MPI
     libgeopm_la_SOURCES +=$(mpi_source_files)
--- a/src/Controller.hpp
+++ b/src/Controller.hpp
@@ -14,6 +14,12 @@
 #include <map>
 #include <set>
 
+
+extern "C"
+{
+    int geopmctl_main(int argc, char **argv);
+}
+
 namespace geopm
 {
     class Comm;
--- a/src/geopmctl_main.cpp
+++ b/src/geopmctl_main.cpp
@@ -17,15 +17,11 @@
 
 #include "geopm_version.h"
 #include "geopm_error.h"
+#include "Controller.hpp"
 
 enum geopmctl_const {
     GEOPMCTL_STRING_LENGTH = 128,
 };
-
-extern "C"
-{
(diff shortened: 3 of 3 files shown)


bgeltz:
I sorted this all out today.  This PR isn't necessary.

This PR started as a workaround for building the GEOPM Runtime in spack with the Service built with gcc@8.5.0 and the Runtime built with oneapi.  Compile time issues were encountered relating to missing symbols for string streams and GLIBCXX_3.4.26.  These build time issues were resolved with this PR, but the missing symbol issues came back at runtime.

The previous thinking was that the fact that we build C binaries that require linking against our C++ libraries (i.e. libgeopm.so, libgeopmd.so) was problematic because of issues resolving the required symbols in libstdc++.so with the C compiler (gcc).  This was all a red herring.

The real issue here is that the oneapi compilers rely on a system GCC install to be functional.  Because I was building with oneapi (which was pulling in the system installed GCC toolchain @ 7.5.0) but attempting to link against my spack built gcc@8.5.0 for among other things libstdc++, the missing symbols issue reproduced.

The fix is to inform the oneapi compilers about the desired GCC toolchain to resolve all the symbol issues.  The LLVM options ``--gcc-toolchain`` and ``--gcc-install-dir`` allow for this specification:
https://discourse.llvm.org/t/add-gcc-install-dir-deprecate-gcc-toolchain-and-remove-gcc-install-prefix/65091

The spack docs describe this process in detail for the older classic compilers:
https://spack.readthedocs.io/en/latest/getting_started.html#command-line-configuration
https://spack.readthedocs.io/en/latest/build_systems/intelpackage.html#integrating-external-compilers
However the newer oneapi docs do not mention this: https://spack.readthedocs.io/en/latest/build_systems/inteloneapipackage.html

The resolution here is to modify the compilers.yaml to specify the spack installed GCC toolchain location.  Example:
```
- compiler:
    spec: oneapi@=2024.0.0
    paths:
      cc: <PATH_TO>/icx
      cxx: <PATH_TO>/icpx
      f77: <PATH_TO>/ifx
      fc: <PATH_TO>/ifx
    flags:
      cflags: --gcc-install-dir=<OUTPUT from 'spack location -i gcc@8.5.0'>/lib/gcc/x86_64-pc-linux-gnu/8.5.0
      cxxflags: --gcc-install-dir=<OUTPUT from 'spack location -i gcc@8.5.0'>/lib/gcc/x86_64-pc-linux-gnu/8.5.0
      fflags: -gcc-name=<OUTPUT from 'spack location -i gcc@8.5.0'>/bin/gcc
    operating_system: sles15
    target: x86_64
    modules: []
    environment: {}
    extra_rpaths: []
```

PR #3360, opened 2024-02-22
Bump ABI version for all shared object libraries to 3.0.0

bgeltz:
- Based on https://github.com/geopm/geopm/pull/3110
- https://github.com/semver/semver/blob/master/semver.md


bgeltz:
Will address backwards compatibility issues with #3361

Issue #3363, opened 2024-02-29
As an LDMS user, I would like to check whether Intel GPU RAS errors are correctable or not


sidjana:
**Is your feature request related to a problem? Please describe.**
For sysadmins, one of the next steps towards monitoring GPU RAS error counters is to analyze whether the errors are correctable or not. This issue suggests an addition of an IOGroup signal that flags this status. 

**Describe the solution you'd like**
Add a new LevelZero IOGroup RAS signal that tracks whether the RAS error sets are correctable or not

Relates to https://github.com/intel-innersource/applications.analyzers.geopm.geopm/issues/84

PR #3380, opened 2024-03-22
LevelZero support for correctable & uncorrectable RAS counters

sidjana:
- Relates to #3363 

The past support for Level-Zero RAS counters lacked awareness of RAS error types (correctable v/s uncorrectable). Only correctable counters were supported. 

This PR introduces the awareness of both types of RAS errors and extends the list of LevelZero IOGroup signals to include both types.


cmcantalupo (dismissed):

libgeopmd/src/LevelZero.cpp
```diff
@@ -159,12 +159,12 @@ namespace geopm
         // TODO: When additional device types such as FPGA, MCA, and Integrated GPU are supported by GEOPM
         // This should be changed to a more general loop iterating over type and caching appropriately
         for (unsigned int gpu_idx = 0; gpu_idx < m_num_gpu; gpu_idx++) {
-            ras_domain_cache(gpu_idx);
             frequency_domain_cache(gpu_idx);
             power_domain_cache(gpu_idx);
             perf_domain_cache(gpu_idx);
             engine_domain_cache(gpu_idx);
             temperature_domain_cache(gpu_idx);
+            ras_domain_cache(gpu_idx);
```
dannosliwcd, line 167: If order matters, can you add a comment about why `ras_domain_cache` needs to be last?

libgeopmd/src/LevelZero.cpp
```diff
@@ -481,53 +481,80 @@ namespace geopm
...
+		// Cache the RAS errorset handle into index - [subdevice Id][error type]
+                else if (property.type == ZES_RAS_ERROR_TYPE_CORRECTABLE ||
+                         property.type == ZES_RAS_ERROR_TYPE_UNCORRECTABLE) {
+                   int ras_idx = property.subdeviceId * M_NUM_ERROR_TYPE +
+                                 (property.type == ZES_RAS_ERROR_TYPE_CORRECTABLE ?
+                                  M_ERROR_TYPE_CORRECTABLE : M_ERROR_TYPE_UNCORRECTABLE);
+                   m_devices.at(device_idx).subdevice.ras_domain.at(ras_idx) = handle;
+                   ++ras_handle_count;
+#ifdef GEOPM_DEBUG
+		   std::cerr << "Level-0 RAS: Inserted handle for subdevice# "<<property.subdeviceId<<", error type# "<<property.type<<", device# "<<device_idx<<"\n";
```
dannosliwcd:
```suggestion
		   std::cerr << "Note: <geopm> Level-0 RAS: Inserted handle for subdevice# "<<property.subdeviceId<<", error type# "<<property.type<<", device# "<<device_idx<<"\n";
```
Should make it clear this debug note comes from GEOPM. Or if it was only there for initial development debug, maybe delete it.

libgeopmd/src/LevelZero.cpp
```diff
@@ -605,71 +632,153 @@ namespace geopm
...
                         GEOPM_ERROR_RUNTIME,
                         "LevelZero::" + std::string(__func__) +
                         ": Sysman failed to get RAS counters",
                         __LINE__);
-
-        std::array<uint64_t, ZES_MAX_RAS_ERROR_CATEGORY_COUNT> temp;
-        for (unsigned int err_idx = 0; err_idx < ZES_MAX_RAS_ERROR_CATEGORY_COUNT; err_idx++)
-            temp[err_idx] = pState.category[err_idx];
-        return temp;
+	return pState.category.at(errorcat);
```
dannosliwcd: This looks like we fully populate the `pState.category` array then return just a single element for the requested category. Does that cancel any benefits we get from combining RAS queries in GEOPM batch reads?

libgeopmd/src/LevelZero.cpp
sidjana: The only reason why ras_domain_cache is placed at the end is due to the undocumented ongoing coding practice of adding the "most recently added L0 signal" towards the end of the caching order. No technical reason.

libgeopmd/src/LevelZero.cpp
sidjana: Good catch. This was debugging code intended only for initial development. Deleting line.

libgeopmd/src/LevelZero.cpp
sidjana: Note that the full array - `pState.category` is populated by L0 sysman, not PlatformIO. Each time a single element needs to be accessed, the entire array needs to be referenced. This overhead of populating all of contents of the `category` array is embedded within the execution model of L0. In fact, this latest commit removes the additional 'for' loop by removing the extra overhead of caching into 'temp' (on top of the aforementioned L0 internal overhead).

dannosliwcd (approved):
Please open a tech debt issue referring back here about the `pState.category` batch_read discussion

dannosliwcd (approved):

PR #3382, opened 2024-03-26
Do not check RAS signals at LevelZero startup

cmcantalupo:
- These signals can take 0.5 seconds each to read
- Errors will be delayed until user tries to read signal.

### DIFF
--- a/service/src/LevelZeroIOGroup.cpp
+++ b/service/src/LevelZeroIOGroup.cpp
@@ -687,18 +687,21 @@ namespace geopm
             for (int domain_idx = 0; domain_idx < m_platform_topo.num_domain(
                                      signal_domain_type(sv.first)); ++domain_idx) {
                 try {
-                    double init_setting = 0;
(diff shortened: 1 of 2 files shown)


service/src/LevelZeroIOGroup.cpp
```diff
@@ -687,18 +687,18 @@ namespace geopm
...
-                    if (init_setting == -1) {
-                        unsupported_signal_names.push_back(sv.first);
-                    }
-                    else {
-                        std::shared_ptr<Signal> signal =
-                                std::make_shared<LevelZeroSignal>(sv.second.m_devpool_func,
-                                                                  domain_idx,
-                                                                  sv.second.m_scalar);
-                        result.push_back(signal);
+                    if (domain_idx == 0 && sv.first.find("_RAS_") == std::string::npos) {
```
cmcantalupo, line 693: Add a comment about why RAS signals are skipped during start up checks

bgeltz (dismissed):

sidjana (approved):

PR #3396, opened 2024-04-16
Fixup obs workflow for docs build

cmcantalupo:


.github/workflows/build.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       run: ./make_sdist.sh
     - name: make docs dist
       working-directory: docs
-      run: make dist
+      run: python3 -m pip install -r requirements.txt && make dist
```
bgeltz: This belongs up with the code on line 147-149.

bgeltz (changes requested):

.github/workflows/build.yml
```diff
@@ -147,6 +147,7 @@ jobs:
       run: |
            python3 -m pip install --upgrade pip setuptools wheel pep517
            python3 -m pip install -r geopmdpy/requirements.txt
```
bgeltz:
```suggestion
```
I think this was just for the docs.

bgeltz (approved):

PR #3409, opened 2024-04-17
Enable user to specify options to dpkg_buildpackage

cmcantalupo:
Enables upload to launchpad


dannosliwcd (approved):
Looks good. I was going to suggest mentioning the new env var in our packaging documentation, but I don't think there is such a doc to update. Might be a nice section to have in the developer guide.

Issue #3388, opened 2024-04-10
Run integration tests on target hardware


bgeltz:
> estimate 1
Parent [feature/story]: #3384

Done criteria:
- [x] Tests pass
- [x] Any necessary code changes to get the tests to pass are merged to dev
- [x] Any updates to internal scripting that drives the nightly integration testing are in place
- [x] integration/apps builds complete successfully

PR #3416, opened 2024-04-26
Fix SysfsIOGroup batch write

bgeltz:
- Relates to #3388.
- IOGroup was initializing IOUring during batch_write(), and on certain systems this resulted in an exception being thrown.


cmcantalupo (approved):
Love github's new "ignore whitespace" option

dannosliwcd (approved):

PR #3417, opened 2024-04-29
Ensure that the msr module is loaded before running the batch perf test

dannosliwcd:
- Also ensure that the msr module is only removed after running the test if it was not present prior to running the test.


cmcantalupo (approved):
LGTM

PR #3418, opened 2024-04-29
Add missing virtual destructor for SSTIO

dannosliwcd:


libgeopmd/src/SSTIO.hpp
```diff
@@ -16,6 +16,8 @@ namespace geopm
     class SSTIO
     {
         public:
+            virtual ~SSTIO() = default;
```
cmcantalupo, line 20:
```suggestion
            SSTIO() = default;
            virtual ~SSTIO() = default;
```

cmcantalupo (approved):

PR #3426, opened 2024-05-02
Increase the problem size for test_cpu_characterization and run in process mode

dannosliwcd:
- The current problem size is too small for our integration testbed, so several runs finish too quickly to measure any energy differences.
- The larger suite this integration executes within runs within a 4-node allocation. This test only needs one node. The application-mode launcher does not currently support that scenario, so use the process-mode launcher instead.


cmcantalupo (approved):
LGTM

Issue #1071, opened 2020-04-27
Extend tutorial integration test to cover plugin-based subdirectories


dianarg:
Done criteria
--------------
- tests of run scripts for tutorial/agent, tutorial/iogroup, tutorial/plugin_load

PR #3430, opened 2024-05-03
Switch example control over to a temporary file rather than stdout/stderr

cmcantalupo:
The existing tutorial was never properly adapted to have the requirements for all controls to have a save/restore method.  This PR addresses this by making the control more realistic.

- Relates to #1071


integration/tutorial/iogroup/ExampleIOGroup.cpp
```diff
@@ -333,42 +373,39 @@ void ExampleIOGroup::write_control(const std::string &control_name, int domain_t
...
+    if (m_control_saver != nullptr) {
+        m_control_saver->restore(*this);
+    }
 }
 
 void ExampleIOGroup::restore_control(const std::string &save_path)
 {
-
+    if (m_control_saver != nullptr) {
+        m_control_saver->restore(*this);
```
cmcantalupo, line 407: need to pass save_path

integration/tutorial/iogroup/ExampleIOGroup.cpp
```diff
@@ -455,7 +483,11 @@ int ExampleIOGroup::signal_behavior(const std::string &signal_name) const
                         GEOPM_ERROR_INVALID, __FILE__, __LINE__);
     }
     // All example signals are time based and increase monotonically
-    return M_SIGNAL_BEHAVIOR_MONOTONE;
+    int result = M_SIGNAL_BEHAVIOR_MONOTONE;
+    if (signal_name != "TMP_FILE_CONTROL" && signal_name != "EXAMPLE::TMP_FILE_CONTROL") {
+        result = M_SIGNAL_BEHAVIOR_VARIABLE;
+    }
```
dannosliwcd: Is this the opposite of the intended behavior? I think the control file is the only variable one.

integration/tutorial/iogroup/Makefile
```diff
@@ -12,4 +12,17 @@ libgeopmiogroup_example_iogroup.so.2.0.0: ExampleIOGroup.o
 clean:
 	rm -f ./*.o ./*.d ./*.so*
 
-.PHONY: clean
+check: libgeopmiogroup_example_iogroup.so.2.0.0
```
dannosliwcd, line 15: Nice! I guess this can be tagged as related to #1071

Does this need to return-code checking enabled so it fails when any of the intermediate steps fail?

dannosliwcd: Do we need to invoke this somewhere else to make it execute as part of nightly builds?

integration/tutorial/iogroup/README.md
```diff
@@ -209,6 +211,13 @@ the GEOPM install directory, <GEOPM_INSTALL_DIR>/lib/geopm.
 5. Run with geopmread and geopmwrite
 ------------------------------------
 
+First set up the example control file with the value 43.21:
```
dannosliwcd, line 214:
```suggestion
First set up the example control file with any value (for example, 43.21):
```

integration/tutorial/iogroup/Makefile
```diff
@@ -12,4 +12,17 @@ libgeopmiogroup_example_iogroup.so.2.0.0: ExampleIOGroup.o
...
+	rm -f /tmp/geopm_example_control.$$(id -u)
+	GEOPM_PLUGIN_PATH=$$PWD geopmread | grep -q TMP_FILE_CONTROL || true # File does not exist, but control is listed
+	echo foobar > /tmp/geopm_example_control.$$(id -u)
+	GEOPM_PLUGIN_PATH=$$PWD geopmread | grep -q TMP_FILE_CONTROL || true # File has bad value, but control is listed
+	echo 42 > /tmp/geopm_example_control.$$(id -u)
+	GEOPM_PLUGIN_PATH=$$PWD geopmread | grep -q TMP_FILE_CONTROL # File has good value, but control is not listed
```
dannosliwcd: Were the the ` || true` parts just to let debugging proceed past failures? Should it check for an expected grep failure instead?

libgeopm/include/geopm/Environment.hpp
```diff
@@ -116,16 +118,16 @@ namespace geopm
             int num_proc(void) const override;
             bool do_ctl_local(void) const override;
         protected:
-            void parse_environment(void);
-            bool is_set(const std::string &env_var) const;
-            std::string lookup(const std::string &env_var) const;
+            void GEOPM_PRIVATE parse_environment(void);
+            bool GEOPM_PRIVATE is_set(const std::string &env_var) const;
+            std::string GEOPM_PRIVATE lookup(const std::string &env_var) const;
```
dannosliwcd: Would that make it impossible for implementers outside of libgeopm.so to call these functions when extending this class?

Maybe a non-issue for this particular class, but I don't know whether we want to hide `protected` symbols generally.

dannosliwcd (approved):

PR #3062, opened 2023-08-14
Add a frequency rebalancer agent

dannosliwcd:
This set of changes introduces the frequency_balancer agent, which  adjusts for imbalance by assigning more CPU
  frequency resources cores that spend less MPI time per epoch. The changes include integration tests that check for energy savings, and also for performance improvement if SST is available on the test system.

PR #3432, opened 2024-05-04
Address issues discovered during RC integration testing

bgeltz:
- Relates to #3388.

### DIFF
--- a/geopmpy/geopmpy/io.py
+++ b/geopmpy/geopmpy/io.py
@@ -809,6 +809,7 @@ class RawReportCollection(object):
         self._reports_df = pandas.DataFrame()
         self._app_reports_df = pandas.DataFrame()
         self._epoch_reports_df = pandas.DataFrame()
+        self._unmarked_reports_df = pandas.DataFrame()
         self._meta_data = None
         self.load_reports(report_paths, dir_name, dir_cache, verbose, do_cache)
 
@@ -872,7 +873,10 @@ class RawReportCollection(object):
                 except KeyError:
                     pass # No regions in cached report
                 self._app_reports_df = pandas.read_hdf(self._report_h5_name, 'app_report')
-                self._unmarked_reports_df = pandas.read_hdf(self._report_h5_name, 'unmarked_report')
+                try:
+                    self._unmarked_reports_df = pandas.read_hdf(self._report_h5_name, 'unmarked_report')
+                except KeyError:
+                    pass
                 try:
                     self._epoch_reports_df = pandas.read_hdf(self._report_h5_name, 'epoch_report')
                 except KeyError:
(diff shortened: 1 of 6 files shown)


geopmpy/geopmpy/io.py
```diff
@@ -890,13 +894,13 @@ def load_reports(self, reports, dir_name, dir_cache, verbose, do_cache):
...
-                        self._app_reports_df.to_hdf(self._report_h5_name, 'app_report', format='table', append=True)
-                        self._unmarked_reports_df.to_hdf(self._report_h5_name, 'unmarked_report', format='table', append=True)
-                        self._epoch_reports_df.to_hdf(self._report_h5_name, 'epoch_report', format='table', append=True)
+                        self._app_reports_df.to_hdf(self._report_h5_name, 'app_report', format='table')
+                        if len(self._reports_df) > 0:
+                            self._reports_df.to_hdf(self._report_h5_name, 'report', format='table', append=True)
+                        if len(self._unmarked_reports_df) > 0:
+                            self._unmarked_reports_df.to_hdf(self._report_h5_name, 'unmarked_report', format='table', append=True)
+                        if len(self._epoch_reports_df) > 0:
+                            self._epoch_reports_df.to_hdf(self._report_h5_name, 'epoch_report', format='table', append=True)
```
dannosliwcd, line 903: Why are the `if(len)` checks needed now (or were they needed all along)?

It seems like it has something to do with the fact that we initialize these through multiple maybe-executed lines because of a string-related issue.

integration/test/geopm_test_launcher.py
```diff
@@ -128,7 +129,7 @@ def run(self, test_name, include_geopm_policy=True, add_geopm_args=[], enable_af
             argv = ['dummy', detect_launcher(), '--geopm-ctl', self._pmpi_ctl,
                                                 '--geopm-agent', self._agent_conf.get_agent(),
                                                 '--geopm-profile', test_name,
-                                                '--geopm-program-filter', self._app_conf.get_exec_path()]
+                                                '--geopm-program-filter', os.path.basename(self._app_conf.get_exec_path())]
```
dannosliwcd, line 132: Should I apply something similar here as well? 
* Add a `app_conf.get_program_filter()` function, defaulting to exec path but allowing individual appconf implementations to override it https://github.com/geopm/geopm/pull/3062/files#diff-76dd7a1a9349bf0d8cf85212926694ed5006d12dcb6a8b64e04c04a4da7a42ea
* Use `get_program_filter()` in launch_util.py https://github.com/geopm/geopm/pull/3062/files#diff-343359e818df706352ab39f88c63ca674e14f518dcc4eacc1fbd88feea99520d

Or could we do something similar to those changes as part of this PR so it applies for all users of appconf?

bgeltz: Chris and I discussed this and thought it would be a good idea to make the change in the launcher.py.  If I did it there, all these downstream usages would not have to change.  Let's discuss.

geopmpy/geopmpy/io.py
bgeltz: I was having trouble with test_time.py and running with --skip-launch.  The issue was that I was trying to query for the various dataframes with the getters for RawReportCollection, only in some cases I was getting an exception that they didn't exist, and in some cases I was getting back an empty dataframe.

It turns out it was all related to the code that tries to load the HDF5 cache of reports.  If it was the first time and the data was actually parsed and cached, it was all OK, but if the test was trying to read from the cache and do the same operation, it failed.

Let me double check on this today and ensure this is all in fact needed.

dannosliwcd (approved):

libgeopm/src/Environment.cpp
```diff
@@ -382,10 +382,13 @@ namespace geopm
         bool result = false;
         if (is_set("GEOPM_PROGRAM_FILTER")) {
             auto valid_names = geopm::string_split(lookup("GEOPM_PROGRAM_FILTER"), ",");
+            // Extract just the basename for any path provided as input
+            std::transform(valid_names.begin(), valid_names.end(), valid_names.begin(),
+                           [](const std::string& vn) {
+                               return vn.substr(vn.find_last_of("/") + 1);});
```
cmcantalupo, line 388: Does this work when only the base name is provides as input?

PR #3437, opened 2024-05-07
Fix CPU-CA integration tests

dannosliwcd:
- Modify the minife appconf to use a configuration with better baseline performance.
- Merge the integration suites that create input data for the CPU-CA and tests that evaluate the CPU-CA. This avoids duplicate effort to duplicate core and uncore frequency sweeps in both sets of tests.


integration/test/test_cpu_characterization.py
```diff
@@ -23,31 +23,23 @@
...
 from integration.test import geopm_test_launcher
+from integration.experiment.energy_efficiency import cpu_activity
 from integration.experiment.monitor import monitor
 from integration.experiment.uncore_frequency_sweep import uncore_frequency_sweep
 from integration.experiment.uncore_frequency_sweep import gen_cpu_activity_constconfig_recommendation
 from integration.apps.arithmetic_intensity import arithmetic_intensity
+from integration.apps.minife import minife
 
 @util.skip_unless_workload_exists("apps/arithmetic_intensity/ARITHMETIC_INTENSITY/bench_sse")
+@util.skip_unless_workload_exists("apps/minife/miniFE_openmp-2.0-rc3/src/miniFE.x")
```
dannosliwcd: Move to the individual test cases that use this app

bgeltz (approved):

PR #3438, opened 2024-05-07
Fix static analysis issues

cmcantalupo:
- Fixes race condition for error reporting
- Fixes dereference of invalid pointer

### DIFF
--- a/libgeopm/src/DomainNetMap.cpp
+++ b/libgeopm/src/DomainNetMap.cpp
@@ -94,7 +94,7 @@ namespace geopm
         // make sure that there no unexpected keys in the json
         for (const auto &key : nnet_json.object_items()) {
             if (M_EXPECTED_KEYS.find(key.first) == M_EXPECTED_KEYS.end()) {
-                throw Exception("DomainNetMapImp::" + std::string(__func__) + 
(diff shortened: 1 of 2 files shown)


libgeopm/src/DomainNetMap.cpp
dannosliwcd, line 1: Looks like this still matches `docs/json_schemas/domainnetmap_neural_net.schema.json` and looks like it catches all of the users of `signal_it` and `delta_it`

libgeopmd/src/Exception.cpp
dannosliwcd, line 1: Looks like this protects the equality check on `error_value` that static analysis highlighted.

Does the function call that reads `error_value` also need to be protected?

cmcantalupo: no, message_fixed(int) does not use any dynamic data, only M_VALUE_MESSAGE_MAP which is declared const:

https://github.com/geopm/geopm/pull/3438/files#diff-c9e9373602ecc3349eb99be2066655930ddf32fbabc7ec214bcbf472bb5c974bR32

dannosliwcd (dismissed):

dannosliwcd (approved):

PR #3454, opened 2024-05-09
Fix broken package builds

dannosliwcd:
- Resolves a race where builds from tarballs may fail to create some of the generated files.
- Adds GH automation to test that deb builds succeed in each of the subprojects


libgeopmd/Makefile.am
```diff
@@ -321,14 +322,17 @@ else
...
+
+$(msr_cpp_files): $(gen_src_dir)/%.cpp: $(top_srcdir)/json_data/%.json $(gen_src_dir) $(top_srcdir)/src/json_data.cpp.in
 	sed -e '/@JSON_CONTENTS@/ {' \
 	    -e "r $<" \
 	    -e 'd' -e '}' \
 	    -e "s/@JSON_IDENTIFIER@/$(lastword $(subst _, ,$*))_msr_json/g" \
 	$(top_srcdir)/src/json_data.cpp.in > $@
 
-$(sysfs_cpp_files): $(top_builddir)/gen-src/%.cpp: $(top_srcdir)/json_data/%.json $(top_srcdir)/src/json_data.cpp.in
+$(sysfs_cpp_files): $(gen_src_dir)/%.cpp: $(top_srcdir)/json_data/%.json $(gen_src_dir) $(top_srcdir)/src/json_data.cpp.in
```
cmcantalupo, line 335:
```suggestion
$(sysfs_cpp_files): $(gen_src_dir)/%.cpp: $(top_srcdir)/json_data/%.json $(top_srcdir)/src/json_data.cpp.in | $(gen_src_dir)
```

cmcantalupo (commented):
I think adding that | should avoid unnecessary rebuilds if the directory time stamp changes

cmcantalupo (approved):

libgeopmd/Makefile.am
cmcantalupo: Lets deal with this issue when it becomes a problem

PR #3463, opened 2024-05-10
Bump ABI version to 2.1.0 and make plugin load backward compatible

cmcantalupo:
Addresses a number of issues besides:

- Add C++ interfaces for getting version string and ABI numbers.
- Do not include config.h explicitly, use compile time flag
- Get rid of public mark up from MSRIOGroup.hpp


libgeopm/src/Profile.cpp
```diff
@@ -87,6 +86,10 @@ namespace geopm
             return;
         }
         connect();
+#ifndef GEOPM_DEBUG
+        // When debug not enabled, use of m_region_ids is optimized out
```
bgeltz, line 90: Isn't it obvious that m_region_ids is optimized out looking at where/how it's used (i.e. all the GEOPM_DEBUG checks)?

I think what's not obvious here is why you have to manually insert HASH_UNMARKED (because we never call (create) region() on it).

libgeopmd/src/MSRIOGroup.hpp
```diff
@@ -16,15 +16,13 @@
...
-#include "geopm_public.h"
 #include "geopm_time.h"
 
 #include "geopm/IOGroup.hpp"
 
 extern "C"
 {
-    int GEOPM_PUBLIC
-        geopm_allowlist(size_t result_max, char *result);
+    int geopm_allowlist(size_t result_max, char *result);
```
bgeltz, line 25: Why were all these changes in MSRIOGroup.cpp necessary?

libgeopmd/src/MSRIOGroup.hpp
```diff
@@ -206,18 +191,15 @@ namespace geopm
...
                                      double scalar, int units, const std::string &aggregation,
                                      const std::string &description, int behavior,
                                      const std::function<std::string(double)> &format_function);
             // Add a bitfield of an MSR as an available control
-            void GEOPM_PRIVATE
-                add_msr_field_control(const std::string &msr_field_name, int domain_type,
+            void add_msr_field_control(const std::string &msr_field_name, int domain_type,
                                       uint64_t msr_offset, int begin_bit, int end_bit,
                                       int function, double scalar, int units,
                                       const std::string &description);
```
bgeltz:
```suggestion
            void add_msr_field_control(const std::string &msr_field_name, int domain_type,
                                       uint64_t msr_offset, int begin_bit, int end_bit,
                                       int function, double scalar, int units,
                                       const std::string &description);
```

libgeopm/src/Profile.cpp
cmcantalupo: This is to avoid clang warnings about unused member variables.

libgeopmd/src/MSRIOGroup.hpp
cmcantalupo: MSRIOGroup.hpp was moved from public to private, but the public markings were left behind:

https://github.com/geopm/geopm/pull/3462

this is a fixup to that change ^^^

libgeopmd/src/geopm_version.cpp
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        catch (const std::invalid_argument &ex) {
+            throw logic_error;
+        }
+        catch (const std::out_of_range &ex) {
+            throw logic_error;
+        }
+        GEOPM_DEBUG_ASSERT(abi_nums.size() == 3, logic_error.what());
+        return {abi_nums[0] - abi_nums[2], abi_nums[2], abi_nums[3]};
```
bgeltz:
```suggestion
        return {abi_nums[0] - abi_nums[2], abi_nums[2], abi_nums[1]};
```

bgeltz (changes requested):

bgeltz (dismissed):

bgeltz (approved):

PR #3484, opened 2024-05-21
Add versioning to web page

cmcantalupo:


docs/source/conf.py
```diff
@@ -12,6 +12,7 @@
 import sys
 import glob
 import sphinx_rtd_theme
+from geopmdpy import __version__ as version
```
bgeltz, line 15: Where is this used?

bgeltz (changes requested):

docs/source/conf.py
cmcantalupo: sphinx magic, version is one of the many variables you can set in conf.py.  We we using "logo_only" before which was suppressing the version in the side bar.

bgeltz (approved):

PR #3511, opened 2024-06-06
Fix single tile GPU part interactions with Level Zero

cmcantalupo:
- Support for systems that do not have perf factor
- Fixes https://github.com/intel-innersource/applications.analyzers.geopm.geopm/issues/90
- Some assumptions about Level Zero reporting were not valid for single tile GPUs

### DIFF
--- a/docs/source/geopm_pio_levelzero.7.rst
+++ b/docs/source/geopm_pio_levelzero.7.rst
@@ -143,7 +143,7 @@ Signals
     GPU Compute Hardware chip energy in joules.
 
     *  **Aggregation**: sum
-    *  **Domain**: gpu_chip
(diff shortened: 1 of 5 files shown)


lhlawson (commented):
The basic change seems sound, though there may be some confusion on systems that support a GPU level power reading and a GPU subdevice level power reading, same for a GPU perf factor and a GPU subdevice perf factor.   
We try and clarify that with the _CORE_ signal (ex: GPU_CORE_ENERGY vs GPU_ENERGY) but a quick test would help confirm these are separate: Are GPU_CORE_POWER & GPU_POWER returning different values on a single tile GPU part?  (alternatively a multi-tile part with ZE_AFFINITY_MASK=0.0)

Also have these changes been tested on a 2 tile GPU part?

dannosliwcd (approved):

Issue #3492, opened 2024-05-22
Perform integration tests on customer system


bgeltz:
> estimate 1
Parent [feature/story]: #3384 

Done criteria:
- [x] Multi-node job launch process is understood, functional, and documented.
- [ ] Any issues arising from running are dealt with.
- [x] Process to install v3.1 RPMs is understood, functional, and documented.
- [ ] (Non-beta) Integration tests pass clean.

PR #3522, opened 2024-06-21
Make geopmbench adhere to GEOPMBENCH_NO_MPI

bgeltz:
- Relates to #3492.

### DIFF
--- a/libgeopm/src/geopmbench_main.cpp
+++ b/libgeopm/src/geopmbench_main.cpp
@@ -111,9 +111,13 @@ static int main_imp(int argc, char **argv)
 
     const int ERROR_HELP = -4096;
 #ifdef GEOPM_ENABLE_MPI
-    err = MPI_Init(&argc, &argv);
-    if (!err) {
-        err = MPI_Comm_rank(MPI_COMM_WORLD, &rank);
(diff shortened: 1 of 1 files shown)


dannosliwcd (dismissed):

libgeopm/src/geopmbench_main.cpp
```diff
@@ -110,10 +110,14 @@ static int main_imp(int argc, char **argv)
 "\n";
 
     const int ERROR_HELP = -4096;
+    const geopm::GEOPMBenchConfig &config = geopm::geopmbench_config();
+    bool is_mpi_enabled = config.is_mpi_enabled();
 #ifdef GEOPM_ENABLE_MPI
```
cmcantalupo:
```suggestion
#ifdef GEOPM_ENABLE_MPI
    const geopm::GEOPMBenchConfig &config = geopm::geopmbench_config();
    bool is_mpi_enabled = config.is_mpi_enabled();

```

cmcantalupo: Without this change i think we will have unused variable errors when mpi is disabled.

cmcantalupo (commented):
Think we need to move the #ifdef

cmcantalupo (approved):

Issue #3523, opened 2024-06-24
Service restart upon increase in number of open files


sidjana:
**Describe the bug**
When running the geopmd service with CPUs > 256, the service crashes and reboots. This leads to incorrect behavior within clients of the service due to the disconnection with the message bus. 

**GEOPM version**
3.1.1.dev1+gcd5d65344

**Expected behavior**
The service should not abruptly crash and impact the message clients it is connected to. 

**Actual behavior**
When large number of file descriptors remain open (bound to happen for CPUs with large core count, when dedicating a single descriptor for each CPU), the systemd service tends to restart without any warning. This leads to disconnection of connection between the service and the message bus client.

**Error messages produced**
$> geopmread <signal-name> <domain-type> <domain-id>
Runtime error: SDBus: Failed to call sd-bus function sd_bus_call_method(), error: -110 name: org-freedesktop.DBus.Error.NoReply: Message Recipient disconnected from message bus without replying at src/SDBus.cpp:32

PR #3524, opened 2024-06-24
Increase max open file descriptor limit

sidjana:
- Relates to #3523 bug report from github issues

[Brief description of the change]
The limit for maximum number of open file descriptors by the geopmd service has been increased to 16384 to support processors with large cpu count (> 256 cpus)


cmcantalupo (dismissed):

cmcantalupo (changes requested):
Original syntax was correct see:
https://unix.stackexchange.com/questions/345595/how-to-set-ulimits-on-service-with-systemd

dannosliwcd (approved):

cmcantalupo (approved):

PR #3525, opened 2024-06-24
Enable support for NVML and LevelZero with one binary

### LINKED ISSUE #3509: levelzero support for debian packages
Our existing debian packages hardcode the dependencies for GPUs to Nvidia.  We need to provide an option for enabling levelzero.

- [ ] Option exists to enable levelzero for debian packages.
- [ ] Launchpad config exists that has these packages?

### LINKED ISSUE #3515: As a user I would like to have GPU support for all supported flavors in the same binary
I'm always frustrated when I need to rebuild GEOPM to target the GPUs I have.  GPU can either support levelzero or nvml today, but not both simultaneously.

The binaries that are produced by GEOPM (e.g. geopmread, geopmwrite) support both Intel GPUs and Nvidia GPUs without requiring a rebuild and configure changes.

n/a

**Additional context**
Debian builds today are set to build for Nvidia.
The OBS builds only enable LevelZero in the supplementary configurations.

cmcantalupo:
- Fixes #3515 
- Fixes #3509

### DIFF
--- a/libgeopmd/src/GPUTopo.cpp
+++ b/libgeopmd/src/GPUTopo.cpp
@@ -16,18 +16,30 @@ namespace geopm
 {
     static std::unique_ptr<GPUTopo> make_unique_gpu_topo(void)
     {
-        std::unique_ptr<GPUTopo> result;
-        try {
+        std::unique_ptr<GPUTopo> result = geopm::make_unique<GPUTopoNull>();
+        std::unique_ptr<GPUTopo> NVMLTopo = geopm::make_unique<GPUTopoNull>();
+        std::unique_ptr<GPUTopo> LevelZeroTopo = geopm::make_unique<GPUTopoNull>();
 #ifdef GEOPM_ENABLE_NVML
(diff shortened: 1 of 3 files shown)


libgeopmd/src/GPUTopo.cpp
```diff
@@ -16,18 +16,30 @@ namespace geopm
...
+#ifdef GEOPM_ENABLE_LEVELZERO
+        try {
+            LevelZeroTopo = geopm::make_unique<LevelZeroGPUTopo>();
         }
         catch (const Exception &ex) {
-            result = geopm::make_unique<GPUTopoNull>();
+
+        }
+#endif
+        if (NVMLTopo->num_gpu() != 0) {
```
bgeltz: Add an exception to see if num_gpu() > 0 for both NVMLTopo and LevelZero.  We don't support  platforms that have both presently.

bgeltz (changes requested):

bgeltz (approved):

Issue #3542, opened 2024-07-29
As an end-user I would like to be able to use geopmsession on a multi-node run to gather summary statistics


asmaalrawi:
Relates to #408 

I can't easily use geopmsession to monitor aggregate statistics at scale for multi-node runs. (e.g. if I want to monitor maximum power consumption per-node) in a way that scales well with number of nodes.

A new python class (based on the Session class) reads signals provided by the end user in a file. When the job is launched, mpi is used to read signals per-node and ingest them into a data frame. At the end of the run, statistics are exported in a json or csv file.

Alternatives considered: 

- Per-node files are handled in post-processing and consolidated into a single file. This would be data-intensive and thus would not scale well.
- geopmsession is extended with a new command line option for stats aggregation output rather than trace output
- A new user signal is created for any statistics that are needed
- Use the geopm runtime whenever large-scale aggregate statistics are desired

cmcantalupo:
Remaining work:
https://github.com/geopm/geopm/pull/3543/

- [x] Documentation additions for geopmsession.1.rst
- [x] Add doc strings and doxygen to all public interfaces
- [x] Fix broken unit tests by updating mocks (False as new last argument to run_read())
- [x] Unit test coverage for StatsCollector.cpp
- [x] ~Unit test coverage for report logic in session.py~ Create tech debt issue around missing integration tests for geopmsession CLI: https://github.com/geopm/geopm/issues/2843
- [x] ~Unit test coverage for MPI logic in session.py~
- [x] Provide dates in ISO format with nano-seconds included 
- [x] Create tech debt issue, or provide context for StatsCollector (no singleton)
- [x] Create tech debt issue and put behind beta flag, or provide API the C api as below with adjusted python bindings

The changes to the geopm_stats_collector*() APIs could look something like
```C
int GEOPM_PUBLIC
    geopm_stats_collector_create(size_t num_requests, const struct geopm_request_s *requests,
        struct geopm_stats_collector_s **collector);

int GEOPM_PUBLIC
    geopm_stats_collector_update(struct geopm_stats_collector_s *collector);

// if *max_report_size is zero, update it with the required size for the report and do not modify report 
int GEOPM_PUBLIC
    geopm_stats_collector_report_yaml(const struct geopm_stats_collector_s *collector,
        size_t *report_size, char *report_yaml);

int GEOPM_PUBLIC
    geopm_stats_collector_clear(struct geopm_stats_collector_s *collector);

int GEOPM_PUBLIC
[...]

PR #3543, opened 2024-07-29
Added multi-node statistic capabilities to geopmsession

asmaalrawi:
- Fixes to #3542 

A new python class (based on the Session class) reads signals provided by the end user in a file. When the job is launched, mpi is used to read signals per-node and accumulate moments in C++. At the end of the run, statistics are exported in a yaml file.

```
$ printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | geopmsession -t 2 -p 0.01 --print-header --report-out=report.yaml --trace-out=trace.csv --mpi-report
$ cat report.yaml
Hosts:
  saruman:
    TIME:
      count: 201
      first: 0.189681
      last: 2.04973
      min: 0.189681
      max: 2.04973
      mean: 1.05533
      std: 0.572718
    CPU_POWER:
      count: 201
      first: 1.02625e+09
      last: 73.4791
      min: 63.4933
      max: 1.02625e+09
      mean: 5.10579e+06
      std: 7.23861e+07
    CPU_FREQUENCY_STATUS:
      count: 200
      first: 1.055e+09
      last: 1e+09
      min: 1e+09
      max: 1.055e+09
      mean: 1.00399e+09
      std: 1.39072e+07
```

### DIFF
--- a/docs/source/geopm_time.3.rst
+++ b/docs/source/geopm_time.3.rst
@@ -3,11 +3,7 @@
 
 
 geopm_time(3) -- helper methods for time
-==========================================
-
-
-
-
+========================================
 
 
 Synopsis
@@ -55,13 +51,13 @@ C++ interfaces:
 
 .. code-block:: c++
 
-       struct geopm_time_s geopm::time_zero(void);
-
        struct geopm_time_s geopm::time_curr(void);
 
        struct geopm_time_s geopm::time_curr_real(void);
 
        std::string geopm::time_curr_string(void);
+
+       struct geopm_time_s geopm::time_zero(void);
 
        void geopm::time_zero_reset(const geopm_time_s &zero);
 
@@ -74,6 +70,25 @@ header defines GEOPM interfaces for measuring time in seconds relative to a
 fixed arbitrary reference point. The ``geopm_time_s`` structure is used to hold
 time values.
 
+The default clock used is Linux ``CLOCK_MONTONIC_RAW``.  In general, these
+interfaces use this as the time reference.  This clock is the most precise for
+calculating time differences.
+
+There are some times when absolute time accuracy is preferred over time
+difference accuracy. For these cases, the GEOPM provides time functions that use
+``CLOCK_REALTIME`` as the reference time.  There is a C accessor for real time:
+``geopm_time_real()``, and there is also a C++ accessor:
+``geopm::time_curr_real()``.  Only ``geopm_time_s`` values created with these
+accessors may be passed to ``geopm_time_to_iso_string()`` which provides a
+nano-second precision absolute date string.
+
+The last time reference is useful when comparing against the values returned by
+PlatformIO for the ``TIME`` signal.  The reference time (0.0) for this signal is
+a ``CLOCK_MONOTONIC_RAW`` time value that can be accessed in C with
+``geopm_time_zero()`` and in C++ with ``geopm::time_zero()``.  There is also a
+way to reset this reference with ``geopm_time_zero_reset()`` and
+``geopm:time_zero_reset()`` by providing a ``CLOCK_MONOTONIC_RAW`` time value.
+
 
 ``geopm_time_string()``
   Fills *buf* with the current date and time as a string.  The
@@ -81,14 +96,13 @@ time values.
   must be at least 26 characters as required by `asctime_r(3) <https://man7.org/linux/man-pages/man3/asctime_r.3.html>`_.
 
 ``geopm_time()``
-  Sets *time* to the current time in seconds since system boot
-  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
-  approximate date string.
+  Sets *time* to the current time using Linux CLOCK_MONOTONIC_RAW clock. This
+  can be used with geopm_time_to_string() to get a an approximate date string.
 
 ``geopm_time_real()``
-   Sets *time* to the current time in seconds since the epoch (CLOCK_REALTIME).
-   This can be used with geopm_time_real_to_iso_string() for an accurate ISO
-   8601 standard date string representation.
+  Sets *time* to the current time using the Linux CLOCK_REALTIME clock.
+  This can be used with geopm_time_real_to_iso_string() for an accurate ISO
+  8601 standard date string representation.
 
 ``geopm_time_diff()``
   Returns the difference in seconds between *begin* and *end*.
@@ -108,12 +122,12 @@ time values.
   See also: ``geopm::time_zero_reset()``.
 
 ``geopm_time_to_string()``
-  Convert a time representation returned by ``geopm_time()`` or
-  ``geopm::time_curr()`` into an approximate date string.
+  Convert a CLOCK_MONOTONIC_RAW time representation returned by ``geopm_time()``
+  or ``geopm::time_curr()`` into an approximate date string.
 
 ``geopm_time_real_to_iso_string()``
-  Convert a time representation returned by ``geopm_time_real()`` or
-  ``geopm::time_curr_real()`` into an ISO 8601 standard date string with
+  Convert a CLOCK_REALTIME time representation returned by ``geopm_time_real()``
+  or ``geopm::time_curr_real()`` into an ISO 8601 standard date string with
   nanosecond resolution.
 
 ``geopm_time_string()``
@@ -126,18 +140,17 @@ time values.
   See also: ``geopm::time_zero_reset()``.
 
 ``geopm::time_curr()``
-  Returns the current time in seconds since system boot
-  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
-  approximate date string.
+  Returns the current time using Linux CLOCK_MONOTONIC_RAW clock. This can be
+  used with geopm_time_to_string() to get a an approximate date string.
 
 ``geopm::time_curr_real()``
-  Returns the current time in seconds since the epoch (CLOCK_REALTIME).
-  This can be used with geopm_time_real_to_iso_string() for an accurate ISO
-  8601 standard date string representation.
+  Returns the current time using the Linux CLOCK_REALTIME clock.  This can be
+  used with geopm_time_real_to_iso_string() for an accurate ISO 8601 standard
+  date string representation.
 
 ``geopm::time_curr_string()``
   Returns an accurate ISO 8601 standard date string representation of
-  the current (real) time.
+  the current time using Linux CLOCK_REALTIME clock..
 
 ``geopm::time_zero_reset()``
   Override the reference time for the ``geopm_time_zero()`` and
--- a/docs/source/geopm_time.3.rst
+++ b/docs/source/geopm_time.3.rst
@@ -17,6 +17,7 @@ Synopsis
 
 Link with ``-lgeopm`` **(MPI)** or ``-lgeopm`` **(non-MPI)**
 
+C interfaces:
 
 .. code-block:: c
 
@@ -24,6 +25,8 @@ Link with ``-lgeopm`` **(MPI)** or ``-lgeopm`` **(non-MPI)**
                                            char *buf);
 
        static inline int geopm_time(struct geopm_time_s *time);
+
+       static inline int geopm_time_real(struct geopm_time_s *time);
 
        static inline double geopm_time_diff(const struct geopm_time_s *begin,
                                             const struct geopm_time_s *end);
@@ -39,12 +42,35 @@ Link with ``-lgeopm`` **(MPI)** or ``-lgeopm`` **(non-MPI)**
 
        int geopm_time_zero(struct geopm_time_s *zero_time);
 
+       static inline int geopm_time_to_string(const struct geopm_time_s *time, int buf_size, char *buf);
+
+       int geopm_time_real_to_iso_string(const struct geopm_time_s *time, int buf_size, char *buf);
+
+       static inline int geopm_time_string(int buf_size, char *buf);
+
+       static inline double geopm_time_since(const struct geopm_time_s *begin);
+
+
+C++ interfaces:
+
+.. code-block:: c++
+
+       struct geopm_time_s geopm::time_zero(void);
+
+       struct geopm_time_s geopm::time_curr(void);
+
+       struct geopm_time_s geopm::time_curr_real(void);
+
+       void geopm::time_zero_reset(const geopm_time_s &zero);
+
+
 Description
 -----------
 
-The `geopm_time.h <https://github.com/geopm/geopm/blob/dev/libgeopmd/include/geopm_time.h>`_ header defines GEOPM interfaces for measuring time
-in seconds relative to a fixed arbitrary reference point. The ``geopm_time_s``
-structure is used to hold time values.
+The `geopm_time.h <https://github.com/geopm/geopm/blob/dev/libgeopmd/include/geopm_time.h>`_
+header defines GEOPM interfaces for measuring time in seconds relative to a
+fixed arbitrary reference point. The ``geopm_time_s`` structure is used to hold
+time values.
 
 
 ``geopm_time_string()``
@@ -53,7 +79,14 @@ structure is used to hold time values.
   must be at least 26 characters as required by `asctime_r(3) <https://man7.org/linux/man-pages/man3/asctime_r.3.html>`_.
 
 ``geopm_time()``
-  Sets *time* to the current time.
+  Sets *time* to the current time in seconds since system boot
+  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
+  approximate date string.
+
+``geopm_time_real()``
+   Sets *time* to the current time in seconds since the epoch (CLOCK_REALTIME).
+   This can be used with geopm_time_real_to_iso_string() for an accurate ISO
+   8601 standard date string representation.
 
 ``geopm_time_diff()``
   Returns the difference in seconds between *begin* and *end*.
@@ -67,9 +100,51 @@ structure is used to hold time values.
 ``geopm_time_since()``
   Returns the number of seconds elapsed between the current time and *begin*.
 
+``geopm_time_zero()``
+  Return the time since the first call to this interface or
+  the ``geopm::time_zero()`` interface by the calling process.
+  See also: ``geopm::time_zero_reset()``.
+
+``geopm_time_to_string()``
+  Convert a time representation returned by ``geopm_time()`` or
+  ``geopm::time_curr()`` into an approximate date string.
+
+``geopm_time_real_to_iso_string()``
+  Convert a time representation returned by ``geopm_time_real()`` or
+  ``geopm::time_curr_real()`` into an ISO 8601 standard date string with
+  nan-second resolution.
+
+``geopm_time_string()``
+  Provide a date string for the current time in the format returned by
+  geopm_time_to_string().
+
+``geopm::time_zero()``
+  Return the time since the first call to this interface or the
+  ``geopm_time_zero()`` interface by the calling process.
+  See also: ``geopm::time_zero_reset()``.
+
+``geopm::time_curr()``
+  Returns the current time in seconds since system boot
+  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
+  approximate date string.
+
+``geopm::time_curr_real()``
+  Returns the current time in seconds since system boot
+  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
+  approximate date string.
+
+``geopm_time_real()``
+   Returns the current time in seconds since the epoch (CLOCK_REALTIME).
+   This can be used with geopm_time_real_to_iso_string() for an accurate ISO
+   8601 standard date string representation.
+
+``geopm::time_zero_reset()``
+   Override the reference time for the ``geopm_time_zero()`` and
+   ``geopm::time_zero()`` with the value *zero*.
+
+
 Structure Type
 --------------
-
 
 This structure is part of the global **C** namespace.
 This structure is used to abstract the ``timespec`` on Linux from other representations of time.
@@ -83,13 +158,6 @@ This is like a ``struct timeval`` but has *nanoseconds* instead of *microseconds
            struct timespec t;
        };
 
-Singleton Accessor
-------------------
-
-
-``geopm_time_zero()``
-  Is the accessor for the ``TimeZero`` singleton.
-  It records the earliest time request for the process.
 
 See Also
 --------
--- a/docs/source/geopm_time.3.rst
+++ b/docs/source/geopm_time.3.rst
@@ -114,7 +114,7 @@ time values.
 ``geopm_time_real_to_iso_string()``
   Convert a time representation returned by ``geopm_time_real()`` or
   ``geopm::time_curr_real()`` into an ISO 8601 standard date string with
-  nan-second resolution.
+  nanosecond resolution.
 
 ``geopm_time_string()``
   Provide a date string for the current time in the format returned by
--- a/docs/source/geopm_time.3.rst
+++ b/docs/source/geopm_time.3.rst
@@ -75,7 +75,7 @@ interfaces use this as the time reference.  This clock is the most precise for
 calculating time differences.
 
 There are some times when absolute time accuracy is preferred over time
-difference accuracy. For these cases, the GEOPM provides time functions that use
+difference accuracy. For these cases, GEOPM provides time functions that use
 ``CLOCK_REALTIME`` as the reference time.  There is a C accessor for real time:
 ``geopm_time_real()``, and there is also a C++ accessor:
 ``geopm::time_curr_real()``.  Only ``geopm_time_s`` values created with these
--- a/docs/source/geopm_time.3.rst
+++ b/docs/source/geopm_time.3.rst
@@ -61,6 +61,8 @@ C++ interfaces:
 
        struct geopm_time_s geopm::time_curr_real(void);
 
+       std::string geopm::time_curr_string(void);
+
        void geopm::time_zero_reset(const geopm_time_s &zero);
 
 
@@ -129,18 +131,17 @@ time values.
   approximate date string.
 
 ``geopm::time_curr_real()``
-  Returns the current time in seconds since system boot
-  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
-  approximate date string.
+  Returns the current time in seconds since the epoch (CLOCK_REALTIME).
+  This can be used with geopm_time_real_to_iso_string() for an accurate ISO
+  8601 standard date string representation.
 
-``geopm_time_real()``
-   Returns the current time in seconds since the epoch (CLOCK_REALTIME).
-   This can be used with geopm_time_real_to_iso_string() for an accurate ISO
-   8601 standard date string representation.
+``geopm::time_curr_string()``
+  Returns an accurate ISO 8601 standard date string representation of
+  the current (real) time.
 
 ``geopm::time_zero_reset()``
-   Override the reference time for the ``geopm_time_zero()`` and
-   ``geopm::time_zero()`` with the value *zero*.
+  Override the reference time for the ``geopm_time_zero()`` and
+  ``geopm::time_zero()`` with the value *zero*.
 
 
 Structure Type
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -82,9 +82,9 @@ Options
 
     Stop the session when the given process PID ends
 
---print-header  .. _header option:
+--omit-header  .. _header option:
 
-    Print a CSV header before printing any sampled values
+    Do not print the CSV header before printing any sampled values
 
 --delimiter .. _delimiter DELIMITER .. option:
 
@@ -195,6 +195,20 @@ An example report is shown below:
            max: 1.055e+09
            mean: 1.00394e+09
            std: 1.36988e+07
+
+
+Gathering Reports using MPI
+---------------------------
+The ``--mpi-report`` command line option can be used to aggregate reports using
+an MPI communicator.  This can be helpful when running sessions on more than one
+compute node.  To use this option, the user must install the optional ``mpi4py``
+package using their OS package manager or PyPi.  When running in this way the
+``geopmsession`` command line tool must be launched with a mpi launch wrapper
+like mpiexec or mpirun.  The user should run this command specifying one
+``geopmsession`` process per compute node.  When using this option, trace output
+is disabled.  The aggregated report is generated only by the "rank 0" process of
+the geopmsession MPI communicator.
+
 
 See Also
 --------
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -180,6 +180,7 @@ An example report is shown below:
    host: "cluster-node-11"
    sample-time-first: "2024-08-13T13:22:20.719602459-0700"
    sample-time-total: 9.87273
+   sample-count: 2001
    sample-period-mean: 0.00493637
    sample-period-std: 0.000535206
    metrics:
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -174,34 +174,36 @@ An example report is shown below:
 
 .. code-block:: yaml
 
-   host: "123.45.6.7"
-   time-begin: "2024-08-13T09:31:39.443889240-0700"
-   duration: 9.87292
+   host: "cluster-node-11"
+   sample-time-first: "2024-08-13T13:22:20.719602459-0700"
+   sample-time-total: 9.87273
+   sample-period-mean: 0.00493637
+   sample-period-std: 0.000535206
    metrics:
      TIME:
        count: 2001
-       first: 0.177136
-       last: 10.0501
-       min: 0.177136
-       max: 10.0501
-       mean: 5.05102
-       std: 2.8874
+       first: 0.187639
+       last: 10.0604
+       min: 0.187639
+       max: 10.0604
+       mean-arithmetic: 5.06119
+       std: 2.88731
      CPU_POWER:
        count: 2000
-       first: 167.782
-       last: 77.8053
-       min: 72.904
-       max: 167.782
-       mean: 77.6184
-       std: 3.95088
+       first: 74.91
+       last: 77.5113
+       min: 73.1154
+       max: 104.44
+       mean-arithmetic: 77.3586
+       std: 2.12181
      CPU_FREQUENCY_STATUS:
        count: 2001
-       first: 1.035e+09
+       first: 1.02625e+09
        last: 1e+09
        min: 1e+09
-       max: 1.07e+09
-       mean: 1.00116e+09
-       std: 7.18921e+06
+       max: 1.05625e+09
+       mean-arithmetic: 1.00109e+09
+       std: 7.25891e+06
 
 
 Gathering Reports using MPI
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -166,35 +166,34 @@ An example report is shown below:
 
 .. code-block:: yaml
 
-   hosts:
-     localhost:
-       time-begin: "2024-08-06T11:43:33.795328712-0700"
-       time-end: "2024-08-06T11:43:35.670129776-0700"
-       metrics:
-         TIME:
-           count: 201
-           first: 0.175729
-           last: 2.05037
-           min: 0.175729
-           max: 2.05037
-           mean: 1.05502
-           std: 0.574284
-         CPU_POWER:
-           count: 200
-           first: 87.2023
-           last: 70.0256
-           min: 67.9781
-           max: 97.1403
-           mean: 74.3168
-           std: 5.06087
-         CPU_FREQUENCY_STATUS:
-           count: 201
-           first: 1.035e+09
-           last: 1e+09
-           min: 1e+09
-           max: 1.055e+09
-           mean: 1.00394e+09
-           std: 1.36988e+07
+host: "123.45.6.7"
+time-begin: "2024-08-09T16:39:13.378279712-0700"
+time-end: "2024-08-09T16:39:23.385100904-0700"
+metrics:
+  TIME:
+    count: 2001
+    first: 0.207557
+    last: 10.0613
+    min: 0.207557
+    max: 10.0613
+    mean: 5.06262
+    std: 2.8867
+  CPU_POWER:
+    count: 2000
+    first: 87.0568
+    last: 83.0846
+    min: 77.2518
+    max: 99.1282
+    mean: 81.4213
+    std: 1.86577
+  CPU_FREQUENCY_STATUS:
+    count: 2001
+    first: 1.05875e+09
+    last: 1e+09
+    min: 1e+09
+    max: 1.1275e+09
+    mean: 1.00198e+09
+    std: 1.20525e+07
 
 
 Gathering Reports using MPI
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -82,7 +82,7 @@ Options
 
     Stop the session when the given process PID ends
 
---omit-header  .. _header option:
+--no-header  .. _header option:
 
     Do not print the CSV header before printing any sampled values
 
@@ -157,7 +157,8 @@ Multiple signals may be specified by separating them with a newline.
 
 .. code-block:: none
 
-    echo -e 'TIME board 0\nCPU_FREQUENCY_STATUS package 0\nCPU_FREQUENCY_STATUS package 1\nCPU_ENERGY package 0\nCPU_ENERGY package 1' | geopmsession
+    printf 'TIME board 0\nCPU_FREQUENCY_STATUS package 0\nCPU_FREQUENCY_STATUS package 1\nCPU_ENERGY package 0\nCPU_ENERGY package 1' |\
+        geopmsession --no-header
     70.250978379,2434090909.090909,2775000000,198575.8842163086,88752.19470214844
 
 Reading a set of signals and getting summary statistics
@@ -168,7 +169,8 @@ unspecified, no statistics will be gathered.
 
 .. code-block:: none
 
-    printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | geopmsession -t 2 -p 0.01 --print-header --report-out=-
+    printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' |\
+        geopmsession -t 10 -p 0.005 --report-out=- --trace-out=/dev/null
 
 An example report is shown below:
 
@@ -214,7 +216,7 @@ an MPI communicator.  This can be helpful when running sessions on more than one
 compute node in an MPI enabled environment.  The user must install the optional
 ``mpi4py`` package to use the ``--enable-mpi`` command line option .  This can
 be done using the OS package manager or PyPi.  When running in this way the
-``geopmsession`` command line tool must be launched with a mpi launch wrapper
+``geopmsession`` command line tool must be launched with an MPI launch wrapper
 like ``mpiexec`` or ``mpirun``.  The user should run this command specifying one
 ``geopmsession`` process per compute node.  When using this option, trace output
 to stdout is disabled.  The aggregated report is created by the "rank 0" process
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -178,36 +178,36 @@ An example report is shown below:
 .. code-block:: yaml
 
    host: "cluster-node-11"
-   sample-time-first: "2024-08-13T13:22:20.719602459-0700"
-   sample-time-total: 9.87273
+   sample-time-first: "2024-08-14T18:23:58.545153099-0700"
+   sample-time-total: 9.99976
    sample-count: 2001
-   sample-period-mean: 0.00493637
-   sample-period-std: 0.000535206
+   sample-period-mean: 0.00499988
+   sample-period-std: 5.43807e-05
    metrics:
      TIME:
        count: 2001
-       first: 0.187639
-       last: 10.0604
-       min: 0.187639
-       max: 10.0604
-       mean-arithmetic: 5.06119
-       std: 2.88731
+       first: 0.0825453
+       last: 10.0823
+       min: 0.0825453
+       max: 10.0823
+       mean-arithmetic: 5.08268
+       std: 2.88873
      CPU_POWER:
        count: 2000
-       first: 74.91
-       last: 77.5113
-       min: 73.1154
-       max: 104.44
-       mean-arithmetic: 77.3586
-       std: 2.12181
+       first: 71.9161
+       last: 54.9997
+       min: 36.8681
+       max: 75.2159
+       mean-arithmetic: 50.1323
+       std: 6.61714
      CPU_FREQUENCY_STATUS:
        count: 2001
-       first: 1.02625e+09
-       last: 1e+09
+       first: 1.69773e+09
+       last: 1.75341e+09
        min: 1e+09
-       max: 1.05625e+09
-       mean-arithmetic: 1.00109e+09
-       std: 7.25891e+06
+       max: 2.49659e+09
+       mean-arithmetic: 1.5542e+09
+       std: 3.72332e+08
 
 
 Gathering Reports using MPI
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -225,6 +225,82 @@ to stdout is disabled.  The aggregated report is created by the "rank 0" process
 of the geopmsession MPI communicator.
 
 
+.. code-block:: shell-session
+
+   $ printf "TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0" |\
+         srun -n 2 -N 2 geopmsession -t 10 -p 0.005  -r- -o
+	 /dev/null --enable-mpi
+
+An example report is shown below:
+
+.. code-block:: yaml
+
+   host: "cluster-node-11"
+   sample-time-first: "2024-08-14T17:50:00.895968647-0700"
+   sample-time-total: 9.99973
+   sample-count: 2001
+   sample-period-mean: 0.00499987
+   sample-period-std: 5.22108e-05
+   metrics:
+     TIME:
+       count: 2001
+       first: 0.0849912
+       last: 10.0847
+       min: 0.0849912
+       max: 10.0847
+       mean-arithmetic: 5.08514
+       std: 2.88873
+     CPU_POWER:
+       count: 2000
+       first: 58.2713
+       last: 63.4941
+       min: 41.1496
+       max: 89.1348
+       mean-arithmetic: 55.8502
+       std: 6.41338
+     CPU_FREQUENCY_STATUS:
+       count: 2001
+       first: 2.17727e+09
+       last: 1.75341e+09
+       min: 1e+09
+       max: 2.58636e+09
+       mean-arithmetic: 1.55674e+09
+       std: 3.7101e+08
+
+   ---
+
+   host: "cluster-node-12"
+   sample-time-first: "2024-08-14T17:50:01.033367154-0700"
+   sample-time-total: 10.0003
+   sample-count: 2001
+   sample-period-mean: 0.00500015
+   sample-period-std: 5.06152e-05
+   metrics:
+     TIME:
+       count: 2001
+       first: 0.0846359
+       last: 10.0849
+       min: 0.0846359
+       max: 10.0849
+       mean-arithmetic: 5.085
+       std: 2.88887
+     CPU_POWER:
+       count: 2000
+       first: 60.455
+       last: 70.6912
+       min: 46.394
+       max: 89.6428
+       mean-arithmetic: 61.7341
+       std: 5.20186
+     CPU_FREQUENCY_STATUS:
+       count: 2001
+       first: 1.70568e+09
+       last: 1.69773e+09
+       min: 1e+09
+       max: 2.56136e+09
+       mean-arithmetic: 1.54734e+09
+       std: 3.63195e+08
+
 See Also
 --------
 
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -8,7 +8,7 @@ Synopsis
 .. code-block:: bash
 
    usage: geopmsession [-h] [-v] [-t TIME] [-p PERIOD] [--pid PID]
-                       [--print-header] [-n] [-d DELIMITER] [-r REPORT_OUT]
+                       [--print-header | -n] [-d DELIMITER] [-r REPORT_OUT]
                        [-o TRACE_OUT] [--enable-mpi]
 
 Read a signal
--- a/docs/source/geopmsession.1.rst
+++ b/docs/source/geopmsession.1.rst
@@ -92,18 +92,25 @@ Options
 
 --report-out .. _reportout REPORT_OUT option:
 
-    Output summary statistics into a yaml file. Note if ``--report-out=-``
-    is specified, the report will output to stdout.
+    Output summary statistics into a yaml file. Note if ``--report-out=-`` is
+    specified, the report will output to stdout. When used with the
+    ``--enable-mpi`` option, reports from all hosts will be combined using the
+    ``---`` line separator, and the output is written (stdout or to file) solely
+    by the MPI process "rank 0".
 
 --trace-out .. _traceout TRACE_OUT option:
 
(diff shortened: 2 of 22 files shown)


geopmdpy/geopmdpy/session.py
```diff
@@ -14,8 +14,13 @@
 from . import topo
 from . import pio
 from . import loop
+from . import stats
 from . import __version_str__
 
+try:
+    from mpi4py import MPI
```
bgeltz, line 23: Add to requirements.txt?

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,62 @@
...
+
+def collector_update():
+    global _dl
+    err = _dl.geopm_stats_collector_update()
+    if err < 0:
+        raise RuntimeError('geopm_stats_collector_update() failed: {}'.format(error.message(err)))
+
+def collector_report():
+    global _dl
+    report_max = 262144
```
bgeltz: Please comment on where this came from.

bgeltz (dismissed):
AFAICT still TODO:
Docs: RST updates, python docstings
New tests
Fix broken tests
requirements.txt updates

bgeltz:
Other opens:

> A new python class (based on the Session class) reads signals provided by the end user in a file.

Does this accept a file?  An example showing as much should be included in the docs.

---

```
    CPU_POWER:
      count: 201
      first: 1.02625e+09
      last: 73.4791
      min: 63.4933
      max: 1.02625e+09
      mean: 5.10579e+06
      std: 7.23861e+07
```
These numbers look odd.  Why are the min and last 6 to 9 orders of magnitude off compared to the other stats?

asmaalrawi:
> These numbers look odd. Why are the min and last 6 to 9 orders of magnitude off compared to the other stats?

I'm guessing we still have the issue where you get a random first reading on a signal based on a difference (POWER). This would mess up first/max/mean/std but leave count/min/last unaltered.

cmcantalupo:
A suggested future feature of the report:  Enable selecting on a parameter like hint or hash.  The current report looks like:

```
$ printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' |  geopmsession -t 2 -p 0.01 --print-header --report-out=report.yaml -d\| --trace-out=trace.csv
$ cat report.yaml
hosts:
  saruman:
    time-begin: "Thu Aug 01 11:50:42 2024"
    time-end: "Thu Aug 01 11:50:43 2024"
    metrics:
      TIME:
        count: 201
        first: 0.165067
        last: 2.04619
        min: 0.165067
        max: 2.04619
        mean: 1.05035
        std: 0.575056
      CPU_POWER:
        count: 200
        first: 109.982
        last: 73.0664
        min: 67.2973
        max: 109.982
        mean: 74.319
        std: 6.90355
      CPU_FREQUENCY_STATUS:
        count: 201
        first: 1.02625e+09
        last: 1e+09
        min: 1e+09
        max: 1.0525e+09
        mean: 1.00371e+09
        std: 1.32582e+07
        
```


in the future we may consider adding a "--report-select" feature where the output from `--report-select=REGION_HINT` might look like (made up data):

```
$ printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\nREGION_HINT board 0\n' |  geopmsession -t 2 -p 0.01 --print-header --report-out=report.yaml -d\| --trace-out=trace.csv --report-select=REGION_HINT-board-0
$ cat report.yaml
hosts:
  saruman:
    time-begin: "Thu Aug 01 11:50:42 2024"
    time-end: "Thu Aug 01 11:50:43 2024"
    metrics:
      TIME:
        count: 201
        first: 0.165067
        last: 2.04619
        min: 0.165067
        max: 2.04619
        mean: 1.05035
        std: 0.575056
      CPU_POWER:
        count: 200
        first: 109.982
        last: 73.0664
        min: 67.2973
        max: 109.982
        mean: 74.319
        std: 6.90355
      CPU_FREQUENCY_STATUS:
        count: 201
        first: 1.02625e+09
        last: 1e+09
        min: 1e+09
        max: 1.0525e+09
        mean: 1.00371e+09
        std: 1.32582e+07
      REGION_HINT:
        count: 201
        first: 1
        last: 1
        min: 0
        max: 4
        mean: 3.01
    select:
      REGION_HINT:
        value: 1
          metrics:
            TIME:
              count: 100
              first: 0.165067
              last: 1.04619
              min: 0.165067
              max: 1.04619
              mean: 0.55035
              std: 0.275056
            CPU_POWER:
              count: 100
              first: 109.982
              last: 73.0664
              min: 67.2973
              max: 109.982
              mean: 74.319
              std: 6.90355
            CPU_FREQUENCY_STATUS:
              count: 100
              first: 1.02625e+09
              last: 1e+09
              min: 1e+09
              max: 1.0525e+09
              mean: 1.00371e+09
              std: 1.32582e+07
        value: 4
          metrics:
            TIME:
              count: 101
              first: 0.165067
              last: 1.04619
              min: 0.165067
              max: 1.04619
              mean: 0.55035
              std: 0.275056
            CPU_POWER:
              count: 100
              first: 109.982
              last: 73.0664
              min: 67.2973
              max: 109.982
              mean: 74.319
              std: 6.90355
            CPU_FREQUENCY_STATUS:
              count: 101
              first: 1.02625e+09
              last: 1e+09
              min: 1e+09
              max: 1.0525e+09
              mean: 1.00371e+09
              std: 1.32582e+07
```

libgeopmd/include/geopm_time.h
```diff
@@ -14,6 +14,7 @@
 #ifndef __cplusplus
 #include <stdbool.h>
 #else
+#include <string>
```
cmcantalupo, line 17: delete

geopmdpy/geopmdpy/session.py
cmcantalupo: Since it is optional and also has an unspecified dependency on some flavor of MPI development packages being installed, I've opted to leave it out.

libgeopmd/src/RuntimeStats.hpp
```diff
@@ -0,0 +1,54 @@
...
+                double last;
+                double min;
+                double max;
+                double m_1;
+                double m_2;
+                double m_3;
+                double m_4;
+            };
+            const std::vector<std::string> m_metric_names;
+            std::vector<stats_s> m_moments;
```
dannosliwcd: Maybe m_stats instead? Otherwise `m_moments[1]` or `m_moments[#].first` read like "first moment"

libgeopmd/src/RuntimeStats.cpp
```diff
@@ -0,0 +1,181 @@
...
+    double RuntimeStats::std(int metric_idx) const
+    {
+        check_index(metric_idx, __func__, __LINE__);
+        double result = NAN;
+        if (m_moments[metric_idx].count > 1) {
+            result = std::sqrt(
+                         (m_moments[metric_idx].m_2 -
+                          m_moments[metric_idx].m_1 *
+                          m_moments[metric_idx].m_1 / m_moments[metric_idx].count) /
+                         (m_moments[metric_idx].count - 1));
```
dannosliwcd: I was expecting the second term inside the sqrt to be minus mean-squared:
```
 - m_moments[metric_idx].m_1 * m_moments[metric_idx].m_1
   / (m_moments[metric_idx].count * m_moments[metric_idx].count)
```

But I'm not sure if I'm interpreting m_1 and m_2 correctly. Need to return after reviewing the rest of this class

libgeopmd/src/RuntimeStats.cpp
```diff
@@ -0,0 +1,181 @@
...
+        }
+        return result;
+    }
+
+    double RuntimeStats::skew(int metric_idx) const
+    {
+        throw Exception("RuntimeStats::" + std::string(__func__) + " not yet implemented",
```
dannosliwcd: We should probably just drop these functions from the interface until they have an implementation.

libgeopmd/src/RuntimeStats.cpp
```diff
@@ -0,0 +1,181 @@
...
+                }
+                moments_it->last = ss;
+                if (moments_it->min > ss) {
+                    moments_it->min = ss;
+                }
+                if (moments_it->max < ss) {
+                    moments_it->max = ss;
+                }
+                double mm = ss;
+                moments_it->m_1 += mm;
```
dannosliwcd, line 144: I guess `m_#` is (catching up on wikipedia just now) is a running accumulation of `count * #th raw sample moment`

libgeopmd/src/RuntimeStats.cpp
dannosliwcd: Coming back here, I think we have
```
std = sqrt(N*moment_2 - N*moment_1 * N*moment_1 / N) / (N-1)
    = sqrt(N*moment_2 - N*moment_1^2) / (N-1)
    = sqrt(N) * sqrt(moment_2 - moment_1^2) / (N-1)
```

But I think we're aiming for this?
```
std = sqrt(moment_2 - moment_1^2)
```

I think I'm a bit lost. Can you point me to which formula I should be using?

dannosliwcd (commented):
Partial review. I looked at RuntimeStats so far

libgeopmd/src/RuntimeStats.hpp
cmcantalupo: changed to m_metric_stats

libgeopmd/src/RuntimeStats.cpp
cmcantalupo: This issue with your formula is the bias:

https://en.wikipedia.org/wiki/Unbiased_estimation_of_standard_deviation

I can't find my references.  It can be derived, but you use up a degree of freedom by using m_1 and m_2. This is where the n-1 is coming from.

asmaalrawi: The n-1 is used in situations where you wish to extrapolate, i.e. you are estimating the stdev of a distribution based on a set of samples. If you wish to get the stdev of a set of samples without wishing to extrapolate, you can use n. It's not clear to me which is more appropriate here. Is the purpose of the reporter to report statistics of the samples that we've taken, or to understand the characteristics of the hw/wkld?

dannosliwcd: What about the extra `sqrt(n)`?

dannosliwcd: I think the problem is that I misaligned parentheses when translating from the code to my first line above. Instead I think it translates to:

```
std = sqrt( (N*moment_2 - N*moment_1 * N*moment_1 / N) / (N-1) ) <-- I accidentally moved N-1 outside sqrt earlier
    = sqrt( (N*moment_2 - N*moment_1^2 ) / (N-1) )
```

Which works out for sample standard deviation

dannosliwcd:
How does this react to sigint?

E.g.
```
geopmsession -t 9e9 --report-out=report.yaml --trace-out=/dev/null
...
*ctrl-C*
```

Will the user still receive a report of whatever happened up until the interrupt?

docs/source/geopm_time.3.rst
```diff
@@ -53,7 +81,14 @@ structure is used to hold time values.
   must be at least 26 characters as required by `asctime_r(3) <https://man7.org/linux/man-pages/man3/asctime_r.3.html>`_.
 
 ``geopm_time()``
-  Sets *time* to the current time.
+  Sets *time* to the current time in seconds since system boot
+  (CLOCK_MONOTONIC_RAW). This can be used with geopm_time_to_string() to get a an
+  approximate date string.
```
dannosliwcd: Are "CLOCK_MONOTONIC_RAW" and "seconds since system boot" necessary interface-level details? I.e., do we want users to rely on that always being the implementation?

cmcantalupo:
We don't have any signal handling in geopmsession, so you would get no output if you sent a sig int

https://github.com/geopm/geopm/issues/3551

docs/source/geopm_time.3.rst
```diff
@@ -67,10 +102,51 @@ structure is used to hold time values.
...
   Returns the number of seconds elapsed between the current time and *begin*.
 
+``geopm_time_zero()``
+  Return the time since the first call to this interface or
+  the ``geopm::time_zero()`` interface by the calling process.
+  See also: ``geopm::time_zero_reset()``.
+
+``geopm_time_to_string()``
+  Convert a time representation returned by ``geopm_time()`` or
+  ``geopm::time_curr()`` into an approximate date string.
```
dannosliwcd: Does this function work with all `geopm_time_s`, or were those generated by `geopm_time_real()` intentionally omitted from this description?

docs/source/geopm_time.3.rst
```diff
@@ -67,10 +102,51 @@ structure is used to hold time values.
...
+
+``geopm_time_real_to_iso_string()``
+  Convert a time representation returned by ``geopm_time_real()`` or
+  ``geopm::time_curr_real()`` into an ISO 8601 standard date string with
+  nan-second resolution.
```
dannosliwcd:
```suggestion
  nanosecond resolution.
```

docs/source/geopm_time.3.rst
```diff
@@ -67,10 +102,51 @@ structure is used to hold time values.
...
+  approximate date string.
+
+``geopm::time_curr_real()``
+  Returns the current time in seconds since the epoch (CLOCK_REALTIME).
+  This can be used with geopm_time_real_to_iso_string() for an accurate ISO
+  8601 standard date string representation.
+
+``geopm::time_curr_string()``
+  Returns an accurate ISO 8601 standard date string representation of
+  the current (real) time.
```
dannosliwcd: Since this specifies "(real)", should it be `geopm::time_curr_real_string()`?

Do we even need any of the `*_string()` variants other than `geopm_time_to_string()`?

docs/source/geopmsession.1.rst
```diff
@@ -84,6 +86,25 @@ Options
...
+--trace-out .. _traceout TRACE_OUT option:
+
+    Output trace data into a CSV file. Note if ``--trace_out=-`` is specified,
+    the trace will output to stdout. To avoid gathering trace data,
+    set this parameter to ``/dev/null``
+
+--mpi-report .. _mpireport option:
+
+    Gather reports over mpi and write to a single file
+
```
dannosliwcd: The intent/value behind this option might be unclear for people exploring this documentation and who aren't already thinking of multi-node use cases.

Also are there any documentable expectations about how `geopmsession` is launched when this option is used? E.g., do we expect no more than one rank per node?

docs/source/geopmsession.1.rst
```diff
@@ -8,6 +8,8 @@ Synopsis
 .. code-block:: none
 
     geopmsession [-h] [-v] [-t TIME] [-p PERIOD] [--pid PID] [--print-header]
```
dannosliwcd: Would like to discuss (don't need to resolve in this PR. I just want to know whether I should forget this or open a feature request):

What are your thoughts on changing the default behavior for headers in the future?

`--print-header` feels like the common use case to me. Could we change the default behavior to include headers in a future release (plus add `--no-print-header` and emit some kind of warning in releases leading up to that)?

docs/source/geopmsession.1.rst
```diff
@@ -132,6 +152,49 @@ Multiple signals may be specified by separating them with a newline.
...
 
+Reading a set of signals and getting summary statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Summary statistics may be output to stdout by setting ``--report-out=-``.
+Otherwise, the statistics will be output to the specified file path. If
+unspecified, no statistics will be gathered.
+
+.. code-block:: none
+
+    printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | geopmsession -t 2 -p 0.01 --print-header --report-out=- "TIME","CPU_POWER","CPU_FREQUENCY_STATUS"
```
dannosliwcd: Is `"TIME","CPU_POWER","CPU_FREQUENCY_STATUS"` an unintended addition here? If not, then posititional args need to be added to the command's synopsis.

docs/source/geopmsession.1.rst
```diff
@@ -132,6 +152,49 @@ Multiple signals may be specified by separating them with a newline.
...
+           mean: 74.3168
+           std: 5.06087
+         CPU_FREQUENCY_STATUS:
+           count: 201
+           first: 1.035e+09
+           last: 1e+09
+           min: 1e+09
+           max: 1.055e+09
+           mean: 1.00394e+09
+           std: 1.36988e+07
```
dannosliwcd: Maybe follow with a brief summary about how to interpret this. It doesn't need to be a full breakdown, but mention something like the three requested signals from the command each have a section that summarizes observations on each host across the time span bound by time-begin and time-end.

docs/source/geopmsession.1.rst
```diff
@@ -132,6 +152,49 @@ Multiple signals may be specified by separating them with a newline.
...
+.. code-block:: none
+
+    printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | geopmsession -t 2 -p 0.01 --print-header --report-out=- "TIME","CPU_POWER","CPU_FREQUENCY_STATUS"
+
+An example report is shown below:
+
+.. code-block:: yaml
+
+   hosts:
+     localhost:
```
dannosliwcd: `localhost` might be misinterpreted as a literal string people would expect to see in multi-node reports (i.e., hosts are `localhost`, `123.45.6.7`, `...`). Maybe use some other placeholder hostname that doesn't typically have special meaning.

docs/source/geopmsession.1.rst
dannosliwcd: This could be an opportunity to note why power has a different count from the other signals.

geopmdpy/geopmdpy/pio.py
```diff
@@ -600,7 +600,7 @@ def start_batch_server(client_pid, signal_config, control_config):
         RuntimeError: Failure to start the batch server.
 
     """
-
+    global _dl
```
dannosliwcd, line 603: I didn't think this would be necessary since we aren't assigning the `_dl` variable to a new target in this scope. Why is it needed?

geopmdpy/geopmdpy/session.py
dannosliwcd: You could add an optional dependency if you think there's value in one-liner pip installs.

E.g., `pip install geopmdpy[mpi]` could install `geopmdpy` as normal, but also add `mpi4py`.

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

If you do decide to try that out, try it on python3.6 first in case it's another one of those packaging things that doesn't work well until we drop 3.6 support.

geopmdpy/geopmdpy/session.py
```diff
@@ -97,12 +111,15 @@ def run_read(self, requests, duration, period, pid, out_stream):
...
-
         for sample_idx in loop.TimedLoop(period, num_period):
             pio.read_batch()
-            signals = [pio.sample(handle) for handle in signal_handles]
-            line = self.format_signals(signals, requests.get_formats())
-            out_stream.write(line)
+            if out_stream is not None:
+                signals = [pio.sample(handle) for handle in signal_handles]
+                line = self.format_signals(signals, requests.get_formats())
+                out_stream.write(line)
```
dannosliwcd, line 216: Should we be flushing here? Example where it might matter:

`geopmsession -t 123 -p 1 <<< 'TIME board 0' | tee geopm-trace.csv` appears to do nothing for a while until the write buffer fills up enough to flush automatically

geopmdpy/geopmdpy/session.py
```diff
@@ -168,15 +191,27 @@ def run(self, run_time, period, pid, print_header,
...
+            self.run_read(self._requests, run_time, period, pid, out_stream, stats_collector)
+            if do_stats:
+                report = stats_collector.report_yaml()
+        rank = 0
+        if do_stats and mpi_comm is not None:
+            report_list = mpi_comm.gather(report, root=0)
+            rank = mpi_comm.Get_rank()
+            if rank == 0:
+                tmp_list = [report_list[0]]
+                report_list = [rr.replace('Hosts:\n', '') for rr in report_list[1:]]
```
dannosliwcd: It seems brittle to modify already-formatted report strings in place.

I guess we could avoid the back-and-forth between strings and their structured interpretations by gathering the reports as `dict`s, then build a final report like `dict(host={report_host: report_host_data for report in report_list for report_host, report_host_data in report['host'].items()})`, then have pyyaml dump that dict to `report_stream`. But if we always plan to keep the report strings formatted this way, then that might be unnecessary.

geopmdpy/geopmdpy/session.py
```diff
@@ -323,38 +358,73 @@ def get_formats(self):
...
-    parser.add_argument('-p', '--period', dest='period', type=float, default = 0.0,
+    parser.add_argument('-p', '--period', dest='period', type=float, default=0.0,
                         help='When used with a read mode session reads all values out periodically with the specified period in seconds')
     parser.add_argument('--pid', type=int,
                         help='Stop the session when the given process PID ends')
     parser.add_argument('--print-header', action='store_true',
                         help='Print a CSV header before printing any sampled values')
-    args = parser.parse_args()
+    parser.add_argument('-d', '--delimiter', dest='delimiter', default=',',
+                        help='Delimiter used to separate values in CSV output')
```
dannosliwcd:
```suggestion
                        help='Delimiter used to separate values in CSV output. Default: %(default)s')
```

geopmdpy/geopmdpy/session.py
```diff
@@ -323,38 +358,73 @@ def get_formats(self):
...
+    parser.add_argument('--report-out', dest='report_out', default=None,
+                        help='Output summary statistics into a yaml file')
```
dannosliwcd:
```suggestion
                        help='Output summary statistics into a yaml file. Default: No summary statistics are generated.')
```

geopmdpy/geopmdpy/session.py
```diff
@@ -323,38 +358,73 @@ def get_formats(self):
...
+    parser.add_argument('--trace-out', dest='trace_out', default='-',
+                        help='Output trace data into a CSV file')
```
dannosliwcd:
```suggestion
                        help='Output trace data into a CSV file. Default: trace data is printed to stdout.')
```

geopmdpy/geopmdpy/session.py
```diff
@@ -323,38 +358,73 @@ def get_formats(self):
...
+    parser.add_argument('--mpi-report', dest='mpi_report', action='store_true',
+                        help='Gather reports over mpi and write to a single file')
```
dannosliwcd:
```suggestion
    parser.add_argument('--mpi-report', dest='mpi_report', action='store_true',
                        help='Gather reports over mpi and write to a single file specified by '
                        'MPI_REPORT. Default: Reports are generated local to each host executing %(prog)s')
```

docs/source/geopmsession.1.rst
```diff
@@ -84,6 +86,25 @@ Options
...
 
+--delimiter .. _delimiter DELIMITER .. option:
+
+    Delimiter used to separate values in CSV output
+
+--report-out .. _reportout REPORT_OUT option:
+
+    Output summary statistics into a yaml file. Note if ``--report-out=-``
+    is specified, the report will output to stdout.
+
```
dannosliwcd: How does this interact with `--mpi-report`?

dannosliwcd (commented):
Partial review. Will resume starting with stats.py

geopmdpy/geopmdpy/session.py
dannosliwcd: Also relates to https://github.com/geopm/geopm/issues/3543#issuecomment-2276474324, where we get no report if a run is interrupted. For traces, the lack of signal handling means we get truncated output on non-interactive runs (which typically do not default to line-buffered)

docs/source/geopmsession.1.rst
cmcantalupo: Thanks...will fix the copypasta

geopmdpy/geopmdpy/pio.py
cmcantalupo: Perhaps you are right. It is unclear to me if _dl is modified when you calls to its attributes.  Perhaps this is actually not relevant to the global syntax. The pattern is followed in pio.py

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Args:
+            signal_config (list((str, int, int))): List of requested
+                signals where each tuple represents
+                (signal_name, domain_type, domain_idx).
+
+        """
+        global _dl
+        self._collector_ptr = None
+        signal_config = [(rr[0], rr[1], rr[2]) for rr in signal_config]
```
dannosliwcd: This seems like it would have no effect. Is it working around something?

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,152 @@
...
+        num_signal = len(signal_config)
+        if num_signal == 0:
+            raise RuntimeError('Collector creation failed: length of input is zero')
```
dannosliwcd:
```suggestion
            raise ValueError('Collector creation failed: length of input is zero')
```

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,152 @@
...
+
+        signal_config_carr = gffi.gffi.new(f'struct geopm_request_s[{num_signal}]')
```
dannosliwcd:
```suggestion
        signal_config_carr = gffi.gffi.new('struct geopm_request_s[]', num_signal)
```
Not sure if this really matters. The cffi docs recommend it this way because it causes unnecessary re-parsing of their internally-cached first-arg strings. Since we'll likely only call this once per process that parser cache won't ever hit anyways.

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,152 @@
...
+        report_max = gffi.gffi.new("size_t *")
+        report_max[0] = 0
+        _dl.geopm_stats_collector_report_yaml(self._collector_ptr, report_max, gffi.gffi.NULL)
+        report_cstr = gffi.gffi.new("char[]", report_max[0])
+        err = _dl.geopm_stats_collector_report_yaml(self._collector_ptr, report_max, report_cstr)
+        if err < 0:
+            raise RuntimeError('geopm_stats_collector_report_yaml() failed: {}'.format(error.message(err)))
+        return gffi.gffi.string(report_cstr).decode()
+
+    def report_json(self):
```
dannosliwcd: I think this is unused+untested. Since it just generates then immediately parses yaml to generate json, it doesn't seem to add much value. Maybe it can be removed

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        self._check_ptr('report_json')
+        result_yaml = self.report_yaml()
+        result_obj = yaml.safe_load(result_yaml)
+        return json.dumps(result_obj)
+
+    def reset(self):
+        """Reset statistics
+
+        Called by user to zero all statistics galthered.  This may be called
```
dannosliwcd:
```suggestion
        Called by user to zero all statistics gathered.  This may be called
```

geopmdpy/geopmdpy/stats.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Return a yaml string containing a report that shows all statistics
+        gathered by calls to update().
+
+        """
+        global _dl
+        self._check_ptr('report_yaml')
+        report_max = gffi.gffi.new("size_t *")
+        report_max[0] = 0
+        _dl.geopm_stats_collector_report_yaml(self._collector_ptr, report_max, gffi.gffi.NULL)
```
dannosliwcd, line 119: If we start doing scenarios with multiple reports generated per session, then we might consider caching the expected `report_max` inside the Collector instance so we don't need to run the yaml-generating code twice per report.

I think the current implementation is good for our current single-report-per-session case. Maybe just add a note to look into this in the issue tracker for the multi-report-per-session work.

dannosliwcd (commented):
Partial review. Will continue at TestSession.py

geopmdpy/geopmdpy/stats.py
cmcantalupo: otherwise len(signal_config) is unknown.  The text is actually parsed on this line

dannosliwcd: I think the docstring needs to be updated then. It currently says signal_config arg is expected to already be this type/shape.

libgeopmd/include/geopm_stats_collector.h
```diff
@@ -0,0 +1,92 @@
...
+///        be de-allocated with geopm_stats_collector_free()
+///
+/// @returns 0 upon success, or error code upon failure
+int GEOPM_PUBLIC
+    geopm_stats_collector_create(size_t num_requests, const struct geopm_request_s *requests,
+                                 struct geopm_stats_collector_s **collector);
+
+/// @brief Update a stat collector with new values
+///
+/// User is expected to call PlatformIO::read_batch() prior to calling this
```
dannosliwcd, line 40: Since this class depends on changing PlatformIO state, should its constructor take a pointer to PlatformIO?

libgeopmd/src/StatsCollector.cpp
```diff
@@ -0,0 +1,195 @@
...
+    StatsCollector::StatsCollector(const std::vector<geopm_request_s> &requests)
+        : StatsCollector(requests, platform_io())
+    {
+
+    }
+
+    StatsCollector::StatsCollector(const std::vector<geopm_request_s> &requests, PlatformIO &pio)
+        : m_pio(pio)
+        , m_stats(std::make_shared<RuntimeStats>(register_requests(requests)))
+        , m_time_begin(geopm::time_curr_string())
```
dannosliwcd: Wouldn't the report's beginning time be the time of whichever `PlatformIO::batch_read()` precedes the first call to this collector's `update()`?

libgeopmd/src/StatsCollector.cpp
```diff
@@ -0,0 +1,195 @@
...
+            *sample_it = m_pio.sample(pio_idx);
+            ++sample_it;
+        }
+        m_stats->update(sample);
+    }
+
+    std::string StatsCollector::report_yaml(void) const
+    {
+        std::ostringstream result;
+        std::string time_end_str = geopm::time_curr_string();
```
dannosliwcd, line 102: Similarly, would end time be the time of the previous batch_read?

dannosliwcd (commented):
Will resume in StatsCollector.cpp

libgeopmd/src/StatsCollector.cpp
```diff
@@ -0,0 +1,195 @@
...
+        err = err < 0 ? err : GEOPM_ERROR_RUNTIME;
+    }
+    return err;
+}
+
+int geopm_stats_collector_update(struct geopm_stats_collector_s *collector)
+{
+    int err = 0;
+    try {
+        geopm::StatsCollector *collector_cpp = reinterpret_cast<geopm::StatsCollector *>(collector);
```
dannosliwcd, line 176: You mentioned you disliked all of the casts. I guess one way you could avoid them is to define `StatsCollector` as a `struct` instead of a `class`, and forward-declare it in the C header. That said, I think this approach seems fine.

libgeopmd/src/StatsCollector.cpp
```diff
@@ -0,0 +1,195 @@
...
+int geopm_stats_collector_report_yaml(const struct geopm_stats_collector_s *collector,
+                                      size_t *max_report_size, char *report_yaml)
+{
+    int err = 0;
+    try {
+        const geopm::StatsCollector *collector_cpp = reinterpret_cast<const geopm::StatsCollector *>(collector);
+        std::string report_str = collector_cpp->report_yaml();
+        if (report_str.size() < *max_report_size) {
+            strncpy(report_yaml, report_str.c_str(), report_str.size());
+        }
```
dannosliwcd: Looks good. Talking through it:
* `std::string::size` excludes the terminator byte (i.e., it's equivalent to `strlen`)
* `strncpy` does not guarantee a nul bytes if the string doesn't contain it within the given `n` bytes.
* The first call to this function tells the user to allocate `string::size + 1` bytes. 👍
* User ignores the recommendation and allocs `string::size` bytes instead 😮
* This check evaluates as  `if (report_yaml.size < report_yaml.size) ==> if (false)` so we avoid the copy and return an error 👍

This would be good to cover with a test.

libgeopmd/src/StatsCollector.hpp
```diff
@@ -0,0 +1,64 @@
...
+            /// User specifies a vector of PlatformIO signal requests to be
+            /// accumulated.  The report will generate statistics about each
+            /// signal request.
+            ///
+            /// @param [in] requests All signals for monitoring and reporting
+            StatsCollector(const std::vector<geopm_request_s> &requests);
+            /// @brief Test constructor used to mock PlatformIO
+            StatsCollector(const std::vector<geopm_request_s> &requests, PlatformIO &pio);
+            /// @brief Default virtual destructor
+            virtual ~StatsCollector() = default;
```
dannosliwcd: This doesn't need to be virtual since nothing else is virtual in this class. No vtable necessary.

libgeopmd/src/StatsCollector.hpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#ifndef STATSCOLLECTOR_HPP_INCLUDE
+#define STATSCOLLECTOR_HPP_INCLUDE
+
+#include <vector>
+#include <string>
+#include <memory>
+
+struct geopm_request_s;
```
dannosliwcd: Doesn't the user of StatsCollector need to have a fully geopm_request_s definition in order to construct the collector?

libgeopmd/test/StatsCollectorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+        void SetUp();
+        std::shared_ptr<MockPlatformIO> m_pio_mock;
+};
+
+void StatsCollectorTest::SetUp()
+{
+    m_pio_mock = std::make_shared<MockPlatformIO>();
+}
+
+TEST_F(StatsCollectorTest, empty_report)
```
dannosliwcd, line 35: Checks that the report contains `hosts`, `<the test runner's hostname>`, `time-begin`, `time-end`, `metrics`, and optionally more fields in that order, with the right amount of leading whitespace, and with trailing colon characters.

libgeopmd/test/StatsCollectorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+        if (eb_it == expected_begin.end()) {
+            break;
+        }
+        auto line_split = geopm::string_split(line, ":");
+        EXPECT_EQ(*eb_it, line_split[0]);
+        ++eb_it;
+    }
+}
+
+TEST_F(StatsCollectorTest, time_report)
```
dannosliwcd, line 63: Tests that two sequential reports over a triangle wave contain the same header as the previous test, plus has metrics for count, first, last, min, max, mean, std, in that order. Tests, with clear in between report generation, that the correct count/first/last/min/max/mean/sample-stddev are returned for `[0, 1]`

libgeopmd/test/StatsCollectorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+            EXPECT_EQ(*eb_it, line_split[0]);
+            ++eb_it;
+        }
+        EXPECT_NE(std::string::npos, report.find("count: 2\n"));
+        EXPECT_NE(std::string::npos, report.find("first: 0\n"));
+        EXPECT_NE(std::string::npos, report.find("last: 1\n"));
+        EXPECT_NE(std::string::npos, report.find("min: 0\n"));
+        EXPECT_NE(std::string::npos, report.find("max: 1\n"));
+        EXPECT_NE(std::string::npos, report.find("mean: 0.5\n"));
+        EXPECT_NE(std::string::npos, report.find("std: 0.707107\n"));
```
dannosliwcd: This expects that we round to 6 decimals, but the implementation doesn't specify the precision. Does this depend on the ostringstream implementation, or is that standard?

geopmdpy/geopmdpy/session.py
```diff
@@ -323,38 +358,74 @@ def get_formats(self):
...
     parser.add_argument('--print-header', action='store_true',
                         help='Print a CSV header before printing any sampled values')
-    args = parser.parse_args()
+    parser.add_argument('-d', '--delimiter', dest='delimiter', default=',',
+                        help='Delimiter used to separate values in CSV output. Default: %(default)s')
+    parser.add_argument('--report-out', dest='report_out', default=None,
+                        help='Output summary statistics into a yaml file. Default: No summary statistics are generated.')
+    parser.add_argument('--trace-out', dest='trace_out', default='-',
+                        help='Output trace data into a CSV file. Default: trace data is printed to stdout.')
+    parser.add_argument('--mpi-report', dest='mpi_report', action='store_true',
```
avilcheslopez: Minor note: I'd like to suggest giving this option a name that describes what it does, which is to gather reports, and not name it based on how the functionality is implemented (unless there's a strong reason). Just by looking at the option name earlier, I was wondering what it was for.

docs/source/geopmsession.1.rst
cmcantalupo: Still TBD

libgeopmd/include/geopm_stats_collector.h
cmcantalupo: I'm hoping we can deal with that by adding a context for batch requests in the PlatformIO API.  This context would be local to the user of the batch interface.

libgeopmd/src/TimeZero.cpp
```diff
@@ -79,4 +100,40 @@ extern "C"
...
+    int geopm_time_real_to_iso_string(const struct geopm_time_s *time, int buf_size, char *buf)
+    {
+        int err = 0;
+        struct tm local;
+        size_t buf_size_u = buf_size;
+        localtime_r(&(time->t.tv_sec), &local);
+        char *buf_ptr = buf;
+        if (buf_size_u == 0) {
+            return EINVAL;
+        }
```
avilcheslopez: Minor suggestion: perhaps move this to the beginning of the function. It would be better to do the input validation prior to letting the function start its work.

geopmdpy/geopmdpy/stats.py
cmcantalupo: The issue with doing that in the C interface is the const-ness of the collector pointer in the interface.  We could cache in python however.

libgeopmd/test/StatsCollectorTest.cpp
cmcantalupo: 6 is the default:

https://en.cppreference.com/w/cpp/io/ios_base/precision

geopmdpy/geopmdpy/stats.py
dannosliwcd: We could also mark the cache as a `mutable` member

docs/source/geopmsession.1.rst
```diff
@@ -80,10 +82,30 @@ Options
...
+--trace-out .. _traceout TRACE_OUT option:
+
+    Output trace data into a CSV file. Note if ``--trace_out=-`` is specified,
+    the trace will output to stdout. To avoid gathering trace data,
+    set this parameter to ``/dev/null``
+
+--enable-mpi .. _enablempi option:
+
+    Gather reports over MPI and write to a single file. Append MPI rank to trace
+    output file if specified (trace output to stdout not permitted). Requires
```
cmcantalupo, line 115: Append hostname instead of mpi rank to trace files when mpi enabled to be compatible with existing convention.

geopmdpy/geopmdpy/session.py
```diff
@@ -10,12 +10,121 @@
...
+        request_raw = self._comm.bcast(request_raw, root=0)
+        if self._rank != 0:
+            result = ReadRequestQueue(StringIO(request_raw))
+        return result
+
+    def open_trace_stream(self):
+        if not self.do_trace():
+            return None
+        else:
+            return open(f'{self._trace_path}.{self._rank}', "w")
```
cmcantalupo: go back to socket.gethostname()

libgeopmd/src/StatsCollector.cpp
```diff
@@ -0,0 +1,194 @@
...
+        collector_cpp->reset();
+    }
+    catch (...) {
+        err = geopm::exception_handler(std::current_exception());
+        err = err < 0 ? err : GEOPM_ERROR_RUNTIME;
+    }
+    return err;
+}
+
+int geopm_stats_collector_free(geopm_stats_collector_s *collector)
```
avilcheslopez: Minor nit for consistency:

```suggestion
int geopm_stats_collector_free(struct geopm_stats_collector_s *collector)
```

In all other places the `struct` keyword is included.

libgeopmd/src/RuntimeStats.hpp
```diff
@@ -0,0 +1,110 @@
...
+            void check_index(int metric_idx, const std::string &func, int line) const;
+            struct stats_s {
+                stats_s &operator=(const stats_s &other);
+                uint64_t count;
+                double first;
+                double last;
+                double min;
+                double max;
+                double m_1;
+                double m_2;
```
cmcantalupo: add doc string about what m_1 and m_2 are.

libgeopmd/src/StatsCollector.cpp
```diff
@@ -0,0 +1,194 @@
...
+        else {
+            err = ENOBUFS;
+            std::ostringstream err_str;
+            size_t report_str_size = report_str.size() + 1;
+            if (*max_report_size != 0) {
+                err_str << "geopm_stats_collector_report_yaml(): max_report_size is too small, provided: "
+                        << *max_report_size << " required: " << report_str_size;
+            }
+            else {
+                err_str << "geopm_stats_collector_report_yaml(): max_report_size required: " << report_str_size;
```
avilcheslopez: Q: why treat this as an error if the interface spec says this function can be called with a size value of zero to query the actual size required? Returning an error to the caller might let them think the query was unsuccessful.

geopmdpy/geopmdpy/session.py
```diff
@@ -162,21 +280,32 @@ def run(self, run_time, period, pid, print_header,
...
-                                   requests to read.
+                                        requests to read.
 
             out_stream (typing.IO): Stream where output from will be
-                               printed.
+                                    printed.
 
         """
-        requests = ReadRequestQueue(request_stream)
+        if session_io is not None:
```
cmcantalupo, line 291: If session_io is not None assert that request_stream and report_stream are all None.

libgeopmd/src/StatsCollector.cpp
cmcantalupo: First in constructor

m_time_idx = pio.push_signal("TIME", 0, 0)

At first update(): get real time and subtract pio.sample(m_time_idx) and store this as m_time_begin.  

Also do not report end time, but rather, provide seconds since start time in the report yaml by calling pio.sample(m_time_idx) at report generation time.

libgeopmd/src/StatsCollector.cpp
cmcantalupo: create a collector using mocked platformio and use existing logic in C++ unit tests to exercise the C wrappers.

bgeltz:
``
printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | mpiexec -np 2 geopmsession -t 10 -p 0.005  --report-out=report.yaml --trace-out=/dev/null --enable-mpi
``

docs/source/geopmsession.1.rst
```diff
@@ -132,6 +160,64 @@ Multiple signals may be specified by separating them with a newline.
...
+.. code-block:: none
+
+    printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | geopmsession -t 2 -p 0.01 --print-header --report-out=-
+
+An example report is shown below:
+
+.. code-block:: yaml
+
+   host: "123.45.6.7"
+   time-begin: "2024-08-13T09:31:39.443889240-0700"
```
bgeltz: First sample happened at this time.  This is "time zero" + TIME[first]

~~sample-time-begin?
time-of-first-sample?
first-sample-time?
time-sample-begin?
time-begin-sample?~~

sample-time-first

docs/source/geopmsession.1.rst
```diff
@@ -132,6 +160,64 @@ Multiple signals may be specified by separating them with a newline.
...
+   duration: 9.87292
```
bgeltz: ~~sample-duration?
total-sample-time?
last-sample-time?
time-sample-total?
report-time?~~

sample-time-total
- This is the difference between time first and last.

docs/source/geopmsession.1.rst
```diff
@@ -132,6 +160,64 @@ Multiple signals may be specified by separating them with a newline.
...
+
+.. code-block:: none
+
+    printf 'TIME board 0\nCPU_POWER board 0\nCPU_FREQUENCY_STATUS board 0\n' | geopmsession -t 2 -p 0.01 --print-header --report-out=-
+
+An example report is shown below:
+
+.. code-block:: yaml
+
+   host: "123.45.6.7"
```
bgeltz:
```suggestion
   host: "host1"
```

geopmdpy/geopmdpy/stats.py
cmcantalupo: https://github.com/geopm/geopm/issues/3553

libgeopmd/src/StatsCollector.cpp
cmcantalupo: My thinking is that success is determined by the report being fully copied into the buffer.  If the intent is to populate the `max_report_size` value, the return code can be ignored.  On the other hand, if a user has submitted a buffer of zero length due to a bug, but the intent is to populate the buffer, the return code should be non-zero so the user is alerted to their bug.

docs/source/geopmsession.1.rst
cmcantalupo: https://github.com/geopm/geopm/issues/3554

geopmdpy/geopmdpy/session.py
cmcantalupo: resolved with --- separator.

geopmdpy/geopmdpy/stats.py
cmcantalupo: moved out of stats.py

docs/source/geopm_time.3.rst
```diff
@@ -74,21 +70,39 @@ header defines GEOPM interfaces for measuring time in seconds relative to a
 fixed arbitrary reference point. The ``geopm_time_s`` structure is used to hold
 time values.
 
+The default clock used is Linux ``CLOCK_MONTONIC_RAW``.  In general, these
+interfaces use this as the time reference.  This clock is the most precise for
+calculating time differences.
+
+There are some times when absolute time accuracy is preferred over time
+difference accuracy. For these cases, the GEOPM provides time functions that use
```
dannosliwcd:
```suggestion
difference accuracy. For these cases, GEOPM provides time functions that use
```

docs/source/geopmsession.1.rst
```diff
@@ -50,61 +45,75 @@ Get Help or Version
...
+index.
 
-This command can be used to read signals by opening a session with the
-geopm service.  The user specifies which signals to read with standard
-input. When no command line options are provided, a single read of the
-signals requested though standard input is made and the results are
-printed to the screen.
+Unless the ``-n`` option is specified, a descriptive header is written
+first and then the output from reading the requests.  Unless ``-p`` is
+specified, only one line of output after the header is printed..
```
dannosliwcd:
```suggestion
specified, only one line of output after the header is printed.
```

docs/source/geopmsession.1.rst
```diff
@@ -50,61 +45,75 @@ Get Help or Version
...
 --pid PID  .. _pid option:
 
-    Stop the session when the given process PID ends
+    Stop the session when the given process PID ends.
 
---print-header  .. _header option:
+--print-header  .. _printheader option:
 
-    Print a CSV header before printing any sampled values
+    Deprecated now this option is the default, see --no-header.
```
dannosliwcd:
```suggestion
    Deprecated. Now this option is the default, see --no-header.
```

docs/source/geopmsession.1.rst
```diff
@@ -50,61 +45,75 @@ Get Help or Version
...
 
---delimiter .. _delimiter DELIMITER .. option:
+-n, --no-header  .. _header option:
 
-    Delimiter used to separate values in CSV output
+    Do not print the CSV header before printing any sampled values.
```
dannosliwcd: Should either be mutexed or specify what happens when both are provided.

docs/source/geopmsession.1.rst
```diff
@@ -50,61 +45,75 @@ Get Help or Version
...
---trace-out .. _traceout TRACE_OUT option:
+-r, --report-out  .. _reportout REPORT_OUT option:
 
-    Output trace data into a CSV file. Note if ``--trace_out=-`` is specified,
-    the trace will output to stdout. To avoid gathering trace data,
-    set this parameter to ``/dev/null``
+    Output summary statistics into a yaml file. Note if ``--report-out=-`` is
+    specified, the report will output to stdout. When used with the
+    ``--enable-mpi`` option, reports from all hosts will be combined using the
+    ``---`` line separator, and the output is written (stdout or to file) solely
```
dannosliwcd:
```suggestion
    ``---`` document separator, and the output is written (stdout or to file) solely
```

docs/source/geopmsession.1.rst
```diff
@@ -135,9 +144,9 @@ The polling period must be shorter than the timeout specified.
 
 A 100ms polling period with a 300ms timeout is shown below:
 
-.. code-block:: none
+.. code-block:: bash
```
dannosliwcd:
```suggestion
.. code-block:: shell-session
```
and in other locations that include the leading prompt

docs/source/geopmsession.1.rst
```diff
@@ -123,7 +132,7 @@ core zero:
 
 .. code-block:: bash
```
dannosliwcd:
```suggestion
.. code-block:: shell-session
```

docs/source/geopmsession.1.rst
```diff
@@ -158,43 +168,61 @@ Summary statistics may be output to stdout by setting ``--report-out=-``.
...
+an MPI communicator.  This can be helpful when running sessions on more than one
+compute node in an MPI enabled environment.  The user must install the optional
+``mpi4py`` package to use the ``--enable-mpi`` command line option .  This can
+be done using the OS package manager or PyPi.  When running in this way the
+``geopmsession`` command line tool must be launched with an MPI launch wrapper
+like ``mpiexec`` or ``mpirun``.  The user should run this command specifying one
+``geopmsession`` process per compute node.  When using this option, trace output
+to stdout is disabled.  The aggregated report is created by the "rank 0" process
+of the geopmsession MPI communicator.
+
```
dannosliwcd, line 226: An example would be nice, even if just for one specific mpiexec flavor.

geopmdpy/geopmdpy/session.py
```diff
@@ -185,33 +282,37 @@ def run(self, run_time, period, pid, print_header,
...
+                                    printed.
 
         """
-        self._requests = ReadRequestQueue(request_stream)
-        do_stats = report_stream is not None
+        if session_io is not None:
+            if request_stream is not None:
+                raise RuntimeError('Using a request_stream is incompatible with using session_io')
+            if report_path is not None:
+                raise RuntimeError('Using a report_path is incompatible with using session_io')
```
dannosliwcd:
```suggestion
            if request_stream is not None:
                raise ValueError('Using a request_stream is incompatible with using session_io')
            if report_path is not None:
                raise ValueError('Using a report_path is incompatible with using session_io')
```

geopmdpy/geopmdpy/session.py
```diff
@@ -185,33 +282,37 @@ def run(self, run_time, period, pid, print_header,
...
+            requests = session_io.get_request_queue()
+            do_stats = session_io.do_report()
+        else:
+            requests = ReadRequestQueue(request_stream)
+            do_stats = report_path is not None
         self.check_read_args(run_time, period)
+        signal_config = [(rr[0], rr[1], rr[2]) for rr in requests]
```
dannosliwcd:
```suggestion
        signal_config = list(requests)
```
(if the goal is just to convert the iterable to a list)

libgeopmd/src/StatsCollector.cpp
```diff
@@ -61,6 +65,28 @@ namespace geopm
...
+                throw Exception("StatsCollector::update(): geopm_time_real_to_iso_string() call failed",
+                                err, __FILE__, __LINE__);
+            }
+            m_time_begin_str = time_begin_cstr;
+        }
+        else {
+            double time_delta = m_time_sample - time_last;
+            m_time_delta_m_1 += time_delta;
+            m_time_delta_m_2 += time_delta * time_delta;
+        }
```
dannosliwcd: If you move `++m_sample_count`, can you go back to N for mean and N-1 for the stddev denominator?
```suggestion
        double time_last = m_time_sample;
        m_time_sample = m_pio.sample(m_time_pio_idx);
        if (m_time_begin_str == "") {
            m_time_begin = m_time_sample;
            double time_curr = m_pio.read_signal("TIME", GEOPM_DOMAIN_BOARD, 0);
            geopm_time_s time_curr_real = geopm::time_curr_real();
            geopm_time_s time_begin_real;
            geopm_time_add(&time_curr_real, m_time_sample - time_curr, &time_begin_real);
            char time_begin_cstr[NAME_MAX];
            int err = geopm_time_real_to_iso_string(&time_begin_real, NAME_MAX, time_begin_cstr);
            if (err != 0) {
                throw Exception("StatsCollector::update(): geopm_time_real_to_iso_string() call failed",
                                err, __FILE__, __LINE__);
            }
            m_time_begin_str = time_begin_cstr;
        }
        else {
            ++m_sample_count;
            double time_delta = m_time_sample - time_last;
            m_time_delta_m_1 += time_delta;
            m_time_delta_m_2 += time_delta * time_delta;
        }
```

libgeopmd/src/StatsCollector.cpp
```diff
@@ -74,29 +100,47 @@ namespace geopm
...
+        double time_delta_std = 0;
+        if (m_sample_count > 1) {
+            time_delta_mean = m_time_delta_m_1 / (m_sample_count - 1);
+        }
+        if (m_sample_count > 2) {
+            time_delta_std = std::sqrt(
+                (m_time_delta_m_2 -
+                 m_time_delta_m_1 *
+                 m_time_delta_m_1 / (m_sample_count - 1)) /
+                (m_sample_count - 2));
```
dannosliwcd: Should have a comment explaining why we skip a sample

libgeopmd/src/StatsCollector.cpp
cmcantalupo: I'm not sure, to me seems more confusing.  Maybe i need to document that m_sample_count is the number of times sample() has been called? 

Because we are doing time differences we loose a degree of freedom compared to the total number of time samples that have been made (m_sample_count).  This extra degree of freedom lost is why the denominators is smaller.  If m_sample_count were used in the future to take the average of any other sampled value it would be off by one.  In this use case (of course), the code change is equivalent.

docs/source/geopmsession.1.rst
```diff
@@ -50,60 +45,73 @@ Get Help or Version
 Description
 -----------
 
-Command line interface for the GEOPM service read features.
+Command line interface for the GEOPM service batch read features. The input to
+the command line tool has one request per line. A request for reading is made
+of up of three strings separated by white space. The first string is the signal
```
avilcheslopez, line 50: Minor wording issue. My guess is this was the intended line:

```suggestion
up of three strings separated by white space. The first string is the signal
```

docs/source/geopmsession.1.rst
```diff
@@ -50,60 +45,73 @@ Get Help or Version
...
-    set this parameter to ``/dev/null``
+    the trace will output to stdout which is also the default behavior. To avoid
+    gathering trace data, set this parameter to ``/dev/null``.  When used with
+    the ``--enable-mpi`` option, trace file names will be appended with the
+    hostname combined with the ``-`` separator.  It is not possible to write the
+    trace output to stdout when specifying ``--enable-mpi``, this will result in
+    an error.
 
---enable-mpi .. _enablempi option:
+--enable-mpi  .. _enablempi option:
```
avilcheslopez, line 112: Name suggestion: --gather-reports (or something similar)

My concern is that the name `--enable-mpi` is not very intuitive, as you don't know right away for what MPI is being enabled (the actual functionality or feature you're enabling).

Or is it that it's important to make it clear how report gathering is being implemented? If so, perhaps a name like `--mpi-report-gather` could work.

Or is the idea there will be multiple MPI-related features and you want this option to enable all of them?

dannosliwcd (approved):

Issue #3546, opened 2024-07-31
Gaudi client


cmcantalupo:
- [x] #2449
- [x] #3560
- [x] #3559
- [x] intel-innersource/applications.analyzers.geopm.geopm/issues/98
- [x] Write up proposal for Gaudi use case
---
WW 39 goals:
- [ ] #3575
- [x] Compare the performance of node exporter v/s geopm prom exporter v/s geopm exporter
- [x] Compare the list of supported metrics between the node exporter and the geopm*exporters
- [x] Identify missing/misnamed signals in list of metrics exposed by geopm exporter (e.g., sum, mean)
- [ ] ~In the list of metrics summarized by the geopm exporter, add entries for metric sum (counter * rate)~
- [x] In the list of metrics summarized by the geopm exporter, change metric name _mean_arithmetic_ to _mean_
- [x] Create a new grafana dashboard that supports visualizations similar to those plotted w/ plotly on saruman
- [x] Add better unit test coverage of extractor.py
- [x] Fill in all missing doc strings

Issue #3545, opened 2024-07-31
As a user of geopmsession I would like to be able to read telemetry from Gaudi


cmcantalupo:
**Is your feature request related to a problem? Please describe.**
I run GEOPM on a Gaudi system but there are telemetry metrics missing.

**Describe the solution you'd like**
Support for gathering data from some of the upstream kernel driver implementations through hwmon
- [x] POC for WW35
- [x] Merged to dev for future sprint

**Describe alternatives you've considered**
Direct interface with sysfs or device driver interfaces

**Additional context**
Cross reference hl-smi for testing purposes:

https://docs.habana.ai/en/latest/Management_and_Monitoring/Embedded_System_Tools_Guide/System_Management_Interface_Tool.html

dannosliwcd:
In addition to hwmon, they have a /sys/class/accel interface: https://www.kernel.org/doc/html/latest/admin-guide/abi-testing.html#abi-file-testing-sysfs-driver-habanalabs

PR #3556, opened 2024-08-14
Include i915 hwmon entries in SysfsIOGroup

### LINKED ISSUE #2449: As a GPU Researcher I would like the ability to monitor and control GPU power
I'm always frustrated when I want to set power limits on an Intel GPU using GEOPM and can't.  

Support for the sustained power limit outlined in the oneAPI spec

Using direct sysfs calls.

dannosliwcd:
- Relates to #3545 (I think the same / a similar pattern can be used for `/dev/class/accel` drivers that are used in that case)
- Relates to #3546
- Fixes #2449. This doesn't use the solution requested in that issue, but it does make the request obsolete since this change includes `GPU_POWER_LIMIT_CONTROL`, `GPU_POWER_TIME_WINDOW_CONTROL`, and `GPU_POWER_LIMIT_DEFAULT` controls/signals.

### DIFF
--- a/libgeopmd/src/DrmSysfsDriver.cpp
+++ b/libgeopmd/src/DrmSysfsDriver.cpp
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-
 #include "DrmSysfsDriver.hpp"
 
 #include <fcntl.h>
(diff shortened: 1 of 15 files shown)


libgeopmd/src/DrmSysfsDriver.cpp
```diff
@@ -64,27 +119,52 @@ namespace geopm
...
     {
     }
 
-    int DrmSysfsDriver::domain_type(const std::string &) const
+    int DrmSysfsDriver::domain_type(const std::string &name) const
     {
-        return m_domain;
+        // So far, all of the supported i915 DRM signals are tile-scoped and
+        // most of the i915 hwmon signals are card-scoped.
+        if (signal_name_is_from_hwmon(name) && !string_ends_with(name, TILE_SIGNAL_NAME_SUFFIX)) {
```
cmcantalupo: The names of the functions in geopm/Helper.hpp are kind of generic, I like to add the geopm:: namespace when using, but NBD
```suggestion
        if (signal_name_is_from_hwmon(name) && !geopm::string_ends_with(name, TILE_SIGNAL_NAME_SUFFIX)) {
```

libgeopmd/src/DrmSysfsDriver.hpp
```diff
@@ -37,8 +37,8 @@ namespace geopm
             static std::unique_ptr<IOGroup> make_plugin(void);
         private:
             const std::map<std::string, SysfsDriver::properties_s> M_PROPERTIES;
+            const std::map<std::pair<geopm_domain_e, int>, std::string> M_DRM_HWMON_DIR_BY_GEOPM_DOMAIN;
```
cmcantalupo, line 42: A comment describing the key for this map might  be handy.  It is made more obvious by using the geopm_domain_e type rather than an int, as is commonly done in the code base with our enums (i.e. i like the style change).  But, I still was confused about the pair that was the key without looking into the implementation a bit.

libgeopmd/src/DrmSysfsDriver.cpp
```diff
@@ -22,6 +21,57 @@
 static const std::string DRM_DIRECTORY = "/sys/class/drm";
 static const std::string GPU_CARD_PREFIX = "card";
 static const std::string GPU_TILE_PREFIX = "gt";
+static const std::string HWMON_PREFIX = "hwmon";
+static const std::string HWMON_NAME_CARD = "i915\n";
+static const std::string HWMON_NAME_TILE_PREFIX = "i915_gt";
+static const std::string GPU_SIGNAL_NAME_SUFFIX = "::GPU";
+static const std::string TILE_SIGNAL_NAME_SUFFIX = "::GPU_CHIP";
+
+static std::map<std::pair<geopm_domain_e, int>, std::string> map_geopm_index_to_hwmon_path(const std::string &drm_directory, int num_tiles_per_gpu)
```
cmcantalupo, line 30: Seems like we should be able to derive the PlatformTopo information about GPUs from this class map.  Maybe we could wrap the map in an accessor class and reuse it in the PlatformTopo imp?

libgeopmd/src/DrmSysfsDriver.hpp
dannosliwcd: Adding comment: `Map of (GEOPM signal domain, GEOPM signal index) pairs to hwmon sysfs directory paths`

cmcantalupo (approved):
LGTM

libgeopmd/src/DrmSysfsDriver.cpp
dannosliwcd: Opened #3559

PR #3561, opened 2024-08-15
Make DrmSysfsDriver compatible with /sys/class/accel

### LINKED ISSUE #3560: Extract common bits between `/sys/class/drm` and `/sys/class/accel` out of `DrmSysfsDriver`

dannosliwcd:
- /sys/class/accel exposes a DRM interface, but uses a different name so
  that users don't mistake accelerators for render-focused graphics card
- This change reuses our existing /sys/class/drm code, but makes the
  target directory an IOGroup name configurable, and registers both
  accel and drm instances with PlatformIO.
- Resolves #3560


cmcantalupo (approved):
LGTM

PR #3562, opened 2024-08-16
Force geopmsession to terminate when time has elapsed

cmcantalupo:


integration/service/test/test_cron_use_case.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
dannosliwcd: Either discard or use in a checked test.

If adding to a test, please automate with a shorter report period

dannosliwcd (changes requested):

dannosliwcd (approved):

PR #3564, opened 2024-08-16
Create a test that verifies geopmsession reports

cmcantalupo:
- TODO: Improve python script that verifies the generated report

### DIFF
--- a/integration/service/test/geopm_report_cron_plot.py
+++ b/integration/service/test/geopm_report_cron_plot.py
@@ -2,25 +2,24 @@
 #  SPDX-License-Identifier: BSD-3-Clause
 #
 
-import yaml
+from yaml import load as yaml_load
+try:
+    from yaml import CSafeLoader as SafeLoader
+except ImportError:
+    from yaml import SafeLoader
 import sys
 import os
-from glob import glob
 from pandas import DataFrame, to_datetime, to_timedelta
 import plotly.graph_objs as go
 from dash import Dash, dcc, html
 from argparse import ArgumentParser
 
 class CronReport:
-    def __init__(self, reports='geopm-report-*.yml'):
-        if type(reports) is str:
-            report_names = glob(reports)
-        else:
-            report_names = reports
+    def __init__(self, report_paths):
         report_list = []
-        for rn in report_names:
+        for rn in report_paths:
             with open(rn) as fid:
-                report = yaml.load(fid, Loader=yaml.CSafeLoader)
+                report = yaml_load(fid, Loader=SafeLoader)
             if report is None:
                 sys.stderr.write(f'Warning: {rn} is empty, could not be parsed\n')
                 continue
@@ -36,6 +35,8 @@ class CronReport:
                 for m_key, m_value in metrics.items():
                     df_row[m_key] = m_value
                 report_list.append(df_row.copy())
+        if len(report_list) == 0:
+            raise RuntimeError(f'No files could be parsed')
         self._df = DataFrame(report_list)
         self._df.sort_values('sample-time-first', inplace=True)
         if len(self._df) == 0:
@@ -58,19 +59,38 @@ class CronReport:
 
 
     def calculate_energy(self, domain, begin_date, end_date):
+        """Return an estimate of energy for all reports that begin within the time range
+        specified.  Only includes time prior to the beginning of the last report
+        that matches.
+
+        """
+        # This calculation uses only the start time of each report and the
+        # average power measured between report start times.  This avoids any
+        # double counting in case reports overlap, and avoids undercounting when
+        # there are gaps between the reports.
+        #
+        # TODO A side effect is that power measurements from the last report are
+        # ignored, and not used in the calculation, this could be corrected.
+        # Also consider that partial reports are also not used in this method.
         domain = domain.upper()
         metric_name = f'{domain}_POWER'
         time = self.get_stat(begin_date, end_date, metric_name, 'sample-time-first')
         time_total = self.get_stat(begin_date, end_date, metric_name, 'sample-time-total')
         power = self.get_stat(begin_date, end_date, metric_name, 'mean-arithmetic')
         kwh_factor = 1.0 / 3.6e6
-        # TODO include energy from last report as well
         return kwh_factor * sum((power.shift(1) * (time - time.shift(1)).dt.total_seconds()).dropna())
 
     def dataframe(self):
         return self._df
 
     def date_range(self):
+        """Return oldest and newest report 'sample-time-first'.
+
+        Does not infer the ISO date that the last report extends to based on
+        'sample-time-total'.
+        """
+        # TODO: Consider returning last report's last sample time as the second
+        # value instead of the last report's first sample time
         return (min(self._df['sample-time-first']),
                 max(self._df['sample-time-first']))
 
@@ -161,7 +181,7 @@ class CronReport:
         fig.update_layout(title=f'Energy - kWh ({begin_string} - {end_string})', hovermode='x')
         return fig
 
-    def run_webpage(self, begin_date, end_date, port):
+    def run_webpage(self, begin_date, end_date, port, host_mask):
         app = Dash('geopmsession')
         style={
             'textAlign': 'center',
@@ -179,7 +199,7 @@ class CronReport:
             dcc.Graph(id='energy-pie',
                       figure=self.plot_energy(begin_date, end_date)),
         ])
-        app.run(host='0.0.0.0', port=port)
+        app.run(host=host_mask, port=port)
 
     def static_webpage(self, begin_date, end_date, output_html):
         if (not output_html.endswith('.html')):
@@ -198,12 +218,14 @@ class CronReport:
 
 def get_parser():
     parser = ArgumentParser(description=main.__doc__)
-    parser.add_argument('report', nargs='*')
+    parser.add_argument('report', nargs='+')
     static_group = parser.add_mutually_exclusive_group()
     static_group.add_argument('-s', '--static-html', dest='static_html', default=None,
                               help='Do not run web server, output static html into directory provided')
     static_group.add_argument('-p', '--port', dest='port', type=int, default=10850,
-                              help='Port opened by server to provide visualization')
+                              help='Port opened by server to provide visualization. Default: %(default)s')
+    parser.add_argument('-m', '--host-mask', dest='host_mask', default='0.0.0.0',
+                        help='Restrict connections based on IP address. Default: no restrictions')
     return parser
 
 def main():
@@ -217,17 +239,10 @@ def main():
     err = 0
     try:
         args = get_parser().parse_args()
-        if len(args.report) == 0:
-            args.print_help()
-            return -1
-        if len(args.report) == 1:
-            files = args.report[0]
-        else:
-            files = args.report
-        cron_report = CronReport(files)
+        cron_report = CronReport(args.report)
         begin_date, end_date = cron_report.date_range()
         if args.static_html is None:
-            cron_report.run_webpage(begin_date, end_date, args.port)
+            cron_report.run_webpage(begin_date, end_date, args.port, args.host_mask)
         else:
             cron_report.static_webpage(begin_date, end_date, args.static_html)
     except RuntimeError as ee:
--- a/integration/service/test/geopm_report_cron_plot.py
+++ b/integration/service/test/geopm_report_cron_plot.py
@@ -1,0 +1,242 @@
+#  Copyright (c) 2015 - 2024 Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+import yaml
+import sys
+import os
+from glob import glob
+from pandas import DataFrame, to_datetime, to_timedelta
+import plotly.graph_objs as go
+from dash import Dash, dcc, html
+from argparse import ArgumentParser
+
+class CronReport:
+    def __init__(self, reports='geopm-report-*.yml'):
+        if type(reports) is str:
+            report_names = glob(reports)
+        else:
+            report_names = reports
+        report_list = []
+        for rn in report_names:
+            with open(rn) as fid:
+                report = yaml.load(fid, Loader=yaml.CSafeLoader)
+            if report is None:
+                sys.stderr.write(f'Warning: {rn} is empty, could not be parsed\n')
+                continue
+            df_row_keys = set(report) - {'metrics'}
+            df_row = dict()
+            for df_key in df_row_keys:
+                if df_key == 'sample-time-first':
+                    df_row[df_key] = to_datetime(report[df_key])
+                else:
+                    df_row[df_key] = report[df_key]
+            for metric_name, metrics in report['metrics'].items():
+                df_row['metric-name'] = metric_name
+                for m_key, m_value in metrics.items():
+                    df_row[m_key] = m_value
+                report_list.append(df_row.copy())
+        self._df = DataFrame(report_list)
+        self._df.sort_values('sample-time-first', inplace=True)
+        if len(self._df) == 0:
+            raise RuntimeError('Could not load reports')
+        self._colors = dict(
+            mean_marker='rgb(247, 240, 25)',
+            min_marker='rgb(82, 115, 222)',
+            max_marker='rgb(217, 85, 85)',
+            std_marker='rgb(68, 68, 68)',
+            std_fill='rgba(68, 68, 68, 0.5)',
+            range_fill='rgba(68, 68, 68, 0.3)',
+            pie_line='rgb(0, 0, 0)',
+            pie_gpu='red',
+            pie_cpu='orange',
+            pie_dram='yellow',
+            background='rgb(41, 128, 185)',
+            page_text='rgb(252, 252, 252)',
+        )
+        self._font_family = 'Lato, proxima-nova, Helvetica Neue, Arial, sans-serif'
+
+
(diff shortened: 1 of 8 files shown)


integration/service/test/test_cron_use_case_plot.py
```diff
@@ -0,0 +1,67 @@
...
+report_names = glob('geopm-report-*.yaml')
+report_list = []
+for rn in report_names:
+    with open(rn) as fid:
+        report = yaml.load(fid, Loader=yaml.SafeLoader)
+    if report is None:
+        sys.stderr.write(f'Warning: {rn} could not be parsed\n')
+        continue
+    df_row_keys = ['host', 'sample-time-first', 'sample-time-total',
+                   'sample-count', 'sample-period-mean', 'sample-period-std']
```
dannosliwcd: Alternatively, if you think the non-`metrics` fields may ever change in the future, it could be something like `df_row_keys = set(report) - {'metrics'}`

integration/service/test/test_cron_use_case_plot.py
```diff
@@ -0,0 +1,67 @@
+import yaml
+import sys
+from glob import glob
+from pandas import DataFrame, to_datetime
+
+report_names = glob('geopm-report-*.yaml')
+report_list = []
+for rn in report_names:
+    with open(rn) as fid:
+        report = yaml.load(fid, Loader=yaml.SafeLoader)
```
dannosliwcd: It would be nice to use CSafeLoader when available since it parses yaml faster than SafeLoader

integration/service/test/test_cron_use_case.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# */5 * * * * /path/to/test_cron_use_case.sh 300 1 /output/path/geopm-reports
+#
+# In the above example, the cron job runs every 5 minutes (*/5) and the sampler
+# runs for 300 seconds, i.e. 5 minutes.  The reports generated in this way may
+# be visualised with the test_cron_use_case_plot.py.  Note that /path/to and
+# /output/path are intended as place holders for paths that reflect the user's
+# environment.
+#
+# python3 /path/to/test_cron_use_case_plot.py /output/path/geopm-reports/\*/\*.yml
```
cmcantalupo: probably can just leave the glob un-escaped in the example.

integration/service/test/test_cron_use_case_plot.py
```diff
@@ -0,0 +1,182 @@
+import yaml
```
cmcantalupo: needs header

integration/service/test/test_cron_use_case_plot.py
```diff
@@ -0,0 +1,182 @@
...
+        record_mean_time = sample_time_first + to_timedelta(sample_time_total / 2.0, unit='s')
+        record_min_power = self.get_stat(begin_date, end_date, metric_name, 'min')
+        record_max_power = self.get_stat(begin_date, end_date, metric_name, 'max')
+        record_mean_power = self.get_stat(begin_date, end_date, metric_name, 'mean-arithmetic')
+        record_std_power = self.get_stat(begin_date, end_date, metric_name, 'std')
+        record_plus_std = [min((a, b)) for (a, b) in zip(record_mean_power + record_std_power, record_max_power)]
+        record_minus_std = [max((a, b)) for (a, b) in zip(record_mean_power - record_std_power, record_min_power)]
+        fig = go.Figure([
+            go.Scatter(
+                name=f'{domain} Power Std',
```
cmcantalupo: I think we want to swap the order of Std and Min/Max so the min/max is shaded first

integration/service/test/test_cron_use_case.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# The output can be visualized by opening a web page to port 10850 on the server
+# running the plotting script.
+
+set -e
+
+if [[ $# -ne 4 ]]; then
+    echo "Usage: $0 REPORT_PERIOD SAMPLE_PERIOD OUTPUT_DIR"
+    echo "       REPORT_PERIOD: Period of time covered by each report in seconds"
+    echo "       SAMPLE_PERIOD: Period between sample in units of seconds (may be fractional)"
```
cmcantalupo:
```suggestion
    echo "       SAMPLE_PERIOD: Period between samples in units of seconds (may be fractional)"
```

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,189 @@
...
+                      figure=self.plot_power('cpu', begin_date, end_date)),
+            dcc.Graph(id='dram-power',
+                      figure=self.plot_power('dram', begin_date, end_date)),
+            dcc.Graph(id='energy-pie',
+                      figure=self.plot_energy(begin_date, end_date)),
+        ])
+        app.run(host='0.0.0.0', port='10850')
+
+if __name__ == '__main__':
+    if len(sys.argv) == 2:
```
cmcantalupo: Add help text

integration/service/test/test_geopm_report_cron.py
```diff
@@ -0,0 +1,43 @@
...
+
+from geopm_report_cron_plot import CronReport
+from geopmdpy.loop import TimedLoop
+import unittest
+import subprocess # nosec
+import os
+import shutil
+import pathlib
+
+class TestIntegraitonCronReport(unittest.TestCase):
```
cmcantalupo:
```suggestion
class TestIntegrationCronReport(unittest.TestCase):
```

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
+#  Copyright (c) 2015 - 2024 Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+import yaml
+import sys
+import os
+from glob import glob
+from pandas import DataFrame, to_datetime, to_timedelta
+import plotly.graph_objs as go
```
dannosliwcd, line 13: plotly and dash should be added to integration/requirements.txt

integration/service/test/test_geopm_report_cron.py
dannosliwcd: Should this test be optional? If so, skip when plotly cannot be imported

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+    Parses a collection of geopmsession YAML reports and creates an HTML
+    visualiation.  By default it will open a dash server on port 10850.  Can
+    also create static web pages with the ``--static-html`` option.
+
+    """
+    err = 0
+    try:
+        args = get_parser().parse_args()
+        if len(args.report) == 0:
+            args.print_help()
```
dannosliwcd:
```shell-session
$ python3 integration/service/test/geopm_report_cron_plot.py --static-html ./geopm_reports.html
  File "...../integration/service/test/geopm_report_cron_plot.py", line 242, in <module>
    exit(main())
  File "...../integration/service/test/geopm_report_cron_plot.py", line 221, in main
    args.print_help()
AttributeError: 'Namespace' object has no attribute 'print_help'
```

dannosliwcd: Maybe use get_parser().error('reason') https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error

It informs the user our reason for exiting early, then emits usage and exits non-zero rc

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+                    df_row[df_key] = to_datetime(report[df_key])
+                else:
+                    df_row[df_key] = report[df_key]
+            for metric_name, metrics in report['metrics'].items():
+                df_row['metric-name'] = metric_name
+                for m_key, m_value in metrics.items():
+                    df_row[m_key] = m_value
+                report_list.append(df_row.copy())
+        self._df = DataFrame(report_list)
+        self._df.sort_values('sample-time-first', inplace=True)
```
dannosliwcd, line 41:
```
$ python3 integration/service/test/geopm_report_cron_plot.py --static-html ./geopm_reports.html /missing/file
  File "...../integration/service/test/geopm_report_cron_plot.py", line 242, in <module>
    exit(main())
  File "...../integration/service/test/geopm_report_cron_plot.py", line 227, in main
    cron_report = CronReport(files)
  File "...../integration/service/test/geopm_report_cron_plot.py", line 40, in __init__
    self._df.sort_values('sample-time-first', inplace=True)
  File "...../site-packages/pandas/core/frame.py", line 6947, in sort_values
    k = self._get_label_or_level_values(by[0], axis=axis)
  File "...../site-packages/pandas/core/generic.py", line 1844, in _get_label_or_level_values
    raise KeyError(key)
KeyError: 'sample-time-first'
```

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+class CronReport:
+    def __init__(self, reports='geopm-report-*.yml'):
+        if type(reports) is str:
+            report_names = glob(reports)
+        else:
+            report_names = reports
+        report_list = []
+        for rn in report_names:
+            with open(rn) as fid:
+                report = yaml.load(fid, Loader=yaml.CSafeLoader)
```
dannosliwcd: This will fail if libyaml is not installed. The following functions (but not as fast) if libyaml is absent, and uses libyaml when available:

```
from yaml import load
try:
    from yaml import CSafeLoader as SafeLoader
except ImportError:
    from yaml import SafeLoader

...

    with open(report_path) as f:
        report = load(f, Loader=SafeLoader)
```

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+            if report is None:
+                sys.stderr.write(f'Warning: {rn} is empty, could not be parsed\n')
+                continue
+            df_row_keys = set(report) - {'metrics'}
+            df_row = dict()
+            for df_key in df_row_keys:
+                if df_key == 'sample-time-first':
+                    df_row[df_key] = to_datetime(report[df_key])
```
dannosliwcd, line 30: This would likely be faster as a single call after all rows are constructed, but I suspect any speedup won't be noticeable compared to time spent in pyyaml's loader

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+    def calculate_energy(self, domain, begin_date, end_date):
+        domain = domain.upper()
+        metric_name = f'{domain}_POWER'
+        time = self.get_stat(begin_date, end_date, metric_name, 'sample-time-first')
+        time_total = self.get_stat(begin_date, end_date, metric_name, 'sample-time-total')
+        power = self.get_stat(begin_date, end_date, metric_name, 'mean-arithmetic')
+        kwh_factor = 1.0 / 3.6e6
+        # TODO include energy from last report as well
```
dannosliwcd: Is this still to-do?

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+        return kwh_factor * sum((power.shift(1) * (time - time.shift(1)).dt.total_seconds()).dropna())
```
dannosliwcd, line 81: Maybe add a comment explaining why it's insufficient to simply sum each report's `T*P`.

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def dataframe(self):
+        return self._df
+
+    def date_range(self):
+        return (min(self._df['sample-time-first']),
+                max(self._df['sample-time-first']))
```
dannosliwcd, line 95: Would the range go to the final report's start+elapsed time?

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+        sample_time_first = self.get_stat(begin_date, end_date, metric_name, 'sample-time-first')
+        sample_time_total = self.get_stat(begin_date, end_date, metric_name, 'sample-time-total')
+        record_mean_time = sample_time_first + to_timedelta(sample_time_total / 2.0, unit='s')
+        record_min_power = self.get_stat(begin_date, end_date, metric_name, 'min')
+        record_max_power = self.get_stat(begin_date, end_date, metric_name, 'max')
+        record_mean_power = self.get_stat(begin_date, end_date, metric_name, 'mean-arithmetic')
+        record_std_power = self.get_stat(begin_date, end_date, metric_name, 'std')
+        record_plus_std = [min((a, b)) for (a, b) in zip(record_mean_power + record_std_power, record_max_power)]
+        record_minus_std = [max((a, b)) for (a, b) in zip(record_mean_power - record_std_power, record_min_power)]
+        fig = go.Figure([
```
dannosliwcd, line 113: I don't see any examples manually constructing a subplot list in https://plotly.com/python/subplots/

Was the manual construction needed to enable something? It seems like the helper figure-builders could add some nice features like synchronizing the time axis across subplots.

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+            dcc.Graph(id='gpu-power',
+                      figure=self.plot_power('gpu', begin_date, end_date)),
+            dcc.Graph(id='cpu-power',
+                      figure=self.plot_power('cpu', begin_date, end_date)),
+            dcc.Graph(id='dram-power',
+                      figure=self.plot_power('dram', begin_date, end_date)),
+            dcc.Graph(id='energy-pie',
+                      figure=self.plot_energy(begin_date, end_date)),
+        ])
+        app.run(host='0.0.0.0', port=port)
```
dannosliwcd: We should make the bound address user-configurable

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+            fid.write('    </body>\n</html>\n')
+
+def get_parser():
+    parser = ArgumentParser(description=main.__doc__)
+    parser.add_argument('report', nargs='*')
+    static_group = parser.add_mutually_exclusive_group()
+    static_group.add_argument('-s', '--static-html', dest='static_html', default=None,
+                              help='Do not run web server, output static html into directory provided')
+    static_group.add_argument('-p', '--port', dest='port', type=int, default=10850,
+                              help='Port opened by server to provide visualization')
```
dannosliwcd:
```suggestion
                              help='Port opened by server to provide visualization. Default: %(default)s')
```

integration/service/test/geopm_report_cron_plot.py
dannosliwcd: Or if 0 reports is always an error, use `nargs='+'`

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,242 @@
...
+    err = 0
+    try:
+        args = get_parser().parse_args()
+        if len(args.report) == 0:
+            args.print_help()
+            return -1
+        if len(args.report) == 1:
+            files = args.report[0]
+        else:
+            files = args.report
```
dannosliwcd: A singleton list is conditionally flattened here, but later code performs type checks to see whether we need to iterate a list of strings or operate on one string. Why not just keep this as a list so the other code does not need to perform those checks?

The current implementation accepts either a single glob spec or 2+ paths. Alternatively it could accept a list of globs.

integration/service/test/geopm_report_cron_plot.py
cmcantalupo: Yes we are just diffing between start times

integration/service/test/geopm_report_cron_plot.py
cmcantalupo: There are a number of benefits to using subplots.  Perhaps we can look into adding these features in a future PR.  It will also clean up the code for generating static html.

integration/service/test/geopm_report_cron_plot.py
```diff
@@ -0,0 +1,257 @@
...
+        time_total = self.get_stat(begin_date, end_date, metric_name, 'sample-time-total')
+        power = self.get_stat(begin_date, end_date, metric_name, 'mean-arithmetic')
+        kwh_factor = 1.0 / 3.6e6
+        return kwh_factor * sum((power.shift(1) * (time - time.shift(1)).dt.total_seconds()).dropna())
+
+    def dataframe(self):
+        return self._df
+
+    def date_range(self):
+        """Return oldest and newest report 'sample-timer-first'.
```
cmcantalupo:
```suggestion
        """Return oldest and newest report 'sample-time-first'.
```

integration/service/test/test_geopm_report_cron.py
```diff
@@ -0,0 +1,69 @@
...
+    _do_skip = True
+_script_dir_name=os.path.dirname(os.path.realpath(__file__))
+
+def skip_unless_import():
+    if _do_skip:
+        return unittest.skip('geopm_report_cron_plot could not be imported, "pip install plotly dash", skipping test.')
+    return lambda func: func
+
+@skip_unless_import()
+class TestIntegraitonCronReport(unittest.TestCase):
```
cmcantalupo:
```suggestion
class TestIntegrationCronReport(unittest.TestCase):
```

dannosliwcd (dismissed):

dannosliwcd (approved):

PR #3569, opened 2024-08-27
Fix static issues

cmcantalupo:
- Fix issues identified by static analysis

### DIFF
--- a/libgeopmd/src/geopm_sched.c
+++ b/libgeopmd/src/geopm_sched.c
@@ -170,6 +170,9 @@ int geopm_sched_proc_cpuset(int num_cpu, cpu_set_t *proc_cpuset)
     int err = pthread_once(&g_proc_cpuset_once, geopm_proc_cpuset_once);
     (void)geopm_sched_proc_cpuset_pid(getpid(), num_cpu, g_proc_cpuset);
     int sched_num_cpu = geopm_sched_num_cpu();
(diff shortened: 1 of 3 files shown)


libgeopmd/src/geopm_sched.c
```diff
@@ -195,16 +202,27 @@ int geopm_sched_woomp(int num_cpu, cpu_set_t *woomp)
...
+    if (sched_num_cpu <= 0) {
+        return GEOPM_ERROR_LOGIC;
+    }
     size_t req_alloc_size = CPU_ALLOC_SIZE(num_cpu);
-
-    if (!err && !g_proc_cpuset) {
-        err = ENOMEM;
+    if (g_proc_cpuset == NULL) {
+        return ENOMEM;
     }
```
dannosliwcd: 1. `g_proc_cpuset` is used on 212 before we check whether it is null on 221
2. `geopm_sched_proc_cpuset_pid(..., g_proc_cpuset)` on 212 repeats work done by `geopm_proc_cpuset_once` on 208. Is that intentional? I expect either the first-time-initialization or the every-time-update could be dropped.

dannosliwcd: Noticing now, I guess the same comment+question apply to `geopm_sched_proc_cpuset`

dannosliwcd (dismissed):

bgeltz (approved):

PR #3570, opened 2024-08-28
Add signal handling for SIGTERM in geopmsession

### LINKED ISSUE #3551: As a user of geopmsession I would like support for cleaner signal handling
I'm always frustrated when I send SIGINT to geopmsession with Ctrl-C that there is no output in most cases.

I'd like to see an exception handler for SIGINT in the geopmsession main.  The handler should flush the trace buffer if it is non-null and generate a report based on the statistics gathered so far if the report buffer it non-null.

We could also consider flushing the trace buffer with every line written or with some other more regular cadence.  

**Additional context**

cmcantalupo:
Resolves #3551.

### DIFF
--- a/geopmdpy/geopmdpy/session.py
+++ b/geopmdpy/geopmdpy/session.py
@@ -13,6 +13,8 @@ import math
 from socket import gethostname
 from io import StringIO
 from argparse import ArgumentParser
+from signal import signal
+from signal import SIGTERM
 from . import topo
 from . import pio
 from . import loop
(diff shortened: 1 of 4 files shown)


geopmdpy/geopmdpy/session.py
```diff
@@ -300,16 +308,33 @@ def run(self, run_time, period, pid, print_header,
...
+        self.report_path = report_path
+        self.session_io = session_io
+
+    def stop(self):
+        if self.out_stream is not None:
+            self.out_stream.flush()
+        if self.do_stats:
+            report = self.stats_collector.report_yaml()
+            if self.session_io is not None:
+                self.session_io.write_report(report)
```
bgeltz, line 332: I don't see ``close_trace_stream()`` being called anywhere.  Should it be here?  This was the only spot I found where there's no ``with open()`` context manager, so might be a file left open.

cmcantalupo: close_trace_stream() was dead code, and actually not correctly implemented.  I have put the method in the run path and fixed it.

bgeltz (approved):

bgeltz:
Was there a bug or issue for this?

PR #3577, opened 2024-09-05
Add a helper function for enabling the fixed counters

### LINKED ISSUE #3549: Frequency reporting from the runtime is broken after stopping run
**Describe the bug**
I tried to use the GEOPM Runtime to profile my MPI application and I expected frequency reporting to work instead no frequency data was present in the reports.

**GEOPM version**
v3.1

**Expected behavior**
All reports always contain frequency data:
```
$ cat sgemm_1_cpu_sweep_3500000000_report.yaml
GEOPM Version: 3.1.1.dev47+g6e066589d
Start Time: Tue Aug 06 16:04:12 2024
Profile: "sgemm_1_cpu_sweep_3500000000"
Agent: monitor
Policy: {}

Hosts:
  x1001c3s2b0n0:
    Application Totals:
...
      frequency (%): 171.956
      frequency (Hz): 3.43911e+09
...
```

**Actual behavior**
```
$ cat sgemm_1_cpu_sweep_3500000000_report.yaml
GEOPM Version: 3.1.1.dev47+g6e066589d
Start Time: Tue Aug 06 16:27:44 2024
Profile: "sgemm_1_cpu_sweep_3500000000"
Agent: monitor
Policy: {}

Hosts:
  x1001c1s6b0n0:
    Application Totals:
...
      frequency (%): 0
      frequency (Hz): 0
...
```

**Error messages produced**
None

**Additional context**
This was encountered during the third run of a CPU frequency sweep.  On the first run, the frequency reporting worked as expected.  The second run was started, then CTRL-C'd to stop it during the first frequency trial.  Then the third run exhibited the problem with 0's reported for the CPU frequency fields.

Once in this state, re-running will always produce 0's in the reports for the CPU frequency fields.  I was able to restore the proper frequency reporting by issuing ```sudo systemctl restart geopm```.

The current suspicions around why this is occurring center around our use of the fixed performance counters.  We program them here:
https://github.com/geopm/geopm/blob/6e066589d961f72055ca32bcd2bed3e062885ed7/libgeopmd/src/MSRIOGroup.cpp#L1056
That programming occurs the first time any signal is pushed.  The control writes that happen in ``MSRIOGroup::enable_fixed_counters()`` occur outside the regular save/restore mechanisms because this code executes under the context of the root user, thus bypassing the save/restore file that is created in ``geopmdpy``.
The only way for the fixed counters to be restored to their initial state is for the Controller to call ```m_platform_io.restore_control()``` as it does at the end of ```Controller::run()```.  There is another call to ```m_platform_io.restore_control()``` in the Controller destructor, but it seems there may be a race that prevents that from resolving this issue as it is gated by ```m_do_restore```.  ```m_do_restore``` is only set true part way through ```Controller::run()```.
It seems if the run was killed before that line was reached, the restore never occurs.

A few potential solutions:
1. Move the ```save_control()``` call to either be directly after IOGroup initialization, or make it a part of IOGroup initialization.  Also, ``enable_fixed_counters()`` would also have to be called during MSRIOGroup construction, and not later like it is now.
2. Ensure ```restore_control()``` is always called as long as ```save_control()``` has completed.
3. Evaluate the need for using the fixed counters at all.  See if we can avoid their use entirely.
   - Current alias mappings to fixed counters:
     - ``CPU_INSTRUCTIONS_RETIRED`` -> ``MSR::FIXED_CTR0:INST_RETIRED_ANY``
     - ``CPU_CYCLES_THREAD`` -> ``MSR::FIXED_CTR1:CPU_CLK_UNHALTED_THREAD``
     - ``CPU_CYCLES_REFERENCE`` -> ``MSR::FIXED_CTR2:CPU_CLK_UNHALTED_REF_TSC``
   - ``THREAD`` and ``REFERENCE`` could be mapped to ``APERF`` and ``MPERF``.  Experiments are needed to ensure this would capture the same data as before re: when the counter is actually counting - through sleep states or not.
   - Unclear if there is another way to get ``INSTRUCTIONS_RETIRED``.

cmcantalupo:
- Fixes #3549 
- Use save/resore feature to guard against the problem.

### DIFF
--- a/libgeopm/src/Controller.cpp
+++ b/libgeopm/src/Controller.cpp
@@ -289,6 +289,15 @@ namespace geopm
         if (m_do_init_control) {
             m_init_control->parse_input(environment().init_control());
         }
+        try {
+            geopm::enable_fixed_counters(m_platform_io);
+        }
+        catch (const geopm::Exception &ex) {
+            // Fixed counter enabling may fail if not on an MSR based platform,
+            // or if user does not have permission to write to the enalbing
+            // MSRs.  When not enabled on MSR based platforms, reads from fixed
+            // counters will return zero.
+        }
     }
 
     Controller::~Controller()
--- a/docs/json_data/msr_data_arch.json
+++ b/docs/json_data/msr_data_arch.json
@@ -1066,7 +1066,7 @@
                     "behavior":  "monotone",
                     "writeable": false,
                     "aggregation": "sum",
-                    "description": "The count of the number of instructions executed."
+                    "description": "The count of the number of instructions executed. Requires geopmwrite -e."
                 }
             }
         },
@@ -1083,7 +1083,7 @@
                     "behavior":  "monotone",
                     "writeable": false,
                     "aggregation": "sum",
-                    "description": "The count of the number of cycles while the logical processor is not in a halt state.  The count rate may change based on core frequency."
+                    "description": "The count of the number of cycles while the logical processor is not in a halt state.  The count rate may change based on core frequency.  Requires geopmwrite -e."
                 }
             }
         },
@@ -1100,7 +1100,7 @@
                     "behavior":  "monotone",
                     "writeable": false,
                     "aggregation": "sum",
-                    "description": "The count of the number of cycles while the logical processor is not in a halt state and not in a stop-clock state.  The count rate is fixed at the TIMESTAMP_COUNT rate."
+                    "description": "The count of the number of cycles while the logical processor is not in a halt state and not in a stop-clock state.  The count rate is fixed at the TIMESTAMP_COUNT rate.  Requires geopmwrite -e."
                 }
             }
         },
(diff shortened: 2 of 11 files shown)


libgeopm/src/Controller.cpp
```diff
@@ -289,6 +289,12 @@ namespace geopm
         if (m_do_init_control) {
             m_init_control->parse_input(environment().init_control());
         }
+        try {
+            geopm::enable_fixed_counters(m_platform_io);
+        }
+        catch (const Exception &ex) {
+
```
bgeltz: Might be work spelling out in a comment either here in the exception block or in the implementation for enable_fixed_counters that this is only necessary for platforms that support the MSRIOGroup.

bgeltz (commented):
This all LGTM.  I'd like to see some data after a frequency sweep with these changes to ensure we don't start having problems like this: https://github.com/geopm/geopm/issues/3549#issuecomment-2274502280

bgeltz:
Consider how this impacts ABI as we get closer to the next release.

cmcantalupo:
```
python3 test/test_frequency_map.py
Warning: test_frequency_map: using existing file "machine.json", delete if invalid

srun -N 1 -n 1 lscpu

Warning: <geopm> geopmpy.io: Trace HDF5 file not detected or older than traces.  Data will be saved to trace_f17ac1eb7646d4b1153bc2d6be43.h5.
.
Ran 1 test in 427.853s

OK
```
Frequency map agent test is working, and report looks clean.

cmcantalupo:
```
$ PYTHONPATH=$HOME/Git/geopm LD_LIBRARY_PATH=$HOME/build/geopm/lib:$LD_LIBRARY_PATH python3 test/test_frequency_map.py
Warning: test_frequency_map: using existing file "machine.json", delete if invalid

srun -N 1 -n 1 lscpu

Warning: <geopm> geopmpy.io: Trace HDF5 file not detected or older than traces.  Data will be saved to trace_f17ac1eb7646d4b1153bc2d6be43.h5.
.
Ran 1 test in 424.364s

OK
```

bgeltz (approved):

PR #3579, opened 2024-09-06
Fix one shot issue with geopmsession

### LINKED ISSUE #3573: Geopmsession's one-shot mode prints two samples
**Describe the bug**
I tried to run geopmsession in one-shot mode (no time limit or period specified). I expected geopm to print a single sample. Instead, it printed two samples.

**GEOPM version**
784219e03

**Expected behavior**
```
$ geopmsession <<< 'TIME cpu 0'
"TIME-cpu-0"
0.001345207
```

**Actual behavior**
```
$ geopmsession <<< 'TIME cpu 0'
"TIME-cpu-0"
0.001345207
0.002139055
```

cmcantalupo:
- Fixes #3573

### DIFF
--- a/geopmdpy/geopmdpy/loop.py
+++ b/geopmdpy/geopmdpy/loop.py
@@ -69,6 +69,7 @@ class TimedLoop:
         self._period = period
         self._max_time = time.time() + period * num_period
         self._stop_iteration = False
+        self._one_shot = (num_period == 0)
 
     def __iter__(self):
         """Set up a timed loop
@@ -98,6 +99,8 @@ class TimedLoop:
 
         self._target_time += self._period
         self._loop_idx += 1
+        if self._one_shot:
+            self._stop_iteration = True
         return result
 
     def wait(self, timeout):


asmaalrawi:
In the case of derivative signals, the first n samples will return nan. E.g. power signals can return a value on the 2nd sample. 

E.g.

```
> geopmsession <<< 'CPU_POWER board 0'
"CPU_POWER"
nan
424.2861563497158

```
What should this behavior look like after this change?

cmcantalupo:
Producing two lines is a new "feature" that came in when I was fixing some other logic (#3562).  The change in this PR will make the output of geopmsession match the v3.1.0 release, which is to print only the nan.  

Maybe not great for those derivative based signals...might consider a different solution for the one-shot use case that looks more like what the geopmread CLI does for derivative signals (https://github.com/geopm/geopm/blob/dev/libgeopmd/src/DerivativeSignal.cpp#L100-L113).   I'd say that is a different issue, but still valid.

bgeltz (approved):

cmcantalupo:
https://github.com/geopm/geopm/issues/3582
Has been created to track the issue @asmaalrawi raised.

PR #3580, opened 2024-09-06
Add a profiler lock protecting the profile region name stack

### LINKED ISSUE #3520: test_multi_app.sh presents region exit without entry error when an old controller instance is still running
**Describe the bug**
I tried to run test_multi_app.sh and I expected it to consistently pass instead it rarely fails.

**GEOPM version**
v3.1

**Expected behavior**
With test_mult_app.sh configured not to produce trace files, and with GEOPM_CTL_LOCAL present:
```
bgeltz@x4709c7s0b0n0:~/output> PATH=${HOME}/build/stress-ng/bin:${PATH} mpiexec -n 1 -ppn 1 ~/geopm/integration/test/test_multi_app.sh
/home/bgeltz/output
stress-ng: info:  [204962] setting to a 10 secs run per stressor
stress-ng: info:  [204962] dispatching hogs: 1 cpu
stress-ng: info:  [204962] skipped: 0
stress-ng: info:  [204962] passed: 1: cpu (1)
stress-ng: info:  [204962] failed: 0
stress-ng: info:  [204962] metrics untrustworthy: 0
stress-ng: info:  [204962] successful run completed in 10.00 secs
Beginning loop of 2 iterations.
Completed loop.

bgeltz@x4709c7s0b0n0:~/output> ls
temp_config.json  test_multi_app_report.yaml-x4709c7s0b0n0
```

**Actual behavior**
Exit before entry error, and no report files output:
```
bgeltz@x4709c7s0b0n0:~/output> PATH=${HOME}/build/stress-ng/bin:${PATH} mpiexec -n 1 -ppn 1 ~/geopm/integration/test/test_multi_app.sh
/home/bgeltz/output
stress-ng: info:  [204894] setting to a 10 secs run per stressor
stress-ng: info:  [204894] dispatching hogs: 1 cpu
Error: <geopm> Invalid argument: ValidateRecord::filter(): Region exit without entry Region=0x00000000: at src/ValidateRecord.cpp:70
Error: <geopm> Invalid argument: ValidateRecord::filter(): Region exit without entry Region=0x00000000: at src/ValidateRecord.cpp:70
stress-ng: info:  [204894] skipped: 0
stress-ng: info:  [204894] passed: 1: cpu (1)
stress-ng: info:  [204894] failed: 0
stress-ng: info:  [204894] metrics untrustworthy: 0
stress-ng: info:  [204894] successful run completed in 10.00 secs
Beginning loop of 2 iterations.
Completed loop.

bgeltz@x4709c7s0b0n0:~/output> ls
temp_config.json
```

**Additional context**
The above runs were done back-to-back with no changes in between.

cmcantalupo:
- Fixes #3520

### DIFF
--- a/geopmdpy/geopmdpy/system_files.py
+++ b/geopmdpy/geopmdpy/system_files.py
@@ -1260,13 +1260,12 @@ class AccessLists(object):
 
 
 class WriteLock(object):
-    """Class for interacting with control lock file
+    """Class for interacting with lock files
 
-    This class provides the interface to query and set the PID that owns the
-    GEOPM Service control write lock.  The state of this lock is stored in the
-    file path:
-
-        /run/geopm/CONTROL_LOCK
+    This class provides the interface to query and set the PID
+    associated with a lock file.  This is used to support the GEOPM
+    Service control write lock and the application profile monitoring
+    lock.
 
     This file is empty when the lock is free, and contains the PID of the
     controlling process when the lock is held.  The class manages the advisory
@@ -1275,6 +1274,10 @@ class WriteLock(object):
     permissions.  Manipulations of the control lock file should be done
     exclusively with the WriteLock object to insure that the advisory lock is
     effective.
+
+    The default name of the lock file is "CONTROL_LOCK", however, this path name
+    can be overridden with the "lock_name" construction argument to support
+    other lock files used by the GEOPM Service.
 
     """
     def __init__(self, run_path, lock_name="CONTROL_LOCK"):
@@ -1285,7 +1288,10 @@ class WriteLock(object):
         securely create the run_path if it does not exist.
 
         Args:
-            run_path: Directory to create control lock within
+            run_path (str): Directory to create control lock within
+
+            lock_name (str): Optional name for lock file within run_path
+                             directory.  Default: "CONTROL_LOCK".
 
         """
         self._RUN_PATH = run_path
--- a/geopmdpy/geopmdpy/system_files.py
+++ b/geopmdpy/geopmdpy/system_files.py
@@ -1277,7 +1277,7 @@ class WriteLock(object):
     effective.
 
     """
-    def __init__(self, run_path):
+    def __init__(self, run_path, lock_name="CONTROL_LOCK"):
         """Set up initial state
 
         The WriteLock must be used within a context manager.  Use a
@@ -1289,7 +1289,7 @@ class WriteLock(object):
 
         """
         self._RUN_PATH = run_path
-        self._LOCK_PATH = os.path.join(self._RUN_PATH, "CONTROL_LOCK")
+        self._LOCK_PATH = os.path.join(self._RUN_PATH, lock_name)
         self._fid = None
         secure_make_dirs(self._RUN_PATH,
                          perm_mode=GEOPM_SERVICE_RUN_PATH_PERM)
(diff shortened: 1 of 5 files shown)


integration/test/test_multi_app.sh
```diff
@@ -25,6 +25,22 @@ GEOPM_REPORT_SIGNALS=TIME@package \
...
+# Check that second controller startup fails with correct error message
+TEMP_LOG=$(mktemp)
+geopmctl &> ${TEMP_LOG}
+if [ $? -eq 0 ]; then
+    echo "Error: Attempt to create second GEOPM Controller succeeded" 1>&2
+    cat ${TEMP_LOG}
+    rm ${TEMP_LOG}
+    exit -1
+fi
+if ! grep -q "requested control of application profile, but the geopm service already has a application profile controlling client" ${TEMP_LOG}; then
```
bgeltz:
```suggestion
if ! grep -q "requested control of application profile, but the geopm service already has an application profile controlling client" ${TEMP_LOG}; then
```

libgeopmd/src/LevelZeroIOGroup.cpp
```diff
@@ -811,6 +811,10 @@ namespace geopm
             for (int domain_idx = 0; domain_idx < m_platform_topo.num_domain(
                                      signal_domain_type(sv.first)); ++domain_idx) {
                 try {
+                    if (sv.first.find("_PERFORMANCE_") != std::string::npos && !m_is_perf_factor_enabled) {
+                        unsupported_signal_names.push_back(sv.first);
+                        break;
+                    }
```
bgeltz, line 817: What's this change doing here?  What does it fix re: this bug?

bgeltz (changes requested):

libgeopmd/src/LevelZeroIOGroup.cpp
cmcantalupo: It is unrelated.  @dannosliwcd sited the issue.  I can open a separate pull pull request it needed.

cmcantalupo: I have updated the commit message just now.
https://github.com/geopm/geopm/pull/3580/commits/1c18bd84ebe12f9522d1214f61acfe3885c9f705

dannosliwcd: I didn't observe any problematic behavior, just found this through unplanned code inspection.

I was digging in this code while looking into adding memory bandwidth signal and found it odd that we check whether the signal's name matches name patterns of unsupported signals _after_ we attempt to read the signal (`sv.second.m_devpool_func(domain_idx);`).

So with the prior implementation, we try to read the signal, then we check whether it's a signal that we know we should avoid reading.

geopmdpy/geopmdpy/system_files.py
```diff
@@ -1277,7 +1277,7 @@ class checks that the lock file is a regular file with restricted
     effective.
 
     """
-    def __init__(self, run_path):
```
dannosliwcd, line 1280: The class-level and function-level docstrings need to be updated

dannosliwcd (commented):
Changes look good overall.

If users see "requested control of application profile, but the geopm service already has an application profile controlling client", what should they do?

bgeltz (dismissed):

bgeltz (approved):

PR #3585, opened 2024-09-10
Make OMPT region detection opt-in

cmcantalupo:
- Overhead from OMPT calls can be significant
- Since only benefits some use cases, change it to opt-in rather than opt-out

### DIFF
--- a/geopmpy/geopmpy/launcher.py
+++ b/geopmpy/geopmpy/launcher.py
@@ -141,7 +141,9 @@ class Config(object):
         parser.add_argument('--geopm-debug-attach', dest='debug_attach', type=str)
         parser.add_argument('--geopm-preload', dest='preload', action='store_true', default=False)
         parser.add_argument('--geopm-hyperthreads-disable', dest='allow_ht_pinning', action='store_false', default=True)
-        parser.add_argument('--geopm-ompt-disable', dest='ompt_disable', action='store_true', default=False)
+        ompt_group = parser.add_mutually_exclusive_group()
+        ompt_group.add_argument('--geopm-ompt-enable', dest='ompt_enable', action='store_true', default=False)
+        ompt_group.add_argument('--geopm-ompt-disable', dest='ompt_disable', action='store_true', default=False)
         parser.add_argument('--geopm-record-filter', dest='record_filter', type=str)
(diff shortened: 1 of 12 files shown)


geopmpy/geopmpy/launcher.py
```diff
@@ -141,7 +141,7 @@ def __init__(self, argv):
         parser.add_argument('--geopm-debug-attach', dest='debug_attach', type=str)
         parser.add_argument('--geopm-preload', dest='preload', action='store_true', default=False)
         parser.add_argument('--geopm-hyperthreads-disable', dest='allow_ht_pinning', action='store_false', default=True)
-        parser.add_argument('--geopm-ompt-disable', dest='ompt_disable', action='store_true', default=False)
+        parser.add_argument('--geopm-ompt-enable', dest='ompt_enable', action='store_true', default=False)
```
dannosliwcd: Could consider guiding users with existing scripts that use the old option. E.g., 
```
ompt_group = parser.add_mutually_exclusive_group()
ompt_group.add_argument(enable...)
ompt_group.add_argument(disable..., help="Deprecated since OMPT regions are disabled by default.")
...
if opts.ompt_disable:
    print('Warning: --geopm-ompt-disable is deprecated since OMPT regions are now disabled by default. The option may be removed in future releases.', file=sys.stderr)
    # don't abort since ignoring this flag achieves that user's desired behavior anyways
    # or don't even warn, and treat "user explicitly asked for the default" as normal behavior
```

cmcantalupo:
```

srun -N 1 -n 1 lscpu


srun -N 1 -n 1 lscpu

test_regions_absent (test.test_omp_outer_loop.TestIntegrationOMPOuterLoop)
Test that the first run's report does NOT contain ... ok
test_regions_present (test.test_omp_outer_loop.TestIntegrationOMPOuterLoop)
Test that the second run's report DOES contain the ... ok

srun -N 1 -n 1 lscpu

test_unmarked_ompt (test.test_ompt.TestIntegration_ompt) ... ok

Ran 3 tests in 38.553s

OK
```

geopmpy/geopmpy/launcher.py
cmcantalupo: I've added in a check in geopmlaunch and also in Environment.cpp

dannosliwcd (approved):

bgeltz (approved):

Issue #3550, opened 2024-08-08
Some GPU signals show up in the report as NAN or extremely large after service restart


bgeltz:
**Describe the bug**
I tried to use the GEOPM Runtime to monitor a simple microkernel that leverages the GPUs and I expected all the fields I added to the report via ```GEOPM_REPORT_SIGNALS``` to provide valid data instead several GPU signals reported ```nan```.

**GEOPM version**
v3.1

**Expected behavior**
The added report signals were:
```
GEOPM_REPORT_SIGNALS="TIME@board,BOARD_POWER@board,BOARD_POWER_LIMIT_CONTROL@board,CPU_POWER@package,CPU_FREQUENCY_STATUS@package,CPU_FREQUENCY_MAX_CONTROL@package,CPU_UNCORE_FREQUENCY_MAX_CONTROL@package,CPU_UNCORE_FREQUENCY_MIN_CONTROL@package,CPU_INSTRUCTIO
NS_RETIRED@package,CPU_PACKAGE_TEMPERATURE@package,CPU_UNCORE_FREQUENCY_STATUS@package,MSR::CPU_SCALABILITY_RATIO@package,MSR::QM_CTR_SCALED_RATE@package,DRAM_POWER@package,GPU_POWER@gpu,GPU_CORE_POWER@gpu_chip,GPU_CORE_FREQUENCY_STATUS@g
pu_chip,GPU_CORE_ACTIVITY@gpu_chip,GPU_UTILIZATION@gpu,GPU_UNCORE_ACTIVITY@gpu_chip,LEVELZERO::GPU_CORE_TEMPERATURE_MAXIMUM@gpu,DRM::THROTTLE_REASON_PL1@gpu,DRM::THROTTLE_REASON_PL2@gpu,DRM::THROTTLE_REASON_PL4@gpu,DRM::THROTTLE_REASON_PR
OCHOT@gpu,DRM::THROTTLE_REASON_RATL@gpu,DRM::THROTTLE_REASON_STATUS@gpu,DRM::THROTTLE_REASON_THERMAL@gpu,DRM::THROTTLE_REASON_VR_TDC@gpu"
```

```
GEOPM Version: 3.1.1.dev47+g6e066589d
Start Time: Thu Aug 08 15:03:17 2024
Profile: "sgemm_1_cpu_sweep_2"
Agent: monitor
Policy: {}

Hosts:
  node1:
    Application Totals:
...
[...]

PR #3586, opened 2024-09-10
DerivativeSignal update so GPU_ACTIVITY reads 0 instead of NAN.

asmaalrawi:
- Relates to #3550 

Update DerivativeSignal to add a new constructor including m_nan_replace, which replaces NANs due to the denominator evaluating to 0. The default behavior is to output NAN (e.g. oversampling).  For GPU utilization/activity signals, the timer does not update if there's no activity on the card, so 0/0 should evaluate to 0.

In the future, the aggregator should also be updated to handle "NAN" appropriately (rather than propagating NAN).

### DIFF
--- a/libgeopmd/src/DerivativeSignal.cpp
+++ b/libgeopmd/src/DerivativeSignal.cpp
@@ -104,7 +104,7 @@ namespace geopm
         double time = m_time_sig->sample();
         size_t history_size = m_history.size();
         // Check if this is the first call ever to sample() (history_size == 0)
-        // Or check if this is the first call to sample() since the last call to read_batch() (last element of history buffer does not match the sampled time)
+        // Or check if this is the first call to sample() since the last call to read_batch()
         if (history_size == 0ULL || ! is_sampled) {
             double signal = m_y_sig->sample();
             m_last_result = compute_next(m_history, m_derivative_num_fit, time, signal, m_nan_replace);
--- a/libgeopmd/src/DerivativeSignal.cpp
+++ b/libgeopmd/src/DerivativeSignal.cpp
@@ -26,6 +26,7 @@ namespace geopm
         , m_is_batch_ready(false)
         , m_sleep_time(sleep_time)
         , m_last_result(NAN)
+        , m_nan_replace(0)
     {
         GEOPM_DEBUG_ASSERT(m_time_sig && m_y_sig,
                            "Signal pointers for time_sig and y_sig cannot be null.");
@@ -75,6 +76,9 @@ namespace geopm
             if (ssxx != 0) {
                 result = ssxy / ssxx;
             }
+            else {
+                result = m_nan_replace;
+	    }
         }
         return result;
     }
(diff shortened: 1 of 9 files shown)


libgeopmd/src/DerivativeSignal.cpp
```diff
@@ -26,6 +26,7 @@ namespace geopm
         , m_is_batch_ready(false)
         , m_sleep_time(sleep_time)
         , m_last_result(NAN)
+        , m_nan_replace(0)
```
dannosliwcd: Since this is intended for only 2 signals, should the default behavior match the current behavior? Then we could just apply the zero-fill for the two signals where we know that's desirable.

dannosliwcd (commented):
Are there any cases resulting from this change where GEOPM traces would contain false data that otherwise would have been reported as indeterminate data that users could find with a nan query?

Can you check how this behaves on traces with `LEVELZERO::GPU_POWER` and `LEVELZERO::GPU_CORE_POWER`? If neighboring GEOPM samples are too close together (I'm not sure whether LevelZero explicitly defines an expected energy-counter update period), then the GPUs will occasionally show 0 W. Maybe that problem only happens if the trace-head or any later 8 (window width selected by LevelZeroIOGroup) samples in a row are too close, so that's less likely.

libgeopmd/src/DerivativeSignal.cpp
asmaalrawi: Other signals are currently using MSR::TIME which is guaranteed to monotonically increase with each sample. When new signals are introduced that use timers that might not increase with each read, we will add a new constructor that allows m_nan_replace to be declared.

asmaalrawi:
Changed this so that default behavior when oversampling is "NAN" and added a constructor that allows each DerivativeSignal to choose a default value to replace any samples where the timestamp hasn't changed.

I tested with GPU_POWER and GPU_CORE_POWER using geopmsession and was not able to sample quickly enough to get a NAN or a 0 after the first sample.

libgeopmd/src/DerivativeSignal.cpp
```diff
@@ -26,7 +26,25 @@ namespace geopm
...
+                                       double nan_replace)
+        : m_time_sig(std::move(time_sig))
+        , m_y_sig(std::move(y_sig))
+        , M_NUM_SAMPLE_HISTORY(num_sample_history)
+        , m_history(M_NUM_SAMPLE_HISTORY)
+        , m_derivative_num_fit(0)
+        , m_is_batch_ready(false)
+        , m_sleep_time(sleep_time)
+        , m_last_result(NAN)
+        , m_nan_replace(nan_replace)
```
cmcantalupo: Avoid the copy/pasted code to maintain by calling the nan_replace constructor in the one with out nan_replace

cmcantalupo (dismissed):

cmcantalupo:
Note this does change the behavior for the theoretical case where a different signal's denominator is not updated.  Previous to this change, the result would be the last calculated value, now it will be NaN

bgeltz:
An issue should still be filed around having the reporter / aggregator ignore NaNs.  As it is, if a NaN is aggregated, the reporter will effectively ignore that signal forever.

asmaalrawi:
> An issue should still be filed around having the reporter / aggregator ignore NaNs. As it is, if a NaN is aggregated, the reporter will effectively ignore that signal forever.

https://github.com/geopm/geopm/issues/3592

cmcantalupo (approved):

PR #3590, opened 2024-09-13
Add new features to geopmsession reports and add Prometheus exporter

cmcantalupo:
- StatsCollector raw data access
- CSV output for geopmsession reports
- geopmsession periodic reporting
- Prometheus exporter as new installed python script


geopmdpy/test/TestCollector.py
```diff
@@ -0,0 +1,45 @@
...
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+
+from socket import gethostname
+import unittest
+from geopmdpy import loop
+from geopmdpy import pio
+from geopmdpy.stats import Collector
+from yaml import load as yaml_load
```
cmcantalupo: need to add yaml to requirements and packaging files.

geopmdpy/geopmdpy/stats.py
```diff
@@ -132,5 +200,9 @@ def reset(self):
         global _dl
         self._check_ptr('reset')
         err = _dl.geopm_stats_collector_reset(self._collector_ptr)
+        self._is_reset = True
         if err < 0:
             raise RuntimeError('geopm_stats_collector_reset() failed: {}'.format(error.message(err)))
+
+    def is_reset(self):
```
cmcantalupo: Consider adding "is_reset()" method to C++ Collector, not just the python interface

geopmdpy/geopmdpy/session.py
```diff
@@ -41,11 +41,15 @@ def term_handler(signum, frame):
...
+def _write_report(report_path, report, do_append=False):
     if report_path == '-':
         sys.stdout.write(report)
     else:
-        with open(report_path, "w") as fid:
+        if do_append:
+            mode = 'a'
+        else:
+            mode = 'w'
+        with open(report_path, mode) as fid:
```
cmcantalupo: Consider keeping the report file descriptor open like we do the trace given the periodic reporting feature

geopmdpy/geopmdpy/stats.py
```diff
@@ -103,24 +130,153 @@ def update(self):
...
+        """
+        report = self.report()
+        result = []
+        header = ['host', 'sample-time-first', 'sample-time-total', 'sample-count', 'sample-period-mean', 'sample-period-std']
+        data = [f'"{report[kk]}"' if type(report[kk]) is str else str(report[kk]) for kk in header]
+
+        metric_stat_names = ['count', 'first', 'last', 'min', 'max', 'mean-arithmetic', 'std']
+        for metric_name in sorted(report['metrics'].keys()):
+            for stat_name in metric_stat_names:
+                data.append(str(report['metrics'][metric_name][stat_name]))
```
cmcantalupo: string conversion should happen in report_csv not in report_table

geopmdpy/geopmdpy/exporter.py
```diff
@@ -0,0 +1,161 @@
...
+#
+
+import sys
+import os
+from time import sleep
+from argparse import ArgumentParser
+from geopmdpy import pio, topo, stats
+from geopmdpy.loop import TimedLoop
+from geopmdpy.session import ReadRequestQueue
+from prometheus_client import start_http_server, Gauge, Counter, Summary
```
cmcantalupo:
```suggestion
from . import pio, topo, stats
from . import loop.TimedLoop as TimedLoop
from . import session.ReadRequestQueue as ReadRequestQueue
from prometheus_client import start_http_server, Gauge, Counter, Summary
```

libgeopmd/src/StatsCollector.cpp
```diff
@@ -18,8 +18,13 @@
 #include "geopm_time.h"
 #include "RuntimeStats.hpp"
 
+
 namespace geopm
 {
+    static_assert(static_cast<size_t>(StatsCollector::NUM_SAMPLE_STATS) == static_cast<size_t>(GEOPM_NUM_SAMPLE_STATS),
+                  "C++ enum NUM_SAMPLE_STATS does not match C enum geopm::StatsCollector::NUM_SAMPLE_STATS");
+    static_assert(static_cast<size_t>(StatsCollector::NUM_METRIC_STATS) == static_cast<size_t>(GEOPM_NUM_METRIC_STATS),
+                  "C++ enum NUM_METRIC_STATS does not match C enum geopm::StatsCollector::NUM_METRIC_STATS");
```
dannosliwcd: Why not define `sample_stats_e` once in the c header (maybe with a `geopm_` prefix) and use that definition from both c/c++ locations?

With the current interface, C users have no way to symbolically get the individual stat array elements.

libgeopmd/src/StatsCollector.cpp
```diff
@@ -227,6 +263,46 @@ int geopm_stats_collector_report_yaml(const struct geopm_stats_collector_s *coll
...
 }
 
+int geopm_stats_collector_report(const struct geopm_stats_collector_s *collector,
+                                 size_t num_requests, geopm_report_s *report)
+{
+    int err = 0;
+    try {
+        const geopm::StatsCollector *collector_cpp = reinterpret_cast<const geopm::StatsCollector *>(collector);
+        geopm::StatsCollector::report_s report_cpp = collector_cpp->report_struct();
+        if (report_cpp.metric_names.size() != num_requests) {
```
dannosliwcd: Would any num_requests >= size be acceptable? The actual count gets written to report->num_metric anyways

libgeopmd/src/StatsCollector.cpp
```diff
@@ -227,6 +263,46 @@ int geopm_stats_collector_report_yaml(const struct geopm_stats_collector_s *coll
...
+            throw geopm::Exception("geopm_stats_collector_report(): Report not correctly allocated num_request provided: " +
+                                   std::to_string(num_requests) + " required: " + std::to_string(report_cpp.metric_names.size()),
+                            GEOPM_ERROR_INVALID, __FILE__, __LINE__);
+        }
+        if (report_cpp.host.size() >= NAME_MAX) {
+            throw geopm::Exception("geopm_stats_collector_report(): Host name too long: " + report_cpp.host,
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        strncpy(report->host, report_cpp.host.c_str(), NAME_MAX - 1);
```
dannosliwcd: Does the destination buffer get left unterminated if strlen == NAME_MAX-1?

`std::string::size` returns the size of a non-nul-terminated string. The longest equivalent strlen accepted by this check is NAME_MAX-1.

Then we copy:
* to report->host (big enough for strlen NAME_MAX-1 + a nul terminator)
* from report_cpp (of strlen at most NAME_MAX-1)
* of at most NAME_MAX-1 bytes, excluding any nul terminator if this limit is reached

I think writing report->host[NAME_MAX-1]='\0' after the copy would avoid that case

libgeopmd/src/StatsCollector.cpp
```diff
@@ -227,6 +263,46 @@ int geopm_stats_collector_report_yaml(const struct geopm_stats_collector_s *coll
...
+        if (report_cpp.sample_time_first.size() >= NAME_MAX) {
```
dannosliwcd, line 302: Same question about size vs len here

libgeopmd/src/StatsCollector.cpp
```diff
@@ -227,6 +263,46 @@ int geopm_stats_collector_report_yaml(const struct geopm_stats_collector_s *coll
...
+            throw geopm::Exception("geopm_stats_collector_report(): Date too long: " + report_cpp.sample_time_first,
+                                   GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        strncpy(report->sample_time_first, report_cpp.sample_time_first.c_str(), NAME_MAX - 1);
+        memcpy(report->sample_stats, report_cpp.sample_stats.data(), sizeof(double) * GEOPM_NUM_SAMPLE_STATS);
+        report->num_metric = report_cpp.metric_names.size();
+        for (size_t metric_idx = 0; metric_idx < report->num_metric; ++metric_idx) {
+            if (report_cpp.metric_names[metric_idx].size() >= NAME_MAX) {
```
dannosliwcd, line 311: Same question about size vs len here

dannosliwcd (commented):
Partial review. Covered 61450ef1514e0ffab0a3d5b4856338fd23d066e7

dannosliwcd (commented):
Partial review. No comments on 50c32a2bc94c532243b198a2cf8a82a5174ede7c

docs/debian/copyright
```diff
@@ -124,6 +124,7 @@ Files:     .gitignore
            source/geopmlaunch.1.rst
            source/geopmpy.7.rst
            source/geopmread.1.rst
+           source/geopmexporter.1.rst
```
bgeltz: This should follow ``geopmendpoint``.  These sections are alphabetized.

docs/source/conf.py
```diff
@@ -209,6 +209,7 @@ def get_description(rst_file_name):
     "geopm_version.3",
     "geopmwrite.1",
     "geopmaccess.1",
+    "geopmexporter.1",
```
bgeltz: Not sure what order these are in, nor why this is in the middle and not at the end.  It's almost like someone tried to alphabetize them until ``geopmaccess.1`` was added?  Unclear.

docs/source/geopmexporter.1.rst
```diff
@@ -0,0 +1,194 @@
...
+   Sum of all samples divided by the number of samples since last Prometheus
+   server scrape
+
+std:
+   Sample standard deviation represented by the following sum over samples:
+
+       ``sqrt(sum((x_i - m) ^ 2) / (n - 1))``
+
+   where ``m`` is the mean, ``x_i`` are the samples, and ``n`` is the number of
+   samples
```
bgeltz, line 102: Does the number of samples keep increasing forever (util the StatsCollector object is destroyed)?

docs/source/geopmexporter.1.rst
```diff
@@ -0,0 +1,194 @@
...
+Prometheus Summary Method
+-------------------------
+
+When the user specifies the ``--summary prometheus`` command line option the
+``Summary`` and ``Counter`` metrics built into Prometheus are used to aggregate
+samples over time.  In this way the batch of signals read from the GEOPM is used
+to update a ``Summary`` metric for GEOPM signals with variable behavior, or a
+``Counter`` metric for signals with monotone behavior.  Each metric provides
+slightly different statistics for the Prometheus server on the port opened by
+``geopmexporter``.  the ``Summary`` metric exports the **count** and **sum**
```
bgeltz:
```suggestion
``geopmexporter``.  The ``Summary`` metric exports the **count** and **sum**
```

docs/source/geopmexporter.1.rst
```diff
@@ -0,0 +1,194 @@
...
+systemd service.  To use ``geopmexporter`` with systemd create a systemd unit
+configuration file called ``geopmexporter.service`` with the following content:
+
+.. code-block:: ini
+
+   [Unit]
+   Description=Prometheus GEOPM Exporter
+   StartLimitIntervalSec=0
+   StartLimitBurst=3
+   Wants=geopm.service
```
bgeltz, line 141: Will the exporter still function if geopm.service is not available?  Wondering if this should be ``Requires`` instead.

docs/source/geopmexporter.1.rst
```diff
@@ -0,0 +1,194 @@
...
+notified of the new service.
+
+.. code-block:: bash
+
+   sudo groupadd --system prometheus
+   sudo useradd -s /sbin/nologin --system -g prometheus prometheus
+   sudo cp geopmexporter.service /etc/systemd/system/
+   sudo systemctl deamon-reload
+   sudo systemctl enable geopmexporter
+   sudo systemctl start geopmexporter
```
bgeltz, line 172: We could also cook up an example of how to do this without using sudo, i.e. a userspace systemd service.  Not sure what all a user would want in this space.

docs/source/geopmsession.1.rst
```diff
@@ -46,14 +47,15 @@ Description
...
+of three strings separated by white space. The first string is the signal name,
+the second string is the domain name, and the third string is the domain index.
+Provide the "``*``" character as the third string to request all domains
+available on the system.
 
-Unless the ``-n`` option is specified, a descriptive header is written
-first and then the output from reading the requests.  Unless ``-p`` is
-specified, only one line of output after the header is printed.
+Unless the ``-n`` option is specified, a descriptive header is written first and
+then the output from reading the requests.  Unless ``-p`` is specified, only one
```
bgeltz: This is confusing as-is.  Suggestion:
```suggestion
If the ``-n`` option is specified a descriptive header is written first and
then the output from reading the requests.  Unless ``-p`` is specified, only one
```

geopmdpy/geopmdpy/exporter.py
```diff
@@ -0,0 +1,201 @@
...
+
+        """
+        _, self._metric_data = self._stats_collector.report_table()
+        self._stats_collector.reset()
+        self._num_fresh = self._num_metric
+        self._is_fresh = len(self._metric_names) * [True]
+
+    def _get_metric(self, metric_idx):
+        """Get metric from cached report and update cache if all metrics have
+        been sampled at least once.
```
bgeltz, line 54: So the cache is *only* updated when everything has been sampled?  What is it that enforces that all metrics are sampled?

geopmdpy/geopmdpy/session.py
```diff
@@ -256,15 +287,29 @@ def check_read_args(self, run_time, period):
             raise RuntimeError('Specified a period greater than 24 hours')
         if period < 0.0 or run_time < 0.0:
             raise RuntimeError('Specified a negative run time or period')
+        if report_samples is not None:
+            if report_samples < 0:
+                raise RuntimeError('Specified report samples is negative')
+            if int(report_samples) != report_samples:
+                raise RuntimeError(f'Specified report samples is non-integer: {report_samples}')
```
bgeltz: Shouldn't these if checks be reversed since you're doing an integer operation with the ``< 0 check``?
```suggestion
            if int(report_samples) != report_samples:
                raise RuntimeError(f'Specified report samples is non-integer: {report_samples}')
            if report_samples < 0:
                raise RuntimeError('Specified report samples is negative')

```

integration/requirements.txt
```diff
@@ -2,3 +2,4 @@ matplotlib>=2.2.2
 docutils>=0.14
 plotly>=5.18.0
 dash>=2.15.0
+seaborn>=0.11.2
```
bgeltz: What is this about?

bgeltz (changes requested):

docs/source/geopmexporter.1.rst
cmcantalupo: Note the memory requirements for the collector do not change as more samples are added.

docs/source/geopmexporter.1.rst
cmcantalupo: In theory, you could be exporting only signals that can be read without the service.  Not very likely, but could be a supported use case (thinking sysfs entries for gaudi that are not privileged.

docs/source/geopmsession.1.rst
cmcantalupo: The -n option skips writing the CSV header.  I'm not sure how to make that clearer, I think the other order is more confusing?

"A descriptive header is written first, unless the ``-n`` option is specified, in which case the header is not printed.  The output from reading the requests is written subsequently."

geopmdpy/geopmdpy/exporter.py
cmcantalupo: The Prometheus server is expected to sample each metric that is registered at least once per update loop.

geopmdpy/geopmdpy/session.py
cmcantalupo: On closer inspection, argparse does the integer check for us. I think the ``if int(report_samples) != report_samples`` is dead code:

```
$ geopmsession -s 1.234
usage: geopmsession [-h] [-v] [-t TIME] [-p PERIOD] [--pid PID] [--print-header | -n] [-d DELIMITER] [-r REPORT_OUT] [-o TRACE_OUT] [--enable-mpi] [-f REPORT_FORMAT] [-s REPORT_SAMPLES]
geopmsession: error: argument -s/--report-samples: invalid int value: '1.234'
```

integration/requirements.txt
cmcantalupo: Just noticed it was required by a python module in integration but not listed...

geopmdpy/geopmdpy/exporter.py
bgeltz: > ...at least once per update loop

If the Prometheus server samples twice in an update loop, isn't there an issue here?  I'm thinking that on that second sample, self._refresh() was called for one metric (the one being sampled a second time) but not another one (where it was previously sampled only once during this update loop) that the data would be inconsistent.  Wouldn't this be a problem for any derivative based signals?

docs/source/geopmsession.1.rst
bgeltz: I had understood the behavior of -n backwards.  I think it's fine.

The only other language I'd consider is "To suppress the printing of the CSV column headers, use -n".

bgeltz (changes requested):
Thanks for addressing my comments.  One more, and this test is failing:

```
FAIL: test_check_read_args (test.TestSession.TestSession)
  File "/home/runner/work/geopm/geopm/geopmdpy/test/TestSession.py", line 153, in test_check_read_args
    with self.assertRaisesRegex(RuntimeError, err_msg):
AssertionError: RuntimeError not raised
```

bgeltz (approved):

PR #3614, opened 2024-10-11
Ignore first batch read if it contains NaN values

### LINKED ISSUE #3582: Why does using geopmsession to read one sample of POWER always print nan?
**Describe the bug**
I tried to run

```
echo CPU_POWER board 0 | geopmession
```
and I expected to get a number instead nan is printed

**GEOPM version**
3.1.0

**Expected behavior**
For derivative signals, more than one sample is required before the batch interface will provide a non-nan result.  This is resolved though the read_signal() api by sampling multiple times for a single call. 

**Actual behavior**
Output is nan for all derivative based signals when using the default options for geopmsession

**Error messages produced**
No error messages produced, but nan is printed to stdout

cmcantalupo:
- Fixes #3582 
- Similar solution is deployed in the exporter:
  https://github.com/geopm/geopm/blob/dev/geopmdpy/geopmdpy/exporter.py#L23

### DIFF
--- a/geopmdpy/test/TestSession.py
+++ b/geopmdpy/test/TestSession.py
@@ -53,7 +53,7 @@ class TestSession(unittest.TestCase):
         format_return_value = "1.234, 2.345, 3.456"
         mock_requests.get_formats.return_value = format_return_value
         signal_handle = list(range(3))
-        signal_expect = num_period * [1.234, 2.345, 3.456]
+        signal_expect = (1 + num_period) * [1.234, 2.345, 3.456]
 
         with mock.patch('geopmdpy.loop.TimedLoop', return_value=[0, 1]) as mock_timed_loop, \
              mock.patch('geopmdpy.pio.push_signal', side_effect=signal_handle) as mock_push_signal, \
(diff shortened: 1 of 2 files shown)


geopmdpy/test/TestSession.py
```diff
@@ -72,6 +72,25 @@ def test_run_read(self):
...
+            self._session.run_read(mock_requests, duration, period, pid=None, out_stream=out_stream)
+            # Check mocks calls
+            calls = [mock.call(*req) for req in pio_requests]
+            mock_push_signal.assert_has_calls(calls)
+            calls = (1 + num_period) * [mock.call()]
+            mock_read_batch.assert_has_calls(calls)
+            calls = num_period * [mock.call(idx) for idx in signal_handle]
+            mock_sample.assert_has_calls(calls)
+            calls = num_period * [mock.call(format_return_value)]
+            out_stream.write.assert_has_calls(calls)
```
avilcheslopez, line 92: Maybe not worth blocking this PR, but I'm noticing a lot of similarity between this block of test/assert and the previous one, so there's opportunity to factor out common code.

avilcheslopez (approved):

bgeltz (approved):

PR #3616, opened 2024-10-12
Warn about required environment variables when using LevelZero

### LINKED ISSUE #3539: GEOPM reports no GPUs are present if the first usage of LevelZeroDevicePool is from the root user
**Describe the bug**
I tried to login to a freshly updated node as root and I expected `geopmread -d` to show GPUs were present. Instead it showed 0 GPUs even though there were 6 on the system.

**GEOPM version**
GEOPM 3.1.0

**Expected behavior**
I expected `geopmread -d` to show GPUs were present. 

**Actual behavior**
Instead it showed 0 GPUs even though there were 6 on the system.

**Error messages produced**
No error messages were produced.

**Additional context**
The root cause ended up being that the first instantiation of LevelZeroDevicePool was in a context where ZES_ENABLE_SYSMAN=1 was not set. The documentation for **geopm_pio_levelzero(7)** specifies that the variable needs to be set. The geopm systemd service sets it automatically, but the first invocation of the device pool (which sets the topology cache) was from a root user, who does not go through the service.

Since the documentation specifies the environment variable is needed, and since we cannot guarantee that it is always set, we should emit a warning when that expectation is not met at the time our topology cache is being created.

cmcantalupo:
- Fixes #3539

### DIFF
--- a/libgeopmd/src/GPUTopo.cpp
+++ b/libgeopmd/src/GPUTopo.cpp
@@ -53,8 +53,8 @@ namespace geopm
         }
 #endif
 #ifdef GEOPM_ENABLE_LEVELZERO
-        if (getenv("ZE_ENABLE_SYSMAN") != nullptr &&
-            std::string(getenv("ZE_ENABLE_SYSMAN")) == "1" &&
+        if (getenv("ZES_ENABLE_SYSMAN") != nullptr &&
+            std::string(getenv("ZES_ENABLE_SYSMAN")) == "1" &&
             getenv("ZE_FLAT_DEVICE_HIERARCHY") != nullptr &&
             std::string(getenv("ZE_FLAT_DEVICE_HIERARCHY")) == "COMPOSITE") {
             try {
--- a/libgeopmd/src/GPUTopo.cpp
+++ b/libgeopmd/src/GPUTopo.cpp
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+#include <unistd.h>
+
 #include <iostream>
 
 #include "geopm/Exception.hpp"
@@ -33,7 +35,10 @@ namespace geopm
         }
 
         try {
-            AccelTopo = std::make_unique<DrmGpuTopo>("/sys/class/accel");
+            std::string accel_path = "/sys/class/accel";
+            if (access(accel_path.c_str(), R_OK) == 0) {
+                AccelTopo = std::make_unique<DrmGpuTopo>(accel_path);
+            }
         }
         catch (const Exception &ex) {
             std::cerr << "Warning: <geopm> Unable to get /sys/class/accel topology. Reason: "
@@ -48,11 +53,16 @@ namespace geopm
         }
 #endif
 #ifdef GEOPM_ENABLE_LEVELZERO
-        try {
-            LevelZeroTopo = std::make_unique<LevelZeroGPUTopo>();
-        }
-        catch (const Exception &ex) {
-            // LevelZeroTopo cannot be created, already set to GPUTopoNull
+        if (getenv("ZE_ENABLE_SYSMAN") != nullptr &&
+            std::string(getenv("ZE_ENABLE_SYSMAN")) == "1" &&
+            getenv("ZE_FLAT_DEVICE_HIERARCHY") != nullptr &&
+            std::string(getenv("ZE_FLAT_DEVICE_HIERARCHY")) == "COMPOSITE") {
+            try {
+                LevelZeroTopo = std::make_unique<LevelZeroGPUTopo>();
+            }
+            catch (const Exception &ex) {
+                // LevelZeroTopo cannot be created, already set to GPUTopoNull
+            }
         }
 #endif
         if (NVMLTopo->num_gpu() != 0 &&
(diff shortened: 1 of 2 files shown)


avilcheslopez (dismissed):
Should the documentation (geopm_pio_levelzero(7)) be updated as well? Now that both paths (direct, through service) set it automatically, it seems the user no longer has to make sure it's set.

Also, with this change, does the service still need to set it (in libgeopmd/geopm.service)? If not, it might be a good idea to just set it in one place in case in the future we need to change that setting somehow.

bgeltz (dismissed):

bgeltz (changes requested):
Re-jig to emit warnings instead of setenv calls.

This violates process expectations about setenv/getenv timing.

libgeopmd/src/GPUTopo.cpp
```diff
@@ -48,11 +53,16 @@ namespace geopm
...
 #ifdef GEOPM_ENABLE_LEVELZERO
-        try {
-            LevelZeroTopo = std::make_unique<LevelZeroGPUTopo>();
-        }
-        catch (const Exception &ex) {
-            // LevelZeroTopo cannot be created, already set to GPUTopoNull
+        if (getenv("ZE_ENABLE_SYSMAN") != nullptr &&
+            std::string(getenv("ZE_ENABLE_SYSMAN")) == "1" &&
+            getenv("ZE_FLAT_DEVICE_HIERARCHY") != nullptr &&
+            std::string(getenv("ZE_FLAT_DEVICE_HIERARCHY")) == "COMPOSITE") {
```
cmcantalupo: This is a behavior change.  Not 100% sure it is for the better...

libgeopmd/src/LevelZero.cpp
```diff
@@ -32,7 +33,30 @@ namespace geopm
...
+                }
+                else {
+                    std::cerr << "                 ";
+                }
+                do_warn = true;
+                std::cerr << "LevelZero features require the environment variable setting \"ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE\".\n";
+            }
+            if (do_warn) {
+                std::cerr << "                 This process may corrupt the topology cache file.  Run the following commands as root to fix:\n"
+                          << "                     rm -f /run/geopm/geopm-topo-cache; ZE_ENABLE_SYSMAN=1 ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE geopmread --cache\n";
```
cmcantalupo: Might be nice to only warn if we are actually creating a topo cache file, but plumbing that turns out to be a bit tricky.

cmcantalupo:
```
cmcantal@saruman:~/Git/geopm/libgeopmd (issue-3539)]$ ./geopmread > /dev/null
cmcantal@saruman:~/Git/geopm/libgeopmd (issue-3539)]$ sudo ./geopmread > /dev/null
Warning: <geopm> LevelZero features require the environment variable setting "ZE_ENABLE_SYSMAN=1".
                 LevelZero features require the environment variable setting "ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE".
                 This process may corrupt the topology cache file if LevelZero enabled GPUs are present.
                 Run the following commands to fix:
                     sudo rm -f /run/geopm/geopm-topo-cache
                     sudo env ZE_ENABLE_SYSMAN=1 ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE geopmread --cache

cmcantal@saruman:~/Git/geopm/libgeopmd (issue-3539)]$ sudo env ZE_ENABLE_SYSMAN=1 ./geopmread > /dev/null
Warning: <geopm> LevelZero features require the environment variable setting "ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE".
                 This process may corrupt the topology cache file if LevelZero enabled GPUs are present.
                 Run the following commands to fix:
                     sudo rm -f /run/geopm/geopm-topo-cache
                     sudo env ZE_ENABLE_SYSMAN=1 ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE geopmread --cache
cmcantal@saruman:~/Git/geopm/libgeopmd (issue-3539)]$ sudo env ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE  ./geopmread > /dev/null
Warning: <geopm> LevelZero features require the environment variable setting "ZE_ENABLE_SYSMAN=1".
                 This process may corrupt the topology cache file if LevelZero enabled GPUs are present.
                 Run the following commands to fix:
                     sudo rm -f /run/geopm/geopm-topo-cache
                     sudo env ZE_ENABLE_SYSMAN=1 ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE geopmread --cache
```

bgeltz (commented):
Changes seem fine.  This should be tested on a L0 enabled system w/appropriate hardware, including building and deploying the service RPMs before we merge.

cmcantalupo:
```
$ geopmread GPU_POWER gpu 0
286.8566902140415
$ sudo geopmread
Warning: <geopm> LevelZero features require the environment variable setting "ZES_ENABLE_SYSMAN=1".
                 LevelZero features require the environment variable setting "ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE".
                 This process may corrupt the topology cache file if LevelZero enabled GPUs are present.
                 Run the following commands to fix:
                     sudo rm -f /run/geopm/geopm-topo-cache
                     sudo env ZES_ENABLE_SYSMAN=1 ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE geopmread --cache
...
```

also a clean syslog on that run.

bgeltz (approved):

PR #3633, opened 2024-10-29
Define GEOPM constants for string lengths

### LINKED ISSUE #3632: Use of NAME_MAX and PATH_MAX requires _POSIX_SOURCE but this may not be defined
**Describe the bug**
I tried to compile against geopm_pio.h without _POSIX_SOURCE defined, and the build failed.

**GEOPM version**
https://github.com/geopm/geopm/commit/05640dfa738d3f67e12822a4cc2c91a1ed85c72c

**Expected behavior**
Do not make assumptions about _POSIX_SOURCE being defined

**Actual behavior**
Buffer sizes are dependent on compiler options when our headers are included

**Error messages produced**
NAME_MAX not defined

cmcantalupo:
- Three new constants: ``GEOPM_NAME_MAX``, ``GEOPM_MESSAGE_MAX``, ``GEOPM_TIME_STRING_MAX``
- Replace all useage of ``NAME_MAX`` and most use of ``PATH_MAX``
- Replace some use of ``NAME_MAX`` with ``PATH_MAX`` when dealing with paths
- Use ``HOST_NAME_MAX`` when dealing with hostnames (defined in limits.h)
- Remove most includes of <limits.h> and <climits>
- Fixes #3632


libgeopm/include/geopm_endpoint.h
```diff
@@ -15,8 +15,8 @@ extern "C" {
 #endif
 
 static const size_t GEOPM_ENDPOINT_AGENT_NAME_MAX = 256;
-static const size_t GEOPM_ENDPOINT_PROFILE_NAME_MAX = 256;
-static const size_t GEOPM_ENDPOINT_HOSTLIST_PATH_MAX = 512;
+static const size_t GEOPM_ENDPOINT_PROFILE_GEOPM_NAME_MAX = 256;
+static const size_t GEOPM_ENDPOINT_HOSTLIST_GEOPM_MESSAGE_MAX = 512;
```
cmcantalupo: fix me

libgeopm/src/Endpoint.cpp
```diff
@@ -210,8 +210,8 @@ namespace geopm
         auto lock = m_sample_shmem->get_scoped_lock();
         struct geopm_endpoint_sample_shmem_s *data = (struct geopm_endpoint_sample_shmem_s *) m_sample_shmem->pointer(); // Managed by shmem subsystem.
 
-        char profile_name[GEOPM_ENDPOINT_PROFILE_NAME_MAX];
-        std::copy(data->profile_name, data->profile_name + GEOPM_ENDPOINT_PROFILE_NAME_MAX, profile_name);
+        char profile_name[GEOPM_ENDPOINT_PROFILE_GEOPM_NAME_MAX];
+        std::copy(data->profile_name, data->profile_name + GEOPM_ENDPOINT_PROFILE_GEOPM_NAME_MAX, profile_name);
```
cmcantalupo: fixme

integration/runtime_examples/test_hw/rapl_pkg_limit_test.c
```diff
@@ -139,8 +137,8 @@ int rapl_pkg_limit_test(double power_limit, int num_rep)
         }
     }
     if (!err) {
-        int outlen = snprintf(outfile_name, NAME_MAX, "rapl_pkg_limit_test_%s.out", hostname);
-        if (outlen >= NAME_MAX) {
+        int outlen = snprintf(outfile_name, GEOPM_NAME_MAX, "rapl_pkg_limit_test_%s.out", hostname);
```
dannosliwcd: Shouldn't this and other filename strings remain `NAME_MAX`?

libgeopmd/include/geopm_public.h
```diff
@@ -14,3 +14,17 @@
...
+ * reasons.
+ */
+#define GEOPM_NAME_MAX 255
+
+/*
+ * String length allocated for GEOPM messages (like C error messages
+ * derived from C++ exceptions).  Has same value as PATH_MAX from
+ * linux/limits.h for historical reasons.
+ */
+#define GEOPM_MESSAGE_MAX 4096
```
dannosliwcd: Since this is used for geopm error message strings should it go in geopm_error.h instead?

libgeopmd/include/geopm_public.h
```diff
@@ -14,3 +14,17 @@
  */
 #define GEOPM_PRIVATE __attribute__((visibility("hidden")))
 #endif
+
+/*
+ * String length allocated for GEOPM names (like signal and control
+ * names). Has same value as NAME_MAX from limits.h for historical
+ * reasons.
+ */
+#define GEOPM_NAME_MAX 255
```
dannosliwcd: If it's mainly for signal/control names, should this go in geopm_pio.h (or maybe a separate header used by all signal/control interfaces (I guess that's just pio and stats_collector right now)?

libgeopmd/include/geopm_public.h
cmcantalupo: Both of these macros are also used for other string lengths than the primary use case.  We could add a geopm_limits.h.  I believe that geopm_public.h is the only public header that defines preprocessor macros.  Perhaps this is reason to leave them in geopm_public.h?  Also note, that we only use C strings to support our public C interfaces and a few calls to system functions.

libgeopm/src/geopmctl_main.c
```diff
@@ -29,7 +28,7 @@ int main(int argc, char **argv)
 {
     int opt;
     int err0 = 0;
-    char error_str[NAME_MAX] = {0};
```
dannosliwcd, line 32: I think there's a pre-existing bug in this file. Line 77 calls `geopm_error_message(err0, error_str, GEOPMCTL_STRING_LENGTH);`. Should this declaration use `GEOPMCTL_STRING_LENGTH` instead?

libgeopmd/src/geopm_sched.c
```diff
@@ -118,9 +118,9 @@ int geopm_sched_proc_cpuset_pid(int pid, int num_cpu, cpu_set_t *cpuset)
     uint32_t *proc_cpuset = NULL;
     FILE *fid = NULL;
 
-    char status_path[NAME_MAX];
-    int nprint = snprintf(status_path, NAME_MAX, "/proc/%d/status", pid);
-    if (nprint == NAME_MAX) {
+    char status_path[GEOPM_NAME_MAX];
+    int nprint = snprintf(status_path, GEOPM_NAME_MAX, "/proc/%d/status", pid);
```
dannosliwcd: I think this one should have been PATH_MAX before since it's intended to be a multi-component path (though I guess incidentally it wouldn't ever exceed NAME_MAX anyways).

libgeopm/src/Agent.cpp
```diff
@@ -395,7 +395,7 @@ int geopm_agent_policy_json_partial(const char *agent_name,
                 if (i > 0) {
                     output_str << ", ";
                 }
-                err = geopm_agent_policy_name(agent_name, i, PATH_MAX, policy_name);
+                err = geopm_agent_policy_name(agent_name, i, GEOPM_MESSAGE_MAX, policy_name);
```
dannosliwcd: This is okay if the longer string is needed for compatibility with the established interface. Just checking though: would GEOPM_NAME_MAX be a better fit since this is a name?

libgeopm/src/geopmagent_main.cpp
```diff
@@ -68,7 +67,7 @@ static int main_imp(int argc, char **argv)
     char error_str[GEOPMAGENT_STRING_LENGTH] = {0};
     char agent_str[GEOPMAGENT_STRING_LENGTH] = {0};
     char policy_vals_str[GEOPMAGENT_STRING_LENGTH] = {0};
-    char output_str[NAME_MAX * GEOPMAGENT_DOUBLE_LENGTH];
+    char output_str[255 * GEOPMAGENT_DOUBLE_LENGTH];
```
dannosliwcd: Maybe replace 255 with a file-local define or static const to indicate the intent of this limit. Looking at how it's used, I guess the longest expected output_str is the policy json, so maybe 255 is the max policy values we expect to handle.

dannosliwcd (approved):

PR #3635, opened 2024-10-29
Create geopmpy.gffi module and move geopmpy.hash into the geopmdpy module

### LINKED ISSUE #3620: The hash module is located in the geopmpy package even though it does not depend on libgeopm
**Describe the bug**
The `geopmpy.hash.hash_str()` function is a wrapper for `geopm_crc32_str` in `geopm_hash.h`, which is part of libgeopmd (not libgeopm). Most other geopmpy modules depend on libgeopm. It would make more sense for the `hash` module to be part of the geopmdpy package since it only depends on libgeopmd.

**GEOPM version**
7acc719

cmcantalupo:
- Fixes #3620
- Fixes ##3624

### DIFF
--- a/geopmpy/geopmpy/gffi.py
+++ b/geopmpy/geopmpy/gffi.py
@@ -1,0 +1,25 @@
+#
+#  Copyright (c) 2015 - 2024 Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+#
+
+'''The gffi module provides a wrapper around the cffi interface
+'''
+
+def get_dl_geopm():
+    '''Get the FFILibrary instance for libgeopm.so
+
+    Returns:
+        FFILibrary: Object used to call functions defined in
+                    libgeopm.so
+
+    '''
+    if isinstance(_dl_geopm, Exception):
+        raise RuntimeError('Attempted to use libgeopm.so, which is not loaded. Make sure it is available '
+                           'in your system installs or in LD_LIBRARY_PATH') from _dl_geopm
+    return _dl_geopm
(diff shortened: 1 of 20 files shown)


geopmpy/geopmpy/gffi.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Returns:
+        FFILibrary: Object used to call functions defined in
+                    libgeopm.so
+
+    '''
+    if isinstance(_dl_geopm, Exception):
+        raise RuntimeError('Attempted to use libgeopm.so, which is not loaded. Make sure it is available '
+                           'in your system installs or in LD_LIBRARY_PATH') from _dl_geopm
+    return _dl_geopm
```
dannosliwcd, line 20: Is this still needed now that we don't need to import everything then later check which parts of everything failed to import?

That is, now we should only import libgeopm from modules that need libgeopm and we should only import libgeopmd from modules that need libgeopmd. If we're only importing needed dependencies, is it safe to fail at time of import now?

This suggestion feels like another broad-reaching change, so I'm not asking to address it in this PR. What are you thoughts on wont-change vs opening an issue to make that change in the future?

dannosliwcd (approved):
LGTM. The only comment is about a suggestion for possible future work. Open an issue, or wont-change?

geopmpy/geopmpy/gffi.py
cmcantalupo: Good question.  Worth at least opening an issue.  The change seems reasonable.

PR #3641, opened 2024-11-04
Add a script to install development snapshot of both geopmdpy and libgeopmd

cmcantalupo:
Prior to the changes to use cffi in API mode, a user could more easily mix a development snapshot of geopmdpy with a release version of libgeopmd.so.  Now, a development snapshot of libgeopmd.so will be required if new APIs have been added since the last release (e.g. features required for the geopmexporter).  

The purpose of this script is to enable users to install a development snapshot of both libgeopmd.so and geopmdpy in a non-system location (e.g. home directory).


geopmdpy/install_user.sh
```diff
@@ -0,0 +1,55 @@
...
+library_dirs         = %(mpi_dir)s/lib
+runtime_library_dirs = %(library_dirs)s" > mpi.cfg
+MPI4PY_BUILD_MPICFG=$PWD/mpi.cfg python3 -m pip install mpi4py
+if [ ! -d geopm ]; then
+    git clone https://github.com/geopm/geopm.git
+fi
+cd geopm
+git checkout dev
+git fetch origin
+git merge --ff-only origin/dev
```
dannosliwcd: This won't work as expected if the pre-existing `geopm` directory doesn't have the github.com/geopm/geopm remote named "origin".

Will the merge only work if the user's local `dev` has no new commits that are absent from origin/dev?

geopmdpy/install_user.sh
```diff
@@ -0,0 +1,55 @@
...
+    INSTALL_PREFIX=$HOME/geopm-build
+fi
+if [ -z "$MPI_DIR" ]; then
+    echo "HELP: To select a different MPI directory location export MPI_DIR"
+    MPI_DIR=$(dirname $(dirname $(which mpicc)))
+fi
+echo "INSTALL_PREFIX=$INSTALL_PREFIX"
+echo "MPI_DIR=$MPI_DIR"
+set -x
+sleep 2
```
dannosliwcd: What's the sleep for?

geopmdpy/install_user.sh
```diff
@@ -0,0 +1,55 @@
...
+    git clone https://github.com/geopm/geopm.git
+fi
+cd geopm
+git checkout dev
+git fetch origin
+git merge --ff-only origin/dev
+cd libgeopmd
+./autogen.sh
+./configure --enable-levelzero --prefix=$INSTALL_PREFIX
+make -j20
```
dannosliwcd: Alternatively `make -j`, or query the available core count, or just `make` (and let the user use `MAKEFLAGS`) so it isn't hard-coded for something that can do 20+ build tasks.

geopmdpy/install_user.sh
```diff
@@ -0,0 +1,55 @@
...
+make install
+export C_INCLUDE_PATH="$INSTALL_PREFIX/include"
+export LIBRARY_PATH="$INSTALL_PREFIX/lib:$INSTALL_PREFIX/lib64"
+cd ../geopmdpy
+./make_sdist.sh
```
dannosliwcd: I think the sdist is unused

cmcantalupo: This is mysteriously sometimes required to create the VERSION file.  I'm not really sure what the environment requirements are to make it unnecessary.  I have seen this on mcfly...

geopmdpy/install_user.sh
cmcantalupo: Just so the message about the environment variables is not missed...Maybe there is a better solution like printing at the end.

geopmdpy/install_user.sh
bgeltz: When make_sdist.py executes, it relies on ``import build`` which is specified in the requirements.txt for geopmdpy.

If this script is meant to be an easy button for end users to install from dev, it should also install the Python requirements that are needed for geopmdpy.  We've had users hit this issue recently.

Additionally, perhaps it would be better to root cause what's going on w/mcfly and see if this call to make_sdist.sh can be avoided entirely.  You'd still need to install the dev python requirements for geopmdpy (assuming they are different from whatever is installed, if GEOPM had previously been installed).

IMO this should all be handled through spack instead of a special script.  That's the proper way to handle this dependency on mpi4py.  We're going to have issues with this approach if/when people start using different Python environments for their target applications.  Users will be confused why things stop working if they don't realize the Python version has changed if they ``module load ..`` anything that requires a different Python version and will not know that this script would need to be re-run.

Also, I'm wondering if this belongs in ``<REPO_ROOT>/integration/config`` instead of the spot it is in now?  This is just a helper script for building dev.  It won't be used for tagged releases.  Not sure it should be featured so prominently in the geopmdpy root dir.

bgeltz (changes requested):

geopmdpy/install_user.sh
```diff
@@ -0,0 +1,83 @@
...
+    printf "Usage: $0 --prefix=INSTALL_PREFIX ...
+
+Overview:
+
+When using pip to install a development version of geopmdpy, the ABI of
+libgeopmd.so used at runtime must provide all the interfaces required by
+geompdpy to avoid an import error.  The easiest way to ensure this is the case
+is to install a version of libgeopmd.so that matches the pip installed version.
+
+This script enables this use by by installing geopmdpy and libgeopmd.so based on
```
avilcheslopez:
```suggestion
This script enables this use by installing geopmdpy and libgeopmd.so based on
```

avilcheslopez (commented):
Would it make sense to also support doing this build & install from the current repo/tree? I'm thinking of a use case where someone might want to do this based on a local branch or local changes in-progress they might have in their repo.

geopmdpy/install_user.sh
```diff
@@ -0,0 +1,99 @@
...
+cd libgeopmd
+./autogen.sh
+./configure $@
+make -j
+make install
+INSTALL_PREFIX=$(grep '^prefix=' config.log | awk -F\' '{print $2}')
+export C_INCLUDE_PATH="$INSTALL_PREFIX/include"
+export LIBRARY_PATH="$INSTALL_PREFIX/lib:$INSTALL_PREFIX/lib64"
+cd ../geopmdpy
+python3 -m pip install .
```
cmcantalupo, line 95: fix all issues with python dependencies between pyproject.toml vs setup.cfg before merging, e.g. why is cffi build dependency missing on some systems, and why is prometheus-client not being installed at runtime?  Likely we need to duplicate more of the contents of pyproject.toml into setup.cfg

cmcantalupo: Alternately pip install -r requirements.txt until we understand why pyproject.toml is being ignored.

bgeltz:
This is failing presently on mcfly for both in-tree and out-of-tree usages.  I'm seeing the same error signature on borealis as well.

Here's in-tree:
```
[bgeltz@mcfly geopmdpy](user-build-geopmdpy)$ python3 -m venv ~/tmp/venv/geopmdpy_test
[bgeltz@mcfly geopmdpy](user-build-geopmdpy)$ source ~/tmp/venv/geopmdpy_test/activate

(geopmdpy_test) [bgeltz@mcfly geopmdpy](user-build-geopmdpy)$ ./install_user.sh --prefix=$HOME/geopm-build --disable-io-uring
...
+ cd ../geopmdpy
+ python3 -m pip install .
Processing /home/bgeltz/git/geopm/geopmdpy
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /home/bgeltz/tmp/venv/geopmdpy_test/bin/python3 /home/bgeltz/tmp/venv/geopmdpy_test/lib64/python3.6/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpiruecmgu
       cwd: /tmp/pip-req-build-axih9p96
  Complete output (31 lines):
    File "setup.py", line 14, in <module>
      version = get_version(f'{script_dir}/..')
    File "/tmp/pip-build-env-tdn9pvym/overlay/lib/python3.6/site-packages/setuptools_scm/__init__.py", line 146, in get_version
      _version_missing(config)
    File "/tmp/pip-build-env-tdn9pvym/overlay/lib/python3.6/site-packages/setuptools_scm/__init__.py", line 107, in _version_missing
      f"setuptools-scm was unable to detect version for {config.absolute_root}.\n\n"
  LookupError: setuptools-scm was unable to detect version for /tmp.

  Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

  For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj

  During handling of the above exception, another exception occurred:

    File "/home/bgeltz/tmp/venv/geopmdpy_test/lib64/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 280, in <module>
      main()
    File "/home/bgeltz/tmp/venv/geopmdpy_test/lib64/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/bgeltz/tmp/venv/geopmdpy_test/lib64/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 114, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-tdn9pvym/overlay/lib/python3.6/site-packages/setuptools/build_meta.py", line 163, in get_requires_for_build_wheel
      config_settings, requirements=['wheel'])
    File "/tmp/pip-build-env-tdn9pvym/overlay/lib/python3.6/site-packages/setuptools/build_meta.py", line 143, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-tdn9pvym/overlay/lib/python3.6/site-packages/setuptools/build_meta.py", line 158, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 18, in <module>
      with open(f'{script_dir}/{package_name}/VERSION', 'r') as fid:
  FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-req-build-axih9p96/geopmdpy/VERSION'
  ----------------------------------------
ERROR: Command errored out with exit status 1: /home/bgeltz/tmp/venv/geopmdpy_test/bin/python3 /home/bgeltz/tmp/venv/geopmdpy_test/lib64/python3.6/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpiruecm
gu Check the logs for full command output.
```

Workaround:
```
$ SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 ./install_user.sh --prefix=$HOME/geopm-build --disable-io-uring
```

We need to sort out why this isn't working as expected without the pretend version.

bgeltz:
I see the error I commented about earlier if I use the default version of pip that comes with a new venv:
```
(geopmdpy_test) [bgeltz@mcfly geopmdpy](dev)$ python3 -m pip --version
pip 20.2.3 from /home/bgeltz/tmp/venv/geopmdpy_test/lib64/python3.6/site-packages/pip (python 3.6)
```

It goes away if I upgrade to 21.3.1:
```
(geopmdpy_test) [bgeltz@mcfly geopmdpy](dev)$ python3 -m pip install --upgrade pip
```

It seems that the initial version of pip in either a fresh venv or the system provided version will result in the above error.

Can we add a minimum pip version to our requirements to force it to be installed and used when we build/install geopmdpy?

bgeltz (approved):

Issue #3642, opened 2024-11-06
geopmread/write requests sporadically fail


bgeltz:
**Describe the bug**
I tried to use geopmwrite to set a control and I expected it to work instead an error was presented.

**GEOPM version**
3.1

**Expected behavior**
```
[bgeltz@node1 ~]$ geopmwrite CPU_UNCORE_FREQUENCY_MIN_CONTROL board 0 2000000000
[bgeltz@node1 ~]$ echo $?
0
```

**Actual behavior**
```
[bgeltz@node1 ~]$ geopmwrite CPU_UNCORE_FREQUENCY_MIN_CONTROL board 0 2000000000
Error: cannot write control: <geopm> Invalid argument: PlatformIOImp::write_control(): control name "CPU_UNCORE_FREQUENCY_MIN_CONTROL" not found: at src/PlatformIO.cpp:828
```

**Error messages produced**
Upon inspection with a debug build, contention for dbus is observed:
```
node1 47232: Warning: <geopm> Failed to load SERVICE IOGroup.  GEOPM may not work properly unless an alternate IOGroup plugin is loaded to provide signals/controls required by the Controller and Agent.

node1 47232: The error was: <geopm> Runtime error: SDBus: Failed to call sd-bus function sd_bus_call_method(), error  :-110 name: org.freedesktop.DBus.Error.Timeout: Connection timed out: at src/SDBus.cpp:31

...

node1 47232: Error: cannot write control: <geopm> Invalid argument: PlatformIOImp::write_control(): control name "CPU_UNCORE_FREQUENCY_MIN_CONTROL" not found: at src/PlatformIO.cpp:828
```

**Additional context**
It seems we're overwhelming dbus with this request.  It's unclear how much other system traffic over dbus was occurring during the time of the failure.  Reproducing the failure is fairly sporadic but seems to be more likely during large scale jobs (i.e. greater than 2048 nodes).

cmcantalupo:
Work around posted here:
https://raw.githubusercontent.com/cmcantalupo/geopm/refs/heads/batch-write/geopmdpy/geopmwritebatch.py

bgeltz:
The above workaround does not address the dbus calls that occur during the initialization of the ServiceIOGroup/PlatformIO.  Setting the ``SYSTEMD_BUS_TIMEOUT`` environment variable to a value greater than 25 (the default) seems to allow the write requests to proceed as expected.
[...]

PR #3643, opened 2024-11-06
Implement geopmread and geopmwrite CLIs in python and add batch write interface

cmcantalupo:
- Relates to #3642


cmcantalupo:
Alternative would be to add a `--batch-config` option to geopmwrite.  This configuration file would be used instead of the positional arguments.

geopmdpy/geopmdpy/writebatch.py
```diff
@@ -0,0 +1,43 @@
...
+        if sys.argv[1] == '--help':
+            print(__doc__)
+            return 0
+        if sys.argv[1] == '--version':
+            print(__version_str__)
+            return 0
+        with open(sys.argv[1]) as input_stream:
+            run(input_stream)
+    else:
+        run(sys.stdin)
```
cmcantalupo:
```suggestion
        run(sys.stdin)
    return 0
```

geopmdpy/geopmdpy/write.py
```diff
@@ -0,0 +1,98 @@
...
+
+"""Command line interface to write controls through PlatformIO
+
+"""
+
+import sys
+from argparse import ArgumentParser
+from geopmdpy import pio
+from geopmdpy import topo
+from geopmdpy import __version_str__
```
cmcantalupo: Relative imports?

cmcantalupo:
TODO:

~The --config option for geopmwrite needs to be documented in geopmwrite.1.rst~

docs/source/geopmwrite.1.rst
```diff
@@ -122,6 +129,12 @@ Options
                 permissions no operation will be performed.  To force the
                 creation of a new cache file, remove the existing cache file
                 prior to executing this command.
+-f, --config    Read control name, control domain, control index and control value
+                from a configuration file rather than using the positional
+                arguments.  These four parameters is provided on each line of
+                the file separated by white space.  The file may have many lines
+                specifying multiple controls to be written. Proving ``-`` for this
+                option specifies to read the configuration from standard input.
```
dannosliwcd:
```suggestion
-f, --config    Read control name, control domain, control index and control value
                from a configuration file rather than using the positional
                arguments.  These four parameters are provided on each line of
                the file separated by white space.  The file may have many lines
                specifying multiple controls to be written. Providing ``-`` for this
                option specifies to read the configuration from standard input.
```

docs/source/geopmwrite.1.rst
```diff
@@ -190,9 +203,15 @@ Set all CPUs on package 0 to 1.5 *GHz* (cpu 1 is on package 0):
    $ geopmread CPU_FREQUENCY_MAX_CONTROL cpu 1
    1.5e9
 
+.. code-block::
```
dannosliwcd, line 210:
```suggestion
Set the uncore frequency of all CPUs to 1.5 GHz. Perform the operation in a single GEOPM batch-write request instead of across multiple `geopmwrite` commands:

.. code-block::
```

geopmdpy/geopmdpy/read.py
```diff
@@ -0,0 +1,96 @@
...
+    else:
+        parser.print_help()
+        return -1
+    return 0
+
+def main():
+    err = 0
+    try:
+        err = run()
+    except RuntimeError as ee:
```
dannosliwcd: What's special about RuntimeError? Why not catch all Exception types or none of them?

geopmdpy/geopmdpy/write.py
```diff
@@ -0,0 +1,107 @@
...
+                              help='Print version and exit.')
+    positional_group = parser.add_argument_group()
+    positional_group.add_argument('CONTROL_NAME', nargs='?',
+                                  help='Name of control')
+    positional_group.add_argument('DOMAIN_TYPE',  nargs='?',
+                                  help='Name of the domain for which the control should be written')
+    positional_group.add_argument('DOMAIN_INDEX', nargs='?', type=int,
+                                  help='Index of the domain, starting from 0')
+    positional_group.add_argument('VALUE', nargs='?', type=float,
+                                  help='Setting to adjust control to')
```
dannosliwcd: These are all `nargs='?'`. What's the case where we want some but not all of them?

geopmdpy/geopmdpy/write.py
```diff
@@ -0,0 +1,107 @@
...
+                              help='print longer description of all controls')
+    parser_group.add_argument('-c', '--cache', action='store_true',
+                              help='Create geopm topo cache if it does not exist')
+    parser_group.add_argument('-f', '--config',
+                              help='Path to configuration file with one write request per line, use "-" for stdin')
+    parser_group.add_argument('-e', '--enable-fixed', action='store_true',
+                              help='enable msr fixed counters')
+    parser_group.add_argument('-v', '--version', action='store_true',
+                              help='Print version and exit.')
+    positional_group = parser.add_argument_group()
```
dannosliwcd: Should the group of control name/type/index/value be exclusive of --config?

geopmdpy/geopmdpy/read.py
cmcantalupo: This is copy pasta from all of our command line interfaces.  The original thinking (maybe not a great thought) was that we use RuntimeError for things that are "bad, but not unexpected if the runtime environment is not set up right", where as other errors are "bad, and unexpected so better emit the stack trace".  

Now, is this true of all our thrown exceptions in the code base, maybe not.

geopmdpy/geopmdpy/read.py
```diff
@@ -0,0 +1,89 @@
...
+        topo.create_cache()
+    elif len(args.REQUEST) == 0:
+        print_signals()
+    elif len(args.REQUEST) == 3:
+        args.REQUEST[2] = int(args.REQUEST[2])
+        signal = pio.read_signal(*args.REQUEST)
+        print(pio.format_signal(signal, pio.signal_info(args.REQUEST[0])[1]))
+    else:
+        parser.error('When REQUEST is specified, all three parameters must be given: SIGNAL DOMAIN_TYPE DOMAIN_INDEX')
+        return -1
```
dannosliwcd: parser.error terminates with exit status code == 2 so the return has no effect

bgeltz (approved):

PR #3649, opened 2024-12-04
Avoid using group database unless needed

### LINKED ISSUE #3642: geopmread/write requests sporadically fail

cmcantalupo:
- Fixes #3642

### DIFF
--- a/geopmdpy/test/TestAccess.py
+++ b/geopmdpy/test/TestAccess.py
@@ -384,9 +384,10 @@ echo {self._signals_expect[0]} >> $1
 
         """
         group_id = 101
+        group_name = 'group_name'
         class MockGrp:
              def __init__(self):
-                 self.gr_name = 'group_name'
+                 self.gr_name = group_name
                  self.gr_passwd = 'x'
                  self.gr_gid = group_id
                  self.gr_mem = []
@@ -394,7 +395,7 @@ echo {self._signals_expect[0]} >> $1
         ret = MockGrp()
         self.assertEqual(expected, _group_id(expected))
         with mock.patch('grp.getgrnam', return_value=ret):
-            self.assertEqual(expected, _group_id('group_name'))
+            self.assertEqual(expected, _group_id(group_name))
 
 if __name__ == '__main__':
     unittest.main()
--- a/geopmdpy/test/TestAccess.py
+++ b/geopmdpy/test/TestAccess.py
@@ -7,9 +7,10 @@
 import unittest
 import os
 import tempfile
+import grp
 from unittest import mock
 
-from geopmdpy.access import Access
+from geopmdpy.access import Access, _group_id
 from geopmdpy.system_files import GEOPM_SERVICE_LOG_REQUEST
 
 class TestAccess(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


geopmdpy/test/TestAccessLists.py
```diff
@@ -203,23 +199,23 @@ def test_get_group_access_default(self):
         self.assertEqual(set(self._controls_expect), set(controls))
 
     def test_get_group_access_named(self):
```
avilcheslopez: Minor note: maybe this test name should change. Same note applies to `test_set_group_access_named`.

Also, perhaps there should be another test that validates the new flexible behavior in which `geopmaccess` accepts either a group name or ID.

cmcantalupo: We don't currently get coverage of geopmdpy.access.main where the CLI is parsed, and the group is converted.  That is a bigger problem...I've updated the names

cmcantalupo: Moved the implementation into function outside of main() called geopmdpy.access._group_id() and wrote a unit test that calls it.

geopmdpy/test/TestAccess.py
```diff
@@ -378,6 +379,22 @@ def test_read_log_controls(self):
...
+             def __init__(self):
+                 self.gr_name = 'group_name'
+                 self.gr_passwd = 'x'
+                 self.gr_gid = group_id
+                 self.gr_mem = []
+        expected = str(group_id)
+        ret = MockGrp()
+        self.assertEqual(expected, _group_id(expected))
+        with mock.patch('grp.getgrnam', return_value=ret):
+            self.assertEqual(expected, _group_id('group_name'))
```
avilcheslopez: Sorry, the slightest hint of duplicate code bugs me. Feel free to disregard.

```suggestion
        group_id = 101
        group_name = 'group_name'
        class MockGrp:
             def __init__(self):
                 self.gr_name = group_name
                 self.gr_passwd = 'x'
                 self.gr_gid = group_id
                 self.gr_mem = []
        expected = str(group_id)
        ret = MockGrp()
        self.assertEqual(expected, _group_id(expected))
        with mock.patch('grp.getgrnam', return_value=ret):
            self.assertEqual(expected, _group_id(group_name))
```

avilcheslopez (approved):

PR #3650, opened 2024-12-06
Fork and use pio, do not popen geopmread/geopmwrite

cmcantalupo:
- Now that geopmread and geopmwrite are python3 implementations avoid creating a new python interpreter by using fork.

### DIFF
--- a/geopmdpy/geopmdpy/system_files.py
+++ b/geopmdpy/geopmdpy/system_files.py
@@ -986,32 +986,31 @@ class ActiveSessions(object):
             self._update_session_file(client_pid)
         else:
            os.rename(sess_path, renamed_path)
-
-def _get_names(exec_name, quiet=True):
-    try:
-        stdout_encoding = sys.stdout.encoding
-        stderr_encoding = sys.stderr.encoding
-    except AttributeError:
-        stdout_encoding = None
-        stderr_encoding = None
-    if stdout_encoding is None:
-        stdout_encoding = locale.getpreferredencoding()
-    if stderr_encoding is None:
-        stderr_encoding = locale.getpreferredencoding()
-
-    try:
-        stderr_encoding = sys.stderr.encoding
-    except AttributeError:
-        stderr_encoding = None
-    if stderr_encoding is None:
-        stderr_encoding = locale.getpreferredencoding()
-    signals_pid = subprocess.Popen(exec_name,
-                                   stdout=subprocess.PIPE,
-                                   stderr=subprocess.PIPE)
-    stdout_bytes, stderr_bytes = signals_pid.communicate()
-    if not quiet:
-        sys.stderr.write(stderr_bytes.decode(encoding=stderr_encoding))
-    return stdout_bytes.decode(encoding=stdout_encoding).split()
+def _get_names():
+    marker = '\n\nBREAK\n\n'
+    pipe_r, pipe_w = os.pipe()
+    pid = os.fork()
+    if pid != 0:
+        os.close(pipe_w)
+        with os.fdopen(pipe_r) as pipe_ro:
+            buffer = pipe_ro.read()
+        os.waitpid(pid, 0)
+        signal_buffer, control_buffer = buffer.split(marker)
+        signals = []
+        controls = []
+        if signal_buffer != '':
+            signals = signal_buffer.split('\n')
+        if control_buffer != '':
+            controls = control_buffer.split('\n')
+        return signals, controls
+    else:
+        os.close(pipe_r)
+        signal_buffer = '\n'.join(pio.signal_names())
+        control_buffer = '\n'.join(pio.control_names())
+        buffer = f'{signal_buffer}{marker}{control_buffer}'
+        with os.fdopen(pipe_w, 'w') as pipe_wo:
+            pipe_wo.write(buffer)
+        os._exit(0)
 
 
 class AccessLists(object):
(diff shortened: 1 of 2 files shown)


geopmdpy/geopmdpy/system_files.py
```diff
@@ -986,32 +986,31 @@ def _load_session_file(self, sess_path):
...
-        stderr_encoding = locale.getpreferredencoding()
-    signals_pid = subprocess.Popen(exec_name,
-                                   stdout=subprocess.PIPE,
-                                   stderr=subprocess.PIPE)
-    stdout_bytes, stderr_bytes = signals_pid.communicate()
-    if not quiet:
-        sys.stderr.write(stderr_bytes.decode(encoding=stderr_encoding))
-    return stdout_bytes.decode(encoding=stdout_encoding).split()
+def _get_names():
+    marker = '\nBREAK\n'
```
cmcantalupo:
```suggestion
    marker = '\nBREAK\n\n'
```

geopmdpy/geopmdpy/system_files.py
```diff
@@ -986,32 +986,31 @@ def _load_session_file(self, sess_path):
...
-    return stdout_bytes.decode(encoding=stdout_encoding).split()
+def _get_names():
+    marker = '\n\nBREAK\n\n'
+    pipe_r, pipe_w = os.pipe()
+    pid = os.fork()
+    if pid != 0:
+        os.close(pipe_w)
+        with os.fdopen(pipe_r) as pipe_ro:
+            buffer = pipe_ro.read()
+        os.waitpid(pid, 0)
```
bgeltz: The return code of the child should be captured here and examined for success.  If there was an error, an exception should be raised to ensure geopmd dies and a nice message is written to the journal.

geopmdpy/geopmdpy/system_files.py
```diff
@@ -986,32 +986,37 @@ def _load_session_file(self, sess_path):
...
-                                   stdout=subprocess.PIPE,
-                                   stderr=subprocess.PIPE)
-    stdout_bytes, stderr_bytes = signals_pid.communicate()
-    if not quiet:
-        sys.stderr.write(stderr_bytes.decode(encoding=stderr_encoding))
-    return stdout_bytes.decode(encoding=stdout_encoding).split()
+def _get_names():
+    marker = '\n\nBREAK\n\n'
+    pipe_r, pipe_w = os.pipe()
+    pid = os.fork()
```
avilcheslopez: Would we want to catch any exceptions raised here to have an opportunity to provide a friendlier error message to the user? As I'm seeing in the documentation, `os.fork()` can raise `OSError`.

This question also applies to other `os` calls that might raise exceptions.

geopmdpy/geopmdpy/system_files.py
```diff
@@ -986,32 +986,37 @@ def _load_session_file(self, sess_path):
...
+    if pid != 0:
+        os.close(pipe_w)
+        with os.fdopen(pipe_r) as pipe_ro:
+            buffer = pipe_ro.read()
+        _, exit_status = os.waitpid(pid, 0)
```
avilcheslopez, line 1000: I'm wondering if it would be a good idea to manage hangs here. Unfortunately, I see `os.waitpid()` doesn't provide a convenient way to specify a timeout. We could leverage the options `os.WNOHANG`, `os.WNOWAIT` to implement such a mechanism ourselves. Although, it might be overkill and it might be better to just let the user `ctrl+c` the process in case of a hang.

geopmdpy/geopmdpy/system_files.py
```diff
@@ -986,32 +986,37 @@ def _load_session_file(self, sess_path):
...
+        if os.WIFEXITED(exit_status):
+            exit_code = os.WEXITSTATUS(exit_status)
+            if exit_code != 0:
+                raise ChildProcessError(f'Child process to read signal and control names failed with exit status: {exit_code}')
+        else:
+            raise ChildProcessError('Child process to read signal and control names terminated abnormally')
```
avilcheslopez: Two notes:

1. I would recommend to let `os` raise `ChildProcessError`, and raise `RuntimeError` (as we're doing in most places in `geopmdpy`). My reasoning is: let `os` decide what merits raising a `ChildProcessError`. If a caller relies on `os`'s documentation of which errors are raised by what error conditions, it might get "confused" when an error is raised for a condition it didn't expect. This is just my thinking, please feel free to disagree; it's certainly not highly consequential.

2. I would recommend a friendlier error message for the user. Exposing the user to a low-level error message exposes implementation details to the user and doesn't give them enough context of what went wrong. I do see `_get_names()` is a lower-level helper function. So if not handling this here, perhaps handling it at its call sites is appropriate (catching its exceptions and providing the friendlier error message there, at a higher-level context).

avilcheslopez (commented):
I'm just curious: why weren't we using `pio` before?

geopmdpy/geopmdpy/system_files.py
cmcantalupo: I'm not sure what situation would cause a hang, but more importantly, I'm not sure what sort of time out would be appropriate.  We know this can be pretty slow when reading from the service with levelzero enabled.

geopmdpy/geopmdpy/system_files.py
cmcantalupo: This is only called in the context of geopmd starting up.  All messages are printed to the syslog for administrators to manage (not end users at a terminal).  This code path will not be executed outside of the context of the geopmd process.

bgeltz (approved):

Issue #3651, opened 2024-12-07
msr_data_spr.json has the wrong scaling factor for DRAM_ENERGY_STATUS:ENERGY


dannosliwcd:
**Describe the bug**
I tried to read DRAM power on an SPR system and it was 1/4 of the expected value.

**GEOPM version**
3.1.0

**Expected behavior**
`geopmread DRAM_POWER package 0` should print the DRAM power consumption on CPU 0.

**Actual behavior**
On SPR, `geopmread DRAM_POWER package 0` prints 1/4 the DRAM power consumption on CPU 0.

**Error messages produced**
None

**Additional context**
The current scaling factor for DRAM_ENERGY_STATUS:ENERGY in msr_data_spr.json is `1.5258789063e-05`. It should be the same as the value in PKG_ENERGY_STATUS (`6.103515625e-05`). 

Even better, the MSRIOGroup could read MSR::RAPL_POWER_UNIT:ENERGY to query the scaling factor from the target platform.

bgeltz:
Additional context on how we got into this state:

``1.5258789063e-05`` got into the codebase during the v1.0 release:
https://github.com/geopm/geopm/commit/2e37b3f5fedf5725004e7f82847eac28563e70d2

Comments indicate that the MSR provided data was incorrect for ``MSR::RAPL_POWER_UNIT:ENERGY``: https://github.com/geopm/geopm/pull/497#issuecomment-457769208

The above commit changed the scalar to ``1.5258789063e-05`` for HSX, KNL, SKX, and SNB.  This value persisted through the JSON re-write of all of this data, and was copied into the SPR description file from SKX when it was created for release v3.0: https://github.com/geopm/geopm/commit/1b8e623c53509663e7ceec6e1a475474a58a05ff

PR #3653, opened 2024-12-10
Add integration test to verify PKG/DRAM ENERGY_STATUS scalars

bgeltz:
- Relates to #3651.
- Provide scalar in MSRIOGroup::signal_description().

### DIFF
--- a/libgeopmd/src/MSRIOGroup.cpp
+++ b/libgeopmd/src/MSRIOGroup.cpp
@@ -425,7 +425,8 @@ namespace geopm
                                                    "\n    alias_for: Temperature derived from PROCHOT and "
                                                    + ts.msr_name,
                                                    IOGroup::M_SIGNAL_BEHAVIOR_VARIABLE,
-                                                   string_format_double};
+                                                   string_format_double,
+                                                   1};
             }
         }
     }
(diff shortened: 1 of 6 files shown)


cmcantalupo (dismissed):

libgeopmd/src/MSRIOGroup.cpp
```diff
@@ -1109,7 +1116,8 @@ namespace geopm
             result += "    units: " + IOGroup::units_to_string(it->second.units) + '\n';
             result += "    aggregation: " + Agg::function_to_name(it->second.agg_function) + '\n';
             result += "    domain: " + m_platform_topo.domain_type_to_name(it->second.domain) + '\n';
-            result += "    iogroup: MSRIOGroup";
+            result += "    iogroup: MSRIOGroup\n";
+            result += "    scalar: " + std::to_string(it->second.scalar);
```
asmaalrawi, line 1120: Provide docs to help user understand "scalar"

bgeltz:
Marking this back as draft for now.
I started working on the docu, and realized that there are other IOGroups that do not provide the scalar: CpuInfoIOGroup, CNLIOGroup, LevelZeroIOGroup, etc.  If we're going to write general documentation for all these, then all IOGroups should support this.

That being said, the scalar is really an implementation detail that the user does not care about.  IMO, the signal description is not the correct spot for this data.  This should be a separate API that the developers have access to for writing this sort of test/verification.  We should add a new API to the IOGroup base class for this support, then make it accessible in Python for the test.

asmaalrawi:
LevelZeroIOGroup uses scalar but it is not included in the description, as I don't think it is useful to the end user for understanding the signal/control.  Is this the desired pattern?

PR #3671, opened 2025-01-15
Update open PBS integration scripts to work with agents other than node power governor

dannosliwcd:


integration/service/open_pbs/power_sweeps/sweep_hacc_board_power.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+set -e
+source ~/.geopmrc
+HACC_RUN_DIR="${HACC_RUN_DIR?-Set HACC_RUN_DIR to the location of run_mpiexec.sh}"
+GEOPM_SOURCE="${GEOPM_SOURCE?-Set GEOPM_SOURCE to the location of geopm source code}"
+source "${GEOPM_SOURCE}/integration/config/run_env.sh"
+PROFILE_NAME='hacc'
+EXPERIMENT_NAME="${PROFILE_NAME}-geopm-power-sweep"
+TRIAL_COUNT="${TRIAL_COUNT-5}"
```
sidjana: Missing colons in the following lines

Bash syntax for default initialization:
var="${var:-default_val}"

cmcantalupo (approved):
LGTM

PR #3697, opened 2025-02-10
Add support for powercap sysfs

### LINKED ISSUE #3538: As a user of PlatformIO I would like support for the Linux powercap sysfs interfaces
I'm always frustrated when I want to measure power and energy or set power caps on platforms that to not support the MSR RAPL interface.  

It would be great if there were an sysfs driver like
https://github.com/geopm/geopm/blob/dev/libgeopmd/src/CpufreqSysfsDriver.cpp 
that provided access to the powercap interfaces.

Most of the implementation is already provided by the SysfsIOGroup, so writing another driver for it seems like the simplest approach.

**Additional context**
https://www.kernel.org/doc/html/next/power/powercap/powercap.html

cmcantalupo:
- Fixes #3538
- [x] Unit tests have been mostly erased, need to add them back
- [x] CI is broken

### DIFF
--- a/libgeopmd/src/IOGroup.cpp
+++ b/libgeopmd/src/IOGroup.cpp
@@ -19,6 +19,7 @@
 #include "geopm/Helper.hpp"
 #include "CpufreqSysfsDriver.hpp"
 #include "DrmSysfsDriver.hpp"
+#include "PowercapSysfsDriver.hpp"
 #ifdef GEOPM_ENABLE_SYSTEMD
 #include "ServiceIOGroup.hpp"
 #endif
@@ -98,6 +99,8 @@ namespace geopm
             // msr driver once it is considered more stable.
             register_plugin(CpufreqSysfsDriver::plugin_name(),
                             CpufreqSysfsDriver::make_plugin);
+            register_plugin(PowercapSysfsDriver::plugin_name(),
+                            PowercapSysfsDriver::make_plugin);
 
 #ifdef GEOPM_ENABLE_CPUID
 #ifdef GEOPM_ENABLE_RAWMSR
@@ -147,6 +150,8 @@ namespace geopm
             // loaded.
             register_plugin(CpufreqSysfsDriver::plugin_name(),
                             CpufreqSysfsDriver::make_plugin);
+            register_plugin(PowercapSysfsDriver::plugin_name(),
+                            PowercapSysfsDriver::make_plugin);
 #ifdef GEOPM_ENABLE_SYSTEMD
             register_plugin(ServiceIOGroup::plugin_name(),
                             ServiceIOGroup::make_plugin);
--- a/libgeopmd/src/IOGroup.cpp
+++ b/libgeopmd/src/IOGroup.cpp
@@ -96,7 +96,7 @@ namespace geopm
         // fail.
         if (geopm::has_cap_sys_admin()) {
             // May want to give Cpufreq and Powercap higher priority than the
-            // non-safe msr driver once it is considered more stable.  The
+            // non-safe msr driver once they are considered more stable.  The
             // msr-safe ioctl() interface is the highest performance one
             // available for these signals/controls
             register_plugin(CpufreqSysfsDriver::plugin_name(),
--- a/libgeopmd/src/IOGroup.cpp
+++ b/libgeopmd/src/IOGroup.cpp
@@ -95,8 +95,10 @@ namespace geopm
         // service is not active then loading the ServiceIOGroup will
         // fail.
         if (geopm::has_cap_sys_admin()) {
-            // May want to give this higher priority than the non-safe
-            // msr driver once it is considered more stable.
+            // May want to give Cpufreq and Powercap higher priority than the
+            // non-safe msr driver once it is considered more stable.  The
+            // msr-safe ioctl() interface is the highest performance one
+            // available for these signals/controls
             register_plugin(CpufreqSysfsDriver::plugin_name(),
                             CpufreqSysfsDriver::make_plugin);
             register_plugin(PowercapSysfsDriver::plugin_name(),
@@ -141,13 +143,13 @@ namespace geopm
 #endif
         }
         else { // not UID 0
-            // Prefer the Service provided signals to the Cpufreq
-            // provided signals in case MSR access is available from
-            // the service.  MSR values for aliases are both more
-            // accurate and can be read faster.  Note this means even
-            // low level signals like CPUFREQ::SCALING_CUR_FREQ will
-            // be read through the service if the ServiceIOGroup is
-            // loaded.
+            // Prefer the Service provided signals/controls to the Cpufreq or
+            // Powercap provided signals/controls in case MSR access is
+            // available from the service.  MSR values for aliases are both more
+            // accurate and can be read faster.  Note this means even low level
+            // signals like CPUFREQ::SCALING_CUR_FREQ or
+            // POWERCAP::CPU_ENERGY_CONSUMED will be read through the service if
+            // the ServiceIOGroup is loaded.
             register_plugin(CpufreqSysfsDriver::plugin_name(),
                             CpufreqSysfsDriver::make_plugin);
             register_plugin(PowercapSysfsDriver::plugin_name(),
--- a/libgeopmd/src/PowercapSysfsDriver.cpp
+++ b/libgeopmd/src/PowercapSysfsDriver.cpp
@@ -1,0 +1,196 @@
+/*
+ * Copyright (c) 2015 - 2025 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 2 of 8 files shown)


libgeopmd/test/PowercapSysfsDriverTest.cpp
```diff
@@ -0,0 +1,200 @@
+/*
+ * Copyright (c) 2015 - 2024 Intel Corporation
```
bgeltz:
```suggestion
 * Copyright (c) 2015 - 2025 Intel Corporation
```

libgeopmd/src/IOGroup.cpp
```diff
@@ -98,6 +99,8 @@ namespace geopm
             // msr driver once it is considered more stable.
             register_plugin(CpufreqSysfsDriver::plugin_name(),
                             CpufreqSysfsDriver::make_plugin);
+            register_plugin(PowercapSysfsDriver::plugin_name(),
+                            PowercapSysfsDriver::make_plugin);
```
avilcheslopez, line 105: Question: since this plugin is going to be registered in both branches of the `if` statement (i.e., regardless of whether the user has cap sysadmin), why not have this statement just once outside of the `if`?

I do see the same pattern was followed for `CpufreqSysfsDriver`, so my question applies to both.

libgeopmd/src/PowercapSysfsDriver.hpp
```diff
@@ -0,0 +1,37 @@
...
+            std::string driver(void) const override;
+            std::map<std::string, properties_s> properties(void) const override;
+            static std::string plugin_name(void);
+            static std::unique_ptr<IOGroup> make_plugin(void);
+
+        private:
+            const std::map<std::string, properties_s> M_PROPERTIES;
+            const std::map<std::string, std::string> M_POWERCAP_RESOURCE_BY_NAME;
+            std::map<std::string, int> m_domain_map;
+            const std::string m_powercap_directory;
```
avilcheslopez: Naming convention for const member

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -0,0 +1,196 @@
...
+        int read_bytes = pread(name_fd, name_buf, sizeof(name_buf) - 1, 0);
+        close(name_fd);
+        if (read_bytes < 0) {
+            throw geopm::Exception("PowercapSysfsDriver failed to read " + name_path,
+                                   errno, __FILE__, __LINE__);
+        }
+        if (read_bytes == sizeof(name_buf) - 1) {
+            throw geopm::Exception("PowercapSysfsDriver truncated read from " + name_path,
+                                   errno, __FILE__, __LINE__);
+        }
```
avilcheslopez, line 52: I'm seeing there is opportunity to factor out common code with `CpufreqSysfsDriver`, since I'm seeing the good majority of the code in this function is practically the same. At the very least, this section of code could be factored out into a helper function somewhere that both drivers could leverage.

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -0,0 +1,196 @@
...
+    int PowercapSysfsDriver::domain_type(const std::string &name) const
+    {
+        if (M_PROPERTIES.find(name) == M_PROPERTIES.end()) {
+            throw Exception("PowercapSysfsDriver: unknown name: " + name,
+                            GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+        }
+        return GEOPM_DOMAIN_PACKAGE;
+    }
+
+    static std::string name_to_resource_key(const std::string &name, int domain_idx)
```
avilcheslopez: Minor recommendation: could these static functions be moved to the top (to facilitate reading the code)? It might just throw someone off a bit when they're going through class functions, and then encounter a static function not part of the class. It might just be me, so feel free to ignore this suggestion.

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -0,0 +1,196 @@
...
+        }
+        return resource_key;
+    }
+
+    static std::string name_to_property_key(const std::string &name)
+    {
+        std::string property_key;
+        auto substr_pos = name.find("POWERCAP::CPU_");
+        if (substr_pos == std::string::npos) {
+            substr_pos = name.find("POWERCAP::DRAM_");
```
avilcheslopez: Minor suggestion: seeing these literal strings (passed to the `find` function) are used in more than one place, I would recommend putting them in a constant.

libgeopmd/test/PowercapSysfsDriverTest.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+TEST_F(PowercapSysfsDriverTest, driver_and_plugin_name_match)
+{
+    EXPECT_EQ("POWERCAP", m_driver->driver())
+        << "Driver name should match the plugin name";
+    EXPECT_EQ("POWERCAP", PowercapSysfsDriver::plugin_name())
+        << "Plugin name should be POWERCAP";
+}
```
avilcheslopez: It seems this is a duplicate of the first test (`iogroup_plugin_name_matches_driver_name`).

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+        int name_fd = open(name_path.c_str(), O_RDONLY);
+        if (name_fd == -1) {
+            throw geopm::Exception("PowercapSysfsDriver failed to open " + name_path,
+                                   errno, __FILE__, __LINE__);
+        }
+        char name_buf[geopm::SysfsDriver::M_IO_BUFFER_SIZE] = {};
+        int read_bytes = pread(name_fd, name_buf, sizeof(name_buf) - 1, 0);
+        close(name_fd);
+        if (read_bytes < 0) {
```
avilcheslopez, line 45: Is it possible that `pread` in this case could ever return 0 bytes? I'm guessing the answer is highly unlikely, but perhaps we should manage that scenario just in case.

libgeopmd/src/IOGroup.cpp
cmcantalupo: The load order effects which aliases have priority.  The reason for the pattern is to ensure the load order (last IOGroup providing a high level alias wins).

avilcheslopez: Is it that you need to make sure `PowercapSysfsDriver` is loaded after `CpufreqSysfsDriver`? If so, seeing how they're both loaded first in each of the `if` branches, you could move those two `register_plugin()` calls up right before the `if` statement, preserving the order between them and with respect to the rest of the plugins.

libgeopmd/src/PowercapSysfsDriver.cpp
cmcantalupo: "The name_buf array is initialized to all zeroes. This is because the {} initializer in C++ initializes the array elements to their default value, and for a char type, the default value is 0 (null character '\0')."

- Copilot

So in the case of read_bytes == 0, the name_buf should begin with '\0' and thus be converted to an empty std::string name below

libgeopmd/src/IOGroup.cpp
cmcantalupo: I tried to update the comments about cpufreq vs msr support to include powercap in description.  Still should consider prefering cpufreq and powercap vs stock msr driver.

libgeopmd/src/PowercapSysfsDriver.cpp
cmcantalupo: Although there are similarities between load_powercap_resource_by_name() and load_cpufreq_cpus_by_resource(), there are some key differences.  The code that is common, is that we are opening all of the files in a directory, reading the contents and doing something depending on the contents.  Much of the work to do that is already factored in geopm::list_directory_files().  

The key differences are the use of the "name" file in the case of powercap, and "affected_cpus" in the cpufreq case to understand the topology.  These file contents are quite different.  I don't see a really easy way to share the code without using std::functions() to be applied to the contents of the files, but that seems overly complex and difficult to read.

libgeopmd/src/IOGroup.cpp
```diff
@@ -94,10 +95,14 @@ namespace geopm
         // service is not active then loading the ServiceIOGroup will
         // fail.
         if (geopm::has_cap_sys_admin()) {
-            // May want to give this higher priority than the non-safe
-            // msr driver once it is considered more stable.
+            // May want to give Cpufreq and Powercap higher priority than the
+            // non-safe msr driver once it is considered more stable.  The
```
cmcantalupo:
```suggestion
            // non-safe msr driver once they are considered more stable.  The
```

bgeltz (approved):

PR #3705, opened 2025-02-20
Add -D option to geopmread and geopmwrite

### LINKED ISSUE #3667: geopmread --domain <signal> does not return the signal's native domain
**Describe the bug**
I tried to read the native domain of DRAM_ENERGY using `geopmread --domain DRAM_ENERGY` and I expected "memory". Instead, it returned an error and usage information

**GEOPM version**

**On tip of dev:**
```

[bgeltz@mcfly2 geopm](dev)$ geopmread --domain DRAM_ENERGY
usage: geopmread [-h] [-v] [-d | -i INFO | -I | -c | REQUEST [REQUEST ...]]
geopmread: error: argument REQUEST: not allowed with argument -d/--domain
```

**on 3.1.0:**

```
[bgeltz@mcfly2 ~]$ which geopmread/usr/bin/geopmread
[bgeltz@mcfly2 ~]$ geopmread --version
3.1.0

Copyright (c) 2015 - 2024 Intel Corporation. All rights reserved.

[bgeltz@mcfly2 ~]$ geopmread --domain DRAM_ENERGY
board                       1
package                     2
core                        44
cpu                         88
memory                      2
package_integrated_memory   0
nic                         0
package_integrated_nic      0
gpu                         0
package_integrated_gpu      0
gpu_chip                    0

```

cmcantalupo:
- Fixes #3667
- Prints the native domain of a given signal or control
- Add wildcard option for geopmsession domain type parameter

### DIFF
--- a/docs/source/geopmread.1.rst
+++ b/docs/source/geopmread.1.rst
@@ -99,28 +99,30 @@ exists.  See the ``--cache`` option below for more information.
 
 Options
 -------
--d, --domain    Print a list of all domains on the system.
--i, --info      Print description of the provided ``SIGNAL_NAME``.
--I, --info-all  Print a list of all available signals with their descriptions,
-                if any.
--c, --cache     Create a cache file for the ``geopm::PlatformTopo`` object if one
-                does not exist or if the existing cache is from a previous boot
-                cycle.  If a privileged user requests this option (e.g. root or
-                if invoked with sudo) the file path will be
-                ``/run/geopm/geopm-topo-cache``. If a non-privileged user requests
-                this option the file path will be ``/tmp/geopm-topo-cache-<UID>``.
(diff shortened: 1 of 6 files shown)


dannosliwcd (commented):
Looks good. What's the motivation for implementing a different solution than the proposed ones on #3667?

cmcantalupo:
> Looks good. What's the motivation for implementing a different solution than the proposed ones on #3667?

Adding the new option rather than reusing -d avoids having conditional logic in the documentation for the -d option (if a positional argument is given it does X otherwise it does Y).  It also avoids conditional documentation for the positional argument (if -d is provided then one positional argument is required, otherwise three/four positional arguments are required and in both cases the first positional argument is the signal/control name, in the other case ...)

In short, the documentation was easier to write and read this way.

dannosliwcd (approved):

PR #3708, opened 2025-02-25
When user specifies --pid to geopmsession, do not also require a timeout

cmcantalupo:
Fix issues with signal handler

### DIFF
--- a/geopmdpy/geopmdpy/session.py
+++ b/geopmdpy/geopmdpy/session.py
@@ -298,8 +298,6 @@ class Session:
                           option is negative.
 
         """
-        if pid is None and period > run_time:
-            raise RuntimeError('Specified a period that is greater than the total run time')
         if period > 86400:
             raise RuntimeError('Specified a period greater than 24 hours')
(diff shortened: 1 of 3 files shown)


geopmdpy/geopmdpy/session.py
```diff
@@ -614,8 +617,8 @@ def get_parser():
                         help='Print version and exit.')
     parser.add_argument('-t', '--time', dest='time', type=float, default=0.0,
                         help='Total run time of the session to be opened in seconds.')
-    parser.add_argument('-p', '--period', dest='period', type=float, default=0.0,
-                        help='When used with a read mode session reads all values out periodically with the specified period in seconds.')
+    parser.add_argument('-p', '--period', dest='period', type=float, default=0.1,
```
dannosliwcd, line 620: Does this break one-shot mode due to the check on line 301?

cmcantalupo: Yes...yes it does

dannosliwcd (approved):

PR #3718, opened 2025-02-28
Add non_mpi_profiling example and test

### LINKED ISSUE #3674: Enable validation of non-MPI app profiling
> estimate 1
Parent [feature/story]: #XXXX

This was all previously tied to test_multi_app.sh.  On dev, that script has been extended to do much more that is required for validation.  Also, the old test plan references the content as it used to be.

Done criteria:
- [ ] Create "new" test_multi_app.sh from Daniel's aurora branch: https://github.com/dannosliwcd/geopm/blob/aurora/integration/test/test_multi_app.sh
- [ ] Update the test plan on that branch to reflect the new script: https://github.com/dannosliwcd/geopm/tree/aurora/integration/test/plan
- [ ] Create a new Python test that uses the new script.
- [ ] Provide test_multi_app.sh example that launches a Python script and profiles it with geopmctl

### LINKED ISSUE #3698: First use of geopmread gets increasingly slower as job scale increases
**Describe the bug**
I tried to use ``geopmread`` on a large scale job (>= 512 nodes) and I expected it to complete successfully instead a Dbus timeout error was encountered.  When I attempted to work around the issue by setting a larger value for ``SYSTEMD_BUS_TIMEOUT``, I observed that the more nodes are in my job the slower ``geopmread``'s first invocation performs.

Note that this bug is about ``geopmread``, but this slowdown can be observed with any interaction with ``geopmd`` after it has been restarted.  E.g. when using ``geopmctl`` to profile an application, ``geopmwrite``, ``geopmsession``, etc.

**GEOPM version**
v3.1

**Expected behavior**
Setup a job scipt as follows.  First, create a script with the following contents to be launched on a compute node:
```
#!/bin/bash

time geopmread BOARD_POWER board 0

```

Setup a batch script with your desired resource manager to launch the job on each node in your allocation.  Both mpiexec and clush were separately used to launch the above script on the compute nodes for this investigation.

Examining the "real" time (if using the bash built-in time) or "Elapsed (wall clock)" time if using /usr/bin/time, the value reported back on each individual node is relatively consistent.

**Actual behavior**
The time reported back increases based on the number of nodes in the allocation.  This happens with both mpiexec and clush.  Beyond a certain point (~512 nodes), you must set ``SYSTEMD_BUS_TIMEOUT`` to a value large enough to encompass the slowdown.  If you do not, you will encounter the following timeout error:
```
Warning: <geopm> Failed to connect with geopmd, running without geopm. Error: <geopm> Runtime error: SDBus: Failed to call sd-bus function sd_bus_call_method(), error:-110 name: org.freedesktop.DBus.Error.Timeout: Connection timed out: at src/SDBus.cpp:31.
```

**Additional context**
Related old issues where similar issues were encountered:
#3642 - In that issue, ``SYSTEMD_BUS_TIMEOUT`` was suggested but was ultimately abandoned since the LDAP issue was discovered.  In reality, I think we were seeing both this issue and the LDAP issue, and I was incorrect to call the ``SYSTEM_BUS_TIMEOUT`` a red herring - https://github.com/geopm/geopm/issues/3642#issuecomment-2515715053

---

Several attempts to isolate the slowdown were made:

First, since LevelZero can be leveraged without privelidge, a build of libgeopmd.so was performed such that the ServiceIOGroup was not loaded, and the LevelZeroIOGroup *was* loaded directly as the user.  This setup does not involve the use of the geopmd service at all, so there would be no dbus communication.  In this case, the script to be executed on the compute nodes was modified as follows:
```
#!/bin/bash

ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE ZES_ENABLE_SYSMAN=1 /usr/bin/time -v geopmread LEVELZERO::GPU_UTILIZATION board 0
```

The time data gathered this way did not show a scale-based slowdown.  All nodes were relatively consistent in their timing.

bgeltz:
- Resolves #3674.
- Resolves #3698.

### DIFF
--- a/integration/test/plan/README.md
+++ b/integration/test/plan/README.md
@@ -1,14 +1,38 @@
-GEOPM Integration Test Plan
-===========================
+# GEOPM Integration Test Plan
 
 This directory contains the test plan for validating specific
 features within GEOPM.
 
-Prerequisites
--------------
+## Prerequisites
 
-Ensure you have completed the creation of your `~/.geopmrc` file
-described in the [integration README](../../README.md#prerequisites).
+The tests described in this plan require GEOPM's systemd Service and Runtime to
+be made available to the user.  A typical configuration involves installing
+latest tagged release of the service globally, and leveraging a user build for
+providing the GEOPM Runtime.  For information on installing the service via
+system packages, see the [Install Guide](https://geopm.github.io/install.html).
+
+### HPC
+
+In an HPC system context, the GEOPM Service should be integrated into the
+compute image and configured appropriately.  The GEOPM Runtime can be made
+available to users via the spack recipe here:
+[geopm-runtime](https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/geopm-runtime).
+If spack has been used to build a geopm-runtime module, the user can simply
+issue `module load geopm-runtime` to setup the necessary environment.  Ensure
+the expected version is in use after loading the module:
+```bash
+[bgeltz@login_node1 ~]$ geopmctl --version
+3.1.0
+
+
+Copyright (c) 2015 - 2024 Intel Corporation. All rights reserved.
+```
+
+### Development Snapshot
+
+Certain test scenarios may require a user-build from the main development
+branch (dev) of GEOPM on GitHub.  The following steps can be used to build all
+lyaers from source when not utilizing spack:
 
 > [!NOTE]
 > When setting up your `~/.geopmrc` file, use the following example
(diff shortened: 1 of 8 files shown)


integration/test/plan/README.md
```diff
@@ -11,6 +11,11 @@ Ensure you have completed the creation of your `~/.geopmrc` file
 described in the [integration README](../../README.md#prerequisites).
 
 > [!NOTE]
+> `~/.geopmrc` is only needed if utilizing `run_env.sh` to setup your
```
cmcantalupo: Introduce the prerecs for a tagged release that is installed globally or via a system installed module.  Then explain how to run this test using a development snapshot.

integration/test/plan/non_mpi_profiling/test_multi_app.sh
```diff
@@ -0,0 +1,42 @@
...
+TEST_NAME=test_multi_app
+export GEOPM_PROFILE=${TEST_NAME}
+export GEOPM_PROGRAM_FILTER=geopmbench,stress-ng
+export LD_PRELOAD=libgeopm.so.2.1.0
+
+# GEOPM_CTL_LOCAL=true \
+GEOPM_REPORT=${TEST_NAME}_report.yaml \
+GEOPM_REPORT_SIGNALS=TIME@package \
+GEOPM_NUM_PROC=2 \
+setsid geopmctl &
```
cmcantalupo: Not sure why setsid would be needed here

integration/test/plan/non_mpi_profiling/test_multi_app.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Some of the initialization regions of geopmbench finish quickly. The
+# test_multi_app.py test checks the execution time of all regions. This sleep
+# ensures that geopmctl is started before starting geopmbench.
+sleep 2
```
cmcantalupo: Now that the runtime for each benchmark is longer, maybe the sleep 2 is not needed?

integration/test/plan/README.md
```diff
@@ -89,13 +98,20 @@ By default, this test is setup to run on a single compute node.  After
...
+     (v3.1 vs. v3.2 differ on this; see below).
    - If the Runtime was built *without* MPI support, the
      `MPI_Init_thread`  region would not be present in the output
-     files.  `geopmbench` can be invoked with the GEOPMBENCH_NO_MPI
-     environment variable to avoid any MPI calls. For more information
+     files.  `geopmbench` can be invoked with the `GEOPMBENCH_NO_MPI`
+     environment variable to avoid MPI calls. For more information
      about the environment configuration of `geopmbench`, see
      [here](https://geopm.github.io/geopmbench.1.html#environment).
+     - `geopmbench ` @ v3.1 will always call `MPI_Init` and `MPI_Finalize`
```
cmcantalupo, line 154: File a tech debt issue so that we remember to delete this comment on next release. Label it appropriately (next-release?)

bgeltz: Done. https://github.com/geopm/geopm/issues/3723

integration/test/test_multi_app_python.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# geopmbench
+#   geopmbench already links against libgeopm, no LD_PRELOAD is needed
+GEOPM_PROGRAM_FILTER=geopmbench numactl --cpunodebind=0 -- geopmbench ${INPUT_FILE} &
+
+# python example
+#   GEOPM @ v3.1 requires the use of LD_PRELOAD
+LD_PRELOAD=libgeopm.so.2.1.0 GEOPM_PROGRAM_FILTER=python3 numactl --cpunodebind=1 -- python3 -c 'import time; time.sleep(120)' &
+#   GEOPM @ v3.2 and beyond provides a gffi module to dl_open libgeopm; LD_PRELOAD is no longer necessary
+# GEOPM_PROGRAM_FILTER=python3 python3 -c 'import geopmpy.gffi, time; time.sleep(120)'
```
cmcantalupo:
```suggestion
# GEOPM_PROGRAM_FILTER=python3 python3 -c 'import geopmpy.gffi, time; time.sleep(120)' &
```

cmcantalupo (approved):

PR #3738, opened 2025-03-13
Add support for xe-drm interface + add high level aliases

### LINKED ISSUE #3732: Any plan to support Xe2+ systems (BMG, xe-drm, etc)
While proven useful in PVC performance triage that GeoPM could capture frequency and power history easily during running of a process, we found out it doesn't support BMG. Any plan for future intel GPUs?

### LINKED ISSUE #3737: As a user of the Intel platforms I would like the SysfsIOGroup with the DrmDeviceDriver to provide high level aliases
When the Drm device driver is used, some signals and controls are exposed that could have high level aliases (in particular frequency signals and controls).  Providing these will make using GPU systems leveraging the drm sysfs entries to provide more of the same signals and controls as systems using LevelZero or NVML for GPU support.

Current high level aliases are limited to:

GPU_POWER_LIMIT_CONTROL
GPU_POWER_LIMIT_DEFAULT
GPU_POWER_TIME_WINDOW_CONTROL
GPU_CHIP_ENERGY

Similar to what was done with the Cpufreq driver for cases where MSRs are not available, provide high level aliases for GPU signals and controls when the LevelZeroIOGroup cannot be loaded or is not linked.  Aliases that are provided by LevelZero but not by DRM (not all of these may be available from sysfs):

GPU_CORE_ACTIVITY
GPU_CORE_ENERGY
GPU_CORE_FREQUENCY_MAX_AVAIL
GPU_CORE_FREQUENCY_MAX_CONTROL
GPU_CORE_FREQUENCY_MIN_AVAIL
GPU_CORE_FREQUENCY_MIN_CONTROL
GPU_CORE_FREQUENCY_STATUS
GPU_CORE_FREQUENCY_STEP
GPU_CORE_POWER
GPU_ENERGY
GPU_POWER
GPU_UNCORE_ACTIVITY
GPU_UTILIZATION

The existing set of DRM signals and controls is fully featured, but a user familiar with signals like GPU_CORE_FREQUENCY_STATUS will be confused to not see it present.  Even worse, if code leverages these high level aliases, it will not function.

**Additional context**

cmcantalupo:
- Fixes #3732
- Fixes #3737
- [x] Add missing attributes listed below to https://github.com/geopm/geopm/blob/dev/docs/json_data/sysfs_attributes_drm.json
    ```
    DIRECT ATTRIBUTES
    freq0/act_freq
    freq0/cur_freq
    freq0/max_freq
    freq0/min_freq
    freq0/rp0_freq
    freq0/rpa_freq
    freq0/rpe_freq
    freq0/rpn_freq
    gtidle/idle_residency_ms
    gtidle/idle_status (MAYBE SKIP)

    HWMON ATTRIBUTES:
    energy2_input
    power2_crit
    ```

- [x] Do minimal integration testing, e.g.
    ```for ss in $(geopmread); do geopmread $ss board 0 >&/dev/null || echo FAIL: $ss; done```
- [x] Fix unit tests
- [x] Add documentation to geopm_pio_syfs man page
- [x] Derive number of numa domains properly


docs/json_data/sysfs_attributes_drm.json
```diff
@@ -154,7 +154,7 @@
             "format": "double",
             "units": "hertz",
             "writeable": false,
-            "alias": ""
+            "alias": "GPU_CORE_FREQUENCY_STATUS"
```
dannosliwcd, line 157: Should we map this to RPS_ACT_FREQ instead? I think the difference would be if the driver told the gpu_chip to target some frequency and the gpu_chip doesn't achieve that frequency. In that case, this signal would show what the driver asked for but the other one would show what the tile achieved.

docs/json_data/sysfs_attributes_drm.json
```diff
@@ -539,6 +539,138 @@
             "units": "none",
             "writeable": false,
             "alias": ""
+        },
+        "ACT_FREQ": {
+            "attribute": "freq0/act_freq",
```
dannosliwcd, line 544: Were you able to figure out what the 0 in `freq0` maps to in the xe driver? I wonder if it can report more than one frequency domain in those different numbers (vs how the i915 driver reports rps freq and base freq).

docs/json_data/sysfs_attributes_drm.json
```diff
@@ -539,6 +539,138 @@
...
+            "attribute": "freq0/cur_freq",
+            "scalar": 1e6,
+            "description": "Current GPU frequency",
+            "doc_domain": "gpu_chip",
+            "aggregation": "average",
+            "behavior": "variable",
+            "format": "double",
+            "units": "hertz",
+            "writeable": false,
+            "alias": "GPU_CORE_FREQUENCY_STATUS"
```
dannosliwcd, line 565: Similar question as in i915. This signal indicates the driver's cached frequency expectation. Do we want to alias act_freq instead?

docs/json_data/sysfs_attributes_drm.json
```diff
@@ -539,6 +539,138 @@
...
+            "behavior": "variable",
+            "format": "double",
+            "units": "hertz",
+            "writeable": false,
+            "alias": "GPU_CORE_FREQUENCY_MAX_AVAIL"
+        },
+        "RPA_FREQ": {
+            "attribute": "freq0/rpa_freq",
+            "scalar": 1e6,
+            "description": "Actual GPU RPA frequency",
```
dannosliwcd:
```suggestion
            "description": "Achievable GPU frequency",
```
Does the `a` in RPa mean achievable?

docs/json_data/sysfs_attributes_drm.json
```diff
@@ -539,6 +539,138 @@
...
+            "behavior": "variable",
+            "format": "double",
+            "units": "hertz",
+            "writeable": false,
+            "alias": ""
+        },
+        "RPE_FREQ": {
+            "attribute": "freq0/rpe_freq",
+            "scalar": 1e6,
+            "description": "GPU RPE frequency",
```
dannosliwcd:
```suggestion
            "description": "Efficient GPU frequency",
```

docs/json_data/sysfs_attributes_drm.json
```diff
@@ -539,6 +539,138 @@
...
+            "writeable": false,
+            "alias": "GPU_CORE_FREQUENCY_MIN_AVAIL"
+        },
+        "IDLE_RESIDENCY": {
+            "attribute": "gtidle/idle_residency_ms",
+            "scalar": 1e-3,
+            "description": "Time spent in idle state",
+            "doc_domain": "gpu_chip",
+            "aggregation": "average",
+            "behavior": "variable",
```
dannosliwcd:
```suggestion
            "behavior": "monotone",
```
Sounds like an incrementing counter. Have you come across docs/implementation/comments to confirm?

Or if it is variable, then I'd guess we need to update the description to "Time spent in idle state `<per or since> <???>`"

bgeltz (dismissed):

docs/json_data/sysfs_attributes_drm.json
```diff
@@ -539,6 +539,138 @@
...
+            "format": "double",
+            "units": "seconds",
+            "writeable": false,
+            "alias": ""
+        },
+        "HWMON::ENERGY2_INPUT": {
+            "attribute": "energy2_input",
```
dannosliwcd, line 652: Would be nice if we can dynamically determine the `energy#_input` mappings based on the `name` attribute

docs/json_data/sysfs_attributes_drm.json
dannosliwcd: Seems hardcoded 

https://github.com/torvalds/linux/blob/b5329d5a35582abbef57562f9fb6cb26a643f252/drivers/gpu/drm/xe/xe_gt_freq.c#L41

and

https://github.com/torvalds/linux/blob/b5329d5a35582abbef57562f9fb6cb26a643f252/drivers/gpu/drm/xe/xe_gt_freq.c#L251

docs/json_data/sysfs_attributes_drm.json
dannosliwcd: Ah, the hwmon docs you mention in the next file specify the # mapping. They state energy2 is package level and energy1 is card level, so we need to update this.

docs/json_data/sysfs_attributes_drm.json
cmcantalupo: The "ACT" signal is sometimes reported as zero but "CUR" seems to be more reliable.

dannosliwcd (approved):

PR #3769, opened 2025-04-04
Fix names of compiler wrappers in integration build scripts

sidjana:
- Fixes name of the intel compiler wrappers in accordance with the latest agama release
- Minor spelling typo fix

Relates to #3775

### DIFF
--- a/integration/config/mcfly_env.sh
+++ b/integration/config/mcfly_env.sh
@@ -20,5 +20,4 @@ export MPIFORT=mpiifort
 export MPIFC=mpiifort
 export MPIF77=mpiifort
 export MPIF90=mpiifort
-export PALS_PMI=pmix
 export GEOPM_LAUNCHER=srun
--- a/integration/config/mcfly_env.sh
+++ b/integration/config/mcfly_env.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+#
+#  Copyright (c) 2015 - 2025 Intel Corporation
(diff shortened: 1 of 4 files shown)


bgeltz (changes requested):
You shouldn't need to do this.  Take a look at:
https://github.com/geopm/geopm/blob/dev/integration/config/australis_env.sh

And the docu here:
https://github.com/geopm/geopm/tree/dev/integration/test/plan#development-snapshot

Just add:
```
export GEOPM_SYSTEM_ENV=${GEOPM_SOURCE}/integration/config/australis_env.sh
```
to your `~/.geopmrc` and everything should work as expected.

sidjana:
Notes from sync on Apr 8'25:
-- goal is to create a new bash file for mcfly 
-- and change the default of build_env.sh to match the settings for australis_env.sh

integration/config/mcfly_env.sh
```diff
@@ -0,0 +1,24 @@
...
+export MPICC=mpiicc
+export MPICXX=mpiicpc
+export FC=ifx
+export F77=ifx
+export F90=ifx
+export MPIFORT=mpiifort
+export MPIFC=mpiifort
+export MPIF77=mpiifort
+export MPIF90=mpiifort
+export PALS_PMI=pmix
```
bgeltz:
```suggestion
```

bgeltz (approved):

PR #3787, opened 2025-04-11
Fix issue #3777

### LINKED ISSUE #3777: GPU_CORE_POWER seemingly incorrect when included in batch reads with TIMESTAMP, ENERGY
**Describe the bug**
I tried to use geopmsession and I expected it to work consistently with GPU signals instead depending on the specific signals specified, GPU_CORE_POWER appears to be reporting incorrect values.

**GEOPM version**
v3.1

**Expected behavior**
Utilizing `geopmsession` with the following signals provides consistent GPU_CORE_POWER data:
```
LEVELZERO::GPU_CORE_ENERGY gpu_chip 0
LEVELZERO::GPU_CORE_ENERGY_TIMESTAMP gpu_chip 0
LEVELZERO::GPU_CORE_POWER gpu 0
```

**Actual behavior**
Utilizing `geopmsession` with the following signals provides seemingly incorrect GPU_CORE_POWER data:
```
LEVELZERO::GPU_CORE_ENERGY gpu_chip 0
LEVELZERO::GPU_CORE_ENERGY_TIMESTAMP gpu_chip 0
LEVELZERO::GPU_CORE_POWER gpu 0
```

**Additional context**
This seems related to the caching mechanism that is used with the energy is read in LevelZero.cpp.  Additionally, the GPU_CORE_ENERGY_TIMESTAMP was recently changed so it could be read outside of a batch context in PR# 3615 which may be related.

cmcantalupo:
- Fixes #3777

### DIFF
--- a/libgeopmd/src/DerivativeSignal.cpp
+++ b/libgeopmd/src/DerivativeSignal.cpp
@@ -37,6 +37,7 @@ namespace geopm
         , m_sleep_time(sleep_time)
         , m_last_result(NAN)
         , m_nan_replace(nan_replace)
+        , m_last_time(NAN)
     {
         GEOPM_DEBUG_ASSERT(m_time_sig && m_y_sig,
                            "Signal pointers for time_sig and y_sig cannot be null.");
@@ -100,14 +101,14 @@ namespace geopm
             throw Exception("setup_batch() must be called before sample().",
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
-        bool is_sampled = m_time_sig->is_sampled();
         double time = m_time_sig->sample();
         size_t history_size = m_history.size();
         // Check if this is the first call ever to sample() (history_size == 0)
         // Or check if this is the first call to sample() since the last call to read_batch()
-        if (history_size == 0ULL || ! is_sampled) {
+        if (history_size == 0ULL || m_last_time != time) {
             double signal = m_y_sig->sample();
             m_last_result = compute_next(m_history, m_derivative_num_fit, time, signal, m_nan_replace);
+            m_last_time = time;
         }
         return m_last_result;
     }
--- a/libgeopmd/src/LevelZeroDevicePool.cpp
+++ b/libgeopmd/src/LevelZeroDevicePool.cpp
@@ -266,12 +266,39 @@ namespace geopm
                                             dev_subdev_idx_pair.second);
     }
 
+    static double convert_active_timestamp(uint64_t value, uint64_t &last_value, uint64_t &rollover_count, uint64_t &original_value)
+    {
+        static const int num_bits = 32;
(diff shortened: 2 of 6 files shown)


libgeopmd/src/DerivativeSignal.cpp
```diff
@@ -100,14 +101,14 @@ namespace geopm
             throw Exception("setup_batch() must be called before sample().",
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
-        bool is_sampled = m_time_sig->is_sampled();
         double time = m_time_sig->sample();
         size_t history_size = m_history.size();
         // Check if this is the first call ever to sample() (history_size == 0)
         // Or check if this is the first call to sample() since the last call to read_batch()
-        if (history_size == 0ULL || ! is_sampled) {
+        if (history_size == 0ULL || m_last_time != time) {
```
asmaalrawi, line 108: Test with GPU ACTIVITY. IIRC, timestamp doesn't update until there's activity on the GPU. So NAN might persist until there's activity on GPU, which causes issues on report, as the average will not reflect NAN samples.

cmcantalupo: With the change to the implementation the activity rate will not change until the timestamp advances (so if it was 1 it will stay one until the timestamp advances).  The nan_replace will not be used.  Seems like we have to use a real timestamp for gating nan usage to work around the fact that time does not advance for the activity counter timestamps.

cmcantalupo: I have put a fix in place specifically for the activity timestamps.  Would be great to test the CA agents with these changes.

asmaalrawi (dismissed):

libgeopmd/src/LevelZeroDevicePool.cpp
```diff
@@ -266,12 +266,35 @@ namespace geopm
...
+        static const uint64_t overflow = (1ULL << num_bits);
+        static const double overflow_d = overflow;
+        static const uint64_t mask = overflow - 1;
+
+        value &= mask;
+        if (original_value > value) {
+            ++rollover_count;
+        }
+        if (value == original_value) {
+            value = last_value + 1;
```
bgeltz, line 285: Add a comment here explaining that this is required because when there is no activity, the counter is not incrementing.
```suggestion
            // There are time and timestamp signals associated with all activity signals.
            // Time and timestamp do not change if there is no activity, however the sample()
            // API requires that the timstamp signal value increase on each call for derivative signals.
            value = last_value + 1;
```

bgeltz (dismissed):

bgeltz (dismissed):

bgeltz (dismissed):

bgeltz (approved):

PR #3814, opened 2025-04-21
Polishing text for the k8 README

sidjana:
Updating text & adding some clarification notes re. the Prometheus framework setup in a containerized environment


sidjana:
In README.md:

Is the reference to GEOPM Access Service Ubuntu packages (line#34) the same as GEOPM software packages (line# 39)? In that case, those relevant lines can be either rewritten or merged to reduce redundancies.

PR #3837, opened 2025-05-01
Release v3.2 packaging

cmcantalupo:
Fixes required to package GEOPM v3.2.0 in OBS and COPR for all target distros.


cmcantalupo:
We will merge these changes from another branch after rebasing.  We will leave the release-v3.2-packaging unchanged to allow the SHA values to persist when referring to the packing files.

Issue #1550, opened 2021-02-23
Model regions may invoke oom killer


bgeltz:
Depending on the number of ranks used, ModelRegions used in e.g. `geopmbench` may invoke the oom killer unexpectedly.  Here's a run log from 1 node and 4 ranks:

```
$ geopmlaunch srun -N1 -n4 geopmbench

srun -N 1 -n 1 --cpu-freq=Performance -- geopmread --cache

srun: Loaded msrsave restore plugin.
srun: Loaded geopm plugin.
srun: Loaded msrsave restore plugin.
srun: Loaded geopm plugin.
Warning: <geopm> geopmpy.launcher: User did not provide OMP_NUM_THREADS, using 20

OMP_PROC_BIND=true GEOPM_CTL=process GEOPM_PROFILE= GEOPM_REPORT=geopm.report KMP_AFFINITY=disabled MV2_ENABLE_AFFINITY=0 KMP_WARNINGS=FALSE srun -N 1 -n 5 --cpu-bind v,mask_cpu:0x2,0xffc00000000ffc,0x3ff0000
00003ff000,0x3ff000000003ff000000,0xffc00000000ffc00000000 --export=LD_PRELOAD=libgeopm.so,ALL --cpu-freq=Performance geopmbench

srun: Loaded msrsave restore plugin.
srun: Loaded geopm plugin.
srun: Warning: can't honor --ntasks-per-node set to 1 which doesn't match the requested tasks 5 with the number of requested nodes 1. Ignoring --ntasks-per-node.
cpu-bind=MASK - mcfly2, task  0  0 [15507]: mask 0x2 set
cpu-bind=MASK - mcfly2, task  3  3 [15510]: mask 0x3ff000000003ff000000 set
cpu-bind=MASK - mcfly2, task  1  1 [15508]: mask 0xffc00000000ffc set
cpu-bind=MASK - mcfly2, task  4  4 [15511]: mask 0xffc00000000ffc00000000 set
cpu-bind=MASK - mcfly2, task  2  2 [15509]: mask 0x3ff000000003ff000 set
[mcfly2:mpi_rank_0][mv2_get_hca_type] **********************WARNING***********************
[mcfly2:mpi_rank_0][mv2_get_hca_type] Failed to automatically detect the HCA architecture.
[mcfly2:mpi_rank_0][mv2_get_hca_type] This may lead to subpar communication performance.
[mcfly2:mpi_rank_0][mv2_get_hca_type] ****************************************************
[...]

PR #3863, opened 2025-05-19
Avoid oversizing the dgemm matrix

cmcantalupo:
- Relates to #1550

### DIFF
--- a/libgeopm/src/DGEMMModelRegion.cpp
+++ b/libgeopm/src/DGEMMModelRegion.cpp
@@ -51,9 +51,12 @@ namespace geopm
         , m_matrix_a(NULL)
         , m_matrix_b(NULL)
         , m_matrix_c(NULL)
-        , m_matrix_size(0)
+        , m_matrix_m_size(8192)
+        , m_matrix_n_size(2048)
+        , m_matrix_k_size(4096)
         , m_pad_size(geopm::hardware_destructive_interference_size)
         , m_num_warmup(4)
+        , m_num_repeat(100)
     {
         m_name = "dgemm";
         m_do_imbalance = do_imbalance;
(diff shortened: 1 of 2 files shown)


libgeopm/src/DGEMMModelRegion.cpp
```diff
@@ -124,28 +159,30 @@ namespace geopm
...
+        double alpha = 2.0;
+        double beta = 3.0;
+        char transa = 'n';
+        char transb = 'n';
         if (m_big_o != 0.0) {
             if (m_verbosity) {
-                std::cout << "Executing " << m_matrix_size << " x " << m_matrix_size << " DGEMM "
-                          << m_num_progress_updates << " times." << std::endl << std::flush;
+                std::cout << "Executing " << m_matrix_m_size << " x "
+                          << m_matrix_n_size << " DGEMM "
```
cmcantalupo: probably should provide the k dimension as well.

libgeopm/src/DGEMMModelRegion.cpp
```diff
@@ -71,43 +73,76 @@ namespace geopm
...
+#endif
+            for (size_t i = 0; i < mem_size_a / sizeof(double); ++i) {
+                m_matrix_a[i] = 2.0 * i;
             }
+            // Allocate B: K x N
+            size_t mem_size_b = sizeof(double) * (m_matrix_k_size + m_pad_size) * m_matrix_n_size;
+            err = posix_memalign((void **)&m_matrix_b, m_pad_size, mem_size_b);
             if (err) {
+                free(m_matrix_a);
+                m_matrix_a = NULL;
```
asmaalrawi, line 128: Do we need to check if m_matrix_a is already NULL to prevent a double free, in case someone is catching the previous exception? Same below for matrix a/b on 139/140.

libgeopm/src/DGEMMModelRegion.cpp
```diff
@@ -71,43 +73,76 @@ namespace geopm
...
 
         num_progress_updates(big_o_in);
 
-        m_matrix_size = (int)pow(4e9 * big_o_in / m_num_progress_updates, 1.0/3.0);
         if (big_o_in && m_big_o != big_o_in) {
-            size_t mem_size = sizeof(double) * (m_matrix_size * (m_matrix_size + m_pad_size));
-            int err = posix_memalign((void **)&m_matrix_a, m_pad_size, mem_size);
-            if (!err) {
-                err = posix_memalign((void **)&m_matrix_b, m_pad_size, mem_size);
+            // Allocate A: M x K
```
asmaalrawi, line 110: Comments clarifying the padding logic would be useful.

libgeopm/src/DGEMMModelRegion.cpp
```diff
@@ -51,7 +51,9 @@ namespace geopm
         , m_matrix_a(NULL)
         , m_matrix_b(NULL)
         , m_matrix_c(NULL)
-        , m_matrix_size(0)
+        , m_matrix_m_size(4096)
```
asmaalrawi: Add documentation somewhere to clarify reasons for these sizes.

libgeopm/src/DGEMMModelRegion.cpp
cmcantalupo: Don't think that control flow is possible.

libgeopm/src/DGEMMModelRegion.cpp
cmcantalupo: I put the comment in the constructor where m_pad_size is defined.

asmaalrawi (approved):

PR #3865, opened 2025-05-20
Close all sessions when service is stopped

### LINKED ISSUE #3864: Stopping the GEOPM service does not end all sessions
**Describe the bug**
I tried to run `systemctl stop geopm`  and I expected that all client sessions would be ended and all control values modified would be restored instead, the service is simply paused until it is started again.

**GEOPM version**
3.2.0

**Expected behavior**
By sending SIGINT for the RestartKillSignal in the systemd unit, and by making the signal handler for SIGINT exhibit the current behavior whereas SIGTERM (default KillSignal value) should end all sessions, and restore controls.  The SIGINT handler path would be the path for exception management as well.

**Actual behavior**
`systemctl stop geopm` is not different than `systemctl restart geopm` WRT save restore.

**Error messages produced**

**Additional context**

cmcantalupo:
- Pause when an exception is raised, or if a restart is signaled.
- Fixes #3864

### DIFF
--- a/docs/source/admin.rst
+++ b/docs/source/admin.rst
@@ -1,10 +1,10 @@
 Service Administrators
 ======================
 
-This guide covers GEOPM's integration with the Linux OS, directories
-influenced by GEOPM, the utilization of files within those directories, and a
-command-line tool for configuring the GEOPM Service. For further details,
-explore the subsequent sections:
+This guide covers GEOPM's integration with the Linux OS, directories influenced
+by GEOPM, the utilization of files within those directories, and a command-line
+tool for configuring the GEOPM Service. For further details, explore the
+subsequent sections:
 
 - :doc:`Install Guide <install>`
 - :doc:`Security Guide <security>`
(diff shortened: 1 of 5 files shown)


geopmdpy/geopmdpy/__main__.py
```diff
@@ -44,13 +53,13 @@ def main_dbus():
...
+            _service = service.GEOPMService()
+            _service.topo_rm_cache()
+            _bus.publish_object("/io/github/geopm", _service)
             _bus.register_service("io.github.geopm")
             _loop.run()
         finally:
-            stop()
+            pause()
 
 def main_grpc():
```
cmcantalupo, line 64: Consider adding a signal handler for grpc code path, could be a future PR

bgeltz (dismissed):

bgeltz (dismissed):

docs/source/admin.rst
```diff
@@ -14,9 +14,39 @@ Linux Integration
...
+
+- **stop**:
+  Stops the GEOPM Access Service. All active client sessions are closed
+  gracefully before the service terminates. This includes restoring any
+  hardware controls (such as power limits) to their default or saved state, and
+  cleaning up session files in `/run/geopm`. The write-mode session (if any) is
+  closed first to ensure hardware state is restored.
+
+- **restart**:
+  Re-iniatialize the service.  Similar to running ``stop`` followed by
```
bgeltz:
```suggestion
  Re-initialize the service.  Similar to running ``stop`` followed by
```

bgeltz (dismissed):

bgeltz (approved):

Issue #3845, opened 2025-05-07
FFNet agent decisions are not changing with workload changes


asmaalrawi:
**Describe the bug**
I tried to use a neural net trained on geopmbench to test ffnet inference on region characterization while running geopmbench. I expected the output of ffnet agent to show at least 95% of trace samples identifying the correct region with >=95% probability. Instead, region probabilities were largely static.

**GEOPM version**
3.2.0

**Expected behavior**
Output logits of ffnet reflect changes in workload behavior

**Actual behavior**
Logits that are output in trace are largely static.

**Additional Context**
FFNet integration tests with skip decorators reproduce this issue. The skip decorators should be removed and tests should pass when this issue is fixed.

PR #3874, opened 2025-05-28
Resize and adjust ffnet tests for new geopmbench, fix erroneous QM_CTR event select issue

asmaalrawi:
Addresses #3845

### DIFF
--- a/integration/experiment/ffnet/gen_hdf_from_fsweep.py
+++ b/integration/experiment/ffnet/gen_hdf_from_fsweep.py
@@ -41,7 +41,7 @@ def process_region_for_packages(region_dict, nodename, app_name, region_id, freq
         nodename: Name of the node
         app_name: Name of the application
         region_id: ID of the region (hash or "0xDEADBEEF")
-        freqs: Dictionary of frequencies
+        freqs: Dictionary of average region frequencies keyed by device, where device is one of "cpu", "gpu", or "uncore" (for cpu uncore)
 
     Returns:
         list: List of region entries for each package
--- a/integration/experiment/ffnet/gen_hdf_from_fsweep.py
+++ b/integration/experiment/ffnet/gen_hdf_from_fsweep.py
@@ -48,21 +48,23 @@ def process_region_for_packages(region_dict, nodename, app_name, region_id, freq
     """
     entries = []
     num_packages = get_package_count(region_dict)
-
     for pkg_idx in range(num_packages):
         conf = {}
-        # node and package
+        # node, package, and runtime
         conf['node'] = nodename
         conf['package'] = pkg_idx
+        conf['runtime-pkg'] = region_dict[f'TIME@package-{pkg_idx}']
 
-        # frequencies
+        # Frequencies - from policy and report status
+        freq_lookup = {"cpu":f"CPU_FREQUENCY_STATUS@package-{pkg_idx}",
+                       "gpu":"GPU_CORE_FREQUENCY_STATUS",
+                       "uncore":"CPU_UNCORE_FREQUENCY_STATUS@package-{pkg_idx}"}
+
         for device in freqs:
             if freqs[device] is not None:
-                conf[f"{device}-frequency"] = freqs[device]
-
-        # Get per-package frequency if available
-        if f'CPU_CORE_FREQUENCY_STATUS-package-{pkg_idx}' in region_dict:
-            conf['cpu-frequency'] = region_dict[f'CPU_CORE_FREQUENCY_STATUS-package-{pkg_idx}']
+                conf[f'{device}-frequency-ctl'] = freqs[device]
+            if freq_lookup[device] in region_dict:
+                conf[f'{device}-frequency'] = float(region_dict[f'{freq_lookup[device]}'])
 
         region_entry = region_dict.copy()
         region_entry['app-config'] = app_name + '-' + region_id
@@ -102,7 +104,7 @@ def process_report_files(input_dir, region_ignore):
                         # Process per-package data
                         reports.extend(process_region_for_packages(region_dict, nodename, app_name, format_region, freqs))
             else:
-                # Handle sweeps done with python infrastructure that does not have regions or a region hash
+                # Handle sweeps with no markup for regions or region hashes (e.g. GPU workloads)
                 region_dict = report["Hosts"][nodename]["Application Totals"]
                 reports.extend(process_region_for_packages(region_dict, nodename, app_name, "0xDEADBEEF", freqs))
 
@@ -192,10 +194,13 @@ def main(output_prefix, frequency_sweep_dirs, region_ignore=None):
                 'package',
                 'app-config',
                 'runtime (s)',
+                'runtime-pkg',
+                'cpu-frequency-ctl',
                 'cpu-frequency',
                 'package-energy (J)',
             ]
     want_columns = [
+            'gpu-frequency-ctl',
             'gpu-frequency',
             'gpu-energy (J)',
             'uncore-frequency',
--- a/integration/experiment/ffnet/gen_hdf_from_fsweep.py
+++ b/integration/experiment/ffnet/gen_hdf_from_fsweep.py
@@ -14,6 +14,63 @@ try:
 except ImportError:
     from yaml import SafeLoader
 
+def get_package_count(region_dict):
+    """Determine the number of packages by counting TIME@package fields in the region_dict.
+
+    Args:
+        region_dict: Dictionary containing region data from the report
+
+    Returns:
+        int: Number of packages detected
+    """
+    package_count = 0
+    while f'TIME@package-{package_count}' in region_dict:
+        package_count += 1
+
+    # Fallback to default if no packages are detected
+    if package_count == 0:
+        package_count = 1  # Assume at least one package
+
+    return package_count
+
+def process_region_for_packages(region_dict, nodename, app_name, region_id, freqs):
+    """Process a region for all packages and return list of region entries.
+
+    Args:
+        region_dict: Dictionary containing region data
+        nodename: Name of the node
+        app_name: Name of the application
+        region_id: ID of the region (hash or "0xDEADBEEF")
+        freqs: Dictionary of frequencies
+
+    Returns:
+        list: List of region entries for each package
+    """
+    entries = []
+    num_packages = get_package_count(region_dict)
+
+    for pkg_idx in range(num_packages):
+        conf = {}
+        # node and package
+        conf['node'] = nodename
+        conf['package'] = pkg_idx
+
+        # frequencies
+        for device in freqs:
+            if freqs[device] is not None:
+                conf[f"{device}-frequency"] = freqs[device]
+
+        # Get per-package frequency if available
+        if f'CPU_CORE_FREQUENCY_STATUS-package-{pkg_idx}' in region_dict:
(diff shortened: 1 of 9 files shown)


libgeopm/src/FFNetAgent.cpp
```diff
@@ -165,7 +165,7 @@ namespace geopm
 
         m_platform_io.write_control("MSR::PQR_ASSOC:RMID", GEOPM_DOMAIN_BOARD, 0, 0);
         m_platform_io.write_control("MSR::QM_EVTSEL:RMID", GEOPM_DOMAIN_BOARD, 0, 0);
-        m_platform_io.write_control("MSR::QM_EVTSEL:EVENT_ID", GEOPM_DOMAIN_BOARD, 0, 2);
+        m_platform_io.write_control("MSR::QM_EVTSEL:EVENT_ID", GEOPM_DOMAIN_BOARD, 0, 3);
```
bgeltz: We should look at anywhere else where this is programmed for other agents and see if it makes sense to use 3 there as well.  AFAICT it's just in the CPUActivityAgent / test_cpu_characterization.

integration/experiment/ffnet/run_ffnet_arithmetic_intensity.py
```diff
@@ -0,0 +1,27 @@
...
+
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    ffnet.setup_run_args(parser)
+    arithmetic_intensity.setup_run_args(parser)
+    args, extra_args = parser.parse_known_args()
+    mach = machine.init_output_dir(args.output_dir)
+    app_conf = arithmetic_intensity.create_appconf(mach, args)
+    ffnet.launch(app_conf=app_conf,
+                            args=args,
+                            experiment_cli_args=extra_args)
```
bgeltz: nit...
```suggestion
    ffnet.launch(app_conf=app_conf,
                 args=args,
                 experiment_cli_args=extra_args)
```

libgeopm/src/FFNetAgent.cpp
asmaalrawi: Offset 2 seems to work better generally speaking for understanding memory work as it's counting offloads b/w LLC -> DRAM and b/w LLC -> lower level caches, but for some reason 3, which only counts LLC -> L1/L2 cache bandwidth, works better with the neural net. I'm wondering if DRAM activity is just taken care of by the DRAM_POWER measurement so it's able to take care of that component of the QM counter offset 2.

integration/test/test_ffnet_nn_scripts.py
```diff
@@ -462,87 +497,148 @@ def test_freqmap_generation(self):
...
-            - For each region, phi=0 power is at most the same as max freq run
-        """
+            - For each region, phi=0 dgemm/stream runtime is within _MIN_PERF_RATIO of max
+              freq run
+            - For each region, phi=0 dgemm/stream energy consumption x _MIN_ENERGY_RATIO
+              is at most max freq run energy consumption
+        """
+        max_freq = "{:.1e}".format(self._cpu_fsweep_experiment_args.max_frequency)
+        max_freq_fname = glob.glob(f"{str(self._nn_sweep_dir)}/geopmbench_frequency_map_cpu{max_freq}*.report")[0]
+        max_freq_report_data = geopmpy.io.RawReport(max_freq_fname)
```
asmaalrawi, line 533: This won't work, as the problem size can be different. It would be better to kick off a single monitor run at the very beginning.

integration/experiment/ffnet/gen_hdf_from_fsweep.py
```diff
@@ -14,6 +14,65 @@
...
+    return package_count
+
+def process_region_for_packages(region_dict, nodename, app_name, region_id, freqs):
+    """Process a region for all packages and return list of region entries.
+
+    Args:
+        region_dict: Dictionary containing region data
+        nodename: Name of the node
+        app_name: Name of the application
+        region_id: ID of the region (hash or "0xDEADBEEF")
```
dannosliwcd, line 43: Is deadbeef for reports without regions?

integration/experiment/ffnet/gen_hdf_from_fsweep.py
```diff
@@ -14,6 +14,65 @@
...
+        conf = {}
+        # node, package, and runtime
+        conf['node'] = nodename
+        conf['package'] = pkg_idx
+        conf['runtime-pkg'] = region_dict[f'TIME@package-{pkg_idx}']
+
+        # Frequencies - from policy and report status
+        freq_lookup = {"cpu":f"CPU_FREQUENCY_STATUS@package-{pkg_idx}",
+                       "gpu":"GPU_CORE_FREQUENCY_STATUS",
+                       "uncore":"CPU_UNCORE_FREQUENCY_STATUS@package-{pkg_idx}"}
```
dannosliwcd, line 61: Is GPU uncore/base not needed?

integration/experiment/ffnet/gen_hdf_from_fsweep.py
```diff
@@ -14,6 +14,65 @@
...
+
+def process_region_for_packages(region_dict, nodename, app_name, region_id, freqs):
+    """Process a region for all packages and return list of region entries.
+
+    Args:
+        region_dict: Dictionary containing region data
+        nodename: Name of the node
+        app_name: Name of the application
+        region_id: ID of the region (hash or "0xDEADBEEF")
+        freqs: Dictionary of frequencies
```
dannosliwcd:
```suggestion
        freqs: Dictionary of average region frequencies keyed by device, where device is one of "cpu", "gpu", or "uncore" (for cpu uncore)
```

integration/experiment/ffnet/gen_hdf_from_fsweep.py
```diff
@@ -14,6 +14,65 @@
...
+        conf['node'] = nodename
+        conf['package'] = pkg_idx
+        conf['runtime-pkg'] = region_dict[f'TIME@package-{pkg_idx}']
+
+        # Frequencies - from policy and report status
+        freq_lookup = {"cpu":f"CPU_FREQUENCY_STATUS@package-{pkg_idx}",
+                       "gpu":"GPU_CORE_FREQUENCY_STATUS",
+                       "uncore":"CPU_UNCORE_FREQUENCY_STATUS@package-{pkg_idx}"}
+
+        for device in freqs:
```
dannosliwcd, line 63: This potentially duplicates work for GPUs, once per CPU package (line 51). AFAIK it's not wrong since it'll just overwrite a value with itself in the duplicate calls. If that's right, then this comment should be okay to ignore.

dannosliwcd: Actually, is it intentional that we gather per-cpu metrics but not per-gpu metrics?

integration/experiment/ffnet/gen_hdf_from_fsweep.py
```diff
@@ -14,6 +14,65 @@
...
+            if freqs[device] is not None:
+                conf[f'{device}-frequency-ctl'] = freqs[device]
+            if freq_lookup[device] in region_dict:
+                conf[f'{device}-frequency'] = float(region_dict[f'{freq_lookup[device]}'])
```
dannosliwcd, line 67: Do we expect all 6 `-frequency` and `-frequency-ctl` values to be populated after this loop, or is a subset okay?

integration/experiment/ffnet/gen_hdf_from_fsweep.py
```diff
@@ -92,14 +139,39 @@ def process_trace_files(sweep_dir, region_ignore):
...
+            # Drop columns from other packages
+            other_pkg_cols = [col for col in trace_df.columns
+                              if "-package-" in col and not col.endswith(f"-package-{pkg_idx}")]
+            pkg_df = trace_df.drop(other_pkg_cols, axis=1)
+            # Rename columns to remove package suffix
+            pkg_df.rename(columns=lambda x: x.replace(f"-package-{pkg_idx}", ""), inplace=True)
+            # Add package identifier
+            pkg_df['package'] = pkg_idx
+            pkg_dfs.append(pkg_df)
+        trace_pkg_df = pd.concat(pkg_dfs, axis=0) if pkg_dfs else pd.DataFrame()
```
dannosliwcd, line 166: `pd.wide_to_long()` might be more concise, but I don't see any problems with this approach.

integration/experiment/ffnet/gen_neural_net.py
```diff
@@ -186,13 +186,13 @@ def main(input_list, output_name="nnet", describe_net="A neural net."):
...
-                   'MSR::QM_CTR_SCALED-package-0',
-                   'MSR::QM_CTR_SCALED_RATE-package-0'],
+            'cpu':['CPU_POWER',
+                   'DRAM_POWER',
+                   'CPU_FREQUENCY_STATUS',
+                   'MSR::QM_CTR_SCALED',
+                   'MSR::QM_CTR_SCALED_RATE',
+                   'MSR::UNCORE_PERF_STATUS:FREQ'
+                   ],
             'gpu':['GPU_CORE_FREQUENCY_STATUS-gpu-0',
```
dannosliwcd, line 196: I guess this contradicts an earlier assumption when I commented asking why we're repeatedly looking at all GPUs for each CPU package. Instead, we're repeatedly looking at GPU 0 for each CPU package?

integration/experiment/ffnet/neural_net_sweep.py
```diff
@@ -160,7 +162,7 @@ def launch(app_conf, args, experiment_cli_args):
                       "# Next, assign resource monitoring ID for QM events to match\n"
                       "MSR::QM_EVTSEL:RMID board 0 0\n"
                       "# Then determine Xeon Uncore Utilization\n"
```
dannosliwcd:
```suggestion
                      "# Then determine L3 Total External Bandwidth\n"
```
For event ID 2

integration/experiment/ffnet/gen_hdf_from_fsweep.py
asmaalrawi: yes

integration/experiment/ffnet/gen_hdf_from_fsweep.py
asmaalrawi: We can't manipulate it, so it won't help to include that info. This is not for nn training, just for fitting frequency vs runtime relationship per-domain. It shouldn't be too hard to expand this if support for that control is added in the future.

integration/experiment/ffnet/gen_neural_net.py
asmaalrawi: Yeah, I decided to push all the gpu-based changes to a later PR as it was taking too long to validate. For now, we're training on a single GPU and generalizing to all of them.

integration/experiment/ffnet/gen_hdf_from_fsweep.py
dannosliwcd: I guess we could add LEVELZERO::GPU_CORE_PERFORMANCE_FACTOR_CONTROL to manipulate base-die frequency

asmaalrawi: Setting up a new issue to add uncore controls.

bgeltz (approved):

PR #3876, opened 2025-05-29
Add a notification when grpc service is ready

cmcantalupo:
- Enable fixed counters in integration test that requires them
- Required for geopm/integration/service/test/test_batch_perf.sh

### DIFF
--- a/geopmdpy/geopmdpy/grpc_service.py
+++ b/geopmdpy/geopmdpy/grpc_service.py
@@ -9,6 +9,8 @@ import pwd
 import grpc
 import subprocess # nosec
 from concurrent import futures
+from sdnotify import SystemdNotifier
+from time import sleep
 from . import geopm_service_pb2_grpc
 from . import geopm_service_pb2
 from . import service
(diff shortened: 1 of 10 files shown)


geopmdpy/geopmdpy/grpc_service.py
```diff
@@ -218,6 +222,9 @@ def run():
     server.start()
 
     with subprocess.Popen('geopmd-proxy') as proxy:
+        while not os.path.exists(grpc_public_socket_path):
```
asmaalrawi, line 225: Add a timeout to prevent infinite loops if socket opening fails.

asmaalrawi: Add to k8/README TimeoutSec=900

integration/k8/README.md
```diff
@@ -77,7 +77,7 @@ set the `Type` to `simple`:
 $ sudo systemctl edit geopm
 $ cat /etc/systemd/system/geopm.service.d/override.conf
 [Service]
-Type=simple
+Type=notify
```
asmaalrawi, line 80:
```suggestion
Type=notify
TimeoutSec=900
```

cmcantalupo: Put it in the default systemd unit file instead

bgeltz (approved):

Issue #3869, opened 2025-05-21
geopmlaunch and geopmctl try to use a disallowed core


bgeltz:
**Describe the bug**
I tried to run one of the runtime integration tests and I expected it to pass instead it failed to launch.

**GEOPM version**
v3.2

**Expected behavior**
```
# Get a 1-node allocation and attempt to run test_trace.py
$ GEOPM_LAUNCHER=pals PYTHONPATH=${HOME}/geopm:${PYTHONPATH} GEOPM_NUM_NODE=1 python3 ~/flare/geopm/integration/test/test_trace.py -v
# Observe it runs to completion, the test assertions pass, and the RC is 0
```

**Actual behavior**
```
[bgeltz@x4302c5s1b0n0 spack]$ GEOPM_LAUNCHER=pals PYTHONPATH=${HOME}/flare/geopm:${PYTHONPATH} GEOPM_NUM_NODE=1 python3 ~/flare/geopm/integration/test/test_trace.py -v

mpiexec --ppn 1 -n 1 lscpu

setUpClass (__main__.TestIntegration_trace) ... ERROR

======================================================================
ERROR: setUpClass (__main__.TestIntegration_trace)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/bgeltz/flare/geopm/integration/test/test_trace.py", line 51, in setUpClass
    launcher.run(cls._test_name, add_geopm_args=geopm_args)
  File "/home/bgeltz/flare/geopm/integration/test/geopm_test_launcher.py", line 167, in run
    launcher.run(stdout=outfile, stderr=outfile)
  File "/home/bgeltz/git/spack/opt/spack/linux-sles15-icelake/gcc-13.3.0/py-geopmpy-3.2.0-iadqioewkgjit4rdxyiecqnkhf3zjxpl/lib/python3.11/site-packages/geopmpy/launcher.py", line 595, in run
    raise subprocess.CalledProcessError(pid.returncode, argv_mod)
subprocess.CalledProcessError: Command '['timeout', '-v', '--preserve-status', '600s', 'mpiexec', '--ppn', '2', '-n', '2', '--cpu-bind', 'list:104:1-103', '--env=LD_PRELOAD=libgeopm.so.2.2.0', 'geopmbench', 'test_trace_app.config']' returned non-zero exit status 127.

----------------------------------------------------------------------
Ran 0 tests in 9.067s

FAILED (errors=1)
```

**Error messages produced**
[...]

PR #3892, opened 2025-06-05
Update geopmctl and launcher.py to only choose cores that are allowed based on the current cgroup policy

bgeltz:
- Relates to #3869.

### DIFF
--- a/geopmpy/geopmpy/launcher.py
+++ b/geopmpy/geopmpy/launcher.py
@@ -106,6 +106,53 @@ def range_str(values):
     return ','.join(result)
 
 
+def _get_cpuset():
+    """
+    Returns a list of valid CPUs for the current process.
+    Ported from C++ Helper::get_cpuset().
+    """
+    proc_path = '/proc/self/cpuset'
+    try:
+        with open(proc_path) as f:
+            cpuset_cgroup = f.read().strip()
+    except Exception as ex:
+        raise RuntimeError(f'<geopm> geopmpy.launcher: Failed to read {proc_path}: {ex}')
+
+    possible_paths = [
+        f'/sys/fs/cgroup/cpuset{cpuset_cgroup}/cpuset.effective_cpus',
+        f'/sys/fs/cgroup{cpuset_cgroup}/cpuset.cpus.effective'
+    ]
+
+    cpus_content = None
+    cpuset_file = None
+    for path in possible_paths:
+        try:
+            with open(path) as f:
+                cpus_content = f.read().strip()
+            cpuset_file = path
+            break
+        except Exception:
+            continue
+
+    if cpus_content is None:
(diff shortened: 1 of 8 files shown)


libgeopm/src/ApplicationSampler.cpp
```diff
@@ -467,7 +468,8 @@ namespace geopm
             }
         }
         for (int core_idx = num_core - 1; core_idx != -1; --core_idx) {
-            if (!is_core_active.at(core_idx)) {
+            if (!is_core_active.at(core_idx) &&
+                (std::find(allowed_cpus.begin(), allowed_cpus.end(), core_idx) != allowed_cpus.end())) {
```
cmcantalupo: Is comparing core_idx to a cpu_idx appropriate?

geopmpy/geopmpy/launcher.py
```diff
@@ -106,6 +106,53 @@ def range_str(values):
...
+                cpus_content = f.read().strip()
+            cpuset_file = path
+            break
+        except Exception:
+            continue
+
+    if cpus_content is None:
+        raise RuntimeError('<geopm> geopmpy.launcher: Failed to read cpuset.cpus.effective from any expected location')
+
+    result = []
```
cmcantalupo, line 140: maybe result should be a set here too?

geopmpy/geopmpy/launcher.py
```diff
@@ -701,6 +752,11 @@ def affinity_list(self, is_geopmctl):
         app_cpu_per_node = app_rank_per_node * self.cpu_per_rank
         # Total number of cores per node
         core_per_node = self.core_per_socket * self.num_socket
+        allowed_cpuset = get_cpuset()
```
cmcantalupo: This line would actually become:

```
allowed_cpuset = list(get_cpuset())
```
to enable indexing in this use case...so

cmcantalupo (commented):
Maybe change the output type for get_cpuset() in python to also be a set?

geopmpy/geopmpy/launcher.py
```diff
@@ -106,6 +106,53 @@ def range_str(values):
     return ','.join(result)
 
 
+def get_cpuset():
+    """
+    Returns a list of valid CPUs for the current process.
+    Ported from C++ Helper::get_cpuset().
```
cmcantalupo, line 112: Would be nice to have google doc style "Returns: " with the type

cmcantalupo: Nevermind.  This is not done elsewhere in this file.  Better to stay with local style i guess

cmcantalupo (approved):
LGTM.  The `del launcher` pattern in the unit test seems a little weird, but it works.  I kind of wonder about increasing the number of test fixtures and then adding mock decorators to each test fixture?

cmcantalupo (approved):

PR #3895, opened 2025-06-09
Use both all and single types of level zero properties

cmcantalupo:


cmcantalupo:
Superseded by https://github.com/geopm/geopm/pull/3910

PR #3904, opened 2025-06-11
get_cpuset: Handle cases when cpuset does not exist

bgeltz:
- Relates to #3869.
- Refactor to fallback to using all cores when cpuset files are not present.

### DIFF
--- a/geopmpy/geopmpy/launcher.py
+++ b/geopmpy/geopmpy/launcher.py
@@ -31,7 +31,7 @@ import tempfile
 
 from collections import OrderedDict
 from . import __version__
-
+from geopmdpy import topo
 
 class Factory(object):
     def __init__(self):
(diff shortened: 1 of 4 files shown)


libgeopmd/src/Helper.cpp
```diff
@@ -463,70 +463,53 @@ namespace geopm
     std::set<int> get_cpuset(int pid)
```
bgeltz: Add get_cpuset_fallback that the Controller uses.
Restore the original implementation of get_cpuset where it will throw if /proc/self/cpuset is unavailable.
Catch the exception in the new fallback function and use topo as it is here to provide the full range of CPUs.

geopmpy/geopmpy/launcher.py
```diff
@@ -111,45 +111,46 @@ def _get_cpuset():
...
+
+                    # Parse CPU ranges and build result list
+                    for part in cpus_content.split(','):
+                        if '-' in part:  # CPU range (e.g., "0-207")
+                            start, end = map(int, part.split('-'))
+                            result.extend(range(start, end + 1))
+                        elif part:  # Single CPU number
+                            result.append(int(part))
+
+                    found_cpuset = True
```
bgeltz: Try to remove found_cpuset.

cmcantalupo (approved):

PR #3915, opened 2025-06-24
fixup rollover for powercap and msr drivers

### LINKED ISSUE #3914: Roll over for the RAPL energy counter is not managed for the powercap driver interface
**Describe the bug**
When using the MSR interface to support CPU energy, the implementation handles counter rollover, however the powercap implementation does not.

**GEOPM version**
v3.2.0

**Expected behavior**
The CPU_ENERGY signal should have the same behavior for both driver implementations.

**Actual behavior**
If the msr device driver is used roll over is handled, but for powercap it is not.

**Error messages produced**
None.

**Additional context**
https://github.com/cmcantalupo/kepler/pull/1#discussion_r2160549295

cmcantalupo:
- Fixes #3914
- Also fixes rollover for MSR signals through read_signal interface (reuse RolloverGenerator)

### DIFF
--- a/libgeopmd/src/CpuinfoIOGroup.cpp
+++ b/libgeopmd/src/CpuinfoIOGroup.cpp
@@ -156,6 +156,11 @@ namespace geopm
         return false;
     }
 
+    bool CpuinfoIOGroup::is_valid_domain(int domain_type) const
+    {
+        return platform_topo().is_nested_domain(GEOPM_DOMAIN_CPU, domain_type);
+    }
+
     int CpuinfoIOGroup::signal_domain_type(const std::string &signal_name) const
     {
         int result = GEOPM_DOMAIN_INVALID;
@@ -180,13 +185,13 @@ namespace geopm
     int CpuinfoIOGroup::push_signal(const std::string &signal_name, int domain_type, int domain_idx)
     {
         if (!is_valid_signal(signal_name)) {
-            throw Exception("CpuinfoIOGroup::push_signal(): " + signal_name +
-                            "not valid for CpuinfoIOGroup",
+            throw Exception("CpuinfoIOGroup::push_signal(): \"" + signal_name +
+                            "\" not valid for CpuinfoIOGroup",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        else if (domain_type != GEOPM_DOMAIN_BOARD) {
-            throw Exception("CpuinfoIOGroup::push_signal(): domain_type " + std::to_string(domain_type) +
-                            "not valid for CpuinfoIOGroup",
+        if (!is_valid_domain(domain_type)) {
+            throw Exception("CpuinfoIOGroup::push_signal(): domain type \"" + std::to_string(domain_type) +
+                            "\" not valid for CpuinfoIOGroup",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
         return std::distance(m_signal_available.begin(), m_signal_available.find(signal_name));
@@ -231,13 +236,13 @@ namespace geopm
     double CpuinfoIOGroup::read_signal(const std::string &signal_name, int domain_type, int domain_idx)
     {
         if (!is_valid_signal(signal_name)) {
-            throw Exception("CpuinfoIOGroup::read_signal(): " + signal_name +
-                            "not valid for CpuinfoIOGroup",
+            throw Exception("CpuinfoIOGroup::read_signal(): \"" + signal_name +
+                            "\" is not valid for CpuinfoIOGroup",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        else if (domain_type != GEOPM_DOMAIN_BOARD) {
-            throw Exception("CpuinfoIOGroup:read_signal(): domain_type " + std::to_string(domain_type) +
-                            "not valid for CpuinfoIOGroup",
+        if (!is_valid_domain(domain_type)) {
+            throw Exception("CpuinfoIOGroup::read_signal(): domain type \"" + std::to_string(domain_type) +
+                            "\" not valid for CpuinfoIOGroup",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
         return m_signal_available.find(signal_name)->second.value;
@@ -293,7 +298,7 @@ namespace geopm
             result =  "    description: " + it->second.description + '\n'; // Includes alias_for if applicable
             result += "    units: " + IOGroup::units_to_string(it->second.units) + '\n';
             result += "    aggregation: " + Agg::function_to_name(it->second.agg_function) + '\n';
-            result += "    domain: " + platform_topo().domain_type_to_name(GEOPM_DOMAIN_BOARD) + '\n';
+            result += "    domain: " + platform_topo().domain_type_to_name(GEOPM_DOMAIN_CPU) + '\n';
             result += "    iogroup: CpuinfoIOGroup";
         }
 #ifdef GEOPM_DEBUG
--- a/libgeopmd/src/MSRFieldSignal.cpp
+++ b/libgeopmd/src/MSRFieldSignal.cpp
@@ -11,6 +11,7 @@
 #include "geopm_debug.hpp"
 #include "geopm/Helper.hpp"
 #include "MSR.hpp"  // for enums
+#include "RolloverGenerator.hpp"
 
 namespace geopm
 {
@@ -23,13 +24,12 @@ namespace geopm
         , m_shift(begin_bit)
         , m_num_bit(end_bit - begin_bit + 1)
         , m_mask(((1ULL << m_num_bit) - 1) << begin_bit)
-        , m_subfield_max((1ULL << m_num_bit) - 1)
         , m_function(function)
         , m_scalar(scalar)
-        , m_last_field(0)
-        , m_num_overflow(0)
         , m_is_batch_ready(false)
+        , m_rollover_gen(std::make_shared<RolloverGenerator>())
     {
+        m_rollover_gen->set_factor(static_cast<double>((1ULL << m_num_bit) - 1) + 1.0);
         /// @todo: some of these are not logic errors if MSR data
         /// comes from user input files or if this interface is
         /// public. Alternatively, checks for these at the json
@@ -52,13 +52,10 @@ namespace geopm
         }
     }
 
-    double MSRFieldSignal::convert_raw_value(double val,
-                                     uint64_t &last_field,
-                                     int &num_overflow) const
+    double MSRFieldSignal::convert_raw_value(double val) const
     {
         uint64_t field = geopm_signal_to_field(val);
         uint64_t subfield = (field & m_mask) >> m_shift;
-        uint64_t subfield_last = (last_field & m_mask) >> m_shift;
         double result = NAN;
 
         uint64_t float_y, float_z;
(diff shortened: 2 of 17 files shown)


cmcantalupo:
🤖 

Suggestions

    Consider initializing last_energy on first use to prevent a false overflow:
    C++

    if (overflow_count == 0 && last_energy == 0) {
        last_energy = current_energy;
    }

    Optionally, clarify in comments how thread-local storage is intended to be used.

Summary:
Your rollover logic is sound and should work correctly for unsigned counters. The main thing to double-check is initialization of last_energy to avoid a spurious overflow on the first sample, and to ensure that thread-local storage matches your usage patterns.

Let me know if you want a more detailed code suggestion or review of the surrounding context!

libgeopmd/src/CpuinfoIOGroup.cpp
```diff
@@ -184,11 +184,6 @@ namespace geopm
                             "not valid for CpuinfoIOGroup",
                             GEOPM_ERROR_INVALID, __FILE__, __LINE__);
         }
-        else if (domain_type != GEOPM_DOMAIN_BOARD) {
```
bgeltz, line 187: Seems the docu already had this set to CPU:
https://geopm.github.io/geopm_pio_cpuinfo.7.html

Since these were previously only allowed to be read at the BOARD level, we should test to make sure the other domains are working as expected re: aggregation.

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -105,7 +107,18 @@ namespace geopm
         : M_PROPERTIES{SysfsDriver::parse_properties_json(plugin_name(), powercap_sysfs_json())}
         , M_POWERCAP_RESOURCE_BY_NAME(load_powercap_resource_by_name(powercap_directory))
         , M_POWERCAP_DIRECTORY(powercap_directory)
+        , m_rollover_factor(0.0)
     {
+        try {
+            std::string factor_path = POWERCAP_DIRECTORY;
+            factor_path += "/intel-rapl/intel-rapl:0/max_energy_range_uj";
```
bgeltz: Can't you get at this with ``M_PROPERTIES``, ``name_to_property_key()``, and the newly defined ``MAX_ENERGY_RANGE`` instead of duplicating the JSON data here?

libgeopmd/src/CpuinfoIOGroup.cpp
cmcantalupo: Yeah this check is *actually* the responsibility of PlatformIO.  It only ever makes requests of an IOGroup at the native domain for a signal or control.  These exceptions were both incorrect (board vs CPU) and also redundant.

libgeopmd/src/RolloverGenerator.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <string>
+#include <cmath>
+#include "RolloverGenerator.hpp"
+#include "geopm/Exception.hpp"
+
+namespace geopm
+{
+    RolloverGenerator::RolloverGenerator(int bit_width)
+        : m_last_value(0.0)
+        , m_rollover_factor(pow(2, bit_width))
```
bgeltz: Why set ``m_rollover_factor`` in the initializer list, and also provide the ``set_factor()`` method?  Why both?

libgeopmd/src/CpuinfoIOGroup.cpp
cmcantalupo: Checked, and LGTM, all the signals are frequencies

libgeopmd/src/RolloverGenerator.cpp
cmcantalupo: The constructor takes a bit width, but in the case of powercap, we only get the max value in microjoules.  We could definitly consider having the constructor take the rollover factor directly.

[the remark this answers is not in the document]
cmcantalupo: doesn't work for bit_width == 64, but yet another reason to put get rid of the bit_width constructor.

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -150,10 +162,18 @@ namespace geopm
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
         double scaling_factor = prop_it->second.scaling_factor;
-        return [scaling_factor](const std::string &content) {
+        std::shared_ptr<RolloverGenerator> rollover_ptr = nullptr;
+        if (signal_name.find("ENERGY") != std::string::npos) {
+            rollover_ptr = std::make_shared<RolloverGenerator>(64);
+            rollover_ptr->set_factor(m_rollover_factor);
+        }
+        return [scaling_factor, rollover_ptr](const std::string &content) {
```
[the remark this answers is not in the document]
cmcantalupo: we want a copy...rollover_ptr will go out of scope. I'm not sure if we can std::move() in the capture?

libgeopmd/src/RolloverGenerator.cpp
```diff
@@ -0,0 +1,33 @@
...
+    {
+        m_rollover_factor = rollover_factor;
+    }
+
+    double RolloverGenerator::update(double value)
+    {
+        if (m_last_value > value) {
+            ++m_num_overflow;
+        }
+        double rollover = m_rollover_factor * m_num_overflow;
```
bgeltz: This could be inside the if statement above as a different member variable to avoid running this calculation during every update() call.

libgeopmd/src/PowercapSysfsDriver.hpp
```diff
@@ -6,11 +6,15 @@
 #ifndef POWERCAPSYSFSDRIVER_HPP_INCLUDE
 #define POWERCAPSYSFSDRIVER_HPP_INCLUDE
 
+#include <memory>
+#include <vector>
```
bgeltz:
```suggestion
#include <memory>
#include <map>
```

libgeopmd/test/MSRFieldSignalTest.cpp
```diff
@@ -143,19 +143,19 @@ TEST_F(MSRFieldSignalTest, read_overflow)
     result = sig->read();
```
bgeltz, line 144: Comment on 139 should be removed.

libgeopmd/test/RolloverGeneratorTest.cpp
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright (c) 2015 - 2025 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include <RolloverGenerator.hpp>
+
+
+#include "geopm/SaveControl.hpp"
+
+#include <iostream>
```
bgeltz:
```suggestion
```

libgeopmd/test/RolloverGeneratorTest.cpp
```diff
@@ -0,0 +1,98 @@
...
+        int value = idx % 2;
+        EXPECT_EQ(idx, gen.update(value)) << "One bit counter";
+    }
+}
+
+TEST_F(RolloverGeneratorTest, test_two_bit_width)
+{
+    auto gen = RolloverGenerator();
+    gen.set_factor(pow(2, 2));
+    for (int idx = 0; idx < 4; ++idx) {
```
bgeltz:
```suggestion
    for (int idx = 0; idx < 8; ++idx) {
```

libgeopmd/test/RolloverGeneratorTest.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+TEST_F(RolloverGeneratorTest, test_32_bit_width)
+{
+    auto gen = RolloverGenerator();
+    gen.set_factor(pow(2, 32));
+    EXPECT_EQ(1, gen.update(1));
+    double rollover_factor = static_cast<double>(UINT32_MAX) + 1.0;
+    EXPECT_EQ(rollover_factor, gen.update(0));
+    EXPECT_EQ(rollover_factor + 1.0, gen.update(1));
+    EXPECT_NEAR(2.0 * rollover_factor + 1.0, gen.update(0), 1);
```
bgeltz: Could this be EXPECT_EQ?

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -150,10 +174,18 @@ namespace geopm
                             GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
         }
         double scaling_factor = prop_it->second.scaling_factor;
-        return [scaling_factor](const std::string &content) {
+        std::shared_ptr<RolloverGenerator> rollover_ptr = nullptr;
+        if (signal_name.find("ENERGY") != std::string::npos) {
+            rollover_ptr = std::make_shared<RolloverGenerator>();
+            rollover_ptr->set_factor(m_rollover_factor);
+        }
+        return [scaling_factor, rollover_ptr = std::move(rollover_ptr)](const std::string &content) {
```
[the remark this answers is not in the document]
cmcantalupo: there is one lambda per signal and per domain

libgeopmd/test/RolloverGeneratorTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    auto gen = RolloverGenerator();
+    gen.set_factor(pow(2, 64));
+    double rollover_factor = static_cast<double>(UINT64_MAX) + 1.0;
+    EXPECT_EQ(1, gen.update(1));
+    EXPECT_EQ(rollover_factor, gen.update(0));
+    EXPECT_EQ(rollover_factor + 1.0, gen.update(1));
+    double expected = 2.0 * rollover_factor;
+    // IEEE 754 double precision floating point mantissa bit width is 51
+    double error = 1 << (64 - 51);
```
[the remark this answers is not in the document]
cmcantalupo: This is a test....

libgeopmd/src/MSRFieldSignal.cpp
```diff
@@ -75,10 +79,7 @@ namespace geopm
                 result = (1ULL << float_y) * (1.0 + float_z / 4.0);
                 break;
             case MSR::M_FUNCTION_OVERFLOW:
-                if (subfield_last > subfield) {
-                    ++num_overflow;
-                }
-                result = subfield + ((m_subfield_max + 1.0) * num_overflow);
+                result = m_rollover_gen->update(subfield);
```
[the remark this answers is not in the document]
cmcantalupo: Copilot seems to be confused.  There is one MSRFieldSignal per subfield per msr per cpu

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -105,7 +108,28 @@ namespace geopm
...
+         std::string contents;
+         try {
+             contents = geopm::read_file(factor_path);
+         }
+         catch (const Exception &ex) {
+             throw geopm::Exception("PowercapSysfsDriver: Unable to read RAPL rollover file from sysfs: \"" + factor_path + "\"",
+                                    GEOPM_ERROR_RUNTIME, __FILE__, __LINE__);
+         }
+         try {
+             m_rollover_factor = 1e-6 * std::stoll(contents);
```
bgeltz: A comment explaining where the magic ``1e-6`` came from would be helpful.  Is it possible to put a link to the upstream docu for this sysfs entry here?
Also, should this be a constexpr instead of hard coded here?

bgeltz (dismissed):

libgeopmd/src/PowercapSysfsDriver.cpp
```diff
@@ -105,7 +108,35 @@ namespace geopm
         : M_PROPERTIES{SysfsDriver::parse_properties_json(plugin_name(), powercap_sysfs_json())}
         , M_POWERCAP_RESOURCE_BY_NAME(load_powercap_resource_by_name(powercap_directory))
         , M_POWERCAP_DIRECTORY(powercap_directory)
+        , m_rollover_factor(0.0)
     {
+        std::string signal_name = "POWERCAP::CPU_MAX_ENERGY_RANGE";
+        std::string factor_path = attribute_path(signal_name, 0);
+        std::string property_key = name_to_property_key(signal_name);
```
bgeltz, line 115:
```suggestion
        std::string property_key = name_to_property_key(signal_name);
        std::string contents;
```

bgeltz (approved):

Issue #3923, opened 2025-07-16
Complete Spack PR review for v3.2.0


bgeltz:
https://github.com/spack/spack-packages/pull/266

Done criteria:
- [x] All review comments resolved
- [x] PR is accepted by reviewer

PR #3924, opened 2025-07-17
Preserve ACLOCAL_PATH if it was already set

bgeltz:
- Relates to #3923.

### DIFF
--- a/libgeopmd/autogen.sh
+++ b/libgeopmd/autogen.sh
@@ -14,4 +14,4 @@ if [ ! -e VERSION ]; then
 	set -x
     fi
 fi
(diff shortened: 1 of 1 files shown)


libgeopm/autogen.sh
```diff
@@ -12,4 +12,4 @@ if [ ! -e VERSION ]; then
         echo "0.0.0" > VERSION
     fi
 fi
-autoreconf -i -f
```
cmcantalupo, line 15: Not needed for libgeopm this hack is about missing pkgconfig files for grpc

cmcantalupo (approved):

PR #3943, opened 2025-08-20
Add skip decorators for optimization module

cmcantalupo:
Fixes RPM builds in environments where scikit is only available via pip


sidjana:
LGTM!

sidjana (approved):

PR #3961, opened 2025-09-04
Reset pio when last client closes

### LINKED ISSUE #3925: geopmd leaves file handles open after session cleanup
**Describe the bug**
I tried to use ``nvidia-smi -r`` to reset the GPUs after using ``geopmread`` and I expected it to work instead it failed.

**GEOPM version**
3.2

**Expected behavior**
```
node1:~ # nvidia-smi -r
GPU 00000000:42:00.0 was successfully reset.
All done.
```

**Actual behavior**
```
node1:~ # nvidia-smi -r
The following GPUs could not be reset:
  GPU 00000000:42:00.0: In use by another client

1 device is currently being used by one or more other processes (e.g., Fabric Manager, CUDA application, graphics application such as an X server, or a monitoring application such as another instance of nvidia-smi). Please first kill all processes using this device and all compute applications running in the system.
```

**Additional context**
``lsof`` is showing ``geopmd`` holding on to handles in the failure case:
```
...
geopmd  5135 root  15u      CHR            195,255      0t0      535 /dev/nvidiactl
geopmd  5135 root  16u      CHR              195,0      0t0      539 /dev/nvidia0
geopmd  5135 root  17u      CHR              235,0      0t0      562 /dev/nvidia-uvm
geopmd  5135 root  18r      CHR              238,1      0t0      558 /dev/nvidia-caps/nvidia-cap1
geopmd  5135 root  19r      CHR              238,2      0t0      560 /dev/nvidia-caps/nvidia-cap2
geopmd  5135 root  20u      CHR              235,0      0t0      562 /dev/nvidia-uvm
geopmd  5135 root  21u      CHR              195,0      0t0      539 /dev/nvidia0
geopmd  5135 root  22u      CHR              195,0      0t0      539 /dev/nvidia0
...
```

Old issue with similar issues from LevelZero: https://github.com/geopm/geopm/issues/2956

cmcantalupo:
- Fixes #3925

### DIFF
--- a/geopmdpy/geopmdpy/service.py
+++ b/geopmdpy/geopmdpy/service.py
@@ -522,7 +522,7 @@ class PlatformService(object):
             with system_files.WriteLock(self._RUN_PATH, self._PROFILER_LOCK_NAME) as lock:
                 if lock.try_lock() == client_pid:
                     lock.unlock(client_pid)
-        if start_num_client != 0 and self._active_sessions.get_clients() == 0:
+        if start_num_client != 0 and len(self._active_sessions.get_clients()) == 0:
             self._pio.reset()
 
     def _close_session_write(self, lock, pid):
--- a/geopmdpy/geopmdpy/service.py
+++ b/geopmdpy/geopmdpy/service.py
@@ -498,7 +498,9 @@ class PlatformService(object):
             client_pid (int): Linux PID of the client thread
 
         """
-        if client_pid in self._active_sessions.get_clients():
+        all_clients = self._active_sessions.get_clients()
+        start_num_client = len(all_clients)
+        if client_pid in all_clients:
             GLib.source_remove(self._active_sessions.get_watch_id(client_pid))
             batch_pid = self._active_sessions.get_batch_server(client_pid)
             if batch_pid is not None:
@@ -520,6 +522,8 @@ class PlatformService(object):
             with system_files.WriteLock(self._RUN_PATH, self._PROFILER_LOCK_NAME) as lock:
                 if lock.try_lock() == client_pid:
                     lock.unlock(client_pid)
+        if start_num_client != 0 and self._active_sessions.get_clients() == 0:
+            self._pio.reset()
 
     def _close_session_write(self, lock, pid):
         save_dir = os.path.join(self._RUN_PATH, self._SAVE_DIR)


bgeltz:
This isn't working quite yet.  Testing on saruman as root with:
```
# lsof | grep /dev/nvidia | grep geopm
```

I see geopmd holding on to file handles once I do the following in a user terminal:
```
geopmwrite NVML::GPU_CORE_FREQUENCY_MAX_CONTROL gpu 0 1.4e9
```

After logging out of that session (emulating job end) and verifying that `/run/geopm/CONTROL_LOCK` was empty, I checked with `lsof` again and the files were still present.

bgeltz:
I'm able to see open files linger after a fresh service restart by just issuing:
```
$ geopmread NVML::GPU_CORE_FREQUENCY_MAX_CONTROL gpu 0
```

bgeltz:
This still isn't working, even with my fix for ``len(get_clients())``.  The main PID must terminate for these open files to be released properly.  I also tested this sort of change:
```
         if start_num_client != 0 and len(self._active_sessions.get_clients()) == 0:
-            self._pio.reset()
+            sys.exit()
```

Which DOES indeed fix the open files problem and tests fine when starting a test from a fresh service restart with ``geopmwrite``.

However, if you just try a ``geopmread`` this solution presents a new error:
```
$ geopmread NVML::GPU_CORE_FREQUENCY_MAX_CONTROL gpu 0
1530000000
  what():  <geopm> Runtime error: SDBus: Failed to call sd-bus function sd_bus_call_method(), error:-110 name: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying: at src/SDBus.cpp:31
Aborted (core dumped)
```
Seemingly this error is because the main service PID terminates before completing the ``platform_close_session()`` call in the ``ServiceIOGroup`` destructor.

AFAICT if we want to go down this path, we need to defer the ``sys.exit()`` until after all in-progress calls from the client are complete.  Sketch of what that could look like here:
https://github.com/bgeltz/geopm/commit/81053eadc5fac923a3cd134ec4fdf87ea1e94c19
https://github.com/bgeltz/geopm/commits/issue-3925/
Allegedly, ``Glib.idle_add`` would run a callback as soon as the current DBus method returns and the main loop goes back to idle.

PR #3969, opened 2025-09-16
Pin dependencies to avoid known vulnerabilities

cmcantalupo:


.github/workflows/codeql-analysis.yml
```diff
@@ -33,9 +33,11 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+  # Note: If Python dependencies are added to this workflow in future, install with '-c constraints.txt'
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
+      # github/codeql-action/init pinned to commit for v3.24.5
```
bgeltz, line 40: This is redundant.  There is already a comment for this on the next line.  All these generated comments should be removed.

.github/workflows/codeql-analysis.yml
```diff
@@ -47,7 +49,7 @@ jobs:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
     - name: install system dependencies
       run: sudo apt-get update && sudo apt-get install libelf-dev mpich libmpich-dev libomp-18-dev libsystemd-dev liburing-dev libgirepository1.0-dev libcap-dev zlib1g-dev autoconf automake libtool libsqlite3-dev cargo libgrpc++-dev libgrpc-dev libprotobuf-dev protobuf-compiler protobuf-compiler-grpc zstd
-    - name: configure libgeopmd dir
+  - name: configure libgeopmd dir
```
bgeltz, line 52: There was no problem here.

constraints.txt
bgeltz, line 1: Why do we need this file?  Shouldn't this be handled by the requirements.txt?

bgeltz (changes requested):
What about the setup.cfg, pyproject.toml, etc?  This is missing required changes.

PR #4038, opened 2026-06-24
GPUActivityAgent: fix policy file requirement and missing min freq control

### LINKED ISSUE #4037: GPUActivityAgent: policy file required even with all-default policy; min frequency control assumed present
## Summary

Two bugs prevent the `gpu_activity` agent from running on Battlemage (Intel Xe2) GPU systems without extra workarounds.

---

### Description

The `Controller` constructor throws the following error when `GEOPM_POLICY` is not set and no endpoint is configured:

```
Controller(): at least one of policy or endpoint path must be provided.
```

This is enforced unconditionally whenever an agent has any policy values (`m_num_send_down > 0`). However, `GPUActivityAgent` has a single policy value (`GPU_PHI`) with a well-defined default of `0.5`, and `validate_policy()` already converts `NaN → 0.5`. There is no reason for a policy file to be mandatory in this case.

### Expected behavior

If `validate_policy()` can convert an all-`NaN` policy vector to a fully-specified policy with no `NaN` values remaining, the agent should be usable without specifying a policy file. The controller should start with an all-`NaN` policy and let `validate_policy()` fill in the defaults on the first iteration.

---

### Description

`init_platform_io()` unconditionally calls:

```cpp
m_platform_io.control_domain_type("GPU_CORE_FREQUENCY_MIN_CONTROL");
// ...
m_platform_io.push_control("GPU_CORE_FREQUENCY_MIN_CONTROL", ...);
```

Battlemage (Intel Xe2 / Level Zero) does not expose a minimum GPU core frequency control — only a maximum frequency control is available. This causes `init_platform_io()` to throw when the agent is initialized on that hardware.

Additionally, `adjust_platform()` accesses `m_gpu_freq_min_control` unconditionally, which would cause an out-of-bounds access if the min control vector were empty.

### Expected behavior

The agent should check whether `GPU_CORE_FREQUENCY_MIN_CONTROL` is present in `control_names()` before attempting to use it. If absent, the agent should skip pushing the min control and only write `GPU_CORE_FREQUENCY_MAX_CONTROL` during `adjust_platform()`.

---

## Affected hardware

- Intel Battlemage (Xe2) GPUs via Level Zero

## Proposed fix

- **`Controller.cpp`**: Before throwing the "policy file required" error, probe the agent by calling `validate_policy()` on an all-`NaN` vector. If it succeeds and leaves no `NaN` values, allow the controller to proceed without a policy file.
- **`GPUActivityAgent.hpp/cpp`**: Add a `m_has_freq_min_control` boolean. In `init_platform_io()`, check `control_names()` and conditionally push the min control. In `adjust_platform()`, guard all accesses to `m_gpu_freq_min_control` behind this flag.

cmcantalupo:
Two bugs prevented the gpu_activity agent from running on Battlemage (Intel Xe2) GPU systems:

1. Policy file required even when all policy values have defaults

   The Controller constructor threw an error when GEOPM_POLICY was unset, even though GPUActivityAgent's only policy value (GPU_PHI) has a well-defined default of 0.5 and validate_policy() already converts NaN to that default.  The check is relaxed: before throwing, the Controller calls validate_policy() on an all-NaN trial vector; if every entry is filled in, the agent is allowed to run without a policy file.

2. GPU_CORE_FREQUENCY_MIN_CONTROL assumed to always be present

   init_platform_io() unconditionally called control_domain_type() and push_control() for GPU_CORE_FREQUENCY_MIN_CONTROL, which is not exposed by Battlemage (Level Zero only provides a max frequency control).  A new boolean member m_has_freq_min_control is set by checking control_names() at init time.  The min control is only pushed and written when it is available; otherwise only the max frequency control is used.

A new unit test (adjust_platform_no_min_control) verifies the min-control-absent path.

- Fixes #4037

### DIFF
--- a/libgeopm/src/Controller.cpp
+++ b/libgeopm/src/Controller.cpp
@@ -261,9 +261,21 @@ namespace geopm
         , m_do_restore(false)
     {
         if (m_num_send_down > 0 && !(m_do_policy || m_do_endpoint)) {
-            throw Exception("Controller(): at least one of policy or endpoint path"
-                            " must be provided.", GEOPM_ERROR_INVALID,
-                            __FILE__, __LINE__);
+            // Allow proceeding without a policy file if validate_policy() can
+            // fill every NaN entry with a default value.
+            std::vector<double> trial(m_num_send_down, NAN);
+            bool has_all_defaults = false;
+            try {
+                m_agent[0]->validate_policy(trial);
+                has_all_defaults = std::none_of(trial.begin(), trial.end(),
+                                                [](double v) { return std::isnan(v); });
+            }
+            catch (...) {}
+            if (!has_all_defaults) {
+                throw Exception("Controller(): at least one of policy or endpoint path"
+                                " must be provided.", GEOPM_ERROR_INVALID,
+                                __FILE__, __LINE__);
+            }
         }
         // Three dimensional vector over levels, children, and message
         // index.  These are used as temporary storage when passing
--- a/libgeopm/src/GPUActivityAgent.cpp
+++ b/libgeopm/src/GPUActivityAgent.cpp
@@ -44,6 +44,7 @@ namespace geopm
         , M_NUM_CHIP_PER_GPU(M_NUM_GPU == 0 ? 0 : M_NUM_GPU_CHIP / M_NUM_GPU)
         , m_do_write_batch(false)
         , m_do_send_policy(true)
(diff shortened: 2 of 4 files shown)


cmcantalupo:
Back out min control change, it will lead to confusing errors if access list is not set up properly.

bgeltz (approved):

PR #4039, opened 2026-06-26
Add pure-Python gpu_activity session agent for non-HPC deployment

cmcantalupo:
## Summary

This branch ports the C++ `GPUActivityAgent` to the `geopmsession` Python `Agent`
interface so the `gpu_activity` control logic can run **outside the GEOPM HPC
Runtime** — no Controller, no MPI/OpenMP interposition, and no DBus profile
registration. It also relocates the existing pure-Python session agents into
`geopmdpy` so they live next to the interface they depend on.

The agent reimplements only the leaf-level control logic of the C++ agent on top
of `geopmdpy.pio`; the HPC-only tree methods (`split_policy` / `aggregate_sample`
/ ascend / descend) and the agent `Waiter` are dropped, with the session owning
loop cadence via `-p/--period`.

Runnable as:

```
python -m geopmdpy.gpu_activity_agent -t 30 -p 0.02 --phi 0.5
```

## Commits

1. **`geopmpy: add pure-Python gpu_activity session agent`**
   - New `GPUActivityAgent(session.Agent)` implementing the `gpu_activity`
     control loop in Python.
   - `--phi` CLI argument (default `0.5`, validated to `[0.0, 1.0]`) replaces the
     `GPU_PHI` policy value.
   - `run_begin()` resolves the coarsest shared GPU/GPU_CHIP domain, pushes the
     activity/utilization signals and GPU core frequency min/max controls, reads
     the static frequency characterization (CONST_CONFIG → LEVELZERO → midpoint
     fallback), and saves controls for restoration.
   - `update_loop()` applies the phi-biased efficient-to-max interpolation,
     activity/utilization scaling, and clamping, writing the frequency controls
     only when the request changes.
   - `run_end()` prints a summary mirroring the C++ `report_host` fields.

2. **`Derive GPU activity from DRM idle residency when Level Zero is absent`**
   - The Level Zero `GPU_CORE_ACTIVITY` / `GPU_UTILIZATION` signals are not
     exposed by the DRM driver (e.g. the `xe` driver on Battlemage).
   - `run_begin()` now detects the available activity source: prefer
     `GPU_CORE_ACTIVITY`, otherwise derive a GPU busy fraction from the monotonic
     `DRM::IDLE_RESIDENCY` counter as `busy = 1 - delta_idle / delta_time`,
     clamped to `[0, 1]` and treated as fully utilized.
   - Domain resolution, signal push, and the trace config override only reference
     signals that actually exist.

3. **`geopmdpy: move monitor and gpu_activity session agents into geopmdpy`**
   - `geopmpy/monitor.py` → `geopmdpy/monitor_agent.py`
   - `geopmpy/gpu_activity.py` → `geopmdpy/gpu_activity_agent.py`
   - These agents depend only on `geopmdpy` (`pio`, `topo`, `session`,
     `exporter`), so hosting them there removes the cross-package dependency.
   - Internal imports switched to the relative style used elsewhere in
     `geopmdpy`; docstring module-path references updated accordingly. Both
     packages auto-include modules via `packages = <pkg>`, so no setup changes
     are required.

## Out of scope

Ports of the `power_balancer` and `frequency_balancer` agents, which depend on
the multi-node tree model that has no meaning in the single-node session.

## Testing

- Both modules byte-compile and import from their new `geopmdpy` locations.
- `python -m geopmdpy.gpu_activity_agent --phi 2.0` fails fast with the range
  validation error; `--help` shows `--phi` and the agent epilog.
- `python -m geopmdpy.monitor_agent --help` shows `--hi-res`.


bgeltz:
Future todo: test on non-PVC system to evaluate DRM::IDLE_RESIDENCY when GPU_CORE_ACTIVITY is not available.

geopmdpy/integration_test/apps/gpu_activity_benchmark.cpp
```diff
@@ -0,0 +1,261 @@
...
+                  << "\n";
+        if (args.profile == "decode") {
+            return run_decode_profile(queue, args);
+        }
+        return run_compute_profile(queue, args);
+    }
+    catch (const std::exception &ex) {
+        std::cerr << "gpu_activity_benchmark error: " << ex.what() << "\n";
+        return 1;
+    }
```
bgeltz, line 284: Please add comments explaining what this benchmark is doing without the reader having to parse the source.  It's not clear.

geopmdpy/integration_test/README.md
```diff
@@ -0,0 +1,168 @@
...
+# expect: FOM (images/sec): <n>
+```
+
+If the compiler is not on `PATH`, set `GEOPM_GPU_BENCH_CXX` to the compiler
+path.  The resulting binary is cached under `integration_test/apps/build/` by
+default; set `GEOPM_GPU_BENCH_BUILD_DIR` to place it elsewhere.
+
+The older PyTorch/container drivers are still present for optional experiments,
+but they are not used by the integration tests by default and require a local
+image cache or native PyTorch/IPEX install.
```
bgeltz: What older drivers?  Examine this comment and others like it that reference older things and remove or refactor the sentence to be in the present tense.  This code is all new.

geopmdpy/integration_test/apps/gpu_activity_benchmark.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+#include <chrono>
+#include <cmath>
+#include <cstdlib>
+#include <cstring>
+#include <iostream>
+#include <stdexcept>
+#include <string>
+#include <thread>
+#include <vector>
```
bgeltz:
```suggestion
```
Does not appear to be used.
Check other includes.

geopmdpy/integration_test/apps/gpu_activity_benchmark.cpp
bgeltz: Are all the magic numbers that are used to setup the various data structures specific to PVC?  Do you expect modifications to be needed for different HW with more/less GPU memory, etc.?

geopmdpy/integration_test/apps/build_gpu_activity_benchmark.sh
bgeltz, line 1: Any reason this is not a makefile?

geopmdpy/integration_test/README.md
```diff
@@ -0,0 +1,168 @@
...
+| `test_phi1_energy_saving_extreme` | 1.0 | GPU energy below baseline (perf harm not asserted) |
+
+> **phi semantics:** `phi < 0.5` biases toward performance (at `phi = 0`
+> frequency is pinned at F_max — no harm); `phi > 0.5` biases toward energy
+> savings (at `phi = 1` frequency is pinned at F_efficient); `phi = 0.5` spans
+> the full range and is the only setting that dynamically tracks activity.
+
+## Prerequisites (one-time)
+
+Ensure the host has the Intel GPU kernel driver (`i915`/`xe`), Level Zero, and
```
bgeltz, line 81: I see the word "driver" used a bunch of places in this PR, but no where it it spelled out what it means in this context.  Here it's definitely a "kernel driver"  but there are other spots that talk about the "Pytorch driver"...

geopmdpy/integration_test/apps/run_workload.sh
```diff
@@ -0,0 +1,67 @@
...
+if [[ -n "${render_gid}" ]]; then
+    engine_args+=(--group-add "${render_gid}")
+fi
+
+if [[ -n "${GEOPM_GPU_SELINUX_DISABLE:-}" ]]; then
+    engine_args+=(--security-opt label=disable)
+fi
+
+exec "${ENGINE}" "${engine_args[@]}" "${IMAGE}" \
+    python "/apps/${DRIVER}" "$@"
```
bgeltz:
```suggestion
    python3 "/apps/${DRIVER}" "$@"
```
Is it safe with just python?
Even if yes, it's inconsistent with line 52.

geopmdpy/integration_test/apps/run_workload.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+IMAGE="${GEOPM_GPU_WORKLOAD_IMAGE:-intel/intel-extension-for-pytorch:2.8.10-xpu}"
+ENGINE="${GEOPM_GPU_CONTAINER_ENGINE:-docker}"
+
+if [[ "${GEOPM_GPU_WORKLOAD_NATIVE:-0}" == "1" ]]; then
+    exec python3 "${SCRIPT_DIR}/${DRIVER}" "$@"
+fi
+
+engine_args=(run --rm --device /dev/dri -v "${SCRIPT_DIR}:/apps:ro")
```
bgeltz, line 65: Please add a comment explaining why you're removing this device and what you're setting up here.

geopmdpy/integration_test/apps/run_workload.sh
```diff
@@ -0,0 +1,67 @@
...
+
+render_gid="$(getent group render | cut -d: -f3 || true)"
+if [[ -n "${render_gid}" ]]; then
+    engine_args+=(--group-add "${render_gid}")
+fi
+
+if [[ -n "${GEOPM_GPU_SELINUX_DISABLE:-}" ]]; then
+    engine_args+=(--security-opt label=disable)
```
bgeltz, line 75: Why is this necessary?  Comment

geopmdpy/integration_test/_util.py
```diff
@@ -0,0 +1,256 @@
...
+    return lambda obj: obj
+
+
+def skip_unless_workload():
+    """Class/method decorator: skip when the inference workload cannot run.
+
+    The default workload is a locally built SYCL/oneAPI benchmark.  It requires
+    either an existing benchmark binary or a SYCL compiler on ``PATH`` (or in
+    ``GEOPM_GPU_BENCH_CXX``).  The optional Python/container drivers still use
+    the old ``GEOPM_GPU_WORKLOAD_NATIVE`` / container-engine checks.
```
bgeltz: What are the "old checks"?
Isn't all this code new??

bgeltz (changes requested):

geopmdpy/integration_test/TestGPUActivityAgentInference.py
```diff
@@ -0,0 +1,350 @@
...
+    the GPU's own hardware DVFS.  Dynamic control and energy savings at
+    phi=0.5 are asserted.
+
+This is NOT a unit test: it requires a live GEOPM service, a GPU, and the
+inference workload.  It lives in the independent ``integration_test`` directory
+and is run explicitly.  See ``README.md`` in this directory.
+
+Environment overrides (all optional):
+  GEOPM_GPU_WORKLOAD_SEC     Workload timed duration, seconds (default 30).
+    GEOPM_GPU_BATCH_SIZE       Image-profile FoM scaling factor (default 64).
```
bgeltz:
```suggestion
  GEOPM_GPU_WORKLOAD_SEC     Workload timed duration, seconds (default 30).
  GEOPM_GPU_BATCH_SIZE       Image-profile FoM scaling factor (default 64).
```

Keep the help text aligned like this when possible.

geopmdpy/integration_test/TestGPUActivityAgentInference.py
```diff
@@ -0,0 +1,350 @@
...
+        if phi is not None:
+            agent_trace = os.path.join(cls._tmpdir, f'{tag}_agent.csv')
+            agent_report = os.path.join(cls._tmpdir, f'{tag}_agent.yaml')
+            agent_stderr_path = os.path.join(cls._tmpdir, f'{tag}_agent.stderr')
+            agent_cmd = _util.agent_command(
+                phi, cls._workload_sec + _AGENT_MARGIN_SEC, cls._period,
+                agent_trace, agent_report)
+            agent_err = open(agent_stderr_path, 'w')
+            agent_proc = subprocess.Popen(
+                agent_cmd, stdout=subprocess.DEVNULL, stderr=agent_err,
```
bgeltz, line 167: Is stdout going to /dev/null because the app has no useful output?

geopmdpy/integration_test/TestGPUActivityAgentInference.py
```diff
@@ -0,0 +1,350 @@
...
+    workload gives the agent nothing to exploit.
+
+``Serving`` (regime 1: idle gaps)
+    An over-provisioned online-serving profile that idles between
+    requests to a target duty cycle.  Compute activity oscillates, so the
+    agent drops the frequency during the gaps: dynamic control and energy
+    savings at phi=0.5 are asserted.
+
+``Decode`` (regime 2: frequency-insensitive but busy)
+    A batch-1, memory-bandwidth-bound autoregressive decode proxy (the LLM
```
bgeltz, line 35: What's batch-1? Make sure any terms used are clearly defined in comments.

bgeltz (changes requested):
Was this the PR where you had to pin the app to get the test passing regularly?  If yes, it would be good to add an example runscript.

geopmdpy/integration_test/apps/gpu_activity_benchmark.cpp
cmcantalupo: Done in de726797f — added a file-level comment describing the benchmark and its three profiles (steady/serving/decode), plus short comments on `run_compute_profile` and `run_decode_profile` explaining what each exercises.

cmcantalupo: No — the defaults are not PVC-specific. They are generic tuning knobs chosen to produce steady, measurable GPU activity, and every one is overridable via CLI flags and the `GEOPM_GPU_*` environment variables, so no source change is needed to retarget hardware with different memory sizes. I documented each `Args` field and added a note to that effect in de726797f.

geopmdpy/integration_test/README.md
cmcantalupo: Fixed in de726797f — rewrote the paragraph in the present tense with no "older"; it now just states the PyTorch/container drivers are optional alternatives to the default SYCL benchmark.

geopmdpy/integration_test/apps/gpu_activity_benchmark.cpp
cmcantalupo: Fixed in de726797f — removed `<vector>`, and while auditing the rest I also dropped the unused `<cstring>` and `<cmath>`.

geopmdpy/integration_test/apps/run_workload.sh
cmcantalupo: Added a comment in de726797f: these are the base container run args — expose the GPU render nodes (`/dev/dri`) to the container and bind-mount this `apps` directory read-only at `/apps` so the driver script is available inside.

geopmdpy/integration_test/README.md
cmcantalupo: Added a Terminology section in de726797f distinguishing the **kernel driver** (`i915`/`xe`) from the **workload driver** (the programs under `apps/`), and noting that unqualified uses of "driver" below mean the workload driver.

geopmdpy/integration_test/apps/run_workload.sh
cmcantalupo: Done in de726797f — changed to `python3` to match the native path on line 52.

geopmdpy/integration_test/apps/build_gpu_activity_benchmark.sh
cmcantalupo: Agreed — converted to `apps/Makefile` in f053bbbd6. The source-vs-binary staleness check is now an ordinary prerequisite; compiler auto-detection (`GEOPM_GPU_BENCH_CXX`, else `icpx`/`dpcpp` on `PATH`) and the `GEOPM_GPU_BENCH_BUILD_DIR` output override are preserved. The compile runs in the test harness (`setUpClass`) so `run_workload.sh` stays a pure launcher, and the old shell script is removed.

geopmdpy/integration_test/apps/run_workload.sh
cmcantalupo: Commented in de726797f — on SELinux hosts the read-only bind mount is relabeled by default, which can deny access under rootless podman; the opt-in `--security-opt label=disable` avoids that.

geopmdpy/integration_test/_util.py
cmcantalupo: Fixed in de726797f — dropped the "old" wording; the docstring now just says the optional Python/container drivers use the `GEOPM_GPU_WORKLOAD_NATIVE` / container-engine checks below.

geopmdpy/integration_test/TestGPUActivityAgentInference.py
cmcantalupo: Yes — the agent's stdout is only a human-readable end-of-run summary; the test consumes the trace/report files instead, so it is discarded. Added a comment saying so in de726797f (stderr is still captured to a file for diagnostics on failure).

geopmdpy/integration_test/TestGPUActivityAgentInference.py
cmcantalupo: Done in de726797f — fixed the `GEOPM_GPU_BATCH_SIZE` indentation so the env-var help lines align.

geopmdpy/integration_test/TestGPUActivityAgentInference.py
cmcantalupo: Defined inline in de726797f — "batch-1" means one inference request is processed at a time (batch size of one, no request batching), which is the latency-bound regime that under-fills the compute engine.

bgeltz (approved):

PR #4041, opened 2026-06-27
Relax batch write to no-op unadjusted controls (#4033)

### LINKED ISSUE #4033: BatchClient::write_batch returns misleading "server is unresponsive" and permanently kills the session when any pushed control was not adjusted before the first commit
**Describe the bug**
I tried to commit a batched control write via geopm_pio_write_batch() after pushing N control slots with geopm_pio_push_control() and adjusting only K < N of them with geopm_pio_adjust(), and I expected libgeopmd to either commit only the adjusted slots or return a typed per-call error that the caller can recover from. Instead, geopm_pio_write_batch() returned -1 with "BatchClient::write_batch The server is unresponsive: at src/BatchClient.cpp:95", the forked BatchServer process was killed, and every subsequent write_batch from the same process returned the same error, the BatchClient session is permanently dead with no recovery short of restarting the process.

The geopm_pio(3) man page describes write_batch as "Write all pushed controls so that values provided to geopm_pio_adjust() are written to the platform", this implies every pushed slot needs an adjust-provided value, but the precondition is never stated explicitly, the failure mode is not documented, and the client-side error message points at an unrelated cause (network/responsiveness, not input validity).

**GEOPM version**
$ geopmread --version
3.2.2

(Aurora geopm-runtime/3.2.0 module → library reports 3.2.2.)

**Expected behavior**
One of:

1. write_batch commits only the adjusted slots and leaves unadjusted slots untouched (matches the implication of "values provided to adjust are written").
2. write_batch returns the underlying GEOPM_ERROR_INVALID from MSRIOGroup::write_batch as a typed, recoverable per-call error. The BatchClient session must survive the violation so the caller can adjust the missing slot and retry.
3. The doc explicitly states the precondition ("every pushed control must be adjusted at least once before the first write_batch") and the failure mode (server death + permanent session loss).

The misleading "server is unresponsive" message in BatchClient.cpp:94 should be replaced by the underlying error text regardless of which fix is chosen, the server is unresponsive because it died, but the user has no way to know why.

**Actual behavior**
Call chain (read from upstream dev source):

1. Forked BatchServer reaches MSRIOGroup::write_batch (libgeopmd/src/MSRIOGroup.cpp:859-868), which throws GEOPM_ERROR_INVALID because std::any_of(m_is_adjusted, !adjusted) is true.
2. The exception unwinds through BatchServerImp::update_and_write (BatchServer.cpp:269-282) into run_batch (BatchServer.cpp:179-200), which sends M_MESSAGE_QUIT to the client and rethrows, the server process dies.
3. The client's blocking receive_message(M_MESSAGE_CONTINUE) in BatchClient::write_batch (BatchClient.cpp:78-97) receives QUIT instead of CONTINUE, throws, and BatchClient.cpp:94 rewrites the error as "The server is unresponsive".
4. Subsequent write_batch calls from this client fail identically, the server is gone.

m_is_adjusted[i] flips true only via MSRIOGroup::adjust (MSRIOGroup.cpp:885-903) and is never reset by write_batch, so the workaround "pre-adjust every pushed slot to its current value once" works (and is what we use). Single-shot geopm_pio_write_control() (sync path) is unaffected.

Related but distinct: #452 ("MSRIOGroup does not list unadjusted controls on error", open 2018) covers diagnosability of the same precondition in the direct (sync) path; this report is the batch-IPC inheritance + permanent-session-death + misleading-message combination, which #452 does not address.

**Error messages produced**

Standalone reproducer log (full):

=== variant: adjust_all=false (expect FAIL) ===
  pkg0 original (or fallback) = 350 W
  pkg1 original (or fallback) = 350 W
  pushed CPU_POWER_LIMIT_CONTROL[pkg=0] -> batch slot 0
  pushed CPU_POWER_LIMIT_CONTROL[pkg=1] -> batch slot 1
  adjusted 1 of 2 pushed slots
  write_batch -> FAIL
  write_batch rc=-1: <geopm> Runtime error: BatchClient::write_batch The server is unresponsive: at src/BatchClient.cpp:95

=== variant: adjust_all=true (expect PASS) ===
  pkg0 original (or fallback) = 350 W
  pkg1 original (or fallback) = 350 W
  pushed CPU_POWER_LIMIT_CONTROL[pkg=0] -> batch slot 0
  pushed CPU_POWER_LIMIT_CONTROL[pkg=1] -> batch slot 1
  adjusted 2 of 2 pushed slots
  write_batch -> OK

Same trigger inside a geopm::Agent plugin loaded under geopmctl + LD_PRELOAD libgeopm:

[minrepro_batch] init: ctrl=CPU_POWER_LIMIT_CONTROL dom=1 n_dom=2 n_slots=2 adjust_all=false warmup_ticks=20
[minrepro_batch] pushed CPU_POWER_LIMIT_CONTROL[dom=1 idx=0] -> slot 0  original=350
[minrepro_batch] pushed CPU_POWER_LIMIT_CONTROL[dom=1 idx=1] -> slot 1  original=350
[minrepro_batch] tick=20, adjusted 1 of 2 pushed slots, calling write_batch
[minrepro_batch] write_batch FAILED rc=-1 : <geopm> Runtime error: BatchClient::write_batch The server is unresponsive: at src/BatchClient.cpp:95

Debug build: the run above is the release build from the Aurora geopm-runtime/3.2.0 module — not a local --enable-debug build. The call chain in "Actual behavior" was read from upstream dev source rather than a debug-runtime log. Happy to rebuild with GEOPM_GLOBAL_CONFIG_OPTIONS="--enable-debug" ./build.sh and re-run if maintainers want richer runtime diagnostics.

**Additional context**

Hardware: Aurora (ALCF) compute node, 2× Intel Xeon CPU Max 9470C (Sapphire Rapids HBM) + 6× Intel Data Center GPU Max 1550 (Ponte Vecchio). SLES15 SP4. oneAPI 2025.2.0, MPICH 4.3.1, cray-pals 1.8.0. geopm-runtime/3.2.0 (install hash geopm-runtime-3.2.0-ic334tr); BatchServer talks to geopm-service at /soft/libraries/geopm/service/2.0.2-1/.

Process layout: single rank, single thread, the bug is single-process. The plugin variant uses a 1-rank geopm_prof_* workload so geopmctl ticks adjust_platform().

Reproducer code attached (single .cpp, 140 lines, links only against libgeopmd.so.2).

[repro_standalone.cpp](https://github.com/user-attachments/files/29220119/repro_standalone.cpp)

Build & run:

icpx -O2 -std=c++17 repro_standalone.cpp \
     -L/usr/lib64 -l:libgeopmd.so.2 -Wl,-rpath,/usr/lib64 \
     -o repro_standalone
ZES_ENABLE_SYSMAN=1 ./repro_standalone

Exit 0 if reproduced.

cmcantalupo:
Pushing a control via geopm_pio_push_control() without a subsequent geopm_pio_adjust() previously caused MSRIOGroup::write_batch() to throw GEOPM_ERROR_INVALID.  In the non-root batch path this killed the geopmbatch server, and the client surfaced a misleading "server is unresponsive" error that permanently disabled the session.

Treat an unadjusted pushed control as a no-op instead, matching every other IOGroup (LevelZero, NVML, DCGM, Sysfs), which silently skip controls that were never adjusted.  The MSRIO read-modify-write already leaves a field untouched when its accumulated write mask is zero, so the current platform value is preserved.

- BatchServer::update_and_write(): skip invalid (NAN) slots via PlatformIO::is_valid_value() so the server no longer rejects them.
- MSRIOGroup::write_batch(): remove the throw and the now-unused m_is_adjusted bookkeeping; it is a silent no-op for unadjusted controls.
- Document the no-op contract for geopm_pio_write_batch() in geopm_pio.3.rst.
- Add/adjust unit tests for the relaxed behavior.

- Fixes #4033

### DIFF
--- a/docs/source/geopm_pio.3.rst
+++ b/docs/source/geopm_pio.3.rst
@@ -369,7 +369,10 @@ Batch Functions
 
 ``geopm_pio_write_batch()``
   Write all pushed controls so that values provided to
-  ``geopm_pio_adjust()`` are written to the platform.
+  ``geopm_pio_adjust()`` are written to the platform.  Any control that
+  was pushed via ``geopm_pio_push_control()`` but never adjusted is left
+  unchanged: its current value on the platform is preserved (the write is
+  a no-op for that control).
 
 ``geopm_pio_start_batch_server()``
   Creates a batch server with the following signals and controls.
--- a/libgeopmd/src/BatchServer.cpp
+++ b/libgeopmd/src/BatchServer.cpp
@@ -275,7 +275,13 @@ namespace geopm
         double *shmem_buffer = (double *)m_control_shmem->pointer();
         int buffer_idx = 0;
         for (const auto &handle : m_control_handle) {
-            m_pio.adjust(handle, shmem_buffer[buffer_idx]);
+            // An invalid (NAN) slot is a control that the client pushed but
+            // never adjusted.  Skip it so it is preserved at its current
+            // platform value rather than rejected; the IOGroup write_batch()
+            // treats an unadjusted control as a no-op.
+            if (PlatformIO::is_valid_value(shmem_buffer[buffer_idx])) {
+                m_pio.adjust(handle, shmem_buffer[buffer_idx]);
+            }
             ++buffer_idx;
         }
         m_pio.write_batch();
--- a/libgeopmd/src/MSRIOGroup.cpp
+++ b/libgeopmd/src/MSRIOGroup.cpp
@@ -830,7 +830,6 @@ namespace geopm
             result = m_control_pushed.size();
             m_control_pushed.push_back(control);
             control->setup_batch();
-            m_is_adjusted.push_back(false);
 
             if (control_name == "CPU_POWER_LIMIT_CONTROL" || control_name == "MSR::PKG_POWER_LIMIT:PL1_POWER_LIMIT") {
                 int enable_idx = push_control("MSR::PKG_POWER_LIMIT:PL1_LIMIT_ENABLE", domain_type, domain_idx);
@@ -859,10 +858,10 @@ namespace geopm
     void MSRIOGroup::write_batch(void)
     {
         if (m_control_pushed.size() != 0) {
-            if (std::any_of(m_is_adjusted.begin(), m_is_adjusted.end(), [](bool it) {return !it;})) {
-                throw Exception("MSRIOGroup::write_batch() called before all controls were adjusted",
-                                GEOPM_ERROR_INVALID, __FILE__, __LINE__);
-            }
+            // A control that was pushed but never adjusted is preserved at its
+            // current platform value: the per-field write mask remains zero, so
+            // MSRIO::write_batch() leaves those bits untouched.  This silent
+            // no-op matches the behavior of the other IOGroups.
             m_msrio->write_batch();
         }
         m_is_active = true;
@@ -895,11 +894,9 @@ namespace geopm
             // If setting power limit to 0, disable the limit-enable bit, otherwise enable the bit
             double enable_val = (setting != 0.0) ? 1.0 : 0.0;
             m_control_pushed[it->second]->adjust(enable_val);
-            m_is_adjusted[it->second] = true;
         }
 
         m_control_pushed[control_idx]->adjust(setting);
-        m_is_adjusted[control_idx] = true;
     }
 
     double MSRIOGroup::read_signal(const std::string &signal_name, int domain_type, int domain_idx)
--- a/libgeopmd/src/MSRIOGroup.hpp
+++ b/libgeopmd/src/MSRIOGroup.hpp
@@ -212,7 +212,6 @@ namespace geopm
             std::shared_ptr<Cpuid> m_cpuid;
             bool m_is_active;
             bool m_is_read;
-            std::vector<bool> m_is_adjusted;
 
             // time for derivative signals
             std::shared_ptr<geopm_time_s> m_time_zero;
--- a/libgeopmd/test/BatchServerTest.cpp
+++ b/libgeopmd/test/BatchServerTest.cpp
@@ -4,6 +4,7 @@
  */
 
 #include <unistd.h>
+#include <cmath>
 #include "gtest/gtest.h"
 #include "gmock/gmock.h"
 #include "BatchServer.hpp"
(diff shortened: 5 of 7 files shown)


bgeltz (approved):

bgeltz (approved):

cmcantalupo:
```
$ ./repro_standalone
libgeopmd BatchClient minimal reproducer
control under test : CPU_POWER_LIMIT_CONTROL, 2 package slots

=== variant: adjust_all=false (expect FAIL) ===
  pkg0 original (or fallback) = 150 W
  pkg1 original (or fallback) = 150 W
  pushed CPU_POWER_LIMIT_CONTROL[pkg=0] -> batch slot 0
  pushed CPU_POWER_LIMIT_CONTROL[pkg=1] -> batch slot 1
  adjusted 1 of 2 pushed slots
  write_batch -> FAIL
  write_batch rc=-1: <geopm> Runtime error: BatchClient::write_batch The server is unresponsive: at src/BatchClient.cpp:95

=== variant: adjust_all=true (expect PASS) ===
  pkg0 original (or fallback) = 150 W
  pkg1 original (or fallback) = 150 W
  pushed CPU_POWER_LIMIT_CONTROL[pkg=0] -> batch slot 0
  pushed CPU_POWER_LIMIT_CONTROL[pkg=1] -> batch slot 1
  adjusted 2 of 2 pushed slots
  write_batch -> OK

Summary
  variant adjust 1/2 : FAIL (bug reproduced) (child exit=1)
  variant adjust 2/2 : PASS (precondition satisfied) (child exit=0)
Verdict: BUG REPRODUCED — file upstream
$ # INSTALLED NEW VERSION OF GEOPM SERVICE THEN RERAN
$ ./repro_standalone
libgeopmd BatchClient minimal reproducer
control under test : CPU_POWER_LIMIT_CONTROL, 2 package slots

=== variant: adjust_all=false (expect FAIL) ===
  pkg0 original (or fallback) = 150 W
  pkg1 original (or fallback) = 150 W
  pushed CPU_POWER_LIMIT_CONTROL[pkg=0] -> batch slot 0
  pushed CPU_POWER_LIMIT_CONTROL[pkg=1] -> batch slot 1
  adjusted 1 of 2 pushed slots
  write_batch -> OK

=== variant: adjust_all=true (expect PASS) ===
  pkg0 original (or fallback) = 150 W
  pkg1 original (or fallback) = 150 W
  pushed CPU_POWER_LIMIT_CONTROL[pkg=0] -> batch slot 0
  pushed CPU_POWER_LIMIT_CONTROL[pkg=1] -> batch slot 1
  adjusted 2 of 2 pushed slots
  write_batch -> OK

Summary
  variant adjust 1/2 : PASS (unexpected!) (child exit=0)
  variant adjust 2/2 : PASS (precondition satisfied) (child exit=0)
Verdict: NOT REPRODUCED — check env / control availability

PR #4042, opened 2026-07-01
Heatmap agent

cmcantalupo:
Add a pure-Python session Agent (run via 'python -m geopmdpy.heatmap_agent') that samples per-component power, activity, and frequency signals in-process through geopmdpy.pio and renders a normalized component-saturation heatmap (rows=components, cols=time) when the session ends. A component pinned near its own maximum while the others sit idle reads out as the bottleneck.

When CPU_INSTRUCTIONS_RETIRED and CPU_CYCLES_THREAD are both available, the agent derives a per-sample IPC row from their counter deltas and emits it as a trace column to help distinguish compute-bound from memory-bound regions. An external figure-of-merit series can be overlaid with --fom, and an existing geopmsession trace can be re-rendered with --replot-csv without running a session. Heavy plotting dependencies (numpy, matplotlib, pandas) are imported lazily so importing the module needs none of them.

Add TestHeatmapAgent unit tests covering help text, argument parsing, signal-config override, IPC derivation, trace columns, replot, and column matching.

Add a geopm_heatmap_agent(1) man page documenting the synopsis, options, requirements, and limitations, and register it in the docs man-page list.

component_heatmap-ze_peer
component_heatmap-ze_peak

### DIFF
--- a/docs/source/geopm_heatmap_agent.1.rst
+++ b/docs/source/geopm_heatmap_agent.1.rst
@@ -1,0 +1,141 @@
+geopm_heatmap_agent(1) -- render a component-saturation heatmap from GEOPM telemetry
+====================================================================================
+
+Synopsis
+--------
+
+.. code-block:: bash
+
+    python -m geopmdpy.heatmap_agent [-t TIME] [-p PERIOD] [-o TRACE_OUT]
+                                     [-i CONFIG_PATH] [--heatmap-out PATH]
+                                     [--fom FOM_CSV] [--no-plot]
+                                     [--replot-csv TRACE_CSV]
+
+Sample and render (default)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: bash
+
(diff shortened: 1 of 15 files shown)


docs/source/geopm_heatmap_agent.1.rst
```diff
@@ -0,0 +1,141 @@
+geopm_heatmap_agent(1) -- render a component-saturation heatmap from GEOPM telemetry
+====================================================================================
+
+Synopsis
+--------
+
+.. code-block:: bash
+
+    python -m geopmdpy.heatmap_agent [-t TIME] [-p PERIOD] [-o TRACE_OUT]
```
bgeltz: So are we going back to examples with "python" instead of "python3"?  Just noting if so.

bgeltz (dismissed):
Is there a higher level page discussing the Python agents that should include a short blurb about this new one?

bgeltz (approved):

PR #4052, opened 2026-08-17
LevelZero Tools & Metrics (ZET) support

bgeltz:
## Summary

Adds LevelZero Tools & Metrics (ZET) support to GEOPM, exposing per-chip GPU activity signals sourced from the LevelZero metric streamer, and builds out the sampling path so telemetry stays accurate and cheap across a range of controller/monitor sampling periods. Also extends the GPU activity agent to use the new stall signal in its frequency decision and to react to idle GPUs.

## New capabilities

**LevelZero Tools & Metrics (ZET) signals.**
Introduces metric-streamer-backed signals on the existing `GPU_CHIP` domain, including `LEVELZERO::METRIC:XVE_ACTIVE`, `LEVELZERO::METRIC:XVE_STALL`, and `LEVELZERO::METRIC:NUM_REPORTS`, derived from time-based streamer reports via `CalculateMetricValues`. Collection spans all available GPU chips.

**IOGroup / DevicePool plumbing.**
Extends `LevelZeroIOGroup`, `LevelZeroDevicePool`, and the LevelZero implementation to register, read, and reduce the new metric signals, with accompanying mocks and unit tests.

**GPU activity agent: stall-aware frequency selection.**
When the `LEVELZERO::METRIC:XVE_STALL` signal is available, the agent discounts compute activity by the stall fraction (`gpu_core_activity *= (1 - stall)`) before selecting a frequency, so stall-heavy phases request a lower frequency for energy savings. It falls back to compute-activity-only selection when the stall signal is absent.

**GPU activity agent: idle reaction and tracking.**
Adds a debounced idle detector (per-chip idle timer) that drives GPUs to the minimum frequency after sustained idle when energy-biased, plus GPU on-time/on-energy and active-region tracking surfaced in the agent report.

## Sampling path robustness & performance

**Metric quality was coupled to the consumer's sampling period.**
The cost of turning raw streamer reports into metric values scales with the number of reports accumulated in the FIFO, and accumulation grows with the read period. A short period stayed healthy, but a slower consumer let the FIFO back up, stalling the streamer into multi-second `NOT_READY` bursts so telemetry stopped updating. The safe operating point was accidental, not guaranteed.

**Under-drained FIFO from a stale read-size.**
The read-buffer size was reused as the streamer's in/out byte count, so it collapsed to a single report after the first read and never recovered — every later read drained only ~1 report, leaving the FIFO to overflow and produce stale / misaligned data.

**Fix — decouple draining from computing ("read all, process last N").**
Each read now drains the entire FIFO using the full buffer, so freshness no longer depends on the read period, while only the most recent `DEFAULT_MAX_REPORTS_PER_READ` (30) reports are passed to `CalculateMetricValues`, bounding compute cost independent of period.

**Background keep-alive drain thread.**
A background thread drains active streamers on a fixed cadence (`METRIC_DRAIN_PERIOD_US`, 20 ms), independent of the controller loop, keeping the LevelZero metric streamer delivering continuously even when the controller reads infrequently; the controller's own reads count as drains so the two never fight.

**Robustness to dropped / empty data.**
Reads no longer throw on dropped data from the FIFO or on empty reports, avoiding spurious failures under overflow or idle GPUs.

**Per-sample overhead reductions.**
Pass the report buffer by reference instead of by value, and cache the metric name → report-index map rather than rebuilding it each sample.

### DIFF
--- a/libgeopm/src/GPUActivityAgent.cpp
+++ b/libgeopm/src/GPUActivityAgent.cpp
@@ -72,6 +72,17 @@ namespace geopm
         m_resolved_f_gpu_efficient = 0;
         m_f_range = 0;
 
+        for (int domain_idx = 0; domain_idx < M_NUM_GPU; ++domain_idx) {
+            m_gpu_active_region_start.push_back(0.0);
+            m_gpu_active_region_stop.push_back(0.0);
+            m_gpu_active_energy_start.push_back(0.0);
+            m_gpu_active_energy_stop.push_back(0.0);
+
+            m_gpu_on_time.push_back(0.0);
+            m_gpu_on_energy.push_back(0.0);
+            m_prev_gpu_energy.push_back(0.0);
+        }
+
         if (level == 0) {
             init_platform_io();
         }
@@ -144,6 +155,8 @@ namespace geopm
             m_gpu_freq_max_control.push_back(m_control{m_platform_io.push_control("GPU_CORE_FREQUENCY_MAX_CONTROL",
                                                        m_agent_domain,
                                                        domain_idx), NAN});
+            m_gpu_idle_timer.push_back(10);
+            m_gpu_idle_samples.push_back(0);
         }
 
         // We treat energy & time as special cases and only use them at a specific domain.
@@ -159,6 +172,8 @@ namespace geopm
 
         m_freq_gpu_min = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MIN_AVAIL", GEOPM_DOMAIN_BOARD, 0);
         m_freq_gpu_max = m_platform_io.read_signal("GPU_CORE_FREQUENCY_MAX_AVAIL", GEOPM_DOMAIN_BOARD, 0);
+
+        //m_platform_io.write_control("GPU_CORE_FREQUENCY_MIN_CONTROL", GEOPM_DOMAIN_BOARD, 0, m_freq_gpu_min);
 
         const auto ALL_NAMES = m_platform_io.signal_names();
         // F efficient values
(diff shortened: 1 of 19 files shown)


libgeopmd/src/LevelZero.cpp
```diff
@@ -4,19 +4,28 @@
...
-#include "geopm/Agg.hpp"
 #include "geopm/Helper.hpp"
+#include "geopm_debug.hpp"
 
 #include "LevelZeroImp.hpp"
 
+static void __attribute__((constructor)) geopm_levelzero_init(void)
+{
+    setenv("ZES_ENABLE_SYSMAN", "1", 1);
+    setenv("ZET_ENABLE_METRICS", "1", 1);
```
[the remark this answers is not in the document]
bgeltz: This is intentional. `ZES_ENABLE_SYSMAN` and `ZET_ENABLE_METRICS` must be set to `1` before the Level Zero loader/driver reads them at initialization for Sysman and the metrics APIs to function; the overwrite is required so GEOPM's Level Zero support works regardless of the inherited environment. Leaving these to an inherited (possibly unset or `0`) value would break required Level Zero functionality, so we deliberately force them here. Resolving as working-as-intended.

bgeltz: This could be moved to the geopm.service file.

cmcantalupo: I think moving these to the service file seems like a good idea since calling setenv() can disrupt previously obtained getenv() pointers.

libgeopm/src/GPUActivityAgent.cpp
```diff
@@ -412,12 +522,44 @@ namespace geopm
         std::vector<std::pair<std::string, std::string> > result;
 
         result.push_back({"Agent Domain", m_platform_topo.domain_type_to_name(m_agent_domain)});
+        if(m_gpu_stall_activity.size() > 0) {
```
bgeltz:
```suggestion
        if (m_gpu_stall_activity.size() > 0) {
```

libgeopmd/src/LevelZero.cpp
```diff
@@ -513,10 +578,561 @@ namespace geopm
...
+            current["NUM_REPORTS"] = std::vector<double>{ static_cast<double>(num_reports) };
+        }
+
+        // Launch the background sampling thread on first use.
+        metric_thread_start();
+    }
+
+    void LevelZeroImp::metric_thread_start(void)
+    {
+        // Called only from the controller thread (metric_read), so the started
```
bgeltz: This is slightly misleading - yes there is a controller thread but since we don'e have escalated privs in most cases, these requests will be routed through the root running geopmd service and thus satisfied via the geopmbatch server.

cmcantalupo (commented):
Lgtm

sidjana (approved):
LGTM!

Issue #4054, opened 2026-08-24
As a user of the geopmopt command line tool I would like AI assistant guidance that takes me from an uninstalled system to a tuned workload


cmcantalupo:
**Is your feature request related to a problem? Please describe.**

A user who owns a workload or benchmark and understands how to run it, but who
has never used GEOPM, currently faces a steep on-ramp before `geopmopt` becomes
useful to them. They must first determine whether GEOPM is installed, choose
among several install paths, understand the access-list model well enough to get
controls granted, and then learn the signal/control/domain vocabulary and the
`geopmopt` objective interface before they can express what they actually want:
"what hardware settings make my workload fastest, or most energy efficient, on
this machine?"

Each of those steps is documented, but the documentation is organized by
component rather than by that user's journey, and nothing helps them decide
which path applies to their situation.

**Describe the solution you'd like**

Repository-hosted AI assistant tooling, so that an agent working in a checkout
(or with the skills available to it) can guide this user end to end. Two
assistants, because the first is a prerequisite for every GEOPM use case and not
just optimization:

1. **Install assistant** — probe the system under test, recommend exactly one
   install path, perform the install behind confirmation gates, generate a
   least-privilege `geopmaccess` command block for an administrator to review,
   and verify readiness.
2. **`geopmopt` assistant** — interview the user about their workload, probe
   which sweep dimensions the platform actually supports, construct the
   objective and any constraints, run a cheap smoke test before committing to a
   long campaign, and interpret the resulting configuration.

Delivered as `SKILL.md` bundles under `.github/skills/` with thin `.agent.md`
personas under `.github/agents/`, plus a repository-level `AGENTS.md` navigation
guide. The skill format is portable across agent runtimes, and none of these
files are shipped in any package, so there is no packaging impact.

**Describe alternatives you've considered**

- *Documentation alone.* The material largely exists already; the gap is
  decision support, not reference content. Prose cannot probe the platform to
  determine which of the sweep dimensions have real bounds on this machine, or
  which controls this user is permitted to write.
- *A single combined assistant.* Rejected because installation is a prerequisite
  for telemetry collection, `geopmsession` tracing, and Prometheus export as
  well, and should be reusable by those use cases.
- *A wrapper script or TUI.* Higher maintenance burden and does not help with the
  open-ended parts (choosing an objective, writing a metric regex, judging
  whether a campaign converged).

**Additional context**

Work is planned in `plan-for-agents.md`, which breaks this into fifty tasks
[...]

Issue #2776, opened 2023-01-04
geopmd: replace 1-second PID polling with pidfd notification


cmcantalupo:
**Is your feature request related to a problem? Please describe.**

`geopmd` detects client-process death by **polling**. In `PlatformService`
(`geopmdpy/geopmdpy/service.py`), `_watch_client()` registers a GLib timeout
that fires once per second and `check_client()` calls
`psutil.pid_exists(client_pid)` each time; when the PID is gone it calls
`_close_session_completely()`. This has two costs:

1. **Inefficient** — every open session wakes the daemon once per second
   regardless of activity, and `psutil.pid_exists` does a `/proc` stat on each
   wakeup. The cost scales with the number of concurrent sessions.
2. **Up to 1s latency** before a dead client's controls are restored and its
   write lock is released.

It would be more efficient to use `pidfd_open(2)` together with `poll()` to be
notified of PID termination on the actual event instead of polling.

**Describe the solution you'd like**

Use `pidfd_open(2)` + `poll()` instead of polling `/proc`. A pidfd becomes
readable (`POLLIN`) exactly when the target process terminates, so it can be
registered with the daemon's GLib main loop and the death callback fires only on
the real event — zero wakeups while the client lives, and cleanup runs the
instant the client dies (faster than today's up-to-1s poll).

Because `os.pidfd_open()` requires CPython ≥ 3.9 (and the syscall requires Linux
≥ 5.3) while geopm still supports Python 3.6, the pidfd path must be **optional
and runtime-selected**, with the existing 1-second polling retained as a
**permanent fallback**.

**Scope note — PID namespaces are out of scope.** The original request also
raised tracking PIDs across PID namespaces via client-supplied pidfd passing
over the socket (`SCM_RIGHTS`). That is **not** pursued here: on the deployed
gRPC path the host-side `geopmd-proxy` reads `SO_PEERCRED` and forwards a
host-namespace PID to the daemon, so server-side `pidfd_open()` already tracks
containerized clients correctly without any fd passing. Client→daemon pidfd
passing (the part that made #2772 complex) is intentionally excluded from this
work.

**Describe alternatives you've considered**

The current solution resolves the client PID (DBus `GetConnectionUnixProcessID`,
or the proxy-injected PID on the gRPC path) and periodically calls
`psutil.pid_exists(client_pid)`. It works but is inefficient and has up-to-1s
cleanup latency.

**Additional context**

- A sketch was started in PR #2772 but the added process management and new
[...]

PR #4064, opened 2026-09-10
geopmopt assistant - updates post workload validation

sidjana:
Relates to the GEOPM AI assistants effort (#4054).

Ran a real single-node energy-efficiency campaign (OpenMP STREAM, CentOS
Stream 9, 2-socket Xeon 6978P, no GEOPM previously installed) by following
the install and optimize skills exactly. Nine collateral defects turned up
in the skill/script collateral itself (not `docs/source/`), each costing a
detour into geopmdpy source or manual `geopmread`/`geopmaccess`/`ps`
diagnosis that the collateral should have prevented. Eight are fixed in
this PR; the ninth needs an actual `optimizer.py` code change rather than
a collateral fix, so it is tracked separately instead (see below):

- Missing `python3-devel`/`pyconfig.h` prerequisite in `client-venv.md`.
- `CPU_FREQUENCY_GOVERNOR_CONTROL` missing from the default access grant.
- `cpu-power` granting/testing the wrong control name
  (`POWERCAP::CPU_POWER_LIMIT` vs `CPU_POWER_LIMIT_CONTROL`).
- `geopm-check-workload.sh` baseline measured under different conditions
  than the campaign; added `--dimension`/`--venv`.
- `uncore-freq`'s auto-detected max bound can resolve to a hard zero.
- `geopm-sensitivity.sh`'s coarser-step remedy could suggest a step that
  doesn't evenly divide `(max - min)`; it now shifts `min` by the
  remainder (always less than one step) instead of failing outright.
- No troubleshooting entry for the cross-session write-mode lock.
- `geopmopt.agent.md`/`SKILL.md` now require asking the user whether/where
  to save the recommended configuration with `--output-file` before
  running the full campaign.

Subsequent review rounds surfaced further collateral defects, all fixed
here: readiness/probe gates that accepted unusable dimensions (`n/a` vs
degenerate numeric bounds, `step<=0`), `geopmread` used as a proxy for
write access where the granted *control* list was needed, unchecked
`geopmaccess` failures silently degrading to empty lists, companion
controls (the cpu-freq governor, paired `*_MIN_*` controls) not being
validated, long `--dimension` aliases never matching `--list-controls`,
bound signals missing from the generated grant so a freshly granted
dimension still reported `n/a` bounds, and a combined multi-dimension
baseline (`--dimension` is now repeatable) so the campaign's result is
judged against a reference it can actually reproduce.

Three design decisions worth calling out:

- **Default sweep dimensions are `cpu-freq`, `uncore-freq` and `cpu-power`**
  (plus `gpu-freq`/`gpu-power` on GPU workloads). When a user asks to
  optimize a workload, the assistant picks from these.
- **`prefetch` is excluded from that default set**, even where the platform
  supports it. A prefetch level expands into four ordered MSR writes, which
  neither `geopm-check-workload.sh` nor `geopm-sensitivity.sh` performs, so
  a campaign including it has no baseline reference and no sensitivity
  screening. It is swept only on explicit request, and its share of any
  improvement is reported as **unverified**. The collateral makes no claim
  about the firmware's prefetcher state. The access-side collateral still
  treats the four MSR controls as one required set, since `geopmopt` itself
  can sweep `prefetch`.
- **Bounds are resolved at the domain the campaign uses (`board`).**
  `--list-controls` prints *native*-domain bounds, but `ControlGrid`
  resolves a sweep's bounds by reading the bound signals at the requested
  domain, and the two differ for sum-aggregated controls
  (`CPU_POWER_LIMIT_DEFAULT` is one package's limit at `package`, the board
  sum at `board`). Both helpers now read bounds at `board`. Defaulting to
  the native domain instead would have been worse: `add_dimension()`
  creates one optimizer dimension per domain index, so a `cpu`-native
  control would turn a one-knob test into a 480-way search.

Not fixed in this PR:

- `--application-timeout` does not reliably kill the launched process tree.
  `ApplicationEvaluator` still uses `subprocess.run(..., timeout=...)`
  without process-group cleanup. This needs an `optimizer.py` change, not
  a collateral change, so it's tracked in #4065 instead of claimed as fixed
  here.

Also found during review of this PR and tracked separately, since each
needs a core `geopmdpy` change rather than collateral:

- A `geopm-check-workload.sh --dimension` baseline's temporary
  `geopmsession` can leak the write-mode lock/control snapshot if it exits
  abnormally, since it runs as a plain child of the shell rather than its
  own session leader — #4066.
- `geopm-sensitivity.sh`'s coarse-step remainder arithmetic is
  integer-only, so a control with fractional bounds can still be handed a
  sweep that fails `geopmopt`'s divisibility check — #4067.
- `ControlGrid` supports a zero-span (`min == max`) dimension on purpose,
  but `BayesianOptimizer` rejects it, so a pin-one-axis campaign cannot
  run — #4068. The collateral's `min <= max` checks are left as-is
  deliberately, pending that fix.

No change to `plan-for-agents.md`: it is added and then removed again
within this branch's history, netting to no change against `dev`, which
already lacks it.


sidjana:
Follow-up on the [Copilot review](https://github.com/geopm/geopm/pull/4064#pullrequestreview-5183696064) (both the 3 inline comments and the 12 "suppressed" findings in the review body):

### Inline comments — all resolved
- [r3993662122](https://github.com/geopm/geopm/pull/4064#discussion_r3993662122) (`geopm-check-workload.sh:132`, long `--dimension` aliases not normalized before the `--list-controls` lookup) → fixed in `2e6e1d65f`.
- [r3993662187](https://github.com/geopm/geopm/pull/4064#discussion_r3993662187) (`geopm-check-workload.sh:146`, numeric bound check missing from probe/readiness paths) → fixed in `aa06e605d`.
- [r3993662232](https://github.com/geopm/geopm/pull/4064#discussion_r3993662232) (`access-lists.md:129`, "passes every readiness check" no longer accurate) → fixed in `aa06e605d`.

### Suppressed findings — resolved
- `geopm-check-workload.sh:165` — `prefetch` dimension had no mapping under `--dimension` → now explicitly rejected with an explanatory error (`502a3bf2f`) instead of replicating `grid.py`'s 4-MSR expansion here.
- `geopm-install/references/troubleshooting.md:225` — "run a command from the same session to release the lock" was wrong (lock is owned by the session leader, not the client PID) → corrected (`502a3bf2f`).
- `geopm-install/references/access-lists.md:120` — validation example still granted `CPU_POWER_LIMIT_CONTROL` → fixed to `POWERCAP::CPU_POWER_LIMIT` (`502a3bf2f`).
- `geopm-install/scripts/geopm-gen-access.sh:28` — governor comment said omitting it "fails partway"; `geopm-verify-install.sh`'s companion check now fails the gate itself → reworded (`502a3bf2f`).
- `geopm-optimize/references/sweep-dimensions.md:140` — probe/readiness scripts only checked `n/a`, not numeric `max>0`/`min<=max` → fixed in `aa06e605d`.
- `geopm-optimize/scripts/geopm-sensitivity.sh:122` — sensitivity only wrote `*_MAX_CONTROL`, not the paired `*_MIN_CONTROL` `grid.py` pins → both `measure_once()` and `achieved_freq()` now write the companion MIN control (`502a3bf2f`).

### Suppressed findings — intentionally not resolved here
- `geopm-optimize/scripts/geopm-sensitivity.sh:414` — coarse-step remainder calc is integer-only and can still reject some fractional-bound sweeps. Left as-is at the author's request; not changed in this PR.
- `geopm-optimize/scripts/geopm-check-workload.sh:217` and `.github/agents/geopm-optimize.agent.md:61` and `geopm-optimize/SKILL.md:172` — the `--dimension`/`--venv` baseline mode still isn't wired into the prescribed `SKILL.md`/agent procedure (step 3), so following the skill as written still measures unconstrained defaults. Not addressed in this PR.
- `geopm-optimize/scripts/geopm-check-workload.sh:217` (also flagged separately) — the temporary `geopmsession` launched under `--dimension` runs as a plain child of the interactive shell, so the Access Service's write-mode lock/control snapshot can be tied to the shell's session leader rather than `geopmsession` itself, and may not be released/restored if `geopmsession` exits abnormally. Not fixed here; filed as **#4066** with root cause and fix plan (distinct from #2776, which only addresses polling latency for whichever PID is being watched, not the wrong-PID tracking here).
- `geopm-optimize/SKILL.md:231` — the PR's original description claims a fix for `--application-timeout` not killing the launched process tree. No `optimizer.py` code change accompanies this PR; the documentation claiming the symptom was removed from `campaign-design.md`/`troubleshooting.md` (`e79d5ffb0`) and the actual code fix is tracked separately in **#4065**. The PR's top-level description may still need a manual edit to stop listing this as fixed.

Net: all 3 inline comments and 6 of the 12 suppressed findings are resolved on this branch. The remaining 6 are either intentionally deferred to follow-up issues (#4065, #4066) or out of scope for the groups addressed so far (the `--dimension`/`--venv` wiring into `SKILL.md`/agent persona, and the integer-only remainder calc, kept as-is by request).

sidjana:
**Follow-up on the [Copilot review #5192425521](https://github.com/geopm/geopm/pull/4064#pullrequestreview-5192425521)**

### Inline comments — all resolved (`f3c406fba`)
- [r4001024056](https://github.com/geopm/geopm/pull/4064#discussion_r4001024056) (`geopm-verify-install.sh:278`, hard-zero uncore bound misdiagnosed as "every dimension unresolved") → the all-unusable failure message now distinguishes "domain never resolved" from "domain resolved but numeric bounds invalid."
- [r4001024072](https://github.com/geopm/geopm/pull/4064#discussion_r4001024072) (`geopm-check-workload.sh:212`, MIN companion written unconditionally) → now checks platform support (`geopmaccess --all --controls`) before writing the MIN control.
- [r4001024082](https://github.com/geopm/geopm/pull/4064#discussion_r4001024082) (`geopm-probe-controls.sh:92`, "nothing can be swept" message misleading for the bounds-invalid case) → same domain-unresolved vs. bounds-invalid distinction applied here too.
- [r4001024088](https://github.com/geopm/geopm/pull/4064#discussion_r4001024088) (`geopm-sensitivity.sh:142`, `geopmread` conflated "unsupported" with "not granted") → now distinguishes platform support from user access, and fails loudly instead of silently measuring MAX-only when a MIN control is supported but not granted.

### Suppressed findings — resolved (`f3c406fba`)
- `geopm-verify-install.sh:278` / `geopm-probe-controls.sh:92` — added `step>0` to the numeric-bounds predicate (a zero/negative step now correctly excluded from "usable").
- `geopm-install/references/troubleshooting.md:237` and the analogous text in `geopm-optimize/references/troubleshooting.md` — corrected the overstated claim that any background/async execution starts a second session and contends for the write lock; only `setsid`, a new terminal/SSH session, or a service manager actually does.
- `geopm-install/references/troubleshooting.md:210` and `geopm-optimize/references/troubleshooting.md:50` — updated the "nothing is sweepable" guidance to cover the resolved-domain-but-invalid-bounds case (e.g. `uncore-freq max=0`), not just `n/a` domains.
- `geopm-verify-install.sh:38` — the MIN/governor companion check now only requires a MIN control when it's actually supported on the platform (matching `grid.py`'s conditional pairing); the governor companion remains unconditionally required.
- `geopm-probe-controls.sh:149` — the access diagnostic now also checks governor/MIN companions, not just each dimension's primary control, regardless of whether the dimension was already classified usable.
- `geopm-sensitivity.sh:126` — long dashed aliases (`cpu-frequency`, `cpu-uncore-frequency`, `gpu-frequency`) are now normalized before the `--list-controls` lookup, matching `geopm-check-workload.sh`.

### Previously-missed finding — resolved (`c58dbf641`)
- `.github/agents/geopm-optimize.agent.md:61` (and the analogous `SKILL.md:172`) — the guided workflow's baseline step now passes `--dimension`/`--venv` to `geopm-check-workload.sh`, so it actually measures under the same forced governor/pinned-MIN/MAX conditions the campaign uses, instead of unconstrained defaults.

### Not resolved here — filed as a new issue
- `geopm-sensitivity.sh:426` — the coarse-step remedy's remainder calculation is integer-only (`printf "%.0f"` rounds `new_min`), so a control with fractional bounds/step can still get a suggested `--sweep` that fails `geopmopt`'s own divisibility check. Kept as-is in this PR (integer bounds cover the controls actually swept in practice); tracked in **#4067** with a concrete counter-example and a suggested fix.

sidjana:
Summary of tasks achieved
All 15 findings across both reviews are resolved. Grouped by root cause:

1. geopmread used as a writability test (4 findings) — GEOPM grants signals and controls separately, so probing a control with geopmread both accepts readable-but-unwritable controls and rejects writable ones whose signal alias wasn't granted. Replaced with granted-control-list checks (geopmaccess --controls) in:

geopm-sensitivity.sh — MIN companion (r4001024088 follow-up, :160) and cpu-freq governor (:180); a missing governor grant is now a setup error instead of silently measuring a different operating point than the campaign.
geopm-check-workload.sh — --dimension governor write (:199) and MIN companion; now fails clearly rather than baselining under the user's current governor.
2. Wrong failure mode documented for a missing MIN companion (4 findings) — grid.py omits a companion absent from pio.control_names(), which for a service-backed client is the granted list. So granting only MAX produces no permission error; geopmopt drops the companion and silently sweeps a MAX-only cap, losing the intended pinning. Corrected in access-lists.md:132, geopm-gen-access.sh:23, geopm-verify-install.sh:34 (+ its gate failure text), and geopm-probe-controls.sh:189's diagnostic (which had claimed sensitivity/campaign would surface the gap — nothing downstream does).

3. n/a bounds conflated with numerically-invalid bounds (2 findings) — a resolved domain with an n/a bound is an access problem, but was being reported as "not an access-list limitation." Both geopm-probe-controls.sh:138 and geopm-verify-install.sh:310 now count only rows with all bounds present-but-degenerate as invalid-bounds, report n/a-bound rows separately pointing at bounds-signal grants, and geopm-probe-controls.sh now falls through to the access-list diagnosis before exiting instead of hiding the likeliest cause.

4. prefetch dead-end in the guided flow (2 findings) — the probe reports prefetch usable but the helper rejects it, so the prescribed per-dimension baseline ended in a usage error. SKILL.md:113 and geopm-optimize.agent.md:57 now call out the exception and its manual path.

5. Divide-by-zero in the sensitivity remedy (1) — geopm-sensitivity.sh:445 now validates max>0/min<=max/step>0 before the coarse-step arithmetic.

6. Doc/text consistency (2) — sweep-dimensions.md's opening usability criterion now includes the numeric checks (it contradicted its own later section); geopm-check-workload.sh:41's --help prefetch note is self-contained; geopm-probe-controls.sh:206 grammar fixed.

Verification: all five scripts pass bash -n, plus four behavioral smoke tests with stubbed geopmopt/geopmaccess confirming correct classification of bad-numeric-bounds vs n/a-bounds cases and correct governor grant-check pass/fail behavior.

Still open by design: geopm-sensitivity.sh's integer-only coarse-step remainder arithmetic, tracked in [#4067](https://github.com/geopm/geopm/issues/4067).

cmcantalupo (approved):

